PR #67, opened 2024-10-03
Add concept of partially configured arch def

dhower-qc:


dhower-qc:
Changes are added in #68

PR #114, opened 2024-10-17
Merged F extension instructions. Some fields are not populated

AFOliveira:
This PR is only about the F extension. I merged in the info from dhower/F with the exception of operations because they were creating some errors. I haven't taken a close look to ALL fields and therefore some of them have different syntax for source registers. The html builds sucessfully with this and the regress test passes.


dhower-qc (changes requested):
Cool. The errors you are getting with the operation() are probably just because some things have moved around since that dhower/f branch was created. It's likely a quick fix. 

Would you be able to add them back in on this branch? I can then run it locally and fix it up.

AFOliveira:
> Would you be able to add them back in on this branch? I can then run it locally and fix it up.

Sure, give me just a couple minutes and I'll commit them here.

AFOliveira:
@dhower-qc Done!

dhower-qc:
I added some missing files (fp.idl) and fixed up the IDL code where, e.g., function signatures have changed. Should pass now.

arch/inst/F/fabs.s.yaml
```diff
@@ -0,0 +1,25 @@
+# yaml-language-server: $schema=../../../schemas/inst_schema.json
+
+fabs.s:
```
dhower-qc, line 3: fabs.s is a pseudoinstruction. In the UDB scheme, it should be listed as such under fsgnjx.s like so:

``` yaml
fsgnjx.s:
  # ...
  pseduoinstructions:
  - when: "fs1 == fs2"
    to: fabs.s fd, fs1
```

arch/inst/F/fadd.s.yaml
```diff
@@ -0,0 +1,27 @@
...
+fadd.s:
+  long_name: No synopsis available.
+  description: |
+      No description available.
+  definedBy: F
+  assembly: xd, xs1, xs2, rm
+  encoding:
+    match: 0000000------------------1010011
+    variables:
+    - name: rs2
```
dhower-qc, line 12: is it possible to get the names of the variables as "fs2" when they are floating point registers? This would also apply to the assembly string. I know this isn't quite how the ISA manuals treat things, but having the register file explicit helps with machine parsing.

dhower-qc (changes requested):

arch/inst/F/fabs.s.yaml
AFOliveira: Surely! The script does this by itself, I believe the reason for the error here is that I forgot to update riscv-opcodes to its most recent head. In that note, may I also commit the submodule update to it's latest commit?

arch/inst/F/fadd.s.yaml
AFOliveira: Surely, I think this will need to be manually but I'll get to it and fix them.

AFOliveira:
@dhower-qc Another question on this, when an instruction is defined by several extensions, in which folder do we want the .yaml described? Possibly all of them?

arch/inst/F/fabs.s.yaml
AFOliveira: This is now solved in the most recent commit.

arch/inst/F/fadd.s.yaml
AFOliveira: @dhower-qc do you mind taking them on this not fully correct format for now? Just to kind of get all instruction skeletons in place and once that's done come back to this known issues and fix the one by one? I understand that we'll probably be creating some tech debt with this, but given the proximity to the summit, I think it could be beneficial to have all the skeletons in. BTW I'll gladly do whatever option you prefer, I just don't think we'll be able to get all skeletons by the 22nd.

AFOliveira: Done.

AFOliveira:
@dhower-qc Unless I missed or wrongfully assigned any register, this PR is now ready.

dhower-qc (approved):

Issue #163, opened 2024-10-22
MC-1 shows several *h CSRs that aren't present in Priv 1.11 and is missing medeleg


james-ball-qualcomm:
For some reason the medeleg CSR isn't present in MC-1 but the medelegh CSR is present (but shouldn't be).

The following CSRs are shown in the MC-1 doc but weren't in existence in Priv 1.11. They were added in 1.12 or 1.13 (I haven't looked). In any case, UDB treats these *h registers as mandatory. Is that correct?

Here's the *h CSRs I'm talking about:

- mdelegh
- mseccfgh (also mseccfg shows in MC-1 but didn't exist in Priv 1.11)
- mstatush

james-ball-qualcomm:
BTW, mconfigptr is still in the MC-1 doc but shouldn't be there (didn't exist in Priv 1.11). I believe there is an issue for this that is a few weeks old and I thought Derek had fixed this but it isn't in main yet if there is a fix.

james-ball-qualcomm:
The mconfigptr (and probably some of these other CSRs) was supposed to be fixed in https://github.com/riscv-software-src/riscv-unified-db/issues/53 a few weeks ago but it still isn't fixed.

PR #167, opened 2024-10-23
Make CRD gen check ext versions for in scope. Update CSRs with correc…

dhower-qc:
…t version info


dhower-qc:
fixes #163

arch/csr/mseccfg.yaml
james-ball-qualcomm, line 1: I see from this change that a space is required between the operator (>= in this case) and the version number. This seems like a common bug. Can you remove this requirement to have a space or have the schema check for it? Or have an error generated when creating backends or parsing the database?

arch/csr/medelegh.yaml
james-ball-qualcomm, line 1: Don't mdelegh and mseccfgh need the same "definedBy" as you added for mstatush? Namely,

definedBy:
    name: Sm
    version: ">= 1.12"

backends/crd_doc/templates/crd.adoc.erb
james-ball-qualcomm, line 1: Can you explain why you had to switch from in_scope_extensions() to in_scope_ext_reqs()? I want to add comments to the code to explain why one is used vs. the other for future reference.

lib/arch_obj_models/extension.rb
james-ball-qualcomm, line 1: You have a typo: implemente**s**d_instructions. Did you test this?

james-ball-qualcomm (changes requested):
Mostly just questions to better understand your changes but one typo found that is a real bug.

backends/crd_doc/templates/crd.adoc.erb
dhower-qc: When you just ask for the extension list, there is no version info -- it's just a list of extension info. You need the ExtensionRequirement objects (which combines and Extension with a Requirement) in order to filter by, e.g., a "= 1.11" requirement.

lib/arch_obj_models/extension.rb
dhower-qc: I did test it (and you can see it's passing the regressions). I added this function thinking I was going to use it, but then changed my mind. It's still a useful function, which is why I left it. Good job finding the typo.

arch/csr/medelegh.yaml
dhower-qc: Fixed medelegh (>= 1.13). mseccfgh already says >= 1.12

PR #182, opened 2024-10-24
Move mandatory privilege mode option from CRD family to CRD

### LINKED ISSUE #181: Move priv modes present from CRD family to CRD
MC-1 is only M-mode but a future MC release could include optional U and/or S modes.

james-ball-qualcomm:
Fix #181

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,25 @@
+---
+name: Bug report
+about: Report a bug in one or more tools
+title: ''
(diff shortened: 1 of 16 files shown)


ThinkOpenly:
Not to be a pain, but my preference is for commit messages to have subject lines that reflect what is being changed, rather than "Fixes #181". So, something like "Move mandatory privilege mode option from CRD family to CRD" (if I understand the changes correctly).

It makes the `git` history much more helpful, especially `one-line` format.

james-ball-qualcomm:
Premature PR

Issue #177, opened 2024-10-23
MC-1 misa CSR is shown with a bunch of green bits for extensions out of scope


james-ball-qualcomm:
Why are fields like V, U, S, H, G, F, D, B, A shown as green when out of scope? 

![image](https://github.com/user-attachments/assets/983f4a7a-5f04-49cc-8807-9c6dcaa102c1)

dhower-qc:
fixed in #183

PR #183, opened 2024-10-24
Make misa extension bits definedBy their extension so they display correctly in docs. Fix optional calculation

dhower-qc:


dhower-qc:
closes #177

PR #188, opened 2024-10-25
Crypto Extensions

AFOliveira:
In **B** extension, although instructions already have the   `data_independent_timing: true
`, it also make senses to describe the crypto extensions, since they technically also include those instructions, as described in the ISA Manual : 
image

Moreover, I added them in conformity with B extension format [Extension, Subextension, Subextension], e.g. 

```
packw:
...
  definedBy:
    anyOf: [Zbkb, Zk, Zkn, Zks]
```


dhower-qc (approved):
Looks good!

PR #191, opened 2024-10-25
Make support for menvcfg.CBIE == 01 optional

### LINKED ISSUE #169: envcfg.CBIE optional WARL values
https://github.com/riscv-software-src/riscv-unified-db/blob/8b730714117e47fbff73b3dec7f37c26e9ac72ec/arch/csr/menvcfg.yaml#L249

Implementations may upgrade `CBO.INVAL` to `FLUSH`, to prevent un-writing cached dirty data, and thus may only accept writing `2'b00` or `2'b01` to the `xenvcfg.CBIE` field.  This should be an option.

dhower-qc:
closes #169


arch/csr/menvcfg.yaml
```diff
@@ -246,7 +246,9 @@ menvcfg:
       definedBy: Zicbom
       type: RW-R
       sw_write(csr_value): |
-        if (csr_value.CBIE == 0 || csr_value.CBIE == 1 || csr_value.CBIE == 3) {
+        if ((csr_value.CBIE == 0) ||
+            (ALLOW_CBO_INVAL_UPGRADE_TO_FLUSH && (csr_value.CBIE == 1)) ||
+            (csr_value.CBIE == 3)) {
```
ingallsj, line 251: other way around: 3 is the optional value

dhower-qc: ahh, need more coffee today...

PR #201, opened 2024-10-25
small tool to index instruction yaml files

drom:
Added a small PoC JS tool to index instruction yaml files.

```
./index-unifieddb.js . > index.json
```


dhower-qc:
Do you just need an index of the instructions? What about extensions/CSRs/profiles/...?

drom:
good point. fixing

drom:
what do I do with non-yaml files?

drom:
should be reading all yaml, json files now

james-ball-qualcomm (approved):

drom:
Any other work need to be done for merging this PR

dhower-qc:
How should this be used? A few options:

1. We can have it generate the index as a file under Pages, so you would access it (for the tip of main) at, e.g., https://riscv-software-src.github.io/riscv-unified-db/artifacts/index.json
1. We could add the generated index to the repository, along with a pre-commit hook to ensure it stays consistent.

I think 1. will be easier to manage (we don't have any required hooks yet), but has the obvious downside that you don't keep an index around with each commit. 

Thoughts?

dhower-qc (commented):
This file should be under the `backends/` folder. I'd be ok adding it under `backends/arch_gen`, since that backend already creates something similar. A new subfolder would also be fine.

dhower-qc:
> How should this be used? A few options:
> 
> 1. We can have it generate the index as a file under Pages, so you would access it (for the tip of main) at, e.g., https://riscv-software-src.github.io/riscv-unified-db/artifacts/index.json
> 2. We could add the generated index to the repository, along with a pre-commit hook to ensure it stays consistent.
> 
> I think 1. will be easier to manage (we don't have any required hooks yet), but has the obvious downside that you don't keep an index around with each commit.
> 
> Thoughts?

Actually, now that I look closer at the file, it seems like you don't even need the output saved. Is that right?

drom:
Currently I am fetching `yaml` files from `https://raw.githubusercontent.com/riscv-software-src/riscv-unified-db/refs/heads/main/` path. It would be convenient having `index.json` some where under the same hierarchy.
That would mean generating and committing it with every push to git. Maybe git hook.
Yes, please move script anywhere withing repo.

dhower-qc:
I submitted a pull request on your fork. If you think that looks good, update this PR and we'll get it merged

drom:
yes, it is good to merge

PR #204, opened 2024-10-29
Fix broken links

trdthg:
```txt
 * xref:_site/ruby/arch_def/index.html[Ruby database object model documentation]
 * xref:_site/ruby/idl/index.html[IDL Compiler documentation]
```

These two links are duplicates(which are also broken) of the top links, so I guess we can remove them?

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -28,7 +28,7 @@ This repository contains:
 
 === Working examples:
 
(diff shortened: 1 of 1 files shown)


dhower-qc (approved):
Thanks!

PR #228, opened 2024-10-31
Add $mref to inst variables

dhower-qc:
Also enhances arch_gen to expand references.

Supports $ref (JSON Reference) and $mref (merged ref, doesn't ignore other keys in the object)

Difference:
```

(in something.yaml)
key:
  ref_extra: Hello!

obj:
  $ref: path/to/something.yaml#/key
  extra: Hi!

=> obj = { ref_extra: "Hello!" }

obj:
  $mref: path/to/something.yaml#/key
  extra: Hi!

=> obj = { extra: "Hi!", ref_extra: "Hello!" }
```


drom:
`$ref` is immutable reference. `$mref` will create new object similar to the original.
We need to clearly delineate use cases for `$ref` and `$mref`

dhower-qc:
Good point. The way I see it, `$ref` is a pointer, and `$mref` is inheritance. For example, @james-ball-qualcomm is using `$mref` to represent profiles that build on each other (RVA20 -> RVA22).

dhower-qc:
Moving to discussion #247

Issue #237, opened 2024-11-02
Create separate profile PDF for each profile family release (e.g., RVA23) not one PDF for all profiles in a family (e.g., RVA)


james-ball-qualcomm:
Appendices (e.g., instructions, CSRs) are a function of the profile family release each release needs to have its own PDF.

mipsrobert:
As both RVA23 and RVB23 are recently ratified, should not database be updated to reflect ratified specs?
I only see RVA, but not RVB.

james-ball-qualcomm:
Yes, that is the plan.

james-ball-qualcomm:
PR #249 fixed this.

Issue #58, opened 2024-10-02
Reduce duplicate code between Profiles and CRDs


james-ball-qualcomm:
There is a lot of duplication (copy and paste) between the Profiles and CRDs, both in Ruby classes and in the template files. Can we re-factor these to have a common parent class that Profiles and CRDs inherit from and then put duplicate template code into an include file that is included by both Profiles and CRDs?

james-ball-qualcomm:
Also found duplicate code in the profiles and CRD tasks.rake files.

james-ball-qualcomm:
Using "Portfolio" to described a named & versioned grouping of extensions (each with a name and version).

PR #242, opened 2024-11-02
238 missing recommendations in profiles and add extra notes to replace excluded extensions

james-ball-qualcomm:
Passes regression on local machine

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 diag-ditaa-*
 arch/manual/isa/**/riscv-isa-manual
 gen
+gen_expected
 node_modules
 _site
 images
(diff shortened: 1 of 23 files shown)


james-ball-qualcomm:
Passes. Also has commits from the fix for issue #58.

.gitignore
```diff
@@ -10,6 +10,7 @@
 diag-ditaa-*
 arch/manual/isa/**/riscv-isa-manual
 gen
+gen_expected
```
dhower-qc, line 13: What's `gen_expected` for? I don't see any other references to it.

backends/portfolio_doc/tasks.rb
```diff
@@ -0,0 +1,122 @@
+# frozen_string_literal: true
```
dhower-qc, line 1: What's this Rake file for? I think most of it is already present in `backends/crd_doc/tasks.rake`.

dhower-qc (changes requested):

backends/portfolio_doc/tasks.rb
james-ball-qualcomm: It was a work in progress. I've actually deleted it in my latest workspace. So, can we just ignore this for now and approve this PR since I'll be submitting a new PR for #237 soon anyways (which removes this file).

dhower-qc (approved):

PR #259, opened 2024-11-12
fix typos in schema files and a few adocs

kbroch-rivosinc:


schemas/manual_version_schema.json
```diff
@@ -23,7 +23,7 @@
         },
         "authors": {
           "type": "array",
-          "itesm": {
+          "items": {
```
dhower-qc, line 26: Since no good deed goes unpunished, this fix actually uncovered a bug in the data files. You can see the output in the [regression test](https://github.com/riscv-software-src/riscv-unified-db/actions/runs/11807330536/job/32934689932?pr=259).

In `arch/manual/isa/20240411`:

```
    At /volumes/1/authors/0/organization, Expecting object, got SiFive, Inc.
    At /volumes/1/authors/0/url, endpoint is an invalid key
```

The current authors of the second Volume is (line 174):

```yaml
  authors:
  - name: Andrew Waterman
    email: awaterman@sifive.com
    organization: SiFive, Inc.
    url: https://www.sifive.com/
```

but it should be

```yaml
authors:
  - name: Andrew Waterman
    email: awaterman@sifive.com
    organization:
      name: SiFive, Inc.
      url: https://www.sifive.com/
```

Could you please add this fix to the PR so the regression passes? Thanks!

dhower-qc (approved):

PR #270, opened 2024-11-14
Add python env executables, pre-commit, and pyyaml

dhower-qc:


AFOliveira (approved):
I did a dummy python script on the rakefile

```
desc "Run Python with YAML import"
task :run_python do
  puts "Executing Python code..."
  python_code = <<~PYTHON
    import yaml
    import sys
    print("Python executing...", flush=True)
  PYTHON
  
  root_dir = File.expand_path('.')
  venv_python = File.join(root_dir, '.home', '.venv', 'bin', 'python3')
  
  puts "Using Python at: #{venv_python}"
  result = system venv_python, "-c", python_code
  if !result
    puts "Python command failed with exit code: #{$?.exitstatus}"
  end
end

```

It's working just fine! Thanks!

```
Executing Python code...
Using Python at: /home/afonsoo/derek/riscv-unified-db/.home/.venv/bin/python3
Python executing...

```

AFOliveira:
@dhower-qc can you merge this(or I also may if you are ok with it :)) so I can rebase #258 and finalize it?

PR #272, opened 2024-11-14
176 missing description for mip csr

### LINKED ISSUE #176: Missing description for mip CSR
image

james-ball-qualcomm:
Created $copy to copy string values like $ref. Added unit-level test. Fixed mie to be defined by Sm not I extension. Description for mip and mie is now shared mimicing ISA manual.

Fixes #176


arch/csr/mie.yaml
```diff
@@ -5,147 +5,8 @@ mie:
...
-    CSR instruction with a five-bit immediate.
-    ====
-
-    Restricted views of the `mip` and `mie` registers appear as the `sip`
-    and `sie` registers for supervisor level. If an interrupt is delegated
-    to S-mode by setting a bit in the `mideleg` register, it becomes visible
-    in the `sip` register and is maskable using the `sie` register.
-    Otherwise, the corresponding bits in `sip` and `sie` are read-only zero.
+  definedBy: Sm
+  description: "$copy: mip.yaml#/mip/description"
```
dhower-qc, line 9: why doesn't `$ref` work here? I think the intent is that it should, like so:

```yaml
description:
  $ref: "mip.yaml#/mip/description"
```

Further, we don't want $copy to be hidden in a string. Assuming we need to keep `$copy` for some reason, it should look like:

```yaml
description:
  $copy: "mip.yaml#/mip/description"
```

dhower-qc (changes requested):

arch/csr/mie.yaml
james-ball-qualcomm: I would have used $ref if it worked (it doesn't). That's because changing
   description: "foo"
to
   description:
      some-key: "foo"
would be changing "description" from a string to a hash and that would break everything using it (which it does).

james-ball-qualcomm:
I've merged in the changes made to the schema into this branch so it can be merged into main if you are happy with it.

arch/csr/mie.yaml
dhower-qc: we really don't want "$copy: blah" in a string, so at the very least that needs to change. The open question is whether or not we can just use $ref here; to me, it seems like we should.

dhower-qc:
As we discussed on the side, we need to get `$copy` as a key rather than hidden in a string.

james-ball-qualcomm:
I think this is ready to be merged into main after you review it.

dhower-qc (approved):

PR #275, opened 2024-11-17
Move definition of softfloat_roundPackToF32 before it is called.

ghost:
According to https://github.com/riscv-software-src/riscv-unified-db/blob/c55022788590a4601fba3957c2870ca31e0e4157/arch/prose/idl.adoc#compilation-order,

"The order of declaration is important in IDL (like C/C++, unlike Verilog). A variable, constant, or function must be declared before it is used."


dhower-qc:
Thanks for the detailed look! I need to update the IDL documentation -- the sentence you highlight:

> The order of declaration is important in IDL (like C/C++, unlike Verilog). A variable, constant, or function must be declared before it is used.

Is actually not true anymore. I've updated the compiler to load everything in a first pass before going into function bodies, so the syntactic order of functions is no longer important.

Let's keep this PR open, and I'll push the documentation changes in.

PR #301, opened 2024-11-22
Draft for PseudoInstructions

AFOliveira:
I have this format being automatically generated for all pseudo-instructions. Does this reflect the discussion we had on last week's meeting?

The corner case of conditional pseudo-instructions is not automatable.

### DIFF
--- a/arch/inst/I/pseudo/fence.tso.yaml
+++ b/arch/inst/I/pseudo/fence.tso.yaml
@@ -1,0 +1,20 @@
+# yaml-language-server: $schema=../../../schemas/inst_schema.json
+
+$schema: inst_schema.json#
+kind: pseudoinstruction
+name: fence.tso
+long_name: No synopsis available.
+description: |
+      No description available.
+definedBy: I
+assembly: xs1, xd
+origin_instruction: fence
+encoding:
+  match: 100000110011-----000-----0001111
+  variables:
+  - name: rs1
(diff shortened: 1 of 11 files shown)


AFOliveira:
> * `origin_instruction(s?)` should be an array so we can support the multi-inst sequences

Idk about how the backend would be easier, so just asking, but if we define them as you did couldn't we just add another entry? e.g.

```
origin_instructions:
- parent: { $ref: inst/I/add.yaml# }
  encoding:
    variables:
      rs1:
        const: 0
      rs2:
        const: 5
      rd:
        const: 0
- parent: { $ref: inst/I/other.yaml# }
  encoding:
    variables:
      rs1:
        const: 0
      rs2:
        const: 5
      rd:
        const: 0
``` 

> Let's make the instruction reference a $ref (I think we are moving that direction universally in the db).

Ok.


> Instead of capturing encoding/match, could we specify variable values of the parent instruction? I think it captures the same information but might be more informative to a reader. (not sold on this, just thinking out loud). 

I can do this, since it was already being done but in a different way for previous pseudoinstructions definition.

dhower-qc:
> Idk about how the backend would be easier, so just asking, but if we define them as you did couldn't we just add another entry? e.g.

Yep!

dhower-qc:
Closing for now, re-open if it becomes active

Issue #299, opened 2024-11-21
Establish versioning scheme


dhower-qc:
Right now, UDB is assuming that all version numbers follow SemVer. Unfortunately, this is not how RVI versions extensions.

Here is a proposal:

- By default, versions are always backward-compatible. That is, given versions A and B, if B > A, then B is compatible with A (but not vice versa).
- We will establish a mechanism to mark a breaking change in the version list of an extension. If a breaking change is noted, then that version and all higher-numbered versions are incompatible with all lower-numbered versions.

When specifying version requirements, we will can use the `~> A` operator to mean "any version >= A that is not incompatible with A". For example:

```
X ~> 1.2  # With no breaking change, will match 1.2, 1.3, 2.0, 3.1, ...   It will not match 1.0, 1.1
Y ~> 1.1  # Assume that Y version 1.4 is breaking. This will match 1.1, 1.1.1, 1.2, 1.3. It will not match 1.4, 2.0, ...
```

PR #302, opened 2024-11-22
Make ExtensionVersion a proper object in Ruby

dhower-qc:
Previously, we were dealing directly with the YAML


james-ball-qualcomm (approved):

james-ball-qualcomm:
Does this fix #299? Can we now close that issue?

dhower-qc:
No, this just makes the existing ExtensionVersion class more robust. It doesn't change how anything is interpreted.

Issue #314, opened 2024-11-26
Vector Load/Store Segment Instruction mnemonics are missing


BrianAnakPintar:
I was recently testing the results by converting the instructions from UDB to riscv-opcodes' format (#300) and one issue I noticed is that UDB seem to be missing several instructions. I am unsure whether there is a way to obtain these from existing instructions in the UDB or if we simply don't have a way to obtain them:

On the left is from UDB and on the right is riscv-opcodes'
<img width="1219" alt="Screenshot 2024-11-25 at 7 55 42 PM" src="https://github.com/user-attachments/assets/c6b7bd2e-d8bc-4508-9162-31c11e841cde">

There's a handful of these missing instructions, here's a few of them:
```json
"vloxseg8ei64_v": {
        "encoding": "111011-----------111-----0000111",
        "variable_fields": [
            "rs1",
            "vd",
            "vm",
            "vs2"
        ],
        "extension": [
            "rv_v"
        ],
        "match": "0xec007007",
        "mask": "0xfc00707f"
    },
"vsoxseg2ei16_v": {
        "encoding": "001011-----------101-----0100111",
        "variable_fields": [
            "rs1",
            "vm",
            "vs2",
            "vs3"
        ],
        "extension": [
            "rv_v"
        ],
        "match": "0x2c005027",
        "mask": "0xfc00707f"
    },
"vssseg8e8_v": {
        "encoding": "111010-----------000-----0100111",
        "variable_fields": [
            "rs1",
            "rs2",
            "vm",
            "vs3"
        ],
        "extension": [
            "rv_v"
        ],
        "match": "0xe8000027",
        "mask": "0xfc00707f"
    },
```

AFOliveira:
First, thank you so much for the verification.
I was looking to understand issues with the V extension (riscv/riscv-opcodes/issues/322) and your issue helped me to understand this, so thank you @BrianAnakPintar! 
What I now understand is that riscv-opcodes generates this instructions rather then describing them all individually, thus I was not able to fetch them.

Proposed solutions I see for this:

1. Create a back-end to expand instructions just like riscv-opcodes does.
2. Describe them all independently such as @BrianAnakPintar proposed in the beggining of the issue.

I am more inclined towards the 2, since I think it better serves the purpose of having a DB of instructions, but probably the argument to defend 1 is to avoid almost duplicated instructions since they only slightly differ from each other. 

I think doing 2 would be  relatively easy (probably doable until the end of the week).

[...]

Issue #300, opened 2024-11-22
Riscv-opcodes replacement


AFOliveira:
**Is your feature request related to a problem? Please describe.**
We have achieved riscv-opcodes source of information in terms of opcodes and variables, but we don't have it's outputs, which are the most valuable.

**Describe the solution you'd like**
We could create a way of generating the outputs that riscv-opcodes uses. From my experince the encoding.c is used for SPIKE and the .go is used by the Golang language.

**Describe alternatives you've considered**
I know there is someone assigned to generating the riscv-opcodes DB, but if we could produce it's outputs, we could propose to replace it entirely and in a short timespan

AFOliveira:
cc @dhower-qc @ThinkOpenly

ThinkOpenly:
@BrianAnakPintar, did you want to talk about what you are working on here?

BrianAnakPintar:
So there are several outputs generated by riscv-opcodes. Currently I have made a script which reads the instructions yaml files from unified db and produced one of the riscv-opcodes' output. I'm currently verifying to make sure the output is the same as riscv-opcodes's outputs.

Initially we tried recreating the original riscv opcodes DB, but soon found out that most of the data doesn't particularly follow a specific format strictly and could contain a lot of "edge cases". So we tried changing our focus into generating the output files instead which seems to be an easier task as they follow a specific format.

In terms of the outputs here's what is generated by riscv-opcodes.
- [x] instr_dict.json (currently verifying its output)
- [ ] encoding.out.h
- [ ] inst.go
- [ ] instr-table.tex
- [ ] priv-instr-table.tex
- [ ] inst.chisel
- [ ] inst.sverilog 
- [ ] inst.rs
- [ ] inst.spinalhdl

Once I'm done verifying the `json` output, I'm planning to see if we can generate `encoding.out.h` and the `inst.go` from the unified db.

dhower-qc:
IIRC, some (all?) of those files are created from a script in riscv-opcodes using instr_dict.json as input. Do I have that right? And, if so, can we just run that script from the UDB-generated instr_dict.json?

AFOliveira:
[...]

PR #316, opened 2024-11-26
Add V extension missing isntructions

AFOliveira:
Solves #314

PR #21, opened 2024-08-22
Added parse.py, constants.py and Makefile modifications to generate yaml files. 

AFOliveira:
I believe that this type of file organization is a good workaround to not having everything in the upstream repo yet.
This version still has some bugs namely on pseudoinstructions and on some assembly code. I will be working on those and I'll submit a new PR when they're solved.

PR #328, opened 2024-11-29
[Draft] Script to generate instr_dict.json from riscv-opcodes using UDB data

BrianAnakPintar:
This is a draft PR for #300 

I included a `instr_dict.json` which I generated from riscv-opcodes which has data as of Nov 28, 2024

There's 2 main scripts involved here.
1. `generate_instr_dict.py`: This script generates a file called `data.json` which is equivalent of the `instr_dict.json` from riscv-opcodes but using the data from UDB.
2. `sorter.py`: This script generates 2 files which sorts `data.json` and `instr_dict.json` to make it easier to diff the 2 files.

**Known issues:**
I use this [site](https://www.jstoolset.com/diff) to diff the 2 files and here are some immediate issues
1. Sometimes there are more extensions in UDB than there are in riscv-opcodes. For example the instruction `add.uw`. Iirc, this is because 1 extension encapsulates the other extension.
2. Missing instructions #314 
3. Unknown ways to get `variable_fields`. I'm unsure on how to handle some cases like `imm` where it can range from either `immxx` where `xx` is just the size of the range but also deal with cases where `imm` turns into `bimm12hi` and `bimm12lo` for instructions like `beq`. There are more cases of these.
4. `rori` contains incorrect encoding. I'll open a separate pull request for this.

These are currently the roadblocks that I am facing and I can't seem to resolve them. Feel free to give feedback or modify the script if necessary. 

There's also some redundant functions such as `find_first_match` which I previously used but no longer require, but I kept it for now in case anyone finds it useful. I also tried adding types to most of these functions to make it easier to follow through.

### DIFF
--- a/opcodes_maker/generate_instr_dict.py
+++ b/opcodes_maker/generate_instr_dict.py
@@ -1,0 +1,148 @@
+from typing import List, Dict
+import os
+import yaml
+import json
+
+
+def range_size(range_str: str) -> int:
+    try:
+        end, start = map(int, range_str.split('-'))
+        return abs(end - start) + 1
+    except ValueError:
+        return 0;
+
+reg_names = {'qs1', 'qs2', 'qd',
+             'fs1', 'fs2', 'fd'}
+
+def GetVariables(vars: List[Dict[str, str]]):
+    var_names = []
+    for var in vars:
+        var_name = var['name']
+        if var_name in reg_names:
+            # Since strings are immutable.
+            lst_var_name = list(var_name)
+            lst_var_name[0] = 'r'
+            var_name = "".join(lst_var_name)
+        elif var_name == "shamt":
+            size = range_size(var['location'])
+            if size == 5:
+                var_name = "shamtw"
+            elif size == 6:
+                var_name = "shamtd"
+        var_names.append(var_name)
+    var_names.reverse()
+
+    return var_names
+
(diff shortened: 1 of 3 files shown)


AFOliveira:
I think this a great step towards making the UDB an even better source of information!

Answering you points:

1. I think this is supposed to happen, maybe if we are trying to replicate exactly what riscv-opcodes generates, we can simply exclude it on the script - IDK if there are any advantages in doing so.
2. This is supposed to be fixed as soon as #316 is merged.
3. Variable description is still a WiP and therefore all this names might change soon. However, if you want to do it for current version: riscv-opcodes has a mapping in which it does the translation of those fields into more readable (and parseable) names - https://github.com/riscv/riscv-opcodes/blob/9226b0d091b0d2ea9ccad6f7f8ca1283a3b15e88/constants.py#L88C1-L148C1. Moreover, you can either check https://github.com/drom/riscv/blob/master/lib/fieldo.js for more info or my own draft PR in this repo https://github.com/riscv-software-src/riscv-unified-db/pull/21 to see how I did the mapping from riscv-opcodes to current format- might help in encrypting back to riscv-opcodes.  
4. This is a great catch. Thanks!

On another note, when this turns into a non-draft PR, I don't think that .json file should be merged in.

As I previously said, this is truly a great step towards the right direction, thank you!

dhower-qc:
closing from inactivity

Issue #322, opened 2024-11-27
"anyOf" is used where "allOf" should be


ThinkOpenly:
For some instructions, for the set of extensions in which they are "definedBy", there are instances where "anyOf" is used where I believe "allOf" is the correct condition. For example, for `fli.h`:
```
definedBy:
  anyOf: [Zfa, Zfh]
```
However, the unpriv spec (20240411) defines all of the "FLI" instructions in the "Zfa" extension, but for "FLI.H":
> If the Zfh or Zvfh extension is implemented, FLI.H performs the analogous operation [...]

So, "FLI.H" is only implemented if both "Zfa" _and_ "Zfh" are supported.

I think all of these should be "allOf" instead of "anyOf":
```
./arch/inst/Zfh/fli.h.yaml:  anyOf: [Zfa, Zfh]
./arch/inst/Zfh/fcvt.h.s.yaml:  anyOf: [Zfh, Zfhmin]
./arch/inst/Zfh/fmv.x.h.yaml:  anyOf: [Zfh, Zfhmin]
./arch/inst/Zfh/flh.yaml:  anyOf: [Zfh, Zfhmin]
./arch/inst/Zfh/fcvt.h.d.yaml:  anyOf: [D, Zfh]
./arch/inst/Zfh/fmaxm.h.yaml:  anyOf: [Zfa, Zfh]
./arch/inst/Zfh/fleq.h.yaml:  anyOf: [Zfa, Zfh]
./arch/inst/Zfh/fminm.h.yaml:  anyOf: [Zfa, Zfh]
./arch/inst/Zfh/froundnx.h.yaml:  anyOf: [Zfa, Zfh]
./arch/inst/Zfh/fltq.h.yaml:  anyOf: [Zfa, Zfh]
./arch/inst/Zfh/fsh.yaml:  anyOf: [Zfh, Zfhmin]
./arch/inst/Zfh/fcvt.s.h.yaml:  anyOf: [Zfh, Zfhmin]
./arch/inst/Zfh/fround.h.yaml:  anyOf: [Zfa, Zfh]
./arch/inst/Zfh/fcvt.d.h.yaml:  anyOf: [D, Zfh]
./arch/inst/D/fmvh.x.d.yaml:  anyOf: [D, Zfa]
./arch/inst/D/fcvtmod.w.d.yaml:  anyOf: [D, Zfa]
./arch/inst/D/fli.d.yaml:  anyOf: [D, Zfa]
[...]

Issue #319, opened 2024-11-27
Incorrect syntax for some instructions' "definedBy" extensions


ThinkOpenly:
Some instructions' "definedBy" fields appear to be incorrect, using an open comma-separated list, rather than a subfield refinement with "allOf".
```
./arch/inst/F/fmaxm.s.yaml:definedBy: F, Zfa
./arch/inst/F/froundnx.s.yaml:definedBy: F, Zfa
./arch/inst/F/fround.s.yaml:definedBy: F, Zfa
./arch/inst/F/fltq.s.yaml:definedBy: F, Zfa
./arch/inst/F/fleq.s.yaml:definedBy: F, Zfa
./arch/inst/F/fli.s.yaml:definedBy: F, Zfa
./arch/inst/F/fminm.s.yaml:definedBy: F, Zfa
```

I think each of these should be expanded to:
```
definedBy:
  allOf:
  - F
  - Zfa
```

Although, since Unpriv spec Version 20240411 says:
> The Zfa extension depends on the F extension

maybe only the Zfa extension needs to be called out as the "definedBy" extension?

AFOliveira:
This is indeed true and those instructions are missing on the output manual. I'll fix this ASAP. @dhower-qc Do you think  it is worth it to add a test on this to the smoke tests?

dhower-qc:
> Do you think it is worth it to add a test on this to the smoke tests?

definitely!!

PR #340, opened 2024-12-04
Fix definedBy on some instructions

AFOliveira:
As caught and reported by @ThinkOpenly. Solves #319 . Solves #322 .


AFOliveira:
@dhower-qc is this also repeated from what you did on #350? If so I guess we can just close it.

dhower-qc:
Fixed in #350

AFOliveira:
#362 Explains what happened here, I'll fix this PR.

arch/inst/Zfh/fleq.h.yaml
```diff
@@ -7,7 +7,7 @@ long_name: No synopsis available.
 description: |
   No description available.
 definedBy:
-  anyOf: [Zfa, Zfh]
```
dhower-qc, line 10: Shouldn't this remain anyOf?

arch/inst/Zfh/flh.yaml
```diff
@@ -12,7 +12,7 @@ description: |
   `flh` is only guaranteed to execute atomically if the effective address is naturally aligned.
 
 definedBy:
-  anyOf: [Zfh, Zfhmin]
+  allOf: [Zfh, Zfhmin]
```
dhower-qc, line 15: And this one

arch/inst/Zfh/fmv.x.h.yaml
```diff
@@ -5,7 +5,7 @@ kind: instruction
 name: fmv.x.h
 long_name: Move half-precision value from floating-point to integer register
 definedBy:
-  anyOf: [Zfh, Zfhmin]
+  allOf: [Zfh, Zfhmin]
```
dhower-qc, line 8: And this one

arch/inst/Zfh/fsh.yaml
```diff
@@ -15,7 +15,7 @@ description: |
   `fsh` is only guaranteed to execute atomically if the effective address is naturally aligned.
 
 definedBy:
-  anyOf: [Zfh, Zfhmin]
+  allOf: [Zfh, Zfhmin]
```
dhower-qc, line 18: And this one

dhower-qc (changes requested):

arch/inst/Zfh/fleq.h.yaml
ThinkOpenly: This should actually be "allOf", no?

dhower-qc: yes, my bad. I was trying to mark all the [Zfh, Zfhmin] combos. I must have fat fingers...

arch/inst/Zfh/flh.yaml
ThinkOpenly: There is some weirdness with `flh` and the "Zhinx" extension, but I guess we don't include that extension yet.

arch/inst/Zfh/fmv.x.h.yaml
ThinkOpenly: Same "Zhinx" weirdness here.

arch/inst/Zfh/fsh.yaml
ThinkOpenly: Same "Zhinx" weirdness here.

arch/inst/Zfh/flh.yaml
AFOliveira: should I then keep them or leave them?
I actually couldn't find specific information on this is the ISA and riscv-opcodes doesn't even define zfhmin.

dhower-qc: We should keep Zfhmin, but it should be `anyof: [Zfh, Zfhmin]`, since either extension will define `flh` (doesn't require *both* `Zfh` and `Zfhmin`)

Re: @ThinkOpenly's comment on `Zhinx`: Good point. I guess this really should be "anyOf: [Zfh, Zfhmin, Zhinx]", with the `operation()`/`sail()` choosing which version to implement.

AFOliveira: Added all 3.

AFOliveira:
There is an error with missing Zhinx extension:
Should I 
- add Zhinx as an extension to the UDB under the arch/ext folder?
- drop the Zhinx from this PR instructions?

dhower-qc:
Go ahead and add Zhinx, with a new issue to remind us that Zhinx is not implemented (since a simple look for empty operation() won't catch it)

AFOliveira:
@dhower-qc Done!

arch/ext/Zhinx.yaml
```diff
@@ -0,0 +1,22 @@
...
+long_name: Half-precision floating-point instructions using integer registers
+description: |
+  The Zhinx extension provides half-precision floating-point instructions analogous
+  to those in the Zfh extension, except that these instructions use the integer
+  register file rather than the floating-point register file. Specifically, Zhinx
+  adds all the Zfh-extension instructions except the transfer instructions
+  (FLH, FSH, FMV.H.X, and FMV.X.H). The semantics of each Zhinx instruction are
+  identical to the corresponding Zfh instruction, except that occurrences of
+  `f` registers in the original Zfh instructions are replaced by the corresponding
+  `x` registers in Zhinx.
```
ThinkOpenly, line 15: In the commit that adds this content, could you describe where the content originated?
Also, remove the "Signed-off-by" line.

AFOliveira: Done

dhower-qc (approved):

arch/ext/Zhinx.yaml
AFOliveira: Since this is more of global to the repository than to my PR I'll still merge to main and in the future we can sort it with the rest of information of the same kind.

PR #341, opened 2024-12-04
332 create better profile comparisons

james-ball-qualcomm:
Moved comparisons to appendix and have comparisons for releases and profiles.


lib/arch_obj_models/extension.rb
```diff
@@ -631,6 +637,45 @@ def <=>(other)
...
+    (self.optional? && other.mandatory?)
+  end
+
+  # @overload <=(other)
+  #   @param other [ExtensionPresence] An extension presence object
+  #   @return [Boolean] Whether or not this ExtensionPresence is less-than or equal to the other
+  def <=(other)
+    raise ArgumentError, "ExtensionPresence is only comparable to other ExtensionPresence classes" unless other.is_a?(ExtensionPresence)
+    (self < other) || (self == other)
+  end
```
dhower-qc, line 678: FYI, you don't need to implement these comparison functions since you have the `<=>` operator. If you just `include Comparable` in the class, these will be generated on the fly.

dhower-qc (approved):

lib/arch_obj_models/extension.rb
james-ball-qualcomm: Yes, I know. However <=> sorts alphabetically by name and then optional type if present. The other comparisons ignore optional type because no optional type has greater presence than any other. Also, they explicitly make mandatory higher presence than optional and don't rely on their alphabetical order (m happens to come before o).

james-ball-qualcomm: Removed <=> based on discussion with Derek. Now looking to see where I get the wrong sort result.

dhower-qc (approved):

james-ball-qualcomm:
Yay!

PR #348, opened 2024-12-09
Add definedBy validation

AFOliveira:
As mentioned in #319. Please let me know if how I structured it makes sense. I put it under the arch test since I believe it kind of belongs to it, but it maybe could be done separately

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -90,15 +90,37 @@ namespace :test do
   end
   task schema: "gen:arch" do
     validator = Validator.instance
-    puts "Checking arch files against schema.."
+    puts "Checking arch files against schema..."
     arch_files = Dir.glob("#{$root}/arch/**/*.yaml")
     progressbar = ProgressBar.create(total: arch_files.size)
+    
+    incorrect_defined_by_files = []
+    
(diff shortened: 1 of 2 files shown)


dhower-qc:
This is something that *should* be handled by JSON Schema, but it's not...

If you follow the schema, eventually you see that the string should match an "extension_name":

```
    "extension_name": {
      "type": "string",
      "pattern": "^([A-WY])|([SXZ][a-z0-9]+)$"
    },
```

The pattern is supposed to exclude a comma. But it's actually accepting *any* string; see https://stackoverflow.com/a/51959769. The whole thing needs to be wrapped in parens:

```
      "pattern": "^(([A-WY])|([SXZ][a-z0-9]+))$"
```

With that change, I think it solves the problem.

AFOliveira:
@dhower-qc  Alright! And you've done that in #350, right? If so, I'll just close this

dhower-qc:
Actually, I haven't. I fixed the source files, but didn't notice the bug in the schema until now.

dhower-qc:
I just added it to #350, so we can close this now. Thanks!

PR #350, opened 2024-12-10
Big, Ugly Patch

### LINKED ISSUE #299: Establish versioning scheme

### LINKED ISSUE #317: Can "profiles" get a "kind" field, too?
I wrote a short Python script to read in all of the YAML files in the project for the purposes of verifying content, like what extensions are referenced, but never defined. The specific YAML files that reference extensions are (at least) instructions, extensions, CSRs, and profiles. The first three have a "kind" field identifying what each YAML represents, but the latter, "profiles", does not. Current "kind" fields are:
```
$ find . -name '*.yaml' -exec grep 'kind:' {} \; | sort -u
kind: certificate class
kind: certificate model
kind: csr
kind: extension
kind: instruction
```

I'd like to see the YAML files describing profiles also have a kind field like:
```
kind: profile
```

It's actually difficult to work around, because the top-level field is a profile name, so not a consistent key that can be checked for existence.

### LINKED ISSUE #319: Incorrect syntax for some instructions' "definedBy" extensions

### LINKED ISSUE #320: Add missing extensions
I wrote a script that looked for references to extensions for which there was no corresponding extension definition. The following extensions are referenced, but have no definition:

- Q
- Smrnmi
- Zalasr
- Zfa
- Zfbfmin
- Zfh
- Zicond
- Zimop
- Zvfbfmin
- Zvfbfwma
- Zvkg

The script also found these, but there are issues open for adding them already:
- Smdbltrp
- Zabha
- Zacas
- Zawrs
- Sdext

(This script could be a good thing to add to CI, I suppose.)

### LINKED ISSUE #336: UDB doesn't generate an error if a profile or CRD specifies a version of an extension not present in the database
I just noticed that RVA20 has the S extension listed as 1.12 when its yaml file specifies ~> 1.11. Then I looked at the S extension and see it only has version 1.12 and is missing version 1.11. So, it seems like UDB is just using 1.12 since the RVA20 profile yaml says ~> 1.11. However, then the wrong version is reported in the RVA20 PDF. 

I suggest that UDB create an error when a profile or CRD specifies a version number not in the database (especially if it requests a lower version number).

We also need to add version 1.11 to the S extension which hopefully we could include in this GitHub issue so we have it available for testing.

dhower-qc:
A Big, Ugly Patch

 * Performs YAML resolution in Python, using lib/yaml_resolver.py
 * Adds many missing extensions
 * Changes Profile schema to match other objects
 * Implements versioning via lib/version.rb instead of Gem::Version/Requirement
 * Refactors ArchDef into Specification/Config
 * Simplifies ArchDef/Specification by exploiting common format for object schemas
 * New, simplified schema for configs
 * Adds schemas for profile (some are still placeholders)
 * Parallelizes GitHub regression test

fixes #336
fixes #320
fixes #319
fixes #317
fixes #299


james-ball-qualcomm:
Don't see how to add additional comments to my review so I'll just add them here. 

Why did the Profiles get put into their own files instead of inside a ProfileRelease?

james-ball-qualcomm:
Recommend removing any knowledge of Specification and/or ArchDef classes from Config. Derek proposed removing the Config methods that use ArchDef.

james-ball-qualcomm:
What happened to the initializers for Certificate classes? Are those now auto-created by some Ruby magic? If so, can you add comments to the classes that use this since readability is impacted.

lib/arch_obj_models/portfolio.rb
```diff
@@ -184,23 +181,30 @@ def uses_optional_types?
...
-    arch_def_data = arch_def.unconfigured_data
-
-    arch_def_data["mandatory_extensions"] = mandatory_ext_reqs.map do |ext_req|
-      {
-        "name" => ext_req.name,
-        "version" => ext_req.version_requirement.requirements.map { |r| "#{r[0]} #{r[1]}" }
-      }
-    end
-    arch_def_data["params"] = all_in_scope_ext_params.select(&:single_value?).map { |p| [p.name, p.value] }.to_h
+    # build up a config for the certificate
```
james-ball-qualcomm, line 184: Is this code just for the certificate? This is the portfolio file.

lib/config.rb
```diff
@@ -0,0 +1,169 @@
+# frozen_string_literal: true
```
james-ball-qualcomm, line 1: Please add a detailed comment about what this does and how it interacts with other key classes.

james-ball-qualcomm (changes requested):

james-ball-qualcomm:
Can you please update arch/README.adoc since it is stale?

lib/architecture.rb
```diff
@@ -0,0 +1,280 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
dhower-qc: there is a description of the class now

lib/config.rb
dhower-qc: there is a class description now

dhower-qc:
> Can you please update arch/README.adoc since it is stale?

done

dhower-qc:
@AFOliveira, @BrianAnakPintar do you have patches ready? If so, let's merge yours first and I'll handle any conflicts with this patch.

AFOliveira:
only #340 but you fix that in this PR, right?

dhower-qc:
> only #340 but you fix that in this PR, right?

Yes, I do.

PR #351, opened 2024-12-10
Add a base multi-arch container image build and GH Actions

rpsene:
feat: Add base container image and GitHub Actions workflow

- Introduced a base container image with essential dependencies.
- Added GitHub Actions for CI/CD automation.

### DIFF
--- a/.github/workflows/base_container_img.yml
+++ b/.github/workflows/base_container_img.yml
@@ -1,0 +1,50 @@
+name: Container Img Weekly Build and Release
+
+on:
+  schedule:
+    # This will run every week at 00:00 UTC on Monday
+    - cron: "0 0 * * 1"
+  workflow_dispatch:
+
+jobs:
+  build-and-push-docker-image:
+    name: Container Img Weekly Build and Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v3
+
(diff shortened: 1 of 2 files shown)


infrastructure/base-container-img
```diff
@@ -0,0 +1,39 @@
...
+    ruby-dev \
+    bundler \
+    nodejs \
+    npm \
+    ditaa && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# Copy and install Ruby dependencies
+COPY Gemfile Gemfile.lock ./
```
dhower-qc, line 31: We've been installing the Ruby/Node/Python packages on the host (under `${UDB_ROOT}/.home/.gems`, `${UDB_ROOT}/node_modules`, and `${UDB_ROOT}/.home/.venv`, respectively). We then make sure that the UDB_ROOT is bound in the container. The thought was that it would (a) keep the container smaller, and (b) make it easier to add packages as we go (most additions won't need a container rebuild). You can see how all this gets set up in `bin/setup` (which is implicitly called any time you execute a program under `bin/` or the top-level `do` script).

Is there a reason that we should be including them in the container instead?

.github/workflows/base_container_img.yml
```diff
@@ -0,0 +1,50 @@
+name: Container Img Weekly Build and Release
+
+on:
+  schedule:
+    # This will run every week at 00:00 UTC on Monday
+    - cron: "0 0 * * 1"
```
dhower-qc, line 6: See comment on tag/packages: if we stick with the current scheme, we only need to build this when `bin/.container_tag` changes.

dhower-qc (commented):
Thanks! I'm loving that we can build the ARM64 images through Actions, too.

dhower-qc:
we have this automated in #538

PR #352, opened 2024-12-11
fix pre-commit issues from last commit

kbroch-rivosinc:
In the future we need to document the dev environment and also add the github action for running pre-commit so dev's see this problems before they land in main.


dhower-qc:
see #353

dhower-qc (approved):

PR #356, opened 2024-12-13
LLVM verification

AFOliveira:
Following up on #258. 

I've been doing work on what @lenary proposed as a first approach, I think this is an ok first mock-up. Still a WIP since many instructions still have bugs, but if you have any comments or recommendations, please LMK :).

I did not add LLVM as a submodule yet because it may be easier licensing wise to just point at it in the script? Usage is         `python3 riscv_parser.py <tablegen_json_file> <arch_inst_directory>")`. I've also used _jq_ to enhance readibility on the output of `llvm-tblgen -I llvm/include -I llvm/lib/Target/RISCV llvm/lib/Target/RISCV/RISCV.td --dump-json -o <path-to-json-output>`.


ext/auto-inst/parsing.py
```diff
@@ -0,0 +1,328 @@
...
+        # Print the instruction details without separating by category
+        output_stream.write(f"\n{name} Instruction Details\n")
+        output_stream.write("=" * 50 + "\n")
+
+        # Basic Information
+        output_stream.write("\nBasic Information:\n")
+        output_stream.write("-" * 20 + "\n")
+        output_stream.write(f"Name:              {name}\n")
+        output_stream.write(f"Assembly Format:   {safe_get(data, 'AsmString', 'N/A')}\n")
+        output_stream.write(f"Size:              {safe_get(data, 'Size', 'N/A')} bytes\n")
```
lenary, line 140: If an instruction doesn't have a Size, it cannot be encoded, so it's likely not of interest to checking the encoding.

ext/auto-inst/parsing.py
```diff
@@ -0,0 +1,328 @@
...
+        except:
+            output_stream.write("Outputs:           N/A\n")
+
+        # Instruction Properties
+        output_stream.write("\nInstruction Properties:\n")
+        output_stream.write("-" * 20 + "\n")
+        output_stream.write(f"Commutable:        {'Yes' if safe_get(data, 'isCommutable', 0) else 'No'}\n")
+        output_stream.write(f"Memory Load:       {'Yes' if safe_get(data, 'mayLoad', 0) else 'No'}\n")
+        output_stream.write(f"Memory Store:      {'Yes' if safe_get(data, 'mayStore', 0) else 'No'}\n")
+        output_stream.write(f"Side Effects:      {'Yes' if safe_get(data, 'hasSideEffects', 0) else 'No'}\n")
```
lenary, line 168: At some point we probably need a good discussion about these and how we verify them.

Broadly:
- isCommutable will only be set if we can teach LLVM how to commute the instruction - using a hook implemented in C++. We do this quite a lot for e.g. inverting conditions. This is not something the assembler will do, it's done earlier, in code generation though.
- mayLoad and mayStore should be accurate enough, but I guess you'll need to statically analyse the pseudocode to work out if a load or store happens. We really only model loads/stores to conventional memory (and not, e.g., loads for page table walks or permission checks). I'm not sure there's any modelling of ordering. These are used to prevent code motion during code generation.
- hasSideEffects is a catch all for "be very careful with this instruction during codegen", and usually points to other effects that LLVM doesn't model. Generic CSR accesses are part of this (except the floating point CSR, which I think we model correctly), but so are other effects I haven't thought hard about.

ext/auto-inst/parsing.py
```diff
@@ -0,0 +1,328 @@
...
+        encoding_bits = []
+        try:
+            inst = safe_get(data, 'Inst', [])
+            for bit in inst:
+                if isinstance(bit, dict):
+                    encoding_bits.append(f"{bit.get('var', '?')}[{bit.get('index', '?')}]")
+                else:
+                    encoding_bits.append(str(bit))
+            # Reverse the bit order before joining
+            encoding_bits.reverse()
```
lenary, line 195: Of course the orders are swapped. This was just inevitable, right?

ext/auto-inst/parsing.py
```diff
@@ -0,0 +1,328 @@
...
+                instr_name = os.path.splitext(file)[0]
+                # Store lowercase key for easy lookup
+                repo_instructions[instr_name.lower()] = category
+    return repo_instructions
+
+def find_json_key(instr_name, json_data):
+    """
+    Attempt to find a matching key in json_data for instr_name, considering different
+    naming conventions: replacing '.' with '_', and trying various case transformations.
+    """
```
lenary, line 249: I really suggest using the name from `AsmString` rather than this.

lenary (commented):
I think this is a good start, and is heading in the right direction. Some immediate comments before I look again in the next few days.

Is there anything specific you want feedback on?

Which instructions are you running into issues with so far?

Another overall thought I have is: if this is a test/validation suite, could this be structured using pytest? For instance, you could parse the YAML and the JSON to match up instruction descriptions, and use that to create parameterized fixtures (one instance per matched description) - the advantage of this is that you get all of the niceness of pytest's asserts, and pytest's test suite reports, without having to reimplement all the testcase management. This also makes it easier to split the test code that checks the encoding matches from tests that check the assembly strings match (for example, but we can think of others).

ext/auto-inst/parsing.py
AFOliveira: Well, I can either reverse our yaml or LLVM, IDK if it makes any difference though

ext/auto-inst/parsing.py
AFOliveira: ok, I'll change that, thanks for the suggestion!

ext/auto-inst/parsing.py
AFOliveira: Do you think those should be explained in the UDB as well?

AFOliveira:
> I think this is a good start, and is heading in the right direction. Some immediate comments before I look again in the next few days.
> 

Thanks for your feedback!

> Is there anything specific you want feedback on?

Not yet, the point was just some general considerations about the initial approach.

> Which instructions are you running into issues with so far?

I still didnt find a pattern, but I'll try to fix this soon and if I run into any issue I'm not able to solve by myself, I'll try to bring it up, thanks!

> Another overall thought I have is: if this is a test/validation suite, could this be structured using pytest? For instance, you could parse the YAML and the JSON to match up instruction descriptions, and use that to create parameterized fixtures (one instance per matched description) - the advantage of this is that you get all of the niceness of pytest's asserts, and pytest's test suite reports, without having to reimplement all the testcase management. This also makes it easier to split the test code that checks the encoding matches from tests that check the assembly strings match (for example, but we can think of others).

I'll take a look into pytest and see how to port this, thanks for the suggestion!

ext/auto-inst/parsing.py
lenary: I think this probably needs a longer discussion on how denormalised this information should be. i.e., the information is there if you statically-analyse the pseudocode (which should absolutely be something we should be able to do with the pseudocode), but we probably don't want that to be the only way to find out this sort of thing as the pseudocode operations might be a large amount of code.

I don't think you should be exactly matching LLVM's internal representation, but I do think there is the opportunity to denormalise more information that might be generally useful for this kind of tool.

ext/auto-inst/parsing.py
lenary: I don't think you're doing anything wrong, this comment wasn't actionable.

AFOliveira: Got it. Thanks!

ext/auto-inst/parsing.py
AFOliveira: I agree that this needs a more appropriate discussion, I'll think @dhower-qc has been working on instruction representations lately, so he probably also has views on this

lenary: I don't have time for this discussion before mid-January.

Let's try get the things that are already in the YAML done first, i.e.:
- Instruction Encodings
- Extensions and Profiles

AFOliveira:
current known bugs are:
Can't parse yaml complex imm encodings (i.e. 31|7|30-25|11-8) 
Some instructions need to be fixed in the UDB, since this already caught some errors.

I'm working on both now.

lenary:
Somewhere, a `__pycache__` and `*.pyc` is missing from a gitignore. :)

ext/auto-inst/parsing.py
```diff
@@ -244,23 +244,37 @@ def get_repo_instructions(repo_directory):
...
-    }
-
-    for v in variants:
-        if v in json_data:
-            return v
+    # First, normalize the instruction name for comparison
+    instr_name = instr_name.lower().strip()
+    
+    # Search through all entries in json_data
+    for key, value in json_data.items():
```
lenary, line 261: You can use the `!instanceof` metadata to only search through things with an encoding, I pointed this out in the comment about the initial approach. This will save you iterating through aliases (and other data like isel patterns)

AFOliveira: From what I can see `!instanceof` follows this structure, but I'm finding it hard to understand how to use it for the parsing purposes, can you please further explain?

```
{
  "!instanceof": {
    "AES_1Arg_Intrinsic": [
      "int_arm_neon_aesimc",
      "int_arm_neon_aesmc"
    ],
    "AES_2Arg_Intrinsic": [
      "int_arm_neon_aesd",
      "int_arm_neon_aese"
    ],
    "ALUW_rr": [
      "ADDW",
      "ADD_UW",
      "DIVUW",
      "DIVW",
      "MULW",
      "PACKW",
      "REMUW",
      "REMW",
      "ROLW",
      "RORW",
      "SH1ADD_UW",
      "SH2ADD_UW",
      "SH3ADD_UW",
      "SLLW",
      "SRAW",
      "SRLW",
      "SUBW"
    ],
    "ALU_ri": [
      "ADDI",
      "ANDI",
      "ORI",
      "SLTI",
      "SLTIU",
      "XORI"
    ],

```

lenary: So, let's assume you parsed the whole json object, into a python variable called `json`.

`json["!instanceof"]` is a map, from tablegen classes, to a list of definition names that are instances of that class (or its sub-classes). These definition names appear in the top-level json.

You would use code like the following:

```python
for def_name in json["!instanceof"]["RVInstCommon"]:
  def_data = json[def_name]
  ...
```

This saves you having to look at all the tablegen data that is not an instruction (so an alias or a pattern or CSR or something).

Note you'll still have to look at `isPseudo` and `isCodeGenOnly` and potentially exclude items where one or both of those is true.

AFOliveira:
Unit testing is now working.

**TODO:**
Solve complex immediates
Add the optimization @lenary proposed on code structure
Add this to the validation process

AFOliveira:
Can I insert LLVM as a submodule or would this be another licensing problem?

AFOliveira:
This is functional. Things missing is how to address LLVM and adding it the test to Rakefile. I will also see how I can optimize code, i.e. what @lenary proposed on one review.

lenary:
> Can I insert LLVM as a submodule or would this be another licensing problem?

I highly suggest using environment variables to point to LLVM, and skipping these tests if those environment variables are not set.

AFOliveira:
I had LLVM outside of singularity repo thus I was not able to do it, but if I clone inside and run it correctly I think that works as well, I'll set it up!

Rakefile
```diff
@@ -293,6 +309,7 @@ namespace :test do
     Rake::Task["test:lib"].invoke
     Rake::Task["test:schema"].invoke
     Rake::Task["test:idl"].invoke
+    Rake::Task["test:llvm"].invoke
```
lenary, line 312: Can this task be changed into `task :smoke =>

AFOliveira: I can definitely change it, but since @dhower-qc was the one that wrote this in first place, I just want to ask if that's all good for him aswell.

lenary: Yeah, sorry, I was going to give more comments but pressed the wrong button (and submitted this early). I'll work through this review again next week.

AFOliveira: Thank you!

dhower-qc: yes, the task can be changed to use the prerequisite syntax

AFOliveira: done!

AFOliveira:
> No problem with an LLVM submodule.

A problem I can foresee is that CI will have to download it  ~2GB and then to build the table, is that fine or should it be avoided?

dhower-qc:
> > No problem with an LLVM submodule.
> 
> A problem I can foresee is that CI will have to download it ~2GB and then to build the table, is that fine or should it be avoided?

Hmm, that's unfortunate ;)

A few options:

- Add the LLVM check as a nightly test.
- If the LLVM files needed are contained in sub-directories, we could use [partial clone/sparse-checkout](https://github.blog/open-source/git/bring-your-monorepo-down-to-size-with-sparse-checkout/).
- Just check in the table into UDB periodically?

AFOliveira:
More on how to access that information, I couldnt find a sparse checkout that would allow me to still build the table; however, depth 1 gets it down to only around 200MB. Moreover we would need to either build llvm or have it pre installed in the CI server. Therefore, I think the approach I end up liking more is:

> Just check in the table into UDB periodically?

Can I commit it?

lenary:
> More on how to access that information, I couldnt find a sparse checkout that would allow me to still build the table; however, depth 1 gets it down to only around 200MB. Moreover we would need to either build llvm or have it pre installed in the CI server. Therefore, I think the approach I end up liking more is:
> 
> > Just check in the table into UDB periodically?
> 
> Can I commit it?

Is 200MB too much?

You don't need to do a full build to get the information you need, just `llvm-tblgen` which should be significantly smaller/quicker - when building, just do that target rather than all of them.

I'm not very happy with committing LLVM info into this repo, on both license and "it might get out of date" grounds.

AFOliveira:
> Is 200MB too much?

That, IDK :) 

To do `llvm-tblgen` we also need to have corresponding binary installed, or am I not right on this? i.e. when I try that locally I get 

`
Command 'llvm-tblgen' not found, but can be installed with:
sudo apt install llvm
`

And if I dont build it locally, I also have some version mismatch error

lenary:
My point was not to install `llvm-tblgen`, but to do just the parts of the llvm build that are enough to get `llvm-tblgen` as a binary, which should take a small percentage of the time a full LLVM build takes.

AFOliveira:
Pardon me if I am not understanding this correctly, but isn't what you're describing the following sequence of commands?

```
cmake -S llvm -B build -DCMAKE_BUILD_TYPE=Release

cmake --build build --target llvm-tblgen

./build/bin/llvm-tblgen -I llvm/include -I llvm/lib/Target/RISCV llvm/lib/Target/RISCV/RISCV.td --dump-json -o riscv.json
```

AFOliveira:
Well, this still take some time for smoke test, maybe would be better suited for nightly, but @dhower-qc probably has a better take on this, I'm fine with any option on this topic TBH.

AFOliveira:
@dhower-qc  as on out last call was talked, I added LLVM to CI with caching so it doesn't always have to build the table, For local runs, I still kept it optional, it only runs if the user have the file present, is it ok this way?

lenary (approved):
Sorry, I've taken too long to get back to this.

I have one actionable comment, but let's just land this and fix-forward the verification/validation (i.e., new PRs)

Hopefully that makes sense?

ext/auto-inst/parsing.py
```diff
@@ -0,0 +1,274 @@
...
+):
+    """Compare the YAML encoding with the JSON encoding."""
+    if not yaml_match:
+        return ["No YAML match field available for comparison."]
+    if not json_encoding_str:
+        return ["No JSON encoding available for comparison."]
+
+    expected_length = (
+        16 if instr_name.lower().startswith(("c_", "c.", "cm_", "cm.")) else 32
+    )
```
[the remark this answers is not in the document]
AFOliveira: Now you that speak about it, I'm sure we can just count the match field, sure. I'm currently really, busy but I'll make that fix whenever I can!

AFOliveira:
> Sorry, I've taken too long to get back to this.
> 
> I have one actionable comment, but let's just land this and fix-forward the verification/validation (i.e., new PRs)
> 
> Hopefully that makes sense?

Yes, that makes total sense. @dhower-qc mentioned that this was on queue depending on #386, so likely I'll have to adapt to that as well. I guess I'll change both things at the time.

Thank for the review!

AFOliveira:
@dhower-qc should be working now. Remember that the test is optional, so if the users don't have LLVM tablegen generated in their computers, it will just skip all tests.

AFOliveira:
@dhower-qc Can I get an approval on this so I don't need to consistently come back and solve merge conflicts?

dhower-qc (approved):

dhower-qc:
> @dhower-qc Can I get an approval on this so I don't need to consistently come back and solve merge conflicts?

Yep, thanks for the bump

AFOliveira:
@dhower-qc @kbroch-rivosinc the pre-commit is complaining about copyright licenses, should I just insert BSD-3 license?

AFOliveira:
@kbroch-rivosinc Sorry for bothering about this, but is it now a requirement that all files introduced into UDB have an explicit LICENSE somewhere? To fix this I should introduce the BSD-3 reference on the top?

reuse lint...............................................................Failed
- hook id: reuse
- exit code: 1

# MISSING COPYRIGHT AND LICENSING INFORMATION

The following files have no copyright and licensing information:
* ext/auto-inst/parsing.py
* ext/auto-inst/test_parsing.py

PR #358, opened 2024-12-17
Adding support for VSCode rdbg Ruby debugger extension

james-ball-qualcomm:


james-ball-qualcomm:
I don't think the JSON checker should be failing so I'm exploring the idea of changing its checks. See #359.

dhower-qc (approved):

PR #371, opened 2024-12-20
Issue 276 retire mc 1 google doc

james-ball-qualcomm:
Finally merging in improvements to MC100. Now there are 32 & 64 bit versions and MC200 and MC300.


james-ball-qualcomm:
The smoke test is failing because I added a check in the SymbolTable constructor that the cfg_arch parameter is a ConfiguredArchitecture object. However, your IDL unit-level test uses a MockConfiguredArchitecture class that isn't inherited from anything. How do you want me to fix this? There are many places that check that a cfg_arch parameter is a ConfiguredArchitecture class type. 
 
Note that this is in my oldest PR (https://github.com/riscv-software-src/riscv-unified-db/pull/371) that doesn't have the refactored ConfiguredArchitecture (inherited from the new Design class). In my next oldest PR (https://github.com/riscv-software-src/riscv-unified-db/pull/386) where the Design class exists, the SymbolTable constructor is passed a MockDesign class which also isn't inherited from anything (but could be inherit from the Design class). So, I'd recommend that we fix this in my later PR since it will be easier to make MockDesign inherit from Design than MockConfiguredArchitecture inherit from ConfiguredArchitecture.

james-ball-qualcomm:
@dhower-qc, can you please approve this so I don't keep having to merge in other changes from main?

dhower-qc (approved):

PR #376, opened 2024-12-23
Fix CSRRS

AFOliveira:
Caught by check against LLVM on #356 .


dhower-qc:
This is great. On a side note, I recently added a check for conflicting instructions encodings, which would have caught most (but not all) of these. It's in the cpp_hart branch for now, and I'll try to add it as a smoke test to catch future issues.

dhower-qc (approved):

PR #389, opened 2025-01-06
Improved support for devcontainer

james-ball-qualcomm:
Simple corrections to script

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -55,8 +55,9 @@ else
   done
 fi
 
-if [ "${CONTAINER_TYPE}" != "docker" -a "${CONTAINER_TYPE}" != "singularity" ]; then
+if [ "${CONTAINER_TYPE}" != "docker" -a "${CONTAINER_TYPE}" != "singularity" -a "${CONTAINER_TYPE}" != "devcontainer" ]; then
   echo "BAD CONTAINER TYPE: ${CONTAINER_TYPE}"
+  exit 1
 fi
 
 if [ ! -f ${ROOT}/.container-type ]; then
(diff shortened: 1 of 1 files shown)


james-ball-qualcomm:
Looks like this small PR can be approved. @dhower-qc

.gitignore
```diff
@@ -11,6 +11,7 @@
 diag-ditaa-*
 arch/manual/isa/**/riscv-isa-manual
 gen
+gen.*
```
dhower-qc, line 14: what's this for?

bin/setup
```diff
@@ -56,14 +56,28 @@ else
 fi
 
 if [ "${CONTAINER_TYPE}" != "docker" -a "${CONTAINER_TYPE}" != "singularity" -a "${CONTAINER_TYPE}" != "devcontainer" ]; then
-  echo "BAD CONTAINER TYPE: ${CONTAINER_TYPE}"
+  echo "BAD CONTAINER TYPE: ${CONTAINER_TYPE}" 1>&2
+  exit 1
```
dhower-qc, line 60: This, and the change below, are already in the main branch (I added it as part of #393).

dhower-qc (commented):
Since the meat of this is already pushed, we probably just want to abandon this PR.

.gitignore
james-ball-qualcomm: Yes, wanted to talk to you about this one. I'd like to have the ability to keep old copies of the generated files in my workspace for reference. I want them ignored by Git. So, my scheme is to take the gen directory and rename it to gen.<something> so that's why I've added gen.* in .gitignore. If you have another way of solving this, I'm open to it. This is just what I was imagining as a solution.

james-ball-qualcomm: Derek suggested changing this to *.bak in the .gitignore. I'll do that in some future PR.

PR #390, opened 2025-01-06
Change to http from https to work around Qualcomm IT SSL issues

james-ball-qualcomm:
Still need a fix for Python.


dhower-qc:
I think we determined this was a configuration issue with your machine. Re-open if I'm mis-remembering.

PR #396, opened 2025-01-08
convert f-string termination to single-quote to fix syntax errors

### LINKED ISSUE #395: malformed f-strings causing parsing problems in python 3.9
I put in the https://pre-commit.com/#top_level-default_language_version
and specified 3.9 as a conservative version as it is the oldest supported and that ships with RHEL9.

Upon doing this the pre-commit hook `check-ast` flagged several f-strings that look problematic. (a second double-quote terminating the f-string prematurely)

Example:

```
❯ pre-commit run --all-files check-ast
check python ast.........................................................Failed
- hook id: check-ast
- exit code: 1

lib/yaml_resolver.py: failed parsing with CPython 3.9.6:

    Traceback (most recent call last):
      File "/Users/kbroch/.cache/pre-commit/repora9v2e19/py_env-python3.9/lib/python3.9/site-packages/pre_commit_hooks/check_ast.py", line 21, in main
        ast.parse(f.read(), filename=filename)
      File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/ast.py", line 50, in parse
        return compile(source, filename, mode, flags,
      File "lib/yaml_resolver.py", line 269
        ref_obj["$parent_of"].append(f"{obj_file_path}#/{"/".join(obj_path)}")
                                                          ^
    SyntaxError: f-string: expecting '}'
```

kbroch-rivosinc:
fixes #395

### DIFF
--- a/lib/yaml_resolver.py
+++ b/lib/yaml_resolver.py
@@ -214,7 +214,7 @@ def resolve(rel_path : str | Path, arch_root : str | Path, do_checks: bool) -> d
       exit(1)
     fn_name = Path(rel_path).stem
     if do_checks and (fn_name != unresolved_arch_data["name"]):
-      print(f"ERROR: 'name' key ({unresolved_arch_data["name"]}) must match filename ({fn_name} in {arch_root}/{rel_path}", file=sys.stderr)
+      print(f"ERROR: 'name' key ({unresolved_arch_data['name']}) must match filename ({fn_name} in {arch_root}/{rel_path}", file=sys.stderr)
       exit(1)
     resolved_objs[str(rel_path)] = _resolve(unresolved_arch_data, [], rel_path, unresolved_arch_data, arch_root, do_checks)
     return resolved_objs[str(rel_path)]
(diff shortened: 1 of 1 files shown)


lib/yaml_resolver.py
```diff
@@ -214,7 +214,7 @@ def resolve(rel_path : str | Path, arch_root : str | Path, do_checks: bool) -> d
       exit(1)
     fn_name = Path(rel_path).stem
     if do_checks and (fn_name != unresolved_arch_data["name"]):
-      print(f"ERROR: 'name' key ({unresolved_arch_data["name"]}) must match filename ({fn_name} in {arch_root}/{rel_path}", file=sys.stderr)
+      print(f'ERROR: "name" key ({unresolved_arch_data["name"]}) must match filename ({fn_name} in {arch_root}/{rel_path}', file=sys.stderr)
```
ThinkOpenly, line 217: This change is actually changing the output, from:
```
ERROR: 'name' key [...]
```
to:
```
ERROR: "name" key [...]
```
...if anyone cares.
Another approach to fixing this which preserves the output is:
```
print(f"ERROR: 'name' key ({unresolved_arch_data['name']}) must match filename ({fn_name} in {arch_root}/{rel_path}", file=sys.stderr)
```
(single quotes around the "name" dictionary index string.)

lib/yaml_resolver.py
```diff
@@ -266,11 +266,11 @@ def _resolve(obj, obj_path, obj_file_path, doc_obj, arch_root, do_checks):
 
       if "$parent_of" in ref_obj:
         if isinstance(ref_obj["$parent_of"], list):
-          ref_obj["$parent_of"].append(f"{obj_file_path}#/{"/".join(obj_path)}")
+          ref_obj["$parent_of"].append(f'{obj_file_path}#/{"/".join(obj_path)}')
```
ThinkOpenly, line 269: For this and the next two, (this is a nit, but) I would've just put single quotes around the `/` character.

lib/yaml_resolver.py
kbroch-rivosinc: @ThinkOpenly : thanks agree with all of these, I've modified commit to reflect these.

kbroch-rivosinc:
@dhower-qc : I remove setting python to 3.9 in the pre-commit cfg, but I still think these edits are useful and I am unsure why py 3.12 is fine with these.  (My guess is the curl brackets of variable in f-string are parsed differently in > 3.9)

dhower-qc (approved):

PR #398, opened 2025-01-08
Update Zbk* extensions

jmawet:
Updated instructions in Zbkb and Zbkx


ThinkOpenly:
Would you be willing to include the Sail code as well? It can be found at https://github.com/riscv/sail-riscv/blob/master/model/riscv_insts_zbkb.sail.

arch/inst/Zbkb/brev8.yaml
```diff
@@ -23,3 +23,12 @@ access:
...
 operation(): |
+  if (implemented?(ExtensionName::B) && (CSR[misa].Zbkb == 1'b0)) {
+    raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
+  }
+
+  result : xlenbits = EXTZ(0b0);
+  foreach (i from 0 to sizeof(xlen) by 8) {
+    result[i+7..i] = reverse_bits_in_byte(X(rs1)[i+7..i]);
+  };
+  X(rd) = result;
```
dhower-qc, line 34: This looks like Sail. `operation()` needs to be IDL. We can (should) also include the Sail code, but that will go under the `sail()` key.

dhower-qc (changes requested):

arch/inst/Zbkb/brev8.yaml
jmawet: I've moved the existing sail code to the `sail()` key, and updated `operation()` to match the rest to the best of my abilities. Let me know if my new changes are correct

dhower-qc: Looks good now!

dhower-qc (dismissed):

arch/inst/Zbkb/unzip.yaml
```diff
@@ -3,9 +3,11 @@
 $schema: inst_schema.json#
 kind: instruction
 name: unzip
-long_name: No synopsis available.
+long_name: Inverse of Zip.
```
ThinkOpenly, line 6: Page 215 of the ISA spec calls `unzip` "Bit deinterleave", which is arguably more directly informative than "Inverse of Zip".

arch/inst/Zbkb/brev8.yaml
```diff
@@ -3,9 +3,9 @@
 $schema: inst_schema.json#
 kind: instruction
 name: brev8
-long_name: No synopsis available.
+long_name: Reverse the bits in each byte of a source register.
```
ThinkOpenly, line 6: Page 375 of the ISA spec calls `brev8` "Reverse bits in bytes". There's an argument for (ironically) keeping the "long_name" somewhat shorter. This is not a show-stopper if left as-is.

arch/inst/Zbkb/zip.yaml
```diff
@@ -3,9 +3,11 @@
 $schema: inst_schema.json#
 kind: instruction
 name: zip
-long_name: No synopsis available.
+long_name: Gather odd and even bits of the source word into upper/lower halves of the destination.
```
ThinkOpenly, line 6: Page 375 of the ISA spec calls `zip` "Bit interleave".

arch/inst/Zbkx/xperm4.yaml
```diff
@@ -3,9 +3,11 @@
 $schema: inst_schema.json#
 kind: instruction
 name: xperm4
-long_name: No synopsis available.
+long_name: Nibble-wise lookup of indicies into a vector.
```
ThinkOpenly, line 6: Page 376 calls `xperm4` "Crossbar permutation (nibbles)".

(Arguably, it should be "nybbles", but I'm sure there are those that disagree. ;-) )

arch/inst/Zbkx/xperm8.yaml
```diff
@@ -3,9 +3,11 @@
 $schema: inst_schema.json#
 kind: instruction
 name: xperm8
-long_name: No synopsis available.
+long_name: Byte-wise lookup of indicies into a vector.
```
ThinkOpenly, line 6: Page 376 calls `xperm8` "Crossbar permutation (bytes)".

ThinkOpenly (commented):
Sorry, this probably should've been merged, but I took another look since it was sitting there. (Sorry.) I'm not hung up on these changes, so just marking this review as "comment".

arch/inst/Zbkb/unzip.yaml
jmawet: I got this from page 281 of the spec, but I agree "bit deinterleave" is more informative. Could we do "Bit deinterleave (inverse of Zip)" to have both? Or would you prefer I change it to just "Bit Deinterleave"?

arch/inst/Zbkb/brev8.yaml
jmawet: I got this from page 410 of the spec. In my opinion the slightly more verbose name would be better but I'm not the maintainer so I will go with what you want once you confirm

arch/inst/Zbkx/xperm4.yaml
jmawet: I got this from page 426, but I can change it if you'd like

jmawet:
> Sorry, this probably should've been merged, but I took another look since it was sitting there. (Sorry.) I'm not hung up on these changes, so just marking this review as "comment".

So it seems you'd prefer I pick from pages 375/376 with the summary of instructions (right most column in those tables). I've been picking from the pages with the complete instruction details, but I can make the changes if you'd like

arch/inst/Zbkb/unzip.yaml
ThinkOpenly: > I got this from page 281 of the spec

To make sure we're on the same page (no pun intended!), what version of the spec are you using? I've got 20240411.

page | section | section name | string
-- | -- | -- | --
215 | 28.4.5 | Zbkb: Bit-manipulation for Cryptography | Bit deinterleave
262 | 28.5.46 | unzip | Synopsis: Implements the inverse of the zip instruction
375 | 32.3.1 | Zbkb - Bitmanip instructions for Cryptography | Bit deinterleave
423 | 32.4.45 | unzip | Synopsis: Implements the inverse of the zip instruction
438 | 32.6.5 | Zkt listings | Bit deinterleave
440 | 32.7.4 | Bitmanip Instructions for Cryptography | Bit de-interleave

So, other than synopses, the human-readable _name_ appears to fairly consistently "bit deinterleave".

My understanding of the YAML fields is that "name" is the instruction mnemonic (here, "unzip"), and "long_name" is essentially a short human-readable version of the mnemonic. And the synopses in the ISA are short descriptions (generally longer than a translated mnemonic).

Of course, this is all somewhat subjective.

arch/inst/Zbkb/brev8.yaml
ThinkOpenly: Let's go with the shorter human-readable name. Here, "Reverse bits in bytes".

arch/inst/Zbkx/xperm4.yaml
ThinkOpenly: Yeah, "Crossbar permutation (nibbles)" seems to be what the ISA spec uses in two places as the "name" (contrast with "synopsis") for this instruction.

ThinkOpenly:
> So it seems you'd prefer I pick from pages 375/376 with the summary of instructions (right most column in those tables). I've been picking from the pages with the complete instruction details, but I can make the changes if you'd like, just let me know

I think I'm arguing for the shorter human-readable names, where they can be found, in lieu of the text (or a summary of) the synopses.

Note that the RISC-V ISA Manual that is created from this data uses the "long_name" immediately under the mnemonic, and has a "Synopsis" section for each instruction which is, ironically, based on the "description" field. Example: https://riscv-software-src.github.io/riscv-unified-db/manual/html/isa/20240411/insts/brev8.html

arch/inst/Zbkb/unzip.yaml
jmawet: I am using this version: https://drive.google.com/file/d/1uviu1nH-tScFfgrovvFCrj7Omv8tFtkp/view
It seems the google pdf page numbers and the spec page numbers are not exactly aligned. Regardless, I see the page you are referencing, I will make the requested change

jmawet:
alright I will make the changes and update here when done

arch/inst/Zbkb/unzip.yaml
```diff
@@ -24,3 +26,23 @@ access:
...
+
+  for(U32 i=0; i<(xlen()/2-1); i = i+1) {
+    output[i] = input[2*i];
+    output[i+xlen()/2] = input[2*i+1];
+  }
+
+  X[rd] = output;
+
+sail(): |
+
```
ThinkOpenly, line 44: nit: is it OK to have a blank line here?

ThinkOpenly (approved):
There are a few instances where the "sail" field has a blank leading line. Not sure if it's a problem. Fix if you wish. Looks good otherwise, so I'm approving.

dhower-qc (changes requested):
For all of these instructions, they are only defined by the primary extension (Zbkb, Zbkx), not B, Zbb, etc. We'll need to adjust `definedBy` and remove the check at the beginning of each `operation()`

ThinkOpenly:
> For all of these instructions, they are only defined by the primary extension (Zbkb, Zbkx), not B, Zbb, etc. We'll need to adjust `definedBy` and remove the check at the beginning of each `operation()`

Not related to this PR, but to your statement above...

I think, then, that these should only reference the "Zfa" extention, correct?
```
arch/inst/F/fleq.s.yaml:definedBy: { allOf: [F, Zfa] }
arch/inst/F/fli.s.yaml:definedBy: { allOf: [F, Zfa] }
arch/inst/F/fltq.s.yaml:definedBy: { allOf: [F, Zfa] }
arch/inst/F/fmaxm.s.yaml:definedBy: { allOf: [F, Zfa] }
arch/inst/F/fminm.s.yaml:definedBy: { allOf: [F, Zfa] }
arch/inst/F/froundnx.s.yaml:definedBy: { allOf: [F, Zfa] }
arch/inst/F/fround.s.yaml:definedBy: { allOf: [F, Zfa] }
```

I can open a separate PR for these changes.

dhower-qc:
> 
> I think, then, that these should only reference the "Zfa" extention, correct?
> 
> ```
> arch/inst/F/fleq.s.yaml:definedBy: { allOf: [F, Zfa] }
> arch/inst/F/fli.s.yaml:definedBy: { allOf: [F, Zfa] }
> arch/inst/F/fltq.s.yaml:definedBy: { allOf: [F, Zfa] }
> arch/inst/F/fmaxm.s.yaml:definedBy: { allOf: [F, Zfa] }
> arch/inst/F/fminm.s.yaml:definedBy: { allOf: [F, Zfa] }
> arch/inst/F/froundnx.s.yaml:definedBy: { allOf: [F, Zfa] }
> arch/inst/F/fround.s.yaml:definedBy: { allOf: [F, Zfa] }
> ```
> 
> I can open a separate PR for these changes.

That's true.

TLDR; Technically, either `allOf: [F, Zfa]` or just `Zfa` will work since UDB captures (transitive) requirements and Zfa requires F. However, just Zfa is closer to how the spec is written.

arch/inst/Zbkb/unzip.yaml
jmawet: I went ahead and removed the blank lines

jmawet:
> For all of these instructions, they are only defined by the primary extension (Zbkb, Zbkx), not B, Zbb, etc. We'll need to adjust `definedBy` and remove the check at the beginning of each `operation()`

The definedBy is generated automatically, I did not set that. 
That being said, I have gone ahead and made the requested change. They should only be defined by the primary extension now

arch/inst/Zbkb/brev8.yaml
```diff
@@ -3,11 +3,11 @@
...
 kind: instruction
 name: brev8
-long_name: No synopsis available.
+long_name: Reverse bits in bytes
 description: |
-  No description available.
+  This instruction reverses the order of the bits in every byte of a register.
 definedBy:
-  anyOf: [B, Zbkb, Zk, Zkn, Zks]
+  anyOf: [Zbkb]
```
ThinkOpenly, line 10: I worry we're testing your patience, so thank you for sticking with it.

Could you change these two lines to:
```
definedBy: Zbkb
```
...and similarly for the others, below?

jmawet: It seems I cannot:
`arch/inst/Zbkb/unzip.yaml::$.definedBy: ['Zbkb'] is not valid under any of the given schemas`
Seeing the same for all of them. I think the anyOf has to be there

jmawet: done

ThinkOpenly (approved):

dhower-qc (approved):
Looks good!

ThinkOpenly:
nit, but I'd recommend a squash merge here. ;-)

PR #402, opened 2025-01-09
Clarify jalr description and IDL

### LINKED ISSUE #392: `jalr` IDL and description is unclear and potentially invalid
The ISA manual and Sail code specify that `jalr` clears the LSB of the calculated target address before jumping. The IDL code and description in this database do not specify this (not even in the `jump` function, to my knowledge), potentially making these entries invalid.

```
jump(X[rs1] + imm)
```
https://github.com/riscv-software-src/riscv-unified-db/blob/main/arch/inst/I/jalr.yaml#L30

**Expected value**
If I understand IDL correctly, it should be something vaguely like this:
```
jump({(X[rs1] + imm)[XLEN - 1:1], 1'b0})
```
The description should also be updated in my opinion to better clarify this behaviour.

ThePuzzlemaker:
Fixes #392

### DIFF
--- a/arch/inst/I/jalr.yaml
+++ b/arch/inst/I/jalr.yaml
@@ -6,7 +6,7 @@ name: jalr
 long_name: Jump and link register
 description: |
   Jump to an address formed by adding rs1
-  to a signed offset then clearing the least 
+  to a signed offset then clearing the least
   significant bit, and store the return address
   in rd.
 definedBy: I
--- a/arch/inst/I/jalr.yaml
+++ b/arch/inst/I/jalr.yaml
@@ -6,7 +6,9 @@ name: jalr
 long_name: Jump and link register
 description: |
   Jump to an address formed by adding rs1
-  to a signed offset, and store the return address in rd.
+  to a signed offset then clearing the least 
+  significant bit, and store the return address
+  in rd.
 definedBy: I
 assembly: xd, imm(rs1)
 encoding:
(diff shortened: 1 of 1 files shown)


ThinkOpenly (approved):

arch/inst/I/jalr.yaml
```diff
@@ -27,7 +29,7 @@ operation(): |
   XReg returnaddr;
   returnaddr = $pc + 4;
 
-  jump(X[rs1] + imm);
+  jump((X[rs1] + imm) & ~1);
```
dhower-qc, line 32: I know I said this would work, but now I've thought about it more ;)

C/C++-style literals in IDL use the smallest amount of bits necessary to represent the number. So the _unsigned_ literal `1` will be a 1-bit value. Negated, it becomes a 1-bit `0`, and in the `&` operator it will be zero-extended (because it's unsigned) to an XLEN-bit `0`.

Any of the following will work to get what you want:

- `XLEN'1` (Verilog-style literal, read as: XLEN-bit integer with value 1)
- `'1` (width defaults to XLEN in Verilog-style, so `'1` is equivalent to `XLEN'1`).
- `1s` (because it's signed, this is a 2-bit value `01`, and will be sign-extended when it is enlarged in the conjunction)

dhower-qc (changes requested):

arch/inst/I/jalr.yaml
ThinkOpenly: _My 2 cents is that I find the `1s` approach the least understandable to the non-Verilog crowd, like me. `XLEN'1` is pretty explicit. I'd vote for that choice._

ThePuzzlemaker: Makes sense, thanks!

dhower-qc (dismissed):
Thanks!

dhower-qc (approved):

PR #403, opened 2025-01-09
C++ model generator

dhower-qc:


backends/cpp_hart_gen/lib/gen_cpp.rb
```diff
@@ -301,6 +301,8 @@ def gen_cpp(symtab, indent = 0, indent_spaces: 2)
         else
           "#{' ' * indent}__UDB_MUTABLE_GLOBAL(#{text_value})"
         end
+      elsif text_value == "imm"
```
dhower-qc, line 304: actually, this _special case_ is going to apply to all decode variables.

so, we want something like:

```ruby
elsif !var.nil? && var.decode_var?
  "#{' ' * indent}#{text_value}()"
```

dhower-qc:
@ThinkOpenly: This PR is ready now. I added you as a reviewer with the full understanding that a detailed review is an absurd request. Just wanted to give you the opportunity to take a look at whatever detail you want.

arch/inst/B/zext.h.yaml
```diff
@@ -14,7 +14,11 @@ description: |
   [NOTE]
   The *zext.h* instruction is a pseudo-op for `packw` when `Zbkb` is implemented and XLEN == 64.
 definedBy:
-  anyOf: [B, Zbb]
+  # when The bitmanip cryptography extensions are implemented, then zext.h is an alias
+  allOf:
+    - not:
+        anyOf: [Zbkb, Zk, Zkn, Zks]
+    - anyOf: [B, Zbb]
```
ThinkOpenly, line 21: Is this clearly documented in the spec anywhere? What I see is that `zext.h` is defined by "Zbb" extension, or if "Zbb" is not present, then it is a pseudoinstruction for `slli/srli` (which calls into question the two NOTEs, above).

arch/inst/C/c.beqz.yaml
```diff
@@ -29,8 +29,8 @@ operation(): |
   if (implemented?(ExtensionName::C) && (CSR[misa].C == 1'b0)) {
     raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
   }
-  if (X[rs1] != 0) {
-    jump($pc + imm);
+  if (X[xs1 + 8] == 0) {
```
ThinkOpenly, line 32: I wonder if it would read better if `xs1 + 8` were instead a mapping function. Sail uses `creg2reg_idx`, for example (which, interestingly, does a concatenation `0b01 @ creg`). "8" just seems like a magic number, absent some sort of explanation.

This occurs in many places, of course.

arch/inst/C/c.fld.yaml
```diff
@@ -10,9 +10,12 @@ description: |
...
 definedBy:
-  allOf:
-    - C
-    - D
+  anyOf:
+    - allOf:
+        - C
+        - D
+    - Zcd
+base: 32
```
ThinkOpenly, line 18: Is this correct? The spec says "C.FLD is an RV32DC/RV64DC-only instruction".

arch/inst/C/c.slli.yaml
```diff
@@ -11,16 +11,24 @@ definedBy:
...
-    - name: shamt
-      location: 12|6-2
-    - name: rd
-      location: 11-7
-      not: 0
+  RV32:
+    match: 0000----------10
+    variables:
+      - name: shamt
+        location: 6-2
```
ThinkOpenly, line 20: Should have:
```
  not: 0
```
Per the spec, "For RV32C and RV64C, the shift amount must be non-zero".

Same for the RV64 stanza, below.

arch/inst/C/c.slli.yaml
```diff
@@ -11,16 +11,24 @@ definedBy:
...
+      - name: rd
+        location: 11-7
+        not: 0
```
ThinkOpenly, line 23: I don't think "not: 0" is correct here. Since this is a compressed instruction, a "0" register index means `x8`.

Same for the RV64 stanza, below.

arch/inst/C/c.srai.yaml
```diff
@@ -12,15 +12,22 @@ definedBy:
...
-  variables:
-    - name: shamt
-      location: 12|6-2
-    - name: rd
-      location: 9-7
+  RV32:
+    match: 100001--------01
+    variables:
+      - name: shamt
+        location: 6-2
```
ThinkOpenly, line 21: I think both RV32 and RV64 stanzas, "shamt" variables need "not: 0".
The spec says, "C.SRAI is defined analogously to C.SRLI" and (for C.SRLI) "For RV32C and RV64C, the shift amount must be non-zero".

arch/inst/C/c.srli.yaml
```diff
@@ -12,15 +12,22 @@ definedBy:
...
-  variables:
-    - name: shamt
-      location: 12|6-2
-    - name: rd
-      location: 9-7
+  RV32:
+    match: 100000--------01
+    variables:
+      - name: shamt
+        location: 6-2
```
ThinkOpenly, line 21: "shamt" variables need "not: 0" here, too.

arch/inst/Zcmp/cm.mva01s.yaml
```diff
@@ -8,12 +8,12 @@ description: |
...
+  allOf:
+    - not:
+        anyOf:
+          - allOf: [C, D]
+          - Zcd
     - Zcmp
-excludedBy:
-  anyOf:
-    - allOf: [C, D]
-    - Zcd
```
ThinkOpenly, line 16: Can't this just be:
```
definedBy: Zcmp
```
?

"Zcmp" is incompatible with Zcd, so won't the extension support logic cover the incompatibilities already?

arch/inst/Zicsr/csrrs.yaml
```diff
@@ -30,14 +30,17 @@ access:
...
+  Boolean will_write = rs1 != 0;
+  check_csr(csr, will_write, $encoding);
 
   XReg initial_csr_value = CSR[csr].sw_read();
-  XReg mask = X[rs1];
 
-  # set bits using the mask
-  # performing any WARL transformations first
-  CSR[csr].sw_write(initial_csr_value | mask);
+  if (rs1 != 0) {
```
ThinkOpenly, line 38: shall we use `will_write` here, too?

ThinkOpenly (commented):
Finished a review pass. Nothing urgent, but if we postpone anything significant, lets open issues to track.

arch/inst/B/zext.h.yaml
dhower-qc: `zext.h` and `packw` are the same instruction in RV64 and `zext.h` and `pack` are the same instruction in RV32. From my reading of the spec, `pack*` is the real instruction and `zext.h` is the alias.

A note in the `pack` instruction states:

> For RV32, the pack instruction with rs2=x0 is the zext.h instruction. Hence, for RV32, any extension that contains the pack instruction also contains the zext.h instruction (but not necessarily the c.zext.h instruction, which is only guaranteed to exist if both the Zcb and Zbb extensions are implemented).

Similar for `packw`:

> For RV64, the packw instruction with rs2=x0 is the zext.h instruction. Hence, for RV64, any extension that contains the packw instruction also contains the zext.h instruction (but not necessarily the c.zext.h instruction, which is only guaranteed to exist if both the Zcb and Zbb extensions are implemented).

So, what this is trying to captrue is that `zext.h` is only defined as a real instruction when:

    - Zbb is implemented (`anyOf: [B, Zbb]`)
    - `pack*` is not implemented (`anyOf: [Zbkb, Zk, Zkn, Zks]`)

arch/inst/C/c.beqz.yaml
dhower-qc: Done

arch/inst/C/c.fld.yaml
dhower-qc: Good catch.

arch/inst/Zicsr/csrrs.yaml
dhower-qc: Yep.

arch/inst/Zicsr/csrrsi.yaml
```diff
@@ -24,3 +24,16 @@ access:
   vu: always
 data_independent_timing: false
 operation(): |
+  Boolean will_write = uimm != 0;
+  check_csr(csr, will_write, $encoding);
+
+  XReg initial_csr_value = CSR[csr].sw_read();
+
+  if (uimm != 0) {
```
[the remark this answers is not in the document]
dhower-qc: Yep.

arch/inst/Zcmp/cm.mva01s.yaml
dhower-qc: Yea, good point. This is actually stale; we don't use excludedBy anymore. I've removed it.

arch/inst/B/zext.h.yaml
ThinkOpenly: You must be reading a newer version of the spec than I am (20240411). I can't easily find any of the text you reference.
Nevertheless, the bits match as you described.

So, the logic you are trying to express is that `zext.h` is only defined when `pack` is _not_ defined _and_ when "Zbb" is defined, correct?

Can't that be expressed as:
```
defined:
  allOf:
    - Zbb # defines "zext.h"
    - not: Zbkb # defines "pack" and "packw"
```
?

~~Or, is `zext.h` _always_ defined (with "Zbb"), and rs2=0 should be removed from `pack` and `packw`? But only with "Zbb" defined?~~ Ugh.

How about the above? "Zbb" and not "Zbkb"? The rest of the logic should be automatic in processing the extension dependencies and "implies".

dhower-qc: Just as an update: in the process of refactoring this change, I ran into the following problem:

What 'C' implies is dependent on whether or not you also implement 'F' or 'D' (i.e., if it implies Zcf / Zcd). The current schema can't handle that.

I'm thinking through how to get this represented. It's probably going to need something like JSON Schema if/then constructs. Let me know if you have other thoughts.

ThinkOpenly: Venn diagrams? Yeah, some sort of logic is needed. (Is there an analogy to `type` vs `type()`? Maybe we don't need something that fancy.)
```
implies:
- Zcf
  - if: D
- Zcd
  - if: C
```
...and you could make the value of the "if" field a full "requires" logic field, so you could have more complex logic like:
```
implies:
- ABC
  if:
    allOf: [ A, B, C ]
    not: CBA
```

AFOliveira:
@dhower-qc I think you need to approve your own PR to merge

dhower-qc:
I can't ;) Can you?

AFOliveira (approved):
I ran it locally without any problems, although I did not look through all the 658 changed files :)

AFOliveira:
@dhower-qc I removed it from queue only because IDK if you want Paul to see it before going main or if it's fine

AFOliveira:
@dhower-qc Oh, I think you added it just after my review, did I mess up?

PR #411, opened 2025-01-10
Reduce duplicate code between various backend templates

james-ball-qualcomm:


james-ball-qualcomm:
Ooops, meant to make this a draft PR.

james-ball-qualcomm:
I'd like to get this merged into main.

ThinkOpenly:
Not to be a pain, but it would be nice if each commit was reasonably understandable standalone:
- a meaningful subject line for the commit message
- a meaningful description in the body of the commit message

I'm looking at "One more fix" and "Finished" in particular. ;-)

We don't have contribution guidelines, but we probably should start thinking about that. The generations that follow us might need to understand why we made the choices we made. :-)

james-ball-qualcomm:
Included in #584

PR #416, opened 2025-01-14
in ext_pdf, display inst defined by as table when the condition is simple

dhower-qc:


backends/ext_pdf_doc/templates/ext_pdf.adoc.erb
```diff
@@ -182,6 +182,22 @@ Requires::
...
+[cols="1,2"]
+|===
+| Operator | Meaning
+
+| `~> VERSION` | Accepts any version that is _compatible_ with `VERSION`. By default, a version A is compatible with a version B if A's version number is greater than or equal to version B. If that default assumption is ever broken, it will be noted in the list of extension versions.
+| `= VERSION` | Accepts only version `VERSION`.
+| `>= VERSION` | Accepts any version greater than or equal to `VERSION`.
+| `\<= VERSION` | Accepts any version less than or equal to `VERSION`.
+| `> VERSION` | Accepts any version greater than `VERSION`.
+| `< VERSION` | Accepts any version less than `VERSION`.
```
ThinkOpenly, line 199: do you need a backslash before `<` here, as is done 2 lines above?

dhower-qc: maybe? It rendered correctly without one. The `<=` was turning into a Unicode left arrow.

ThinkOpenly: I just noticed that it rendered funny here, and saw the escaped character above. NBD if it renders fine in the result.
image

dhower-qc: Seems like I was running into a ["Character Replacement"](https://docs.asciidoctor.org/asciidoc/latest/subs/replacements/) with `<=`. `<` gets substituted  for `&lt`, which is OK.

PR4NJ41: Hi @dhower-qc, @ThinkOpenly 

Is there any slack or something for LFX related queries ?? I would love to contribute to RISCV. Thank You

dhower-qc: Hi @PR4NJ41! There is no slack channel. We are in the process of starting a group under RISC-V International (riscv.org) to discuss UDB development. In the meantime, UDB is open source so contributions are always welcome. If you'd like, we can have a chat to discuss where help is needed.

PR4NJ41: @dhower-qc I have set up this repository using Docker and can successfully use do commands. However, I am facing some challenges in figuring out how to start contributing. Previously, I worked on a Ruby project during GSOC, and now I’m eager to contribute through the upcoming LFX program.

Issue #425, opened 2025-01-17
schema files have problematic top level "$ref" definitions


kbroch-rivosinc:
Its unclear to me why we want these top level "$ref":

```
~/rvi/repos/riscv-software-src/riscv-unified-db on main                                                                                                           Py 3.12.4 at 05:14:11 PM
❯ jq -r 'select(has("$ref")) | ."$ref"' schemas/*_schema.json

#/$defs/csr_register
#/$defs/ext_data
#/$defs/inst_data
```

iakshkhurana:
Sir please assign me this issue.
Thanks.

PR #426, opened 2025-01-17
remove top level ref keywords

kbroch-rivosinc:
relates to #425


dhower-qc (changes requested):
To pull these out, I think we'll have to move what's being pointed to (e.g., #/$defs/csr_register) to the top level, too.

dhower-qc (approved):

dhower-qc:
I think this is fixed now

PR #436, opened 2025-01-23
Floating point subtraction and description for floating point addition

ShashankVM:
This PR adds the IDL code for floating point subtraction and also updates description for floating point addition.


ThinkOpenly:
If you are willing, could you update the "long_name" for each as well? I suggest:
- fadd.s: "single-precision floating-point addition"
- fsub.s: "single-precision floating-point subtraction"

(These phrases are taken verbatim from the ISA, p.115.)

ShashankVM:
Hi Paul,

Yes, will do it tomorrow. It's night here in India.

Thanks,
Shashank

PR #444, opened 2025-01-26
Add excludedBy statement to all Zcmp instructions

ayosher:
Add excludedBy statement to all Zcmp instructions to make sure it will not be used when Zcd or C + D extensions are included in configuration


ThinkOpenly:
Isn't `excludedBy: Zcd` sufficient?

ayosher:
> Isn't `excludedBy: Zcd` sufficient?

Spec initially defined, that when we have C + D extensions the instructions for double floating point load/store with 16-bit encoding should be included. It was called Zcd only in Zc spec. So, it could be situation when in configuration we see just C and D extensions. Now somebody will add to configuration file explicitly Zcmp (which is wrong).  In such case all Zcmp instructions should not be added to resulted configuration. Therefore, I believe, that just Zcd formally is not enough.

ThinkOpenly:
I opened issue #445 that could help with the C/D/Zcd/Zcmp implications. When C and D are defined, then Zcd should automatically be defined, but I don't think that's the case now.

dhower-qc:
`excludedBy` was removed from the schema (I thought?) after `definedBy` was extended to understand arbitrary Boolean logic, including `not`. The way to specify this now is:

```yaml
definedBy:
  allOf:
    - not:
        anyOf: [ Zcd, allOf: [C, D] ]
    - Zcmp
```

@ayosher, I believe you made this change because there was an issue generating an extension spec. I'll see if I can catch up and figure this out.

ayosher:
> `excludedBy` was removed from the schema (I thought?) after `definedBy` was extended to understand arbitrary Boolean logic, including `not`. The way to specify this now is:
> 
> ```yaml
> definedBy:
>   allOf:
>     - not:
>         anyOf: [ Zcd, allOf: [C, D] ]
>     - Zcmp
> ```
> 
> @ayosher, I believe you made this change because there was an issue generating an extension spec. I'll see if I can catch up and figure this out.

I see it in schema, trying to add any "not" causing error. Maybe it was an intention to remove excludedBy, but currently on main it is there and "not" does not work

dhower-qc:
Ok, I see that its removal happened on my `cpp_hart` branch, which isn't merged in. `not` should still work, though... I'll check out this PR and see what I can find.

dhower-qc:
grrr. 'not' and all the backend logic to use it is also on the `cpp_hart` branch. That branch needs to go through a painful merge with @james-ball-qualcomm's patches before it can go on main. @ayosher , I recommend we rebase your patches on cpp_hart so we can get the docs generated. I will make sure cpp_hart is up to date with main, and then you and I can sync.

ayosher:
> grrr. 'not' and all the backend logic to use it is also on the `cpp_hart` branch. That branch needs to go through a painful merge with @james-ball-qualcomm's patches before it can go on main. @ayosher , I recommend we rebase your patches on cpp_hart so we can get the docs generated. I will make sure cpp_hart is up to date with main, and then you and I can sync.

Maybe you can get me in main first, so we can generate what we need. Once you branch pass all painful merges, I will add another change to port everything to the new thing in another branch under your cpp_hart.

dhower-qc (approved):

PR #449, opened 2025-01-30
RISC-V Opcodes Data Cross-Validation

AFOliveira:
I integrated @BrianAnakPintar with normal UDB directory for this. I also added the generation of riscv-opcodes outputs, such as encodings.out.h that is what @ayosher was looking for if I am not in error. I added a README.md for detailed instructions.

@ayosher Until next meeting I won't have enough bandwidth to test this with SPIKE and Qualcomm new extensions, but to do so, you should just go to the folder `/ext/opcodes_maker` and run `make YAML_DIR=../../cfgs/qc_iu/arch_overlay/inst/` and look for `output/encoding.out.h` as an input for SPIKE.


ThinkOpenly:
Before you merge, I ask that you change:
> Integrate Brain branch ...
to
"Integrate Brian branch ..."

Although, @BrianAnakPintar is a really smart guy. ;-)

AFOliveira:
@ThinkOpenly Ups! I'll do so :)

dhower-qc (commented):
- Can we move this to `backends/` rather than `ext/`?
- Did you mean to include `instr_dict.json`? If so, we should figure out how to make sure the checked-in version is always consistent with the state of the database.

AFOliveira:
1. Sure!
2. If I did it was by accident. Default Path takes us to arch/inst and configurable user path can use whatever folder, like cfgs/qc_ui/inst

dhower-qc:
Cool. Now let's just make sure this gets added to the smoke tests. Currently, that takes two changes:

- Update test:smoke task in Rakefile (for local runs)
- Update .github/workflows/regress.yml (for GitHub runs)

Now that you've told me there is a tool to run GitHub workflows locally, we should be able to get rid of step 1 eventually...

AFOliveira:
@dhower-qc This is only generating riscv-opcodes outputs, therefore, shouldn't this be a separate task and not be inside the smoke test? This does not test anything, only generates. Probably I could put it into some `gen:opcodes_outputs `task?

dhower-qc:
> @dhower-qc This is only generating riscv-opcodes outputs, therefore, shouldn't this be a separate task and not be inside the smoke test? This does not test anything, only generates. Probably I could put it into some `gen:opcodes_outputs `task?

Sorry, I meant test:regress 😄

ayosher:
Should I expect down the road this generation to be integrated in some "do" task?

AFOliveira:
> Should I expect down the road this generation to be integrated in some "do" task?

That is what I was discussing with Derek. I would say yes, this is an output as many others are; however, I'm not entitled to answer it with certainty by myself.

ayosher:
I tried to generate opcodes for qc_iu/Xqci, faced the following error message:
`Successfully processed 157 YAML files
Output written to: output/instr_dict.json
python3 generator.py output/instr_dict.json -c -chisel -spinalhdl -sverilog -rust -go -latex
Traceback (most recent call last):
  File "/quic/riscv-unified-db/backends/opcodes_maker/generator.py", line 33, in <module>
    from c_utils import make_c
ModuleNotFoundError: No module named 'c_utils'
make: *** [Makefile:38: generate] Error 1`

dhower-qc:
absolutely. we should have something like `do gen:opcode_header`.

AFOliveira:
@ayosher riscv-opcodes was not updated in remote branch. My bad. There should not be any errors now.

AFOliveira:
Updated Rakefile. Example usage:
 ```
./do gen:opcode_outputs YAML_DIR=cfgs/qc_iu/arch_overlay/inst/
```

YAML_DIR is optional and can be left blank to default to normal arch/inst. I'll add to regression test later.

**Comment by ayosher** [comment]
> Updated Rakefile. Example usage:
> 
> ```
> ./do gen:opcode_outputs YAML_DIR=cfgs/qc_iu/arch_overlay/inst/
> ```
> 
> YAML_DIR is optional and can be left blank to default to normal arch/inst. I'll add to regression test later.

Tried that. I still have the same issue. Now with ./do way.
I believe I lack some package somewhere - but I have no idea which one...


Successfully processed 157 YAML files
Output written to: /quic/riscv-unified-db/gen/opcodes_outputs/instr_dict.json
/quic/riscv-unified-db/.home/.venv/bin/python3 /quic/riscv-unified-db/backends/opcodes_maker/generator.py /quic/riscv-unified-db/gen/opcodes_outputs/instr_dict.json -c -chisel -spinalhdl -sverilog -rust -go -latex
Traceback (most recent call last):
  File "/quic/riscv-unified-db/backends/opcodes_maker/generator.py", line 33, in <module>
    from c_utils import make_c
ModuleNotFoundError: No module named 'c_utils'
/quic/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/rake-13.2.1/lib/rake/file_utils.rb:65:in `block in create_shell_runner': Command failed with status (1): [/quic/riscv-unified-db/.home/.venv/bin/python3 /quic/riscv-unified-db/backends/opcodes_maker/generator.py /quic/riscv-unified-db/gen/opcodes_outputs/instr_dict.json -c -chisel -spinalhdl -sverilog -rust -go -latex] (RuntimeError)
        from /quic/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/rake-13.2.1/lib/rake/file_utils.rb:57:in `sh'
        from /quic/riscv-unified-db/Rakefile:49:in `block (2 levels) in <top (required)>'
        from /quic/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/rake-13.2.1/lib/rake/task.rb:281:in `block in execute'
        from /quic/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/rake-13.2.1/lib/rake/task.rb:281:in `each'
        from /quic/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/rake-13.2.1/lib/rake/task.rb:281:in `execute'
        from /quic/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/rake-13.2.1/lib/rake/task.rb:219:in `block in invoke_with_call_chain'
        from /quic/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/rake-13.2.1/lib/rake/task.rb:199:in `synchronize'
        from /quic/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/rake-13.2.1/lib/rake/task.rb:199:in `invoke_with_call_chain'
        from /quic/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/rake-13.2.1/lib/rake/task.rb:188:in `invoke'
        from /quic/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/rake-13.2.1/lib/rake/application.rb:188:in `invoke_task'
        from /quic/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/rake-13.2.1/lib/rake/application.rb:138:in `block (2 levels) in top_level'
        from /quic/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/rake-13.2.1/lib/rake/application.rb:138:in `each'
        from /quic/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/rake-13.2.1/lib/rake/application.rb:138:in `block in top_level'
        from /quic/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/rake-13.2.1/lib/rake/application.rb:147:in `run_with_threads'
        from /quic/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/rake-13.2.1/lib/rake/application.rb:132:in `top_level'
        from -e:1:in `<main>'

**Comment by AFOliveira** [comment]
@ayosher sorry for the delay.

> Tried that. I still have the same issue. Now with ./do way. I believe I lack some package somewhere - but I have no idea which one...

I still think its a riscv-opcodes problem, because it seems you can not import a file that did not exist in previous version, please try 

```
git submodule update --init --recursive
git submodule update --remote ext/riscv-opcodes
```


Still, if it does appear the same error, try `pip install pyyaml` since it the only non-standard python package in the repo.

If it still does not work, please ping me again and so I can take a deeper look trying to figure it out.

**Comment by ayosher** [comment]
@AFOliveira  - now it worked, thank you!

**Comment by AFOliveira** [comment]
> @AFOliveira - now it worked, thank you!

Great!

**Review by ayosher — APPROVED** [review]

**Comment by ThinkOpenly** [comment]
Sorry, peeking ahead...

Isn't `fieldo.js` more of an artifact that we should be generating? Or, maybe we don't have this information anywhere yet... hmm.

And, it's javascript, which is a new data representation for this project. Shall we convert it to JSON or YAML before incorporating? I see your use of it adds a requirement for Node.js.

It would be nice to simplify things if possible.

Do we need the field info for this PR, or can we push it to another PR?

**Comment by dhower-qc** [comment]
I agree with @ThinkOpenly; I think the goal is to cherry-pick the data in fieldo incorporated into the YAML schema. 

BTW, relevant to this PR: I stumbled across this repository this morning, which seems to have a plethora of good data in it (see the json folder):  https://github.com/sparcians/mavis. Note they classify instruction formats, but it's called "form" in their schema.

**Comment by AFOliveira** [comment]
@dhower-qc @ThinkOpenly We can definitely convert that .js into .json , .yaml, or even .csv. The problem is not with risc-v unified database data, is is just that the standard definition of information in riscv-opcodes is not super descriptive, it only ties imm name to position, but reality is that several different immediates have the same position, while they specify different information (like left_shift). The fieldo.js @drom did, simply clarifies all data in the riscv-opcodes repository, allowing for easier convertion. 

But this is just a fix for now; the reason we need to convert to riscv-opcodes middle-end is that we can rely on their scripts to generate the outputs. Down the line, we can tweak all those scripts to work with UDB data, but doing it now would only add "tech debt" into the project.

**Comment by AFOliveira** [comment]
> I was under the impression that the scope of this PR was roughly complete about 2 weeks ago, when you added the `README`. Is the `fieldo` data required in this PR, or can it be part of a follow-on PR? What problem does adding that data solve? Maybe I'm thinking more about "scope creep" for this PR versus tech debt.
> 
> There are a lot of hacky things in `yaml_to_json.py`. Are those just "for now" as well?
> 
> (I may be missing the bigger picture.)

The implementation was mostly complete, but handling immediate fields remained unfinished to a significant extent. The riscv-opcodes repository uses a variety of names to describe immediates, each carrying a specific meaning. However, there is no centralized database explaining these meanings. As a result, translating from UDB to riscv-opcodes isn’t straightforward and requires a reference—a kind of "cheat sheet" for immediates—which is precisely what fieldo.js provides.

While I could manually integrate all this into my script using "hacky" workarounds, I opted to reuse the existing fieldo.js to simplify the process. Our goal is to generate intermediate representation that aligns with riscv-opcodes, and since riscv-opcodes itself contains numerous "hacky" elements, we can't completely avoid them—we have to conform to some extent to maintain compatibility.

That said, I still need to review all outputs whenever I get the time to ensure correctness, which may lead to future refactoring. If you're strongly opposed to using fieldo, please be aware that hardcoding this information somewhere seems unavoidable.

**Comment by ThinkOpenly** [comment]
What's "c_utils" ?
```
$ make
python3 generator.py output/instr_dict.json -c -chisel -spinalhdl -sverilog -rust -go -latex
Traceback (most recent call last):
  File "/home/pc/projects/riscv/riscv-unified-db/riscv-unified-db/backends/opcodes_maker/generator.py", line 33, in <module>
    from c_utils import make_c
ModuleNotFoundError: No module named 'c_utils'
```

**Comment by AFOliveira** [comment]
> What's "c_utils" ?
> 
> ```
> $ make
> python3 generator.py output/instr_dict.json -c -chisel -spinalhdl -sverilog -rust -go -latex
> Traceback (most recent call last):
>   File "/home/pc/projects/riscv/riscv-unified-db/riscv-unified-db/backends/opcodes_maker/generator.py", line 33, in <module>
>     from c_utils import make_c
> ModuleNotFoundError: No module named 'c_utils'
> ```

That is the same problem @ayosher was getting a while back in this thread. You need to get most recent riscv-opcode submodule to run this code.

`git submodule update --init --recursive
git submodule update --remote ext/riscv-opcodes`

**Comment by ThinkOpenly** [comment]
> That is the same problem @ayosher was getting a while back in this thread. You need to get most recent riscv-opcode submodule to run this code.

Ooh. I totally missed that. Sorry for the noise.

Should we add something in the `README`?

**Comment by AFOliveira** [comment]
Don't worry! Yes, we surely can! Or even better, I can likely code a more precise error message for that, so it is obvious how to solve it

**Comment by jordancarlin** [comment]
It should also be possible for the script (Makefile in this case?) to fetch the submodule if it isn't already cloned. Several other projects do that (like riscv-gnu-toolchain).

File: backends/opcodes_maker/yaml_to_json.py [hunk 1]
```diff
@@ -0,0 +1,337 @@
...
+    try:
+        for inst_dir in insts:
+            try:
+                convert(inst_dir, inst_dict)
+            except Exception as e:
+                print(f"Warning: Failed to process {inst_dir}: {str(e)}")
+                continue
+
+        with open(output_file, "w") as outfile:
+            json.dump(inst_dict, outfile, indent=4)
```
ThinkOpenly, line 327: I'm not sure how important this is, but it appears that riscv-opcodes emits JSON with the keys sorted. Something like this does that trick for me:
```
insts_sorted = {}
for inst in sorted(inst_dict):
  insts_sorted[inst] = inst_dict[inst]

with open(output_file, "w") as outfile:
  json.dump(insts_sorted, outfile, indent=4)
```
This made the respective outputs much easier to compare.

ThinkOpenly:
...and now I see that you are trying to get changes like these:
```
     "variable_fields": [
       "rd",
       "rs1",
-      "imm"
+      "imm12"
     ],
```
```
     "variable_fields": [
       "rd",
-      "imm"
+      "imm20"
     ],
```
```
     "variable_fields": [
+      "bimm12hi",
       "rs1",
       "rs2",
-      "imm"
+      "bimm12lo"
     ],
```
...by adding `fieldo.js`.

So, the data in `fieldo.js` is needed to generate the riscv-opcodes files correctly, unless and until the same data is represented in UDB.

My preference would be to convert `fieldo.js` to another representation which doesn't require using `node` to convert it, like JSON, out-of-line, and then add that converted file to your commit. Add a rule or a script to do the conversion for when `fieldo.js` gets updated, so _we_ can add the updated `fieldo.js` and then update that converted file easily, if that makes sense.

In this way, other users will not need to resolve that dependence on `node`.

How does that sound?

AFOliveira:
> ...and now I see that you are trying to get changes like these:
> 
> ```
>      "variable_fields": [
>        "rd",
>        "rs1",
> -      "imm"
> +      "imm12"
>      ],
> ```
> 
> ```
>      "variable_fields": [
>        "rd",
> -      "imm"
> +      "imm20"
>      ],
> ```
> 
> ```
>      "variable_fields": [
> +      "bimm12hi",
>        "rs1",
>        "rs2",
> -      "imm"
> +      "bimm12lo"
>      ],
> ```
> 
> ...by adding `fieldo.js`.
> 
> So, the data in `fieldo.js` is needed to generate the riscv-opcodes files correctly, unless and until the same data is represented in UDB.
> 
> My preference would be to convert `fieldo.js` to another representation which doesn't require using `node` to convert it, like JSON, out-of-line, and then add that converted file to your commit. Add a rule or a script to do the conversion for when `fieldo.js` gets updated, so _we_ can add the updated `fieldo.js` and then update that converted file easily, if that makes sense.
> 
> In this way, other users will not need to resolve that dependence on `node`.
> 
> How does that sound?

Paul, thank you for taking the time and figuring it out by yourself, I didn't really express myself the best. Moreover, that's completely fine for me,  of course. I'll checkout converted fieldo.js in a form of .json and find a way to update it in the future.

backends/opcodes_maker/yaml_to_json.py
AFOliveira: I'll also add that, thank you!

AFOliveira:
> It should also be possible for the script (Makefile in this case?) to fetch the submodule if it isn't already cloned. Several other projects do that (like riscv-gnu-toolchain).

The problem here is, if I understand that riscv-opcodes is already fetched, but not on the most recent version, because prior to this PR UDB had it, but it was pointing to some other commit. Anyways, that is a great thing to add to the Makefile, I'm just afraid it won't be enough, thus my inclination to choose doing an apropriate error message.

jordancarlin:
> > It should also be possible for the script (Makefile in this case?) to fetch the submodule if it isn't already cloned. Several other projects do that (like riscv-gnu-toolchain).
> 
> The problem here is, if I understand that riscv-opcodes is already fetched, but not on the most recent version, because prior to this PR UDB had it, but it was pointing to some other commit. Anyways, that is a great thing to add to the Makefile, I'm just afraid it won't be enough, thus my inclination to choose doing an apropriate error message.

Hmm. That is tricker. Maybe it can do a `git submodule update ext/riscv-opcodes --init` automatically so that it checks out the appropriate version of the submodule whether or not it has already been cloned.

backends/opcodes_maker/yaml_to_json.py
```diff
@@ -0,0 +1,342 @@
...
+            else:
+                print(
+                    f"Warning: Skipping immediate field {var_name} with location {location} since no fieldo mapping was found."
+                )
+        else:
+            result.append(var_name)
+    return result
+
+
+def BitStringToHex(bit_str: str) -> str:
```
ThinkOpenly, line 172: I propose this be renamed to `GetMatch` (or `GetMatchMask`?). It's not really converting a bit string to hexadecimal, since it treats "don't care" bits as "0". "10-10-10-10-" would become "924", for example.

backends/opcodes_maker/yaml_to_json.py
```diff
@@ -0,0 +1,342 @@
...
+            if "RV64" in encodings:
+                encodings = encodings["RV64"]
+                rv64_flag = True
+            enc_match = GetEncodings(encodings["match"])
+
+            var_names = []
+            if "variables" in encodings:
+                var_names = GetVariables(encodings["variables"])
+
+            extension = []
```
ThinkOpenly, line 260: This would make more sense to me as a plural, "extensions", since it's a list of extensions.

Same for the function "GetExtension", preferably "GetExtensions".

backends/opcodes_maker/yaml_to_json.py
```diff
@@ -0,0 +1,342 @@
...
+    if n < 32:
+        return "-" * (32 - n) + enc
+    return enc
+
+
+def convert(file_dir: str, json_out: Dict[str, Any]) -> None:
+    try:
+        with open(file_dir) as file:
+            data = yaml.safe_load(file)
+            instr_name = data["name"].replace(".", "_")
```
ThinkOpenly, line 244: Should we validate that the "kind" of YAML is "instruction" here? It'll probably trip over some of the other field accesses, but a check could avoid a false positive. Just a thought.

backends/opcodes_maker/yaml_to_json.py
```diff
@@ -0,0 +1,342 @@
...
+                "extension": extension,
+                "match": match_hex,
+                "mask": match_mask,
+            }
+    except Exception as e:
+        print(f"Error processing file {file_dir}: {str(e)}")
+        raise
+
+
+def read_yaml_insts(path: str) -> List[str]:
```
ThinkOpenly, line 290: This is more like "find_yaml_files". It's not reading any files.

backends/opcodes_maker/yaml_to_json.py
```diff
@@ -0,0 +1,342 @@
...
+
+    insts = read_yaml_insts(args.input_dir)
+    if not insts:
+        parser.error(f"No YAML files found in {args.input_dir}")
+
+    inst_dict = {}
+    output_file = os.path.join(args.output_dir, "instr_dict.json")
+
+    try:
+        for inst_dir in insts:
```
ThinkOpenly, line 320: "inst_dir" is more like "yaml_file"

backends/opcodes_maker/yaml_to_json.py
```diff
@@ -0,0 +1,342 @@
...
+    parser.add_argument("input_dir", help="Directory containing YAML instruction files")
+    parser.add_argument("output_dir", help="Output directory for generated files")
+
+    args = parser.parse_args()
+
+    # Ensure input directory exists
+    if not os.path.isdir(args.input_dir):
+        parser.error(f"Input directory does not exist: {args.input_dir}")
+
+    insts = read_yaml_insts(args.input_dir)
```
ThinkOpenly, line 312: "insts" is more like "yamls" or "yaml_files"

backends/opcodes_maker/fieldo.json
ThinkOpenly, line 1: Do we also want the `fieldo.js` and the script to generate `fieldo.json` from `fieldo.js`? Having the former would allow us to "diff" the two to know if there are any changes that need to be incorporated here.

ThinkOpenly (changes requested):

.pre-commit-config.yaml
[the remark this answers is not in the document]
AFOliveira: There are still some loose ends on this PR, this is one of them. Thanks for the reminder! I'll get it all ready before getting out of draft stage!

backends/opcodes_maker/sorter.py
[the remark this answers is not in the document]
AFOliveira: It was a Brian file to do exactly what your previous comment with sorting did. Thus, I'll also remove it.

backends/opcodes_maker/yaml_to_json.py
AFOliveira: I agree. I'll change it.

backends/opcodes_maker/yaml_to_json.py
AFOliveira: I am assuming that we only point at folders that are instruction-only, but checking that definitely can't hurt. I'll add it.

backends/opcodes_maker/yaml_to_json.py
AFOliveira: Done!

backends/opcodes_maker/fieldo.json
AFOliveira: I don't see a reason for this, given that I simply picked up the conversion that was already going on and checked the .json I was creating into UDB. If you want, I may put them in, but they also won't stay if we want to merge this PR into main, right?

backends/opcodes_maker/yaml_to_json.py
AFOliveira: Done!

backends/opcodes_maker/fieldo.json
ThinkOpenly: > I simply picked up the conversion that was already going on

I'm confused. What is the canonical source for the information in the new JSON file? And, what is the process for maintaining that information?

I was assuming that we need to maintain that information, including updates, based on a "canonical" source JavaScript file being maintained externally, and a conversion script to be maintained here. If so, then we need some easy way to do so, and I was suggesting that we pull a (read-only) version of the JavaScript file here to facilitate answering "has it changed?" (by comparing it with the upstream/external source), so we can re-run the conversion script and update the JSON file. (This could even be automated, I think.)

Were you anticipating a different process? Or, maybe this is "one and done", and we expect to maintain the JSON here going forward? (I'm not necessarily against that, but worry about it being entirely manual, and error-prone. Unless the external source is automatically generated somehow, and we then need to pull that generation process here as well?)

AFOliveira: Oh, ok, now I remember what you are talking about. We did talk about that pulling and converting; however, that source is just written by hand by @drom. Also, I do not think that it changes that often that we will be needing to create a "pull and compare" method. RISC-V Opcodes is not that volatile, and we may just manually update when needed. I think it will be way easier than to bring that file in, compare and use.

ThinkOpenly: One of the high-level goals is to have a "single source of truth" (as you know, of course). Is the expectation that each of these parameter maps will be maintained independently? (Can @drom use the source here as the input to that project?) While the data is expected to remain fairly constant, I still worry about drift. Maybe this isn't an issue we try to resolve today (or ever).

I'm hearing that this will be "one and done": capture the data as it stands today from @drom, and move on with our own copy. This is probably fine. ("Perfection" being the enemy of the "good enough".)

ThinkOpenly: Hmm. So, where is the information going to be maintained, long term? Or, are you saying we won't need it at all long term? The information basically attaches semantics (size, type, position, interpretation) to variable fields by associating specific names with specific use-cases. Is the idea to eventually use the information already in UDB (or still to be added?) to generate the same information?

AFOliveira: The only thing I am using that .json for is to know how to map from UDB immediates to riscv-opcodes immediate - riscv-opcodes uses weird names to describe immediates fields. However, we already have the immediate information in the UDB, we just don't generate the same outputs that riscv-opcodes does from the UDB because converting into riscv-opcodes is only 1 script and generating those outputs is 5 or 6 scripts.

AFOliveira: So we only need this information until we convince users of riscv-opcodes to migrate to UDB - which I am fairly sure is happening faster than riscv-opcodes is being changed more than 2 or 3 times

ThinkOpenly: We may have to carry that information indefinitely, but that may be the cost of consolidation. So, I'm fine if you just want to include the JSON file, and we carry the burden of updating it if/as necessary. Like you said, that content will likely be pretty static.

AFOliveira:
FYI, this is close to working, but C extensions has shown to be a pain - there's like 20+ operands that need to be further defined by hand, so expect this to take a while to get to 100%.

dhower-qc:
@AFOliveira, how does this PR relate to #553 and #526 ?

AFOliveira:
> @AFOliveira, how does this PR relate to #553 and #526 ?

Well, unfortunately, it does not at all. I realized that this task was way to challenging for a "sole" reason: RISC-V opcodes variable naming is some sort of black magic that is almost impossible to solve. This PR serves to show that we are in a comparable state in terms of instructions to what riscv-opcodes has, but I don't think we should integrate it into any workflow, since that there is no clean way reproducing that .json and leaves this code as a really nice spaghetti mess that we don't want to deal with in the future

dhower-qc:
Ok, should we close this then, or is there a value to keeping it open for the time being?

AFOliveira:
Closing it for now

PR #451, opened 2025-02-01
Removing unratified RV128 instructions

### LINKED ISSUE #446: MC100-32 has 128-bit instructions like c.lq and c.sq

dhower-qc:
fixes #446


ThinkOpenly (approved):
Seems fine. If RV128/C was approved, do we have the ability to conditionally include these instructions?

AFOliveira (approved):

AFOliveira:
> Seems fine. If RV128/C was approved, do we have the ability to conditionally include these instructions?
The UDB instruction schema has a field `base: ` which usually until now only defined 32 or 64 when needed, I believe we could follow the same approach for this.

dhower-qc:
Yes, in theory the schema supports RV128. The backend is another story...for better or worse, there are a ton of places we assume 64 is the largest XLEN

Issue #458, opened 2025-02-11
UDB json index


AFOliveira:
Last meeting, @drom requested a checked out index of the UDB in the repo, I have a similar question/idea. Can we turn the index configuration specfic? If we made an index of specific configuration in json, tools could also use it modularly, including @drom tool. Obviously we don't need to check out all indexes.

dhower-qc:
I'll be pushing a PR soon that adds an index during the resolution step. That means every configuration that you resolve will get an index (located in `gen/resolved_arch/<cfg name>/index.yaml`). I'm adding the resolved "unconfig" ("_") to the GitHub deployment, so there will always be an index reflecting `main` at https://riscv-software-src.github.io/riscv-unified-db/resolved_arch/index.yaml.

james-ball-qualcomm:
Can we choose a different name than "_" to mean unconfigured? That isn't obvious to new people what it means.

dhower-qc:
suggestion?

james-ball-qualcomm:
How about "complete_isa"?

james-ball-qualcomm:
Or "complete_arch"

drom:
I need just `index.json` "non resolved" version of index.

dhower-qc:
how about "rv"? We already have "rv32" and "rv64"

Issue #199, opened 2024-10-25
index file


drom:
Do we have the index yaml file that would allow fetching all the rest of yaml files?

dhower-qc:
there is no index; it just assumes every .yaml file under `arch` is included (type is implied by whether it exists in the `inst`, `ext`, etc., subfolder).

You can generate a single file with everything in it with `./do gen:arch` (will appear in `gen/_/arch/arch_def.yaml`

drom:
I am fetching it from the web page, and I can't do `ls internet`. I need to know list of the files and folders.

dhower-qc:
ah, interesting. we could generate the unified description with each push on main (similar to how we keep https://riscv-software-src.github.io/riscv-unified-db/manual/html/landing/index.html updated). Would that work?

drom:
There is no problem having 1k of yaml files, but if we can just regenerate index.yaml file and add it to the same commit. It would help. If it is uses JSON references then it would simplify things:

```json5
{
  arcs: {
    inst: {
      A: {
        'amoadd.d': {'$ref': '#/arcs/inst/A/amoadd.d.yaml'},
        ...

      }
    }
  }
}
```

drom:
I can make quick PoC in JS

drom:
#201

drom:
with index.json I was able to fetch UnifiedDB from GitHub using web client. Here is example:
https://observablehq.com/@drom/risc-v-unifieddb

PR #459, opened 2025-02-11
Add landing for GitHub pages, add index to resolved arch

dhower-qc:
With this, an index of the architecture files for `main` will be generated with CI at https://riscv-software-src.github.io/riscv-unified-db/resolved_arch/index.yaml (and the paths in the index are relative to https://riscv-software-src.github.io/riscv-unified-db/resolved_arch).


dhower-qc:
closes #458

dhower-qc:
closes #199

lib/yaml_resolver.py
```diff
@@ -597,15 +597,19 @@ def resolve_file(
...
-                else f"{args.resolved_dir}/{arch_path}"
-            )
+            resolved_arch_path = f"{abs_resolved_dir}/{arch_path}"
             os.makedirs(os.path.dirname(resolved_arch_path), exist_ok=True)
             resolve_file(
                 arch_path, args.arch_dir, args.resolved_dir, not args.no_checks
             )
 
+        # create index
+        write_yaml(f"{abs_resolved_dir}/index.yaml", arch_paths)
```
AFOliveira, line 613: Pardon me if I'm blatantly missing the point, but are we going to have yaml index instead of .json?

dhower-qc: ha, missed that part. I'll create a JSON file, too.

AFOliveira (approved):
LGTM!

drom:
I need just index.json "non resolved" version of index.
The index should allow fetching of ALL required parts and iterative resolution at later time.

dhower-qc:
In this case, the "non resolved" and "resolved" indices are the same. Resolution of the standard doesn't add or remove any files, and since the index holds relative paths you could use it to search arch/ as well.

Keep in mind that "resolved" means (a) merged with overlay (a no-op for the RVI standard) and (b) $inherits expansion. It does not expand $ref, so I think for your use cases you could use the "resolved" files generated at deployment.

PR #460, opened 2025-02-11
fix(inst): make fence.tso a separate inst, fix fence bug

### LINKED ISSUE #428: Fence.tso is not pseudo for Fence
Currently UDB specified fence.tso as pseudo-instructions under [fence.yaml](https://github.com/riscv-software-src/riscv-unified-db/blob/d791ca4bb7c1938de313642f7dc0f04fde9d6769/arch/inst/I/fence.yaml#L206). However, this is not the case and fence.tso should have a separate file for it.

I believe this is the specified like this because of riscv-opcodes, but the only reason riscv-opcodes specified it like that is it not possible to support for both on current riscv-opcodes format.

Context:
https://github.com/riscv/riscv-opcodes/pull/330
https://github.com/riscv/riscv-isa-manual/issues/1782#issuecomment-2556036171

dhower-qc:
fixes #428


arch/inst/I/fence.tso.yaml
```diff
@@ -0,0 +1,42 @@
+# yaml-language-server: $schema=../../../schemas/inst_schema.json
+
+$schema: "inst_schema.json#"
+kind: instruction
+name: fence
```
ThinkOpenly: `fence.tso`

arch/inst/I/fence.tso.yaml
```diff
@@ -0,0 +1,42 @@
...
+long_name: Memory ordering fence
```
ThinkOpenly, line 6: (Looking for a different long_name than for FENCE...)
"Memory ordering fence, total store ordering"?

arch/inst/I/fence.tso.yaml
```diff
@@ -0,0 +1,42 @@
...
+description: |
```
ThinkOpenly, line 7: Do we want to use `>` here?

arch/inst/I/fence.tso.yaml
```diff
@@ -0,0 +1,42 @@
...
+  in its predecessor set before all store operations in its successor set. This leaves non-AMO store
+  operations in the 'fence.tso's predecessor set unordered with non-AMO loads in its successor set.
+
+  The `rs1` and `rd` fields are unused and ignored.
+
+  In modes other than M-mode, `fence.tso` is further affected by `menvcfg.FIOM`,
+  `senvcfg.FIOM`<% if ext?(:H) %>, and/or `henvcfg.FIOM`<% end %>.
+
+definedBy: I
+assembly: "TODO"
```
ThinkOpenly, line 21: `assembly: ""`

arch/inst/I/fence.tso.yaml
```diff
@@ -0,0 +1,42 @@
...
+access:
+  s: always
+  u: always
+  vs: always
+  vu: always
+operation(): |
+
+  fence_tso();
+
+sail(): ""
```
ThinkOpenly, line 42:
```
sail(): |
  {
    match (pred, succ) {
      (_ : bits(2) @ 0b11, _ : bits(2) @ 0b11) => sail_barrier(Barrier_RISCV_tso),
      (_ : bits(2) @ 0b00, _ : bits(2) @ 0b00) => (),

      _ => { print("FIXME: unsupported fence");
             () }
    };
    RETIRE_SUCCESS
  }

**Review by ThinkOpenly — CHANGES_REQUESTED** [review]

File: arch/inst/I/fence.tso.yaml (hunk 3 above)
Comment (dhower-qc) [reply arch/inst/I/fence.tso.yaml:7]: I don't think so; the line spacing has syntactic meaning for asciidoc, so we don't want to fold anything.

Comment (ThinkOpenly) [reply arch/inst/I/fence.tso.yaml:7]: My understanding of the folding rules is that the text below this would be folded to (hopefully this renders well enough here):
> Orders memory operations.
> 
> `fence.tso` orders all load operations in its predecessor set before all memory operations in its successor set, and all store operations in its predecessor set before all store operations in its successor set. This leaves non-AMO store  operations in the 'fence.tso's predecessor set unordered with non-AMO loads in its successor set.
> 
> The `rs1` and `rd` fields are unused and ignored.
> 
> In modes other than M-mode, `fence.tso` is further affected by `menvcfg.FIOM`,  `senvcfg.FIOM`<% if ext?(:H) %>, and/or `henvcfg.FIOM`<% end %>.

Is that not preferred?

Comment (dhower-qc) [reply arch/inst/I/fence.tso.yaml:7]: I think it actually renders as:

> Orders memory operations.
> `fence.tso` orders all load operations in its predecessor set before all memory operations in its successor set, and all store operations in its predecessor set before all store operations in its successor set. This leaves non-AMO store  operations in the 'fence.tso's predecessor set unordered with non-AMO loads in its successor set.
> The `rs1` and `rd` fields are unused and ignored.
> In modes other than M-mode, `fence.tso` is further affected by `menvcfg.FIOM`,  `senvcfg.FIOM`<% if ext?(:H) %>, and/or `henvcfg.FIOM`<% end %>.

image


Which in asciidoc will make all the text show up in a single paragraph.

**Review by AFOliveira — APPROVED** [review]

**Comment by AFOliveira** [comment]
This likely could be merged as is without any problems, right?

**Comment by dhower-qc** [comment]
Yes, it just needs to go in after cpp_hart

**Review by ThinkOpenly — APPROVED** [review]

**Comment by AFOliveira** [comment]
is this something that was forgotten?

**Comment by dhower-qc** [comment]
Yes, it was forgotten. I have it ready again, but it's failing because of a mismatch with riscv-opcodes.

**Comment by AFOliveira** [comment]
> Yes, it was forgotten. I have it ready again, but it's failing because of a mismatch with riscv-opcodes.

It's a mismatch with LLVM. This is supposed to happen, since our view of the information is on the ISA side, which allows several arguments (in hardware) for a fence, while the pragmatic implementation of fence in a compiler is a full constant expression, since it is the only one that the ISA clearly defines. I will make a patch so the test skips fence.

This also brings me to the question that in the future, "variable" could have a separation between "hardware-only" and "assembly". Since they not always match. This is a great effort and I think should not a be a priority for now.

**Comment by ThinkOpenly** [comment]
> It's a mismatch with LLVM. This is supposed to happen, since our view of the information is on the ISA side, which allows several arguments (in hardware) for a fence, while the pragmatic implementation of fence in a compiler is a full constant expression, since it is the only one that the ISA clearly defines. I will make a patch so the test skips fence.
> 
> This also brings me to the question that in the future, "variable" could have a separation between "hardware-only" and "assembly". Since they not always match.

I'm not following completely. Is this an artifact of having "variables" (`rd`, `rs1`) that aren't actually variables? Does that imply that the definition of `fence.tso` should not show these fields as variables? The spec says:
```
The unused fields in the FENCE instructions--rs1 and rd--are reserved for finer-grain fences in future
extensions. For forward compatibility, base implementations shall ignore these fields, and standard
software shall zero these fields.
```

Given "shall", shall we also restrict the valid values here to only zero?
```
not: 1, 2, 3, [...], 31
```

> This is a great effort and I think should not a be a priority for now.

Shall we open an issue?

dhower-qc:
> Given "shall", shall we also restrict the valid values here to only zero?

The "shall" is that the fields should be ignored, not that they are illegal. If we use `not`, then any non-zero value will be considered an illegal instruction.

AFOliveira:
My view on it is that the spec leaves open to later implementations to move fence around and switch any of those values. However, from the compiler view what I suppose happened is that "shall" is enough for them to assume that every bit is 0, and if someone does a different implementation, then whoever implements that is responsible for updating the compiler

AFOliveira:
> Shall we open an issue

Sure, I will.

AFOliveira:
Smoke test fixed in #687

Issue #209, opened 2024-10-30
Implement F/D extensions in IDL


dhower-qc:
This has already started; F is roughly 1/2 done.

The task is to take the functions that implement F/D instructions in [softfloat3](https://github.com/ucb-bar/berkeley-softfloat-3) and convert them to IDL. This is a good project to take on for someone wanting to learn IDL -- this is mostly just cut/paste, with some cleanup where IDL and C differ.

Shubhf:
Hello @dhower-qc 
Can I try with this issue ?

ShashankVM:
Hi @Shubhf I am already working on this issue and it is assigned to me, completed addition and subtraction. Please look for a different issue. Thanks.

PR #461, opened 2025-02-11
Update code for fcvt.s.w,  fcvt.s.wu, fadd and fsub

ShashankVM:
# fcvt.s.w
- Removed the inline code and replaced it with function call

# fcvt.s.wu
- Added instruction operation, long name and description fields for this file.

# fadd
- Added check_f_ok function call before doing the operation.

# fsub
- Added check_f_ok function call before doing the operation.

Progress towards #209


arch/inst/F/fcvt.s.w.yaml
```diff
@@ -34,20 +34,8 @@ access:
...
-  RoundingMode rounding_mode = rm_to_mode(rm, $encoding);
-
-  if ((int_value & 32'h7fff_ffff) == 0) {
-    X[fd] = (sign == 1) ? packToF32UI(1, 0x9E, 0) : 0;
-  } else {
-    Bits<32> absA = (sign == 1) ? -int_value : int_value;
-    X[fd] = softfloat_normRoundPackToF32( sign, 0x9C, absA, rounding_mode );
-  }
-
+  RoundingMode rounding_mode = rm_to_mode(X[rm], $encoding);
```
dhower-qc, line 37: shouldn't this stay as `rm` instead of `X[rm]`? You want the value of the rm field in the encoding, not the value in the register rm. (this applies to several other spots, too)

dhower-qc (changes requested):

arch/inst/F/fcvt.s.w.yaml
ShashankVM: Yes, will fix this Derek. Thank you for pointing this out.

ShashankVM: Requested changes completed in https://github.com/riscv-software-src/riscv-unified-db/pull/461/commits/f725f1bf31fa47c94b8efc6e99c1221567c81d2f

PR #467, opened 2025-02-27
Vector extension

jmawet:
Adding IDL support and instruction details for Vector Extension

2/27 work in progress, do not merge to main yet. PR is for review purposes only

### DIFF
--- a/arch/inst/V/vsetivli.yaml
+++ b/arch/inst/V/vsetivli.yaml
@@ -3,9 +3,11 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: vsetivli
-long_name: No synopsis available.
+long_name: Set vector configuration (immediate immediate).
 description: |
-  No description available.
+  Vset (i,i) allows rapid configuration of the values in vl and vtype CSRs to match application
+  needs. The vsetvl instruction sets the vtype and vl CSRs based on the arguments, and writes
+  the new value of vl into rd.
 definedBy: V
 assembly: xd, imm
 encoding:
@@ -24,7 +26,19 @@ access:
   vu: always
 data_independent_timing: false
 operation(): |
+  XReg state = vector_state();
+  XReg VLMAX = state.log2_multiplier * VLEN * state.sew;
+  # todo: state.lmul_type
 
+  if (xs1 < VLMAX){
+    CSR[vl] = uimm;
+  } else {
+    CSR[vl] = VLMAX;
+  }
+
+  CSR[vtype] = zimm11;
+  X[rd] = CSR[vl];
+  CSR[vstart] = 0;
 sail(): |
   {
     let VLEN_pow      = get_vlen_pow();
--- a/arch/inst/V/vsetvl.yaml
+++ b/arch/inst/V/vsetvl.yaml
@@ -3,9 +3,11 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: vsetvl
-long_name: No synopsis available.
+long_name: Set vector configuration (register register).
 description: |
-  No description available.
+  Vset (r,r) allows rapid configuration of the values in vl and vtype CSRs to match application
+  needs. The vsetvl instruction sets the vtype and vl CSRs based on the contents of the argument
+  registers rs1 and rs2, and writes the new value of vl into rd.
 definedBy: V
 assembly: xs2, xs1, xd
 encoding:
@@ -24,3 +26,85 @@ access:
   vu: always
 data_independent_timing: false
 operation(): |
+  XReg state = vector_state();
+  XReg VLMAX = state.log2_multiplier * VLEN * state.sew;
+  # todo: state.lmul_type
+
+  if (xs1 < VLMAX){
+    CSR[vl] = X[rs1];
+  } else {
+    CSR[vl] = VLMAX;
+  }
+
+  CSR[vtype] = X[rs2];
+  X[rd] = CSR[vl];
+  CSR[vstart] = 0;
+sail(): |
+  {
+    let VLEN_pow      = get_vlen_pow();
+    let ELEN_pow      = get_elen_pow();
+    let LMUL_pow_ori  = get_lmul_pow();
+    let SEW_pow_ori   = get_sew_pow();
+    let ratio_pow_ori = SEW_pow_ori - LMUL_pow_ori;
+
+    /* set vtype */
+    match op {
+      VSETVLI => {
+        vtype->bits() = 0b0 @ zeros(sizeof(xlen) - 9) @ ma @ ta @ sew @ lmul
+      },
+      VSETVL  => {
+        let rs2 : regidx = sew[1 .. 0] @ lmul;
+        vtype->bits() = X(rs2)
+      }
+    };
+
+    /* check legal SEW and LMUL and calculate VLMAX */
+    let LMUL_pow_new = get_lmul_pow();
+    let SEW_pow_new  = get_sew_pow();
+    if SEW_pow_new > LMUL_pow_new + ELEN_pow then {
+      /* Note: Implementations can set vill or trap if the vtype setting is not supported.
+       * TODO: configuration support for both solutions
+       */
+      vtype->bits() = 0b1 @ zeros(sizeof(xlen) - 1); /* set vtype.vill */
+      vl = zeros();
+      print_reg("CSR vtype <- " ^ BitStr(vtype.bits()));
+      print_reg("CSR vl <- " ^ BitStr(vl));
+      return RETIRE_SUCCESS
+    };
+    let VLMAX = int_power(2, VLEN_pow + LMUL_pow_new - SEW_pow_new);
+
+    /* set vl according to VLMAX and AVL */
+    if (rs1 != 0b00000) then { /* normal stripmining */
+      let rs1_val = X(rs1);
+      let AVL = unsigned(rs1_val);
+      vl = if AVL <= VLMAX then to_bits(sizeof(xlen), AVL)
+           else if AVL < 2 * VLMAX then to_bits(sizeof(xlen), (AVL + 1) / 2)
+           else to_bits(sizeof(xlen), VLMAX);
+      /* Note: ceil(AVL / 2) <= vl <= VLMAX when VLMAX < AVL < (2 * VLMAX)
+       * TODO: configuration support for either using ceil(AVL / 2) or VLMAX
+       */
+      X(rd) = vl;
+    } else if (rd != 0b00000) then { /* set vl to VLMAX */
+      let AVL = unsigned(ones(sizeof(xlen)));
+      vl = to_bits(sizeof(xlen), VLMAX);
+      X(rd) = vl;
+    } else { /* keep existing vl */
+      let AVL = unsigned(vl);
+      let ratio_pow_new = SEW_pow_new - LMUL_pow_new;
+      if (ratio_pow_new != ratio_pow_ori) then {
+        /* Note: Implementations can set vill or trap if the vtype setting is not supported.
+         * TODO: configuration support for both solutions
+         */
+        vtype->bits() = 0b1 @ zeros(sizeof(xlen) - 1); /* set vtype.vill */
+        vl = zeros();
+      }
+    };
+    print_reg("CSR vtype <- " ^ BitStr(vtype.bits()));
+    print_reg("CSR vl <- " ^ BitStr(vl));
+
+    /* reset vstart to 0 */
+    vstart = zeros();
+    print_reg("CSR vstart <- " ^ BitStr(vstart));
+
+    RETIRE_SUCCESS
+  }
--- a/arch/inst/V/vsetvli.yaml
+++ b/arch/inst/V/vsetvli.yaml
@@ -3,9 +3,11 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: vsetvli
-long_name: No synopsis available.
+long_name: Set vector configuration (register immediate).
 description: |
-  No description available.
+  Vset (r,i) allows rapid configuration of the values in vl and vtype CSRs to match application
+  needs. The vsetvl instruction sets the vtype and vl CSRs based on the contents of the argument
+  register rs1 and immediate, and writes the new value of vl into rd.
 definedBy: V
 assembly: xs1, xd, imm
 encoding:
@@ -24,7 +26,19 @@ access:
   vu: always
 data_independent_timing: false
 operation(): |
+  XReg state = vector_state();
+  XReg VLMAX = state.log2_multiplier * VLEN * state.sew;
+  # todo: state.lmul_type
 
+  if (xs1 < VLMAX){
+    CSR[vl] = X[rs1];
+  } else {
+    CSR[vl] = VLMAX;
+  }
+
+  CSR[vtype] = zimm11;
+  X[rd] = CSR[vl];
+  CSR[vstart] = 0;
 sail(): |
   {
     let VLEN_pow      = get_vlen_pow();
--- a/arch/isa/vec.idl
+++ b/arch/isa/vec.idl
@@ -1,0 +1,50 @@
+%version: 1.0
+
+# - test suite?
+# - Vector params (there will be many)
+# - Vector state
+# - Order (might be dictated somewhat by test suite)
+#   - vset*
+#   - integer arith (vadd/vsub/vrsub(.vv,.vx,.vi), compare, min, max)
+#   - integer widening arith ()
+
+# the vector register file
+Bits<VLEN> v[32] = [0, 0, 0, 0, 0, 0, 0, 0,
+                    0, 0, 0, 0, 0, 0, 0, 0,
+                    0, 0, 0, 0, 0, 0, 0, 0,
+                    0, 0, 0, 0, 0, 0, 0, 0];
+
+enum VectorLmulType {
+  Divide
+  Multiply
+}
+
+struct VectorState {
+  Bits<7> sew;
+  VectorMultiplierDirection lmul_type;
+  Bits<2> log2_multiplier;
+}
+
+function vector_state {
+  returns VectorState
+  description {
+    Get the current vector state from CSRs
+  }
+  body {
+    VectorState state;
+
+    state.sew = 7'b1 << (3 + CSR[VTYPE].VSEW);
+    Bits<3> vlmul = CSR[VTYPE].VLMUL;
+    state.lmul_type = CSR[vtype].VLMUL[2] == 1'b1 ? VectorLmulType::Divide : VectorLmulType::Multiply;
+    state.log2_multiplier = CSR[vtype].VLMUL[1:0];
+    if (vlmul == 3'b101) {
+      state.log2_multiplier = 8;
+    } else if (vlmul == 3'b110) {
+      state.log2_multiplier = 4;
+    } else if (vlmul == 3'b111) {
+      state.log2_multiplier = 2;
+    }
+
+    return state;
+  }
+}


jmawet:
I used the vector_state() function from vec.idl (hope I used it right) to get the LMUL and SEW variables, but how do I use lmul_type to modify the operation of state.log2_multiplier? Would I do something like:
if (lmul_type == mul) vlmax = lmul*vlen*sew
elseif (lmul_type == div) vlmax = (1/lmul)*vlen*sew

?

Additionally, how should I create the CSR files for the required CSRs? Like was mentioned in our meeting, those are pretty important and should probably get done first if they don't already exist (I couldn't find any so I assume no?). And then is what I did correct in terms of using the CSRs in the operation() field, syntax wise?

ThinkOpenly:
Shall we mark this as "draft" for now? (There's a button on the right.)

jmawet:
> Shall we mark this as "draft" for now? (There's a button on the right.)

I don't see the button but sure yeah

ThinkOpenly:
> > Shall we mark this as "draft" for now? (There's a button on the right.)
> 
> I don't see the button but sure yeah

There's a "convert to draft" link in the "Reviewers" section for me, at least.

There's also a dropdown for the green "Create pull request" button that offers a "Create draft pull request".

Just FYI. :-)

arch/inst/V/vsetivli.yaml
```diff
@@ -24,7 +26,19 @@ access:
   vu: always
 data_independent_timing: false
 operation(): |
+  XReg state = vector_state();
+  XReg VLMAX = state.log2_multiplier * VLEN * state.sew;
```
dhower-qc, line 30: - Need to be careful here. Multiplication is not widening in IDL, so you are likely going to have the result truncated. This comes up a lot, so I'm going to add widening operators that end in '<' (e.g., `*<`, `+<`, `-<`, `/<`, `<<<`). You can assume they exist.
- This needs to check `state.lmul_type` to determine if the `state.log2_multiplier` should be multiplied or divided.
- I think `state.sew` should be a division

So, for `lmuml_type == VectorLmulType::Multiply`, the equation should be:

`XReg VLMAX = (VLEN * pow(2, state.log2_multiplier)) / state.sew; # don't use`

Which is equivalent to:

`XReg VLMAX = (VLEN <<< state.log2_multiplier) >> (3 + state.vsew);`

arch/inst/V/vsetivli.yaml
```diff
@@ -24,7 +26,19 @@ access:
...
+  # todo: state.lmul_type
 
+  if (xs1 < VLMAX){
```
dhower-qc, line 33: I think you mean `if (uimm < VLMAX)`

arch/inst/V/vsetivli.yaml
```diff
@@ -24,7 +26,19 @@ access:
...
+    CSR[vl] = uimm;
+  } else {
+    CSR[vl] = VLMAX;
```
dhower-qc, line 36: We need to capture the two other cases listed in the spec:

- `ceil(AVL / 2) ≤ vl ≤ VLMAX if AVL < (2 * VLMAX)`

If uimm/AVL is less than 2*VLMAX, then the setting of vl is implementation-dependent, subject to the above constraint. We need to add a parameter to the V extension to capture some probable behaviors, with a fallback on `unpredictable()`. Something like `RVV_VL_WHEN_AVL_LT_DOUBLE_VLMAX`, with possible values "ceil(AVL/2)", "VLMAX", "custom". 

- `vl = VLMAX if AVL ≥ (2 * VLMAX)`

arch/inst/V/vsetivli.yaml
```diff
@@ -24,7 +26,19 @@ access:
...
+  }
+
+  CSR[vtype] = zimm11;
```
dhower-qc, line 39: Need to check for illegal vtype, and set vill if so.

arch/inst/V/vsetvl.yaml
```diff
@@ -24,3 +26,85 @@ access:
   vu: always
 data_independent_timing: false
 operation(): |
+  XReg state = vector_state();
```
dhower-qc, line 29: (Same comments as vsetivli)

dhower-qc (changes requested):

jmawet:
@dhower-qc added at least the skeleton for CSRs, but have some questions on the following:
- sw_read/write
- priv_mode
- type (under field)
- a couple other misc questions (vlenb, reading, field-agnostic write)
See my comments on the files for details

arch/csr/V/vcsr.yaml
```diff
@@ -0,0 +1,19 @@
+# yaml-language-server: $schema=../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: vcsr
+long_name: Vector control and status register
+address: 0x00F
+priv_mode: U #should be URW?
```
dhower-qc, line 8: U is correct

arch/csr/V/vcsr.yaml
```diff
@@ -0,0 +1,19 @@
...
+description: TBA.
+definedBy: V
+fields:
+  VALUE:
+    location_rv32: 31-0
+    location_rv64: 63-0
+    description: |
+      TBA
+    type: RO #??
+    reset_value: 0
```
dhower-qc, line 19: This should have two fields: `VXRM` and `VXSAT`. Reserved bits (the rest of the CSR) do not need to be specified.

Also, I believe reset_value here is `UNDEFINED_LEGAL`

arch/csr/V/vl.yaml
```diff
@@ -0,0 +1,25 @@
...
+    location_rv64: 63-0
+    description: |
+      The vl register holds an unsigned integer specifying the number of elements to be updated with
+      results from a vector instruction, as further detailed in Section Section 31.5.4.
+
+      The number of bits implemented in vl depends on the implementation's maximum vector
+      length of the smallest supported type. The smallest vector implementation with VLEN=32
+      and supporting SEW=8 would need at least six bits in vl to hold the values 0-32
+      (VLEN=32, with LMUL=8 and SEW=8, yields VLMAX=32).
+    type: RO #??
```
dhower-qc, line 24: Should be `RO-H` (read-only with hardware update)

arch/csr/V/vl.yaml
```diff
@@ -0,0 +1,25 @@
+# yaml-language-server: $schema=../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: vl
+long_name: Vector Length
+address: 0xC20
+priv_mode: U #should be URO
```
dhower-qc, line 8: U is correct

arch/csr/V/vlenb.yaml
```diff
@@ -0,0 +1,21 @@
...
+fields:
+  VALUE:
+    location_rv32: 31-0
+    location_rv64: 63-0
+    description: |
+      The value in vlenb is a design-time constant in any implementation.
+      Without this CSR, several instructions are needed to calculate VLEN in bytes, and the code
+      has to disturb current vl and vtype settings which require them to be saved and restored.
+    type: RO #??
+    reset_value: 0 #VLEN/8
```
dhower-qc, line 21: reset_value should be a function here:

```yaml
reset_value(): return VLEN / 8;
```

arch/csr/V/vlenb.yaml
```diff
@@ -0,0 +1,21 @@
...
+definedBy: V
+fields:
+  VALUE:
+    location_rv32: 31-0
+    location_rv64: 63-0
+    description: |
+      The value in vlenb is a design-time constant in any implementation.
+      Without this CSR, several instructions are needed to calculate VLEN in bytes, and the code
+      has to disturb current vl and vtype settings which require them to be saved and restored.
+    type: RO #??
```
dhower-qc, line 20: RO is correct

arch/csr/V/vstart.yaml
```diff
@@ -0,0 +1,59 @@
+# yaml-language-server: $schema=../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: vstart
+long_name: Vector Start Index
+address: 0x008
+priv_mode: U #should be URW
```
dhower-qc, line 8: U is correct.

arch/csr/V/vstart.yaml
```diff
@@ -0,0 +1,59 @@
...
+      arithmetic instruction, instead waiting until the instruction completes to take the
+      interrupt. Such implementations are permitted to raise an illegal instruction exception
+      when attempting to execute a vector arithmetic instruction when vstart is nonzero.
+
+      When migrating a software thread between two harts with different microarchitectures,
+      the vstart value might not be supported by the new hart microarchitecture. The runtime
+      on the receiving hart might then have to emulate instruction execution up to the next
+      supported vstart element position. Alternatively, migration events can be constrained to
+      only occur at mutually supported vstart locations.
+    type: RO #??
```
dhower-qc, line 58: RW-RH (read-write, with hardware update, restricted values)

arch/csr/V/vstart.yaml
```diff
@@ -0,0 +1,59 @@
...
+      CSR, leaving earlier elements in the destination vector undisturbed, and to reset the vstart CSR to
+      zero at the end of execution.
+      All vector instructions, including vset{i}vl{i}, reset the vstart CSR to zero.
+      vstart is not modified by vector instructions that raise illegal-instruction exceptions.
+      The vstart CSR is defined to have only enough writable bits to hold the largest element index (one
+      less than the maximum VLMAX).
+
+      The maximum vector length is obtained with the largest LMUL setting (8) and the smallest
+      SEW setting (8), so VLMAX_max = 8*VLEN/8 = VLEN. For example, for VLEN=256,
+      vstart would have 8 bits to represent indices from 0 through 255.
```
dhower-qc, line 30: [NOTE]

arch/csr/V/vtype.yaml
```diff
@@ -0,0 +1,48 @@
+# yaml-language-server: $schema=../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: vtype
+long_name: Vector Type
+address: 0xC21
+priv_mode: U #should be URO
+length: MXLEN
+description: Provides the default type used to interpret the contents of the vector registrer file.
```
dhower-qc, line 10: typo: registrer

arch/csr/V/vtype.yaml
```diff
@@ -0,0 +1,48 @@
...
+    location_rv64: 62-8
+    description: Reserved if non-zero
+    type: RO #??
+    reset_value: UNDEFINED_LEGAL
+  VMA:
+    location_rv32: 7
+    location_rv64: 7
+    description: Vector mask agnostic
+    type: RO #??
+    reset_value: 0
```
dhower-qc, line 30: UNDEFINED_LEGAL

arch/csr/V/vtype.yaml
```diff
@@ -0,0 +1,48 @@
...
+  VTA:
+    location_rv32: 6
+    location_rv64: 6
+    description: Vector tail agnostic
+    type: RO #??
```
dhower-qc, line 35: RO-H

arch/csr/V/vtype.yaml
```diff
@@ -0,0 +1,48 @@
...
+    reset_value: 0
```
dhower-qc, line 36: UNDEFINED_LEGAL

arch/csr/V/vtype.yaml
```diff
@@ -0,0 +1,48 @@
...
+  VSEW:
+    location_rv32: 5-3
+    location_rv64: 5-3
+    description: Selected element width
+    type: RO #??
```
dhower-qc, line 41: RO-H

arch/csr/V/vxrm.yaml
```diff
@@ -0,0 +1,19 @@
...
+length: MXLEN
+description: TBA.
+definedBy: V
+fields:
+  VALUE:
+    location_rv32: 31-0
+    location_rv64: 63-0
+    description: |
+      TBA
+    type: RO #??
```
dhower-qc, line 18: RO-H

arch/csr/V/vxrm.yaml
```diff
@@ -0,0 +1,19 @@
...
+    reset_value: 0
```
dhower-qc, line 19: UNDEFINED_LEGAL

arch/csr/V/vxsat.yaml
```diff
@@ -0,0 +1,19 @@
...
+length: MXLEN
+description: TBA.
+definedBy: V
+fields:
+  VALUE:
+    location_rv32: 31-0
+    location_rv64: 63-0
+    description: |
+      TBA
+    type: RO #??
```
dhower-qc, line 18: RO-H

arch/csr/V/vxsat.yaml
```diff
@@ -0,0 +1,19 @@
...
+    reset_value: 0
```
dhower-qc, line 19: UNDEFINED_LEGAL

dhower-qc (changes requested):

jmawet:
I have a few remaining questions on 

- idl ceil() function
- implementation-dependent functions
- csr descriptions

Otherwise I think I am about done with the vector CSRs and vset instructions. Almost ready to merge this all into the main repo vector branch

arch/csr/V/vtype.yaml
```diff
@@ -0,0 +1,125 @@
...
+      1XX = Reserved
+
+      [NOTE]
+      elements per vector register:
+      SEW 64 = 2
+      SEW 32 = 4
+      SEW 16 = 8
+      SEW 8  = 16
+
+      It is recommended that at reset, vill is set, and the remaining bits in vtype are zero.
```
ThinkOpenly, line 91: remove

arch/csr/V/vtype.yaml
```diff
@@ -0,0 +1,125 @@
...
+    # I feel like the encoding tables are important here. Is there a better way than the "note" tag?
+    type: RO-H
+    reset_value: 0
+  VLMUL:
+    location_rv32: 2-0
+    location_rv64: 2-0
```
ThinkOpenly, line 97: combine

arch/csr/V/vtype.yaml
```diff
@@ -0,0 +1,125 @@
...
+      effective LMUL (EMUL) of each vector operand is determined by the number of registers
+      required to hold the elements. For example, for a widening add operation, such as add 32-
+      bit values to produce 64-bit results, a double-width result requires twice the LMUL of the
+      single-width inputs.
+
+      LMUL can also be a fractional value, reducing the number of bits used in a single vector register.
+      Fractional LMUL is used to increase the number of effective usable vector register groups when
+      operating on mixed-width values.
+
+      It is recommended that at reset, vill is set, and the remaining bits in vtype are zero.
```
ThinkOpenly, line 122: remove

arch/csr/V/vxrm.yaml
```diff
@@ -0,0 +1,38 @@
+# yaml-language-server: $schema=../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: vxrm
+long_name: Fixed-Point Rounding Mode
```
ThinkOpenly, line 6: Vector Fixed-Point Rounding Mode

arch/csr/V/vxsat.yaml
```diff
@@ -0,0 +1,25 @@
+# yaml-language-server: $schema=../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: vxsat
+long_name: Fixed-Point Saturate Flag
```
ThinkOpenly, line 6: Vector Fixed-Point Saturation Flag

arch/csr/V/vxsat.yaml
```diff
@@ -0,0 +1,25 @@
...
+address: 0x009
+priv_mode: U
+length: MXLEN
+description: Indicates if a fixed-point instruction has had to saturate an output value to fit into a destination format
+definedBy: V
+sw_read(): |
+  return CSR[vcsr].VXSAT;
```
ThinkOpenly, line 13: Do we need a "sw_write" to mask all but the lowest-order bit?

arch/csr/V/vxsat.yaml
```diff
@@ -0,0 +1,25 @@
...
+  VALUE:
+    location_rv32: 31-0
+    location_rv64: 63-0
+    description: |
+      The vxsat CSR has a single read-write least-significant bit (vxsat[0]) that indicates if a fixed-point
+      instruction has had to saturate an output value to fit into a destination format. Bits vxsat[XLEN-1:1]
+      should be written as zeros.
+
+      The vxsat bit is mirrored in vcsr.
+    type: RO-H
```
ThinkOpenly, line 24: Isn't his read-write (RW)?

arch/csr/V/vxrm.yaml
```diff
@@ -0,0 +1,38 @@
...
+      The fixed-point rounding algorithm is specified as follows. Suppose the pre-rounding result is v, and d
+      bits of that result are to be rounded off. Then the rounded result is (v >> d) + r, where r depends on
+      the rounding mode as specified in the following table of vxrm[1:0] values.
+
+      - 00: rnu (round to nearest up), r = v[d-1]
+      - 01: rne (round to nearest even), r = v[d-1] & ( v[d-2:0]!=0 | v[d])
+      - 10: rdn (round down, truncate), r = 0
+      - 11: rod (round to odd), r = !v[d] & v[d-1:0]!=0
+    # I feel like the encoding table is important here. Is there a better way than this list?
+    type: RO-H
```
ThinkOpenly, line 37: Isn't this read-write (RW)?

arch/inst/V/vsetivli.yaml
```diff
@@ -3,9 +3,11 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: vsetivli
-long_name: No synopsis available.
+long_name: Set vector configuration (immediate immediate).
```
ThinkOpenly, line 6: eww. :-)

Suggest "Vector Set Vector Type Immediate and Vector Length Immediate" (not great, I know)

arch/inst/V/vsetivli.yaml
```diff
@@ -3,9 +3,11 @@
...
 description: |
-  No description available.
+  Vset (i,i) allows rapid configuration of the values in vl and vtype CSRs to match application
+  needs. The vsetvl instruction sets the vtype and vl CSRs based on the arguments, and writes
+  the new value of vl into rd.
```
ThinkOpenly, line 10: Set the `vtype` and `vl` CSRs, and write the new value of `vl` into `rd`.

arch/inst/V/vsetivli.yaml
```diff
@@ -24,7 +26,20 @@ access:
   vu: always
 data_independent_timing: false
 operation(): |
+  VectorState state = vector_state();
+  XReg VLMAX = (VLEN <<< state.log2_multiplier) >> (3 + state.vsew);
 
+  if (uimm < VLMAX){
```
ThinkOpenly, line 32: space after `)`

arch/inst/V/vsetvl.yaml
```diff
@@ -3,9 +3,11 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: vsetvl
-long_name: No synopsis available.
+long_name: Set vector configuration (register register).
 description: |
-  No description available.
+  Vset (r,r) allows rapid configuration of the values in vl and vtype CSRs to match application
+  needs. The vsetvl instruction sets the vtype and vl CSRs based on the contents of the argument
+  registers rs1 and rs2, and writes the new value of vl into rd.
```
ThinkOpenly, line 10: Set the `vtype` and `vl` CSRs, and write the new value of `vl` into `rd`.

arch/inst/V/vsetvl.yaml
```diff
@@ -24,3 +26,86 @@ access:
   vu: always
 data_independent_timing: false
 operation(): |
+  VectorState state = vector_state();
+  XReg VLMAX = (VLEN <<< state.log2_multiplier) >> (3 + state.vsew);
+
+  if (xs1 < VLMAX){
```
ThinkOpenly, line 32: space after `)`

arch/inst/V/vsetvl.yaml
```diff
@@ -3,9 +3,11 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: vsetvl
-long_name: No synopsis available.
+long_name: Set vector configuration (register register).
```
ThinkOpenly, line 6: Vector Set Vector Type and Vector Length

arch/inst/V/vsetvli.yaml
```diff
@@ -3,9 +3,11 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: vsetvli
-long_name: No synopsis available.
+long_name: Set vector configuration (register immediate).
```
ThinkOpenly, line 6: Vector Set Vector Type and Vector Length Immediate

arch/inst/V/vsetvli.yaml
```diff
@@ -3,9 +3,11 @@
...
 description: |
-  No description available.
+  Vset (r,i) allows rapid configuration of the values in vl and vtype CSRs to match application
+  needs. The vsetvl instruction sets the vtype and vl CSRs based on the contents of the argument
+  register rs1 and immediate, and writes the new value of vl into rd.
```
ThinkOpenly, line 10: Set the `vtype` and `vl` CSRs, and write the new value of `vl` into `rd`.

arch/inst/V/vsetvli.yaml
```diff
@@ -24,7 +26,20 @@ access:
   vu: always
 data_independent_timing: false
 operation(): |
+  VectorState state = vector_state();
+  XReg VLMAX = (VLEN <<< state.log2_multiplier) >> (3 + state.vsew);
 
+  if (xs1 < VLMAX){
```
ThinkOpenly, line 32: space after `)`

arch/csr/V/vtype.yaml
```diff
@@ -0,0 +1,125 @@
...
+
+      [NOTE]
+      elements per vector register:
+      SEW 64 = 2
+      SEW 32 = 4
+      SEW 16 = 8
+      SEW 8  = 16
+
+      It is recommended that at reset, vill is set, and the remaining bits in vtype are zero.
+    # I feel like the encoding tables are important here. Is there a better way than the "note" tag?
```
ThinkOpenly, line 92: PR #598 has some examples for embedded tables, if that could be helpful
```
          [separator="!"]
          !===
          ! Value ! VSXLEN
          ! 0     ! 32
          ! 1     ! 64
          !===
```

arch/csr/V/vtype.yaml
```diff
@@ -0,0 +1,48 @@
...
+    location_rv64: 5-3
+    description: Selected element width
+    type: RO #??
+    reset_value: 0
+  VLMUL:
+    location_rv32: 2-0
+    location_rv64: 2-0
+    description: Vector register group multiplier
+    type: RO #??
+    reset_value: 0
```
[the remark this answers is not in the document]
ThinkOpenly: how was this resolved?

jmawet:
unresolved and require assistance:
- vill (implementation specific?)
- idl ceil() function (does one exist?)
- sw_write semantics

I've written some questions in the comments regarding the above 3 points. @dhower-qc when you get a chance could you help answer some of them?

./do test:csrs is passing
./do test:idl/smoke is failing due to <<< operation (see comments)

ThinkOpenly:
We probably need to move this forward in some way. @jmawet, is this effort something you are able to continue, should we open it up for collaboration, or maybe find a new owner? There is definitely value here that would be great to get merged.

jmawet:
I am more than happy/able to continue. I have a few questions that need answers before I can proceed. I have been trying to get a hold of @dhower-qc for quite some time now with no response. Once I get a response I can get right back to work

ThinkOpenly:
> I am more than happy/able to continue.

Cool!

> I have a few questions that need answers before I can proceed.

There are a number of unresolved questions from you, a number of open review comments from me, and some changes intermingled, so it's difficult to sort at the moment. May I trouble you to repost your unanswered questions? You might be able to resolve some of the open comments, too? (Or, I can resolve them and do a fresh review?)

jmawet:
unresolved and require assistance:

- vill (implementation specific?)
whether to set vill or not depends on the implementation, there is no global spec for this. So how can we specify it in the instructions?

- idl ceil() function (does one exist?) 
2 questions:
    ceil(AVL / 2) is there a ceil() function in idl? Or should I assume in IDL it rounds down and therefore I should add +1 at the end?
    ceil(AVL / 2) ≤ vl ≤ VLMAX, the value set to vl depends on the implementation. According to this part of the spec, there is a legal range, but not sure how we would define that in idl

- sw_write semantics
for vxsat, the write is handled by vcsr, which does have a sw_write defined already. I assume that would cover it?
and vcsr only allocates 1 bit for vxsat, the next 2 bits are allocated to vxrm. So the sw_write of vcsr for vxsat should cover the "masking"
But I may be wrong, this is just what I am assuming (see more details in my comments)

- idl '<<<' operation
When running ./do test:idl/smoke it complains the '<<<' operation does not exist, but this was something Derek provided to me, so I'm not sure if it's supposed to be valid or not


Besides the questions above, I think you should do a fresh review to make sure all your comments are addressed

ThinkOpenly:
> * vill (implementation specific?)
>   whether to set vill or not depends on the implementation, there is no global spec for this. So how can we specify it in the instructions?

Generally, implementation-defined things require an associated configuration parameter. If there's a blanket case, then one parameter may be sufficient. I see at least one special case:
> The use of `vtype` encodings with LMUL < SEW<sub>MIN</sub>/ELEN is reserved, but implementations can set `vill` if they do not support these configurations.

If there are independent special cases, they likely all need their own parameter.

> * idl ceil() function (does one exist?)
>   2 questions:
>   ceil(AVL / 2) is there a ceil() function in idl? Or should I assume in IDL it rounds down and therefore I should add +1 at the end?

I don't know what built-in functions exist for IDL. I recall there being some sort of reference, but I can't find it now. For future reference, there is an IDL language reference linked from the project's README.

In this case, `ceil` seems like overkill for an integer, and I think this suffices:
`((AVL + 1) / 2)` 

>   ceil(AVL / 2) ≤ vl ≤ VLMAX, the value set to vl depends on the implementation. According to this part of the spec, there is a legal range, but not sure how we would define that in idl

Similarly to above, if there are implementation-defined values, we'll need configuration parameters which define them.

> * sw_write semantics
>   for vxsat, the write is handled by vcsr, which does have a sw_write defined already. I assume that would cover it?

The bits should be defined as one field is an alias of the other. We don't currently inherit any attributes of the alias, so the sw_write would need to still exist.

>   and vcsr only allocates 1 bit for vxsat, the next 2 bits are allocated to vxrm. So the sw_write of vcsr for vxsat should cover the "masking"

sw_write() implementations return the value to be written to the field, so there is no masking required.

> * idl '<<<' operation
>   When running ./do test:idl/smoke it complains the '<<<' operation does not exist, but this was something Derek provided to me, so I'm not sure if it's supposed to be valid or not

A grep isn't finding that operation. I see the reverse, `>>>`, so I presume it's not defined. Need a @dhower-qc update on implementation of `<<<`.

> Besides the questions above, I think you should do a fresh review to make sure all your comments are addressed

OK

arch/csr/V/vtype.yaml
```diff
@@ -0,0 +1,114 @@
...
+      branch on the sign bit.
+
+      If the vill bit is set, then any attempt to execute a vector instruction that depends upon vtype will
+      raise an illegal-instruction exception.
+
+      When the vill bit is set, the other XLEN-1 bits in vtype shall be zero.
+
+      It is recommended that at reset, vill is set.
+    type: RO-H
+    reset_value: 1
```
dhower-qc: Since the spec reads:

`It is recommended that at reset, vill is set.`

The defined reset value is actually UNDEFINED_LEGAL. We could add a param for the reset value when it is known.

jmawet:
ok... progress. 
Tests are able to run, I've been able to fix some of the errors reported, but stuck at this one:
```
/home/jdupac/git/riscv-unified-db/tools/ruby-gems/idlc/lib/idlc/ast.rb:359:in `type_error': In file /home/jdupac/git/riscv-unified-db/gen/spec/rv32/inst/V/vsetivli.yaml (Idl::AstNode::TypeError)
On line 30
In the code:

  28: VectorState state = vector_state();
  29: XReg VLMAX = (VLEN `<< state.log2_multiplier) >> (3 + state.vsew);
  30: XReg AVL = uimm;
  31: XReg CEIL_AVL_OVER_TWO = (AVL + 1) / 2;
  

A type error occurred
  Widening shift amount must be constant (if it's not, the width of the result is unknowable).
```
IDL doesn't seem to like me using the widening operation that depends on the vector state (line 29). 
So either we allow these "unknown widenings" in IDL, or we find a different way of getting the VLMAX value with a constant widening

jmawet:
@dhower-qc ./do test:idl and smoke are both passing now for cfgs rv32/64. Had a decent amount of compilation errors to clean up, but it's all fixed now. 
VLEN and ELEN are added to the v ext parameters as well

all that's missing now is 1 more review and the instruction tests we discussed this week, and we're ready to merge!

spec/std/isa/csr/V/vcsr.yaml
```diff
@@ -0,0 +1,30 @@
...
+length: MXLEN
+description: Allows access to vxrm and vxsat CSRs
+definedBy: V
+fields:
+  VXRM:
+    location: 2-1
+    description: See vxrm.
+    type: RW-RH
+    sw_write(csr_value): |
+      return csr_value.VXRM;
```
dhower-qc, line 24: `vcsr.VXRM` is an alias of `vxrm[1:0]`. That means two things here:

- it should have an `alias: vxrm[1:0]` key/value.
- sw_write needs to forward the write to `vxrm`:

```yaml
sw_write(csr_value): |
  CSR[vxrm].VALUE = csr_value.VXRM;
  return csr_value.VXRM;
```

I see that you defined the alias in the other direction (vxrm aliases vcsr). This is fine functionally, but I think this way better reflects how the spec talks e.g.: 

> The `vxrm` separate CSR can also be accessed via fields in...`vcsr`

Which (to me at least) speaks as if vcsr is the alias.

spec/std/isa/csr/V/vcsr.yaml
```diff
@@ -0,0 +1,30 @@
...
+    reset_value: UNDEFINED_LEGAL
+  VXSAT:
+    location: 0
+    description: See vxsat.
+    type: RW-RH
+    sw_write(csr_value): |
+      return csr_value.VXSAT;
```
dhower-qc, line 33: same

spec/std/isa/csr/V/vl.yaml
```diff
@@ -0,0 +1,30 @@
...
+      The vl register holds an unsigned integer specifying the number of elements to be updated with
+      results from a vector instruction, as further detailed in Section Section 31.5.4.
+
+      [NOTE]
+      The number of bits implemented in vl depends on the implementation's maximum vector
+      length of the smallest supported type. The smallest vector implementation with VLEN=32
+      and supporting SEW=8 would need at least six bits in vl to hold the values 0-32
+      (VLEN=32, with LMUL=8 and SEW=8, yields VLMAX=32).
+    type: RO-H
+    reset_value: 0
```
dhower-qc: UNDEFINED_LEGAL

spec/std/isa/csr/V/vstart.yaml
```diff
@@ -0,0 +1,74 @@
...
+      when attempting to execute a vector arithmetic instruction when vstart is nonzero.
+
+      [NOTE]
+      When migrating a software thread between two harts with different microarchitectures,
+      the vstart value might not be supported by the new hart microarchitecture. The runtime
+      on the receiving hart might then have to emulate instruction execution up to the next
+      supported vstart element position. Alternatively, migration events can be constrained to
+      only occur at mutually supported vstart locations.
+    sw_write(csr_value): |
+      return csr_value.VALUE & (VLEN - 1);
```
dhower-qc: The spec says:

> The vstart CSR is defined to have only enough writable bits to hold the largest element index (one less than the maximum VLMAX).

`VLEN` units are bits, but `vstart` units are elements.

This means that:

- The defined number of bits the largest to hold the biggest `VLMAX`, which happens when LMUL is maxed out and SEW is minimum. `(LMUL_MAX * VLEN / SEW_MIN)`. Assuming both LMUL_MAX and SEW_MIN are 8, the biggest VLMAX is 256.
- When SEW/LMUL have different settings, there will be 'reserved' bits in vstart. It is undefined what happens when you write non-zero into the reserved bits, but there is a recommendation that it takes a takes a trap. This is a parameter that needs to be added.

spec/std/isa/csr/V/vtype.yaml
```diff
@@ -0,0 +1,118 @@
...
+      branch on the sign bit.
+
+      If the vill bit is set, then any attempt to execute a vector instruction that depends upon vtype will
+      raise an illegal-instruction exception.
+
+      When the vill bit is set, the other XLEN-1 bits in vtype shall be zero.
+
+      It is recommended that at reset, vill is set.
+    type: RO-H
+    reset_value: 1
```
dhower-qc: This is only a recommendation. We probably want a parameter to say "I'm following the vtype reset recommendation". If that isn't being followed, reset is UNDEFINED_LEGAL

This applies to all the fields in this CSR

spec/std/isa/csr/V/vxrm.yaml
```diff
@@ -0,0 +1,47 @@
...
+name: vxrm
+long_name: Vector Fixed-Point Rounding Mode
+address: 0x00A
+writable: true
+priv_mode: U
+length: MXLEN
+description: Holds a 2-bit read-write rounding-mode field in the least-significant bits
+definedBy: V
+sw_read(): |
+  return CSR[vcsr].VXRM;
```
dhower-qc, line 17: Swap the alias direction, so vxrm is the CSR with real state.

spec/std/isa/csr/V/vxsat.yaml
```diff
@@ -0,0 +1,31 @@
...
+name: vxsat
+long_name: Vector Fixed-Point Saturate Flag
+address: 0x009
+writable: true
+priv_mode: U
+length: MXLEN
+description: Indicates if a fixed-point instruction has had to saturate an output value to fit into a destination format
+definedBy: V
+sw_read(): |
+  return CSR[vcsr].VXSAT;
```
dhower-qc, line 17: swap alias direction

spec/std/isa/ext/V.yaml
```diff
@@ -50,3 +50,19 @@ params:
...
       assert MSTATUS_VS_LEGAL_VALUES.include?(3) if ext?(:V) && (HW_MSTATUS_VS_DIRTY_UPDATE != "never")
+  RVV_VL_WHEN_AVL_LT_DOUBLE_VLMAX:
+    description: |
+      The value assigned to VL when AVL < 2*VLMAX.
+    schema:
+      type: string
+      enum: ["ceil(AVL/2)", "VLMAX", "custom"]
+  VLEN:
+    description: |
+      The number of bits in a single vector register. VLEN >= ELEN, which must be a power of 2, and must be no greater than 2^16.
```
dhower-qc: We can encode this programmatically once #891 lands. It will look like:

```yaml
requirements:
  idl(): |
    -> (VLEN >= ELEN) && (VLEN <= 'h10000) && (popcount(VLEN) == 1);
  reason: VLEN >= ELEN, which must be a power of 2, and must be no greater than 2^16.
```

For now, perhaps add this but comment it out.

spec/std/isa/ext/V.yaml
```diff
@@ -50,3 +50,19 @@ params:
...
+    schema:
+      type: integer
+  ELEN:
+    description: |
+      The maximum size in bits of a vector element that any operation can produce or consume, ELEN >= 8, which must be a power of 2.
+    schema:
```
dhower-qc, line 77: same as above, let's add the requirement but comment it out

spec/std/isa/isa/vec.idl
```diff
@@ -0,0 +1,53 @@
...
+                    0, 0, 0, 0, 0, 0, 0, 0,
+                    0, 0, 0, 0, 0, 0, 0, 0];
+
+enum VectorLmulType {
+  Divide
+  Multiply
+}
+
+struct VectorState {
+  Bits<7> vsew;
```
dhower-qc: this should be called `sew`, since `vsew` in the spec is the _encoded_ view of SEW, while `VectorState.sew` is the _decoded_ view

spec/std/isa/isa/vec.idl
```diff
@@ -0,0 +1,53 @@
...
+# - Order (might be dictated somewhat by test suite)
+#   - vset*
+#   - integer arith (vadd/vsub/vrsub(.vv,.vx,.vi), compare, min, max)
+#   - integer widening arith ()
+
+# the vector register file
+Bits<VLEN> v[32] = [0, 0, 0, 0, 0, 0, 0, 0,
+                    0, 0, 0, 0, 0, 0, 0, 0,
+                    0, 0, 0, 0, 0, 0, 0, 0,
+                    0, 0, 0, 0, 0, 0, 0, 0];
```
dhower-qc: These have no defined reset state. If you just get rid of the assignment, `v` will have undefined reset state.

dhower-qc (changes requested):

spec/std/isa/ext/V.yaml
```diff
@@ -50,3 +50,33 @@ params:
...
+  RVV_VL_WHEN_AVL_LT_DOUBLE_VLMAX:
+    description: |
+      The value assigned to VL when AVL < 2*VLMAX.
+    schema:
+      type: string
+      enum: ["ceil(AVL/2)", "VLMAX", "custom"]
+  FOLLOW_VTYPE_RESET_RECOMMENDATION:
+    description: |
+      The spec recommends - but does not require - that on reset, vill is set, and remaining bits in vtype and vl are set to 0.
+      If this parameter is set to true, the CSR resets will follow this recommandation. If it is false, the resets will be "UNDEFINED_LEGAL".
```
ThinkOpenly: Typo "recommandation", but perhaps we should use the text from the spec here, which I think is clearer:
```suggestion
      It is recommended that at reset, vtype.vill is set, the remaining bits in vtype are zero, and vl is set to zero.
      If this parameter is set to true, this recommendation is followed. If it is false, at reset the respective fields will be "UNDEFINED_LEGAL".
```

spec/std/isa/inst/V/vsetivli.yaml
```diff
@@ -27,6 +26,72 @@ access:
...
+    CSR[vtype].VTA = 0;
+    CSR[vtype].VSEW = 0;
+    CSR[vtype].VLMUL = 0;
+  } else if (xlen() == 32 && new_vtype[2:0] == 3'b101) {
+    # reserved LMUL in RV32
+    CSR[vtype].VILL = 1;
+    CSR[vtype].VMA = 0;
+    CSR[vtype].VTA = 0;
+    CSR[vtype].VSEW = 0;
+    CSR[vtype].VLMUL = 0;
```
ThinkOpenly: Since the blocks are all identical, let's combine these cases with ORs:
```suggestion
  if ((new_vtype[xlen() - 1] == 1'b1) # software is setting the illegal bit
      || ((new_vtype & 8'd0) != 0)    # reserved bits
      || (new_vtype[5] == 1)          # reserved vsew encoding
      || (new_vtype[2:0] == 3'b100)   # reserved vlmul encoding
      || (xlen() == 32 && new_vtype[2:0] == 3'b101)) # reserved LMUL in RV32
  {
    CSR[vtype].VILL = 1;
    CSR[vtype].VMA = 0;
    CSR[vtype].VTA = 0;
    CSR[vtype].VSEW = 0;
    CSR[vtype].VLMUL = 0;
```

spec/std/isa/inst/V/vsetvl.yaml
```diff
@@ -27,3 +26,144 @@ access:
...
+    CSR[vtype].VTA = 0;
+    CSR[vtype].VSEW = 0;
+    CSR[vtype].VLMUL = 0;
+  } else if (xlen() == 32 && new_vtype[2:0] == 3'b101) {
+    # reserved LMUL in RV32
+    CSR[vtype].VILL = 1;
+    CSR[vtype].VMA = 0;
+    CSR[vtype].VTA = 0;
+    CSR[vtype].VSEW = 0;
+    CSR[vtype].VLMUL = 0;
```
ThinkOpenly: combine these blocks, too.

spec/std/isa/inst/V/vsetvl.yaml
```diff
@@ -27,3 +26,144 @@ access:
...
+    CSR[vtype].VLMUL = new_vtype[2:0];
+  }
+
+  X[xd] = CSR[vl].VALUE;
+  CSR[vstart].VALUE = 0;
+
+# SPDX-SnippetBegin
+# SPDX-FileCopyrightText: 2017-2025 Contributors to the RISCV Sail Model <https://github.com/riscv/sail-riscv/blob/master/LICENCE>
+# SPDX-License-Identifier: BSD-2-Clause
+sail(): |
```
ThinkOpenly, line 76: Where did this come from? If you wrote it, I admire the effort. We probably want to treat Sail content as a unit, and find a programmatic way to bring it all from the canonical Sail code at once (or continually, but not manually). Reluctantly, I suggest that you remove this code and leave the `sail()` method empty here.

spec/std/isa/inst/V/vsetvli.yaml
```diff
@@ -27,6 +26,72 @@ access:
...
+    CSR[vtype].VTA = 0;
+    CSR[vtype].VSEW = 0;
+    CSR[vtype].VLMUL = 0;
+  } else if (xlen() == 32 && new_vtype[2:0] == 3'b101) {
+    # reserved LMUL in RV32
+    CSR[vtype].VILL = 1;
+    CSR[vtype].VMA = 0;
+    CSR[vtype].VTA = 0;
+    CSR[vtype].VSEW = 0;
+    CSR[vtype].VLMUL = 0;
```
ThinkOpenly: combine these blocks, too.

spec/std/isa/csr/V/vcsr.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+$schema: "csr_schema.json#"
+kind: csr
+name: vcsr
+long_name: Vector Control and Status Register
+address: 0x00F
+writable: true
+priv_mode: U
+length: MXLEN
+description: Allows access to vxrm and vxsat CSRs
```
ThinkOpenly: "Allows access" sound like an enable bit, so I suggest
```suggestion
description: Contains aliases to vxrm and vxsat CSRs
```

spec/std/isa/csr/V/vxrm.yaml
```diff
@@ -0,0 +1,46 @@
...
+      [separator="!"]
+      !===
+      ! vxrm[1:0] ! Abbreviation ! Rounding Mode ! Rounding increment, r
+      ! 00 ! rnu ! round-to-nearest-up (add +0.5 LSB) ! v[d-1]
+      ! 01 ! rne ! round-to-nearest-even ! v[d-1] & (v[d-2:0]\!=0 | v[d])
+      ! 10 ! rdn ! round-down (truncate) ! 0
+      ! 11 ! rod ! round-to-odd (OR bits into LSB, aka "jam") ! \!v[d] & v[d-1:0]\!=0
+
+    sw_write(csr_value): |
+      return csr_value.VALUE & (VLEN - 1);
```
ThinkOpenly: We need to preserve just the low-order 2 bits.
```suggestion
      return csr_value.VALUE & 2b'11;
```

spec/std/isa/csr/V/vxsat.yaml
```diff
@@ -0,0 +1,31 @@
...
+    location_rv32: 31-0
+    location_rv64: 63-0
+    description: |
+      The vxsat CSR has a single read-write least-significant bit (vxsat[0]) that indicates if a fixed-point
+      instruction has had to saturate an output value to fit into a destination format. Bits vxsat[XLEN-1:1]
+      should be written as zeros.
+
+      The vxsat bit is mirrored in vcsr.
+    sw_write(csr_value): |
+      return csr_value.VALUE & (VLEN - 1);
```
ThinkOpenly: We need to use only the least significant bit.
```suggestion
      return csr_value.VALUE & 1;
```

arch/inst/V/vsetivli.yaml
ThinkOpenly: Resolving this comment because the current implementation is:
```
  XReg vlen = VLEN;
  XReg vlmax = (vlen << state.log2_multiplier) >> (3 + state.sew);
```
...since `vlen` is of type `XReg`, a shift of the _value_ of `vlen` (which is relatively small; much, much smaller than `XReg` bits, even accounting for all expected values of `log2_multiplier` (2 bits)) will not overflow / get truncated.

spec/std/isa/csr/V/vlenb.yaml
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: vlenb
+long_name: Vector Byte Length
```
ThinkOpenly, line 9:
```suggestion
long_name: Vector Length Bytes
```

ThinkOpenly: nevermind. I see you picked this from the spec.

spec/std/isa/csr/V/vxsat.yaml
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: vxsat
+long_name: Vector Fixed-Point Saturate Flag
```
ThinkOpenly, line 9:
```suggestion
long_name: Vector Fixed-Point Saturation Flag
```

spec/std/isa/csr/V/vcsr.yaml
```diff
@@ -0,0 +1,34 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: vcsr
+long_name: Vector Control and Status Register
```
ThinkOpenly, line 9:
```suggestion
long_name: Vector Control and Status
```
I'm basing this on the spec's convention to call the section for the register "[long_name] (mnemonic) Register", like:
> Vector Type (`vtype`) Register
> Vector Length (`vl`) Register
> Vector Byte Length (`vlenb`) Register

and so on. `vcsr` is listed thus:
> Vector Control and Status (`vcsr`) Register

spec/std/isa/inst/V/vsetvl.yaml
ThinkOpenly: This still needs attention.

Issue #463, opened 2025-02-18
Option for Instructions Representation


AFOliveira:
Currently, UDB already intakes several options in its outputs - as in `Version` and more - can we make some variable to chose between sail and IDL? reality is that when we are generating documents, specially PDF, there is not much of a reason to have both, and it can be confusing, especially if we're trying to make the instructions descriptions widely adopted.

I propose something as  `./do instruction_index CODE=SAIL, IDL`. We can then have default be both, but let users fine-grain the output if it is needed.

james-ball-qualcomm:
Just want to avoid anything that looks like we are choosing between Sail and IDL. I'm including both in my generated PDFs to avoid any criticisms. At some point, hopefully RVI will be aligned on which pseudo-code language to use for official documentation to augment the ISA manuals. Then we could just change the default.

AFOliveira:
> Just want to avoid anything that looks like we are choosing between Sail and IDL. I'm including both in my generated PDFs to avoid any criticisms. At some point, hopefully RVI will be aligned on which pseudo-code language to use for official documentation to augment the ISA manuals. Then we could just change the default.

I understand, but reality is that as of now, RVI is aligned on choosing Sail, right? It just seems that trying to get IDL to go directly with the instruction index, might be too big of a step

james-ball-qualcomm:
A subset of RVI is aligned on Sail (mainly those working on it). The CSC is defining requirements for the pseudo-code language for documentation and also for processor model generation. Then the CSC can evaluate Sail vs. IDL solutions and select the best candidate. Then we'd share this decision with the TSC and see if they can align to it.

AFOliveira:
Oh, ok, I did not know about that. I'm not defending Sail by any means here, I just thought that it was sort of already defined. Any ways, adding a flag so one can choose to only pick one to fit in their document should hurt, right? Since the purpose of UDB is actually allowing for several different outputs, and this requires minimal changes to the ruby code and erb template.

dhower-qc:
I'm fine with a flag (as long as 'both' is also an option)

AFOliveira:
> I'm fine with a flag (as long as 'both' is also an option)

Of course, and I'll make it so that `both` is pre-defined

PR #469, opened 2025-03-03
Add flag to ext_pdf and allow either Sail, IDL or both

AFOliveira:
Solves #463


dhower-qc:
looks like the extension PDF generator regression is stuck?

AFOliveira:
> looks like the extension PDF generator regression is stuck?

I did not notice it, I'll check it out

dhower-qc (approved):

AFOliveira:
I removed the test matrix that was likely messing this PR up. I think we may need to clean the repo gen output in between those generations. I'll keep an issue tracking this, so we ensure that we have proper testing for the changes introduced in the UDB.

PR #471, opened 2025-03-05
Extension B: fix ctz instruction IDL code

ayosher:
IDL code in ctz instruction was wrong, fixing it

### DIFF
--- a/arch/inst/B/ctz.yaml
+++ b/arch/inst/B/ctz.yaml
@@ -30,7 +30,7 @@ operation(): |
     raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
   }
 
-  X[rd] = (xlen() - 1) - $signed(lowest_set_bit(X[rs1]));
+  X[rd] = $signed(lowest_set_bit(X[rs1]));
 
 sail(): |
   {


arch/inst/B/ctz.yaml
```diff
@@ -30,7 +30,7 @@ operation(): |
     raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
   }
 
-  X[rd] = (xlen() - 1) - $signed(lowest_set_bit(X[rs1]));
+  X[rd] = $signed(lowest_set_bit(X[rs1]));
```
ThinkOpenly, line 33: Do we need to retain `$signed`? (I confess my ignorance.) Does that widen the value appropriately? (Note that I opened #473 because `lowest_bit_set` returns an `XReg` (Bits<XLEN>, which I guess wouldn't need widening, but it probably should return Bits<8> or so.)

ayosher: Fixed, also fixed ctzw

ThinkOpenly:
Since you added `ctzw`, I looked at `clz` and `clzw`. They look correct, but they also use `$signed`. I don't think it's needed, and arguably shouldn't be there. Would you mind removing the uses of `$signed` from those two, at your option? Otherwise, I'll create a PR.

ayosher:
> Since you added `ctzw`, I looked at `clz` and `clzw`. They look correct, but they also use `$signed`. I don't think it's needed, and arguably shouldn't be there. Would you mind removing the uses of `$signed` from those two, at your option? Otherwise, I'll create a PR.

I am not sure I fully argee. Please see function highest_set_bit, unlike lowest_set_bit, it returns "-1" in case nothing found. Did I missed something?

dhower-qc:
The `$signed` cast has no effect in this case. There is no extension since `lowest_set_bit` and `X[rd]` are the same width. And the type doesn't follow the value into `X[rd]`, so it's immediately forgotten.

ayosher:
> The `$signed` cast has no effect in this case. There is no sign extension since `lowest_set_bit` and `X[rd]` are the same width. And the type doesn't follow the value into `X[rd]`, so it's immediately forgotten.

This is clear, $signed in IDL code is not only for code itself, it is also part of PDF, it hints that function can return negative result. Once more, I am talking about highest_set_bit only. I suggest have separate PR for leading zero operations (clz/clzw), as was previously suggested. This PR is about ctz and ctzw

dhower-qc (changes requested):
For `ctz`, we have to keep `xlen()` involved here because `xlen()` may not be the same as XLEN if, e.g., `CSR[mstatus].UXL` is mutable.

That said, there is a problem in the upstream code to fix.

The `ctz` instruction is supposed to return XLEN (really `xlen()`) when the input is 0. `lowest_set_bit` returns `xlen()` in that case, so the output of `lowest_set_bit` is exactly what we want in `X[rd]`. We just need to account for the case when `xlen() != XLEN`, e.g.:

```
if (xlen() == 32) {
  X[rd] = lowest_set_bit(X[rs1][31:0]);
} else {
  X[rd] = lowest_set_bit(X[rs1]);
}
```

The change for `ctzw` looks correct to me.

ayosher:
> For `ctz`, we have to keep `xlen()` involved here because `xlen()` may not be the same as XLEN if, e.g., `CSR[mstatus].UXL` is mutable.
> 
> That said, there is a problem in the upstream code to fix.
> 
> The `ctz` instruction is supposed to return XLEN (really `xlen()`) when the input is 0. `lowest_set_bit` returns `xlen()` in that case, so the output of `lowest_set_bit` is exactly what we want in `X[rd]`. We just need to account for the case when `xlen() != XLEN`, e.g.:
> 
> ```
> if (xlen() == 32) {
>   X[rd] = lowest_set_bit(X[rs1][31:0]);
> } else {
>   X[rd] = lowest_set_bit(X[rs1]);
> }
> ```
> 
> The change for `ctzw` looks correct to me.

Fix pushed

dhower-qc (approved):

ThinkOpenly:
> I am not sure I fully argee. Please see function highest_set_bit, unlike lowest_set_bit, it returns "-1" in case nothing found. Did I missed something?

I stand correct. Sorry for the noise. Looks like it's all sorted.

PR #472, opened 2025-03-05
Smrnmi/Smdbltrp extensions fixes

ayosher:
Add missing CSRs and add mnret operation IDL code

### DIFF
--- a/arch/csr/Smrnmi/mnepc.yaml
+++ b/arch/csr/Smrnmi/mnepc.yaml
@@ -1,0 +1,48 @@
+# yaml-language-server: $schema=../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: mnepc
+long_name: Machine Exception Program Counter
+address: 0x741
+priv_mode: M
+length: MXLEN
+description: |
+  Written with the PC of an instruction on an exception or interrupt taken in M-mode.
+
+  Also controls where the hart jumps on an exception return from M-mode.
+definedBy: Sm
+fields:
+  PC:
+    location_rv32: 31-0
+    location_rv64: 63-0
+    description: |
+      When a NMI / double trap is taken into M-mode, `mnepc.PC` is written with the virtual address of the
+      instruction that was interrupted or that encountered the exception.
+      Otherwise, `mnepc.PC` is never written by the implementation, though it may be explicitly written
+      by software.
+
+      On an exception return from M-mode NMI / double trap (from the MNRET instruction),
+      control transfers to the virtual address read out of `mnepc.PC`.
+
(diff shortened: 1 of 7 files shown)


arch/csr/Smrnmi/mnepc.yaml
```diff
@@ -0,0 +1,48 @@
...
+    type: RW-RH
+    sw_write(csr_value): |
+      return csr_value.PC & ~64'b1;
+    reset_value: 0
+sw_read(): |
+  if (implemented?(ExtensionName::C) && CSR[misa].C == 1'b1) {
+    return CSR[mnepc].PC & ~64'b1;
+  } else {
+    return CSR[mnepc].PC;
+  }
```
dhower-qc, line 48: shouldn't this be else `CSR[mnepc].PC & ~64'b11`?

arch/csr/Smrnmi/mnscratch.yaml
```diff
@@ -0,0 +1,18 @@
...
+length: MXLEN
+description: Scratch register for software use in NMI / double trap. Bits are not interpreted by hardware.
+definedBy: Smrnmi
+fields:
+  SCRATCH:
+    location_rv32: 31-0
+    location_rv64: 63-0
+    description: Scratch value
+    type: RW
+    reset_value: 0
```
dhower-qc, line 18: the reset value is undefined according the spec, so this should be `reset_value: UNDEFINED_LEGAL;`

arch/csr/Smrnmi/mnepc.yaml
```diff
@@ -0,0 +1,48 @@
...
+      Because PCs are always word-aligned, bits 1:0 of `mnepc.PC` are always
+      read-only 0.
+
+      [when,"ext?(:C) && MUTABLE_MISA_C == true"]
+      When `misa.C` is clear, bit 1 is masked to zero. Writes to bit 1 are still captured, and
+      may be visible on the next read with `misa.C` is set.
+    type: RW-RH
+    sw_write(csr_value): |
+      return csr_value.PC & ~64'b1;
+    reset_value: 0
```
dhower-qc, line 42: the reset value is undefined according the spec, so this should be reset_value: UNDEFINED_LEGAL;

dhower-qc (changes requested):

arch/csr/Smrnmi/mnepc.yaml
ayosher: Funny thing. You probably right. Please note, this was copy & paste from mepc, so it can have similar issue, which is out of scope of this PR.

dhower-qc (approved):

PR #490, opened 2025-03-14
Instructions ISA manual Appendix

AFOliveira:
This PR is to follow up on latest discussion around instructions Appendix in the ISA Manual - we already can generate this through UDB - although I am using PDF, so I don't think we can "deploy" and would need to do a release. Is this ok for you @dhower-qc or do you think we should turn this html and generate "deployments"?


AFOliveira:
[instructions_appendix.pdf](https://github.com/user-attachments/files/19247543/instructions_appendix.pdf)
Output if anyone wants to check.

ayosher:
Folks,

Would be much more useful to have syntax there in this appendix as well. 
For example, from whole appendix page about lw instruction it is impossible to understand that it should go: "lw, rd, imm(rs1)".
My understanding, that this appendix will be most useful for FW/SW folksto lookup instructions when writing assembly.

dhower-qc:
> Folks,
> 
> Would be much more useful to have syntax there in this appendix as well. For example, from whole appendix page about lw instruction it is impossible to understand that it should go: "lw, rd, imm(rs1)". My understanding, that this appendix will be most useful for FW/SW folksto lookup instructions when writing assembly.

We are planning to have the syntax there, but as part of "Phase II". The current syntax that exists for instructions needs to be vetted first, which is why we are pushing it out of the first phase.

AFOliveira:
@ayosher it is what Derek said about this PT. However, I may create a separate PR where we list all the information we can, maybe I can call them index and unratifiable_index?

dhower-qc (approved):

Issue #478, opened 2025-03-07
RISC-V Opcodes vs UDB gaps


AFOliveira:
RORW has the following variables in UDB:

```
  variables:
    - name: shamt
      location: 24-20
    - name: rs1
      location: 19-15
    - name: rd
      location: 11-7
```

However, the ISA is states that it's rs2 and not shamt. 

![Image](https://github.com/user-attachments/assets/58970a19-d9cb-411a-ad42-14aa01121492)

Aparently, sail code has shamt variable, but fetches it from rs2:
```
sail(): |
  {
    let rs1_val = (X(rs1))[31..0];
    let shamt = (X(rs2))[4..0];
    let result : bits(32) = match op {
      RISCV_ROLW => rs1_val <<< shamt,
      RISCV_RORW => rs1_val >>> shamt
    };
    X(rd) = sign_extend(result);
    RETIRE_SUCCESS
  }

```

I did not PR the change because it also requires IDL code and likely there are a lot of people who may change that with way less effort than me.

AFOliveira:
In PR #300 I am trying to match riscv-opcodes correctly, but there are some decisions to take and questions to made, like the one above. Another one is:

1. On csrrci and csrrsi, there is a variable name _uimm_ ; however, unsigned is something that should be explained on a subsfield, right?

AFOliveira:
@ThinkOpenly @dhower-qc I think you guys may have the answer to this questions

ThinkOpenly:
There are indeed some problems in RORW.

First, `shamt` is for _immediate_ values, not registers.

RORW uses a register, as can be seen in the Sail code. So, the variable name should be "rs2".

The IDL has related issues:
```
  XReg rs1_word = X[rs1][31:0];
  XReg shamt = X[rs1][4:0];
```

The second line is wrong. `shamt` should come from the lower 5 bits of _`rs2`_, not `rs1`.

ThinkOpenly:
@AFOliveira , if you're not in a hurry here, let's open an issue and let one of the new mentees address it, OK?

dhower-qc:
rorw:

In this case, I think we should change the YAML to use a register specifier instead of shamt. We use shamt when the encoding holds an immediate, but that isn't the case here. Note that changing this will also require changing the variable in IDL (from shamt -> rs2).

> On csrrci and csrrsi, there is a variable name uimm ; however, unsigned is something that should be explained on a subsfield, right?

[...]

PR #498, opened 2025-03-17
Fix RORW Instruction Encoding and Shift Logic"

Shehrozkashif:
issue: #478

### DIFF
--- a/arch/inst/B/ctz.yaml
+++ b/arch/inst/B/ctz.yaml
@@ -30,7 +30,11 @@ operation(): |
     raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
   }
 
-  X[rd] = (xlen() - 1) - $signed(lowest_set_bit(X[rs1]));
+  if (xlen() == 32) {
+    X[rd] = lowest_set_bit(X[rs1][31:0]);
+  } else {
+    X[rd] = lowest_set_bit(X[rs1]);
+  }
 
 sail(): |
   {
--- a/arch/inst/B/ctzw.yaml
+++ b/arch/inst/B/ctzw.yaml
@@ -31,7 +31,7 @@ operation(): |
     raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
   }
 
-  X[rd] = (xlen() - 1) - $signed(lowest_set_bit(X[rs1][31:0]));
+  X[rd] = lowest_set_bit(X[rs1][31:0]);
 
 sail(): |
   {


ThinkOpenly:
Can you put _just_ that 2nd commit here, since that's the one that addresses the title of this PR and the referenced issue.

Also, please add a helpful commit message, since that will persist with the repository history, unlike the PR content. The commit should:
- have a meaningful subject
- state the problem
- describe the fix

Something like:
```
Fix RORW operand and IDL

The YAML for the RORW instruction has "shamt" as an operand. "shamt" is used as an immediate
value, but the RORW instruction uses a register value. The RISC-V ISA specification shows the
analogous RORW instruction's operand is named "rs2", which is appropriate here.

Change RORW operand "shamt" to "rs2".

Also, in the IDL for RORW, the "shamt" _value_ should be retrieved from "rs2" (not "rs1",
as is done currently.

Fix the retrieval of the "shamt" value to come from "rs2".
```

ThinkOpenly:
> [Shehrozkashif](https://github.com/Shehrozkashif) [force-pushed](https://github.com/riscv-software-src/riscv-unified-db/compare/8fc15ea91eccf9e526c3add1464e2587966e7381..5a32ead2a4177b025946a58867babd83d8e9561a) the new-feature-branch branch from [8fc15ea](https://github.com/riscv-software-src/riscv-unified-db/commit/8fc15ea91eccf9e526c3add1464e2587966e7381) to [5a32ead](https://github.com/riscv-software-src/riscv-unified-db/commit/5a32ead2a4177b025946a58867babd83d8e9561a)

We now have a single commit. We're partly there. @Shehrozkashif, could you add the commit message I described previously to your commit? (Let me know if you need help.)

AFOliveira:
@Shehrozkashif I think you meant to write the message to @ThinkOpenly here. I deleted your message on the issue to not cause confusion there. Let me know if I am wrong.

ThinkOpenly (approved):

arch/inst/B/rorw.yaml
```diff
@@ -33,7 +33,7 @@ operation(): |
   }
 
   XReg rs1_word = X[rs1][31:0];
-  XReg shamt = X[rs1][4:0];
+  XReg shamt = X[rs2][4:0];
```
Unique-Usman, line 36: instead of shamt, is this not more descriptive if it is rs2_word ?

AFOliveira: Usman, shamt tipically means shift amount in this instructions. I think it is pretty descriptive of what it does, and is also compliant with the ISA that commonly refers to this values as shamt, normally it is related to imm fields, but once this value is fetched from the register, I don't think there are any differences there. Also, there is Sail code for this is in the spec and has been addopted by the authors of the extension, so likely shamt is the best name here.
image

AFOliveira (approved):

AFOliveira:
@Shehrozkashif  I do not know if you put this on the same branch by accident or not, but I don't think this should be on the same Pull Request. Your RORW change is much smaller and should be accepted shortly, while this adding Zcmop extension is a whole other story that will likely require some reviews/iterations to get right.
Also, to give some feedback on what you commited, I do not think you should put all in separate instructions, given that they are all the same instructions and is likely just a lot of duplication. I don't have any strong opinion on this, if you see this in other eyes, but the rest o MOP - may be operations - are currently handled that way.

Shehrozkashif:
@AFOliveira 
Thanks for your feedback! You’re absolutely right—I didn’t intend to mix the RORW change and the Zcmop extension in the same PR. I’ll separate them by keeping the RORW change in this PR and moving the Zcmop extension to a new branch for proper review.

Regarding the Zcmop implementation, I see your point about the duplication. I’ll take another look at how similar operations are handled and refactor accordingly to avoid unnecessary repetition. If you have any specific suggestions or examples from the existing MOP handling, I’d appreciate the guidance.

Thanks again for the review! I’ll make the necessary adjustments.

AFOliveira:
@Shehrozkashif Great! Take a look at arch/inst/Zimop to see how it was previously done. Not saying it's a better approach than yours, but it just keeps conformity!

AFOliveira:
@Shehrozkashif did you manage to get the original commits back?

PR #494, opened 2025-03-14
env bash instead of /bin/bash

kbroch-rivosinc:

Issue #493, opened 2025-03-14
use of /usr/bin/env to select version of bash in scripts


kbroch-rivosinc:
UPDATE:  There's no real bug here, but this change would allow for native dev use, w/o a change to container dev environment. 

~~**Describe the bug**
Unhappy from beginning trying to establish ROOT variable:~~

```
kbroch@kbroch-mac.local:~/rvi/repos/riscv-software-src/riscv-unified-db on  main via  v22.13.1 via 🐍 v3.13.1 via 💎 v3.2.3
❯ bin/setup
realpath: bin/setup[0]: No such file or directory
usage: dirname string [...]
usage: dirname string [...]
cat: /bin/.container-tag: No such file or directory
bin/setup: line 37: [: -v: unary operator expected
bin/setup: line 41: [: -v: unary operator expected
bin/setup: line 43: [: -v: unary operator expected
UDB tools run in a container. Both Docker and Singularity/Apptainer are supported.

1. Docker
2. Singularity

Which would you like to use? (1/2)
1
bin/setup: line 63: /.container-type: Read-only file system
Using docker environment
bin/setup: line 68: [: -v: unary operator expected
Building Docker image...
[+] Building 0.0s (0/0)                                                                                                                docker:desktop-linux
ERROR: invalid tag "riscvintl/udb:": invalid reference format
mkdir: /.home: Read-only file system
docker: Error response from daemon: the working directory 'riscvintl/udb:' is invalid, it needs to be an absolute path.
See 'docker run --help'.
docker: Error response from daemon: the working directory 'riscvintl/udb:' is invalid, it needs to be an absolute path.
See 'docker run --help'.
docker: Error response from daemon: the working directory 'riscvintl/udb:' is invalid, it needs to be an absolute path.
See 'docker run --help'.
docker: Error response from daemon: the working directory 'riscvintl/udb:' is invalid, it needs to be an absolute path.
See 'docker run --help'.
docker: invalid reference format.
See 'docker run --help'.
bin/setup: line 151: /.home/.venv/bin/activate: No such file or directory
docker: invalid reference format.
See 'docker run --help'.
docker: Error response from daemon: the working directory 'riscvintl/udb:' is invalid, it needs to be an absolute path.
See 'docker run --help'.
bin/setup: line 209: /.git/hooks/pre-commit: No such file or directory
chmod: /.git/hooks/pre-commit: No such file or directory
```

**To Reproduce**

```
[...]

PR #500, opened 2025-03-18
add shell script fmt and lint with pre-commit checks and fix errors

### LINKED ISSUE #495: consistent shell script formatting and linting
UDB uses many shell scripts to support development.  I think it is important that these 
be maintained consistently and linting can catch problems that may occur in some development
environments vs. others (see #493)

As usual:

* https://github.com/riscv-software-src/riscv-unified-db/issues/394
* https://github.com/riscv-software-src/riscv-unified-db/issues/408

I'd propose using `pre-commit` tool to run and enforce these.

kbroch-rivosinc:
fixes #495
relates to #493, #494

Currently only flagging shellcheck errors.  There's more warning to work thru if we want later (most around quoting vars)
shfmt runs w/o any corrections so this will just continue that could practice.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,14 +23,14 @@ repos:
       - id: check-yaml
 
   - repo: https://github.com/rbubley/mirrors-prettier
-    rev: v3.4.2
+    rev: v3.5.3
     hooks:
       - id: prettier
         files: \.(json|yml|yaml)$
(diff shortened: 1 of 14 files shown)


ThinkOpenly (approved):

AFOliveira (approved):
Sorry for taking a while, I wanted to test locally. All worked completely fine. LGTM!

dhower-qc (approved):

Issue #516, opened 2025-03-19
Add Zacas (Atomic CAS) extension data


dhower-qc:
Add Zacas.yaml file

syedowaisalishah:
@dhower-qc and @ThinkOpenly , I have completed the synopsis and description for Zacas (Atomic CAS) extension and submitted a PR #518 .  
Please review it when you get the chance.

Issue #517, opened 2025-03-19
Add Zacas (Atomic CAS) extension IDL


dhower-qc:
Add IDL for all Zacas instructions

syedowaisalishah:
@dhower-qc , I am currently implementing the IDL for the Zacas extension  focusing on the AMOCAS.D instruction. My implementation is modeled after the AMOADD.W instruction as a reference. However, I am facing recurring failures in the smoke tests which suggest there might be issues related to memory address translation, atomic operations handling, or exception handling.

Current Implementation

operation(): |
  # Ensure the Zacas extension is implemented and enabled
  if (!implemented?(ExtensionName::Zacas)) {
    raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
  }

  # Calculate the effective address from rs1
  effective_address = X[rs1];

  # Load the 64-bit value from the effective address
  loaded_value = mem_read64(effective_address, aq);

  # Prepare and assign values for comparison and swap
  compare_value_low = rd != 0 ? X[rd] : 0;
  compare_value_high = rd != 0 ? X[rd + 1] : 0;
  swap_value_low = rs2 != 0 ? X[rs2] : 0;
  swap_value_high = rs2 != 0 ? X[rs2 + 1] : 0;

  # Combine high and low parts for full 64-bit values
  compare_value = (compare_value_high << 32) | compare_value_low;
  swap_value = (swap_value_high << 32) | swap_value_low;

  # Perform the comparison and conditional swap
  if (loaded_value == compare_value) {
    mem_write64(effective_address, swap_value, rl);
  }

  # Update the register pair with the loaded value if rd is not x0
  if (rd != 0) {
    X[rd] = loaded_value & 0xFFFFFFFF;
    X[rd + 1] = (loaded_value >> 32) & 0xFFFFFFFF;


I am trying more to resolve issue but still i fail
Could you provide guidance or examples on the correct formatting and structuring of these IDL definitions for handling atomic operations?

dhower-qc:
> // Ensure the Zacas extension is implemented and enabled
> if (!implemented?(ExtensionName::Zacas)) {
> raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
> }

You don't need this; Unlike A, Zacas can't be dynamically disabled. And if your configuration says you don't implement Zacas, any attempt to decode this instruction will trap.

> //Load the 64-bit value from the effective address
> loaded_value = mem_read64(effective_address, aq);

[...]

PR #518, opened 2025-03-19
Add synopsis and description for Zacas (Atomic CAS) extension (#516)

syedowaisalishah:
This PR adds the synopsis and description for the Zacas (Atomic CAS) extension instructions.
- Updated amocas.d.yaml, amocas.q.yaml, amocas.w.yaml.
- Extracted details from the RISC-V Unprivileged ISA Spec.

issue #516.


syedowaisalishah:
@ThinkOpenly Thanks for the detailed feedback! I'll refine the commit message and update the descriptions to align with the ISA spec. Also, I'll fix the operation(): | formatting issue and push the changes shortly.

arch/inst/Zacas/amocas.d.yaml
```diff
@@ -28,3 +64,7 @@ access:
   vu: always
 data_independent_timing: false
 operation(): |
+  temp = mem[X(rs1)]
```
dhower-qc, line 67: We'll need to work on the IDL here. A few things to note:

- the address needs to go through translation (what is in X[rs1] is a virtual address)
- the swap must be atomic, which is going to require a call to what IDL calls a "builtin" function.
- Needs to handle the differences in RV32/64
- Needs to check for x0
- Needs to check all the exception cases

My recommendation is to just leave this empty for now, and we will implement it through #517

arch/inst/Zacas/amocas.d.yaml
```diff
@@ -3,9 +3,45 @@
...
+  **RV64 Operation:**
+  - Atomically loads a 64-bit value from the memory address in rs1.
+  - Compares the loaded value with the 64-bit value stored in rd.
+  - If the comparison is bitwise equal, stores the 64-bit value from rs2
+    back to rs1.
+  - Updates rd with the loaded value.
+
+  **Additional Considerations:**
+  - The first register in a pair must be even-numbered; encodings with
+    odd-numbered registers in rd or rs2 are reserved.
```
dhower-qc, line 29: We don't have the even/odd requirements reflected in the YAML. To do that:

- We need to split the encoding for rv32/rv64, since they are different
- For the rv32 version, need to add `not:` and cover all the odd register numbers

example:

```yaml
encoding:
  RV32:
    match: ...
    variables:
      - name: rd
        location: 11-7
        not: [1, 3, 5, 7, 9, 11, 13, 15, 17, 19, 21, 23, 25, 27, 29, 31]
      ..
  RV64:
    match: ...
    # ...
```

dhower-qc (changes requested):

syedowaisalishah:
@dhower-qc Thanks for the feedback! I'll update the YAML encoding for even/odd constraints, leave operation(): | empty as suggested, and check the failing tests. Will push updates soon.

syedowaisalishah:
@ThinkOpenly and @dhower-qc , I've updated the YAML encoding for even/odd constraints, left operation(): | empty as suggested, and reviewed exception handling. Let me know if any further refinements are needed.

arch/inst/Zacas/amocas.d.yaml
```diff
@@ -3,9 +3,24 @@
...
+  - Compares with the 64-bit value in rd and rd+1.
+  - If equal, stores the 64-bit value from rs2 and rs2+1 to rs1.
+  - Updates rd and rd+1 with the loaded value.
+
+  RV64 Operation:
+  - Loads a 64-bit value from rs1.
+  - Compares with rd.
+  - If equal, stores the 64-bit value from rs2 to rs1.
+  - Updates rd with the loaded value.
+
```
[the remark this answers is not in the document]
ThinkOpenly: This comment is still valid. Try to use the ratified text from the RISC-V ISA Specification when possible. I cut and pasted the above material, and only edited it for specificity (since this text is not associated with all of AMOCAS.W/D/Q, for example).

syedowaisalishah:
@ThinkOpenly and @dhower-qc I've updated the encoding to separate RV32 and RV64 constraints as suggested. I also incorporated the ratified text from the RISC-V ISA Specification for amocas.d,q and w.

arch/inst/Zacas/amocas.d.yaml
```diff
@@ -1,26 +1,68 @@
...
+  register pair is x0, then both halves of the pair read as zero. When the
+  first register of a destination register pair is x0, then the entire
+  register result is discarded and neither destination register is written.
+
+  For RV64, AMOCAS.D atomically loads 64-bits of a data value from address in
+  rs1, compares the loaded value to a 64-bit value held in rd, and if the
+  comparison is bitwise equal, then stores the 64-bit value held in rs2 to the
+  original address in rs1. The value loaded from memory is placed into
+  register rd.
+
```
ThinkOpenly, line 26: Let's insert the intermediate text which also refers to "AMOCAS.D" here:
> Just as for AMOs in the A extension, AMOCAS.D requires that the address held in rs1 be naturally aligned to the size of the operand (i.e., eight-byte aligned for doublewords). And the same exception options apply if the address is not naturally aligned.
> 
> Just as for AMOs in the A extension, the AMOCAS.D optionally provides release consistency semantics, using the aq and rl bits, to help implement multiprocessor synchronization. The memory operation performed by an AMOCAS.D, when successful, has acquire semantics if aq bit is 1 and has release semantics if rl bit is 1. The memory operation performed by an AMOCAS.W/D/Q, when not successful, has acquire semantics if aq bit is 1 but does not have release semantics, regardless of rl.

ThinkOpenly (approved):

dhower-qc (approved):
👍

PR #522, opened 2025-03-20
arch_overlay/qc_iu: Fix bug in qc.c.muliadd/qc.c.mveqz

AntonJohansson:
X[reg] with reg = creg2reg(rd) is used for accessing the source operand rs1.

I believe this is a bug.


ayosher:
Seems you are right

ayosher (commented):
Thanks for fixing this

ayosher (approved):
Thanks for fixing this

PR #526, opened 2025-03-21
Add initial Golang generation support.

AFOliveira:
This PR aims to generate inst.go - currently checked out for reviewing purposes. It doesn't yet allow for subsets of the ISA, but serves as an example that it is much easier to generate the outputs than the riscv-opcodes.

Also, I think I found some CSRs that we are missing, I'll open a issue on that.


AFOliveira:
@markdryan I'll tag you like this since I can't ask for review in any other way. This is still only "All" the RV64 ISA, I'll make compartimentalized on Monday

backends/generators/Go/inst.go
```diff
@@ -0,0 +1,2637 @@
...
+package riscv
+
+import "cmd/internal/obj"
+
+type inst struct {
+	opcode uint32
+	funct3 uint32
+	rs1    uint32
+	rs2    uint32
+    csr    int64
```
markdryan, line 11: The indentation here is weird.  It doesn't matter too much as we will always format this file before adding to the go source tree, but it's easy to fix in the python file.

backends/generators/Go/inst.go
```diff
@@ -0,0 +1,2637 @@
+// Code generated by go_generator.py; DO NOT EDIT.
+package riscv
+
+import "cmd/internal/obj"
+
+type inst struct {
```
markdryan, line 6: Note there is some discussion about changing this structure and the way we split up the encodings.  See [here](https://github.com/riscv/riscv-opcodes/pull/332) and [here](https://go-review.googlesource.com/c/go/+/637940) so this is likely to change in the future.

backends/generators/Go/inst.go
```diff
@@ -0,0 +1,2637 @@
...
+    return &inst{ 0x57, 0x6, 0x0, 0x0, 1056, 0x21 }
+  case AVMVVI:
+    return &inst{ 0x57, 0x3, 0x0, 0x0, 1504, 0x2f }
+  case AVMVVV:
+    return &inst{ 0x57, 0x0, 0x0, 0x0, 1504, 0x2f }
+  case AVMVVX:
+    return &inst{ 0x57, 0x4, 0x0, 0x0, 1504, 0x2f }
+  case AVMVXS:
+    return &inst{ 0x57, 0x2, 0x0, 0x0, 1056, 0x21 }
+  case AVMV1RV:
```
markdryan, line 1745: The ordering of the instructions seems to have changed. In the inst.go generated from riscv-opcodes these instructions occur before AVMVSX.  We try to maintain the same ordering to make it easier to see the what has meaningfully changed from one commit to another.

backends/generators/Go/inst.go
```diff
@@ -0,0 +1,2637 @@
...
+    return &inst{ 0x33, 0x4, 0x0, 0x0, 640, 0x14 }
+  case AZEXTH:
+    return &inst{ 0x3b, 0x4, 0x0, 0x0, 128, 0x4 }
+  case AZIP:
+    return &inst{ 0x13, 0x1, 0x0, 0xf, 143, 0x4 }
+  }
+	return nil
+}
+
+var csrs = map[uint16]string {
```
markdryan, line 2315: The CSRS don't look correct.  Where is FRM and VXRM?  Also, the list in this file includes the 32 bit CSRs, e.g., CYCLEH, which are not useful on riscv64.

markdryan (commented):
I've had a quick look.  Note my comments are based on the version of inst.go generated by the first commit.  It doesn't look like you re-generated inst.go when pushing the second commit, so the comments might not all be relevant.

AFOliveira:
> I've had a quick look. Note my comments are based on the version of inst.go generated by the first commit. It doesn't look like you re-generated inst.go when pushing the second commit, so the comments might not all be relevant.

They are, I did not change much other than filtering yet, thanks!

ThinkOpenly:
I don't have any particular concerns after quickly looking at the code. Is the generated artifact for `inst.go` match the target?

For Spike, same question. I don't think there's a relationship between Golang and Spike... should this be a separate PR?

AFOliveira (commented):
I think my answers did not show because I did not comment here, can you see them @markdryan?

AFOliveira:
> I don't have any particular concerns after quickly looking at the code. Is the generated artifact for `inst.go` match the target?
> 
> For Spike, same question. I don't think there's a relationship between Golang and Spike... should this be a separate PR?



So, for Golang I think we are there or, at least super close (some order may be changed, but that is it). Same for Spike, but I agree, I will remove Spike out of the branch since it just adds confusion, wait for this to be merged and then get Spike on board as well.

dhower-qc (changes requested):
Can you add a tasks.rake in backends/generators/Go with a task that will be displayed with `./do --tasks`?

AFOliveira:
> Can you add a tasks.rake in backends/generators/Go with a task that will be displayed with `./do --tasks`?

I already did locally actually :) Let me just remove Spike, and solve the commit history and I will change this PR from draft to ready to review. Unfortunately, it seems CSRs information in the UDB is the weak-point ATM, although I think we will change this in a matter of weeks

dhower-qc:
> @dhower-qc @kbroch-rivosinc Do you guys have any clue on why the tests are failing? https://github.com/riscv-software-src/riscv-unified-db/actions/runs/14088608154/job/39459215925?pr=526#step:6:142

It looks like you are stuck in the middle here. GitHub thinks it needs to check for C# files, which is true in main. But your branch doesn't have any C# files, which is why I think it's failing. Hopefully after you merge with main the error will resolve.

AFOliveira:
I think this is ready to merge - only problems are:

1. Inherently some instructions order is different than riscv-opcodes generated inst.go, I don't think there is a way to solve this. (also, tiny scope, I think it only happens once in the whole inst.go) 
2. We are missing some essential CSRs, so we can not officially propose this to the Go community, I propose just relying on an issue to track this and merge this PR into main as is. Whenever, we have the same CSRs as RISC-V Opcodes I'll just update every one.

ThinkOpenly:
> 1. Inherently some instructions order is different than riscv-opcodes generated inst.go, I don't think there is a way to solve this. (also, tiny scope, I think it only happens once in the whole inst.go)

Could you post a diff here, so we have a reference?

> 2. We are missing some essential CSRs, so we can not officially propose this to the Go community, I propose just relying on an issue to track this and merge this PR into main as is. Whenever, we have the same CSRs as RISC-V Opcodes I'll just update every one.

OK

AFOliveira:
> Could you post a diff here, so we have a reference?

I had it checked out, but I don't think we want that, how can I post the difference between the two files?

AFOliveira:
> Could you post a diff here, so we have a reference?

https://www.diffchecker.com/srkhmI0D/
The left one is riscv-opcodes and the right is UDB generated.


1. What I was mentioning as order change happens on line 921.
2. There are other differences in instructions but this is a question of how riscv-opcodes describes data vs how UDB does, it may require discussions but has nothing to do with this PR.
3. You may see the CSR diff in the end.

dhower-qc:
Cool. The instructions look pretty good.

We have pretty good issue coverage for the CSRs. 

I notice that the riscv-opcodes version isn't correct for pmpcfgODD; if I understand correctly, this is supposed to be for RV64, and those CSRs don't exist in RV64.

I'd vote to approve this PR since the generation looks great. We can work on the missing data.

AFOliveira:
> Cool. The instructions look pretty good.
> 
> We have pretty good issue coverage for the CSRs.
> 
> I notice that the riscv-opcodes version isn't correct for pmpcfgODD; if I understand correctly, this is supposed to be for RV64, and those CSRs don't exist in RV64.
> 
> I'd vote to approve this PR since the generation looks great. We can work on the missing data.

Nice, let me just add this to the regression test then, just to ensure it is not broken along the way

backends/generators/Go/inst.go
markdryan: This code just orders the CSRs and doesn't affect the instructions.  

I was talking about the whole register move instructions which appear after AVMVSX in your version of inst.go and before in the version of inst.go in the Go project.

I should probably clarify the process I'm using to review this patch.

I [cloned and built go locally](https://go.dev/doc/install/source), then I replaced in the in tree version of go/src/cmd/internal/obj/riscv/inst.go with the version in this PR and formatted the replaced inst.go in place by

```
cd go/src/cmd/internal/obj/riscv
../../../../../bin/go fmt inst.go
```

and then I just did a git diff.  This clearly shows the differences between the existing inst.go and the one generated by this PR.  For example, one of the differences I see is

```
@@ -918,14 +1732,6 @@ func encode(a obj.As) *inst {
                return &inst{0x57, 0x2, 0x0, 0x0, -1792, 0x48}
        case AVMULHUVX:
                return &inst{0x57, 0x6, 0x0, 0x0, -1792, 0x48}
-       case AVMV1RV:
-               return &inst{0x57, 0x3, 0x0, 0x0, -1568, 0x4f}
-       case AVMV2RV:
-               return &inst{0x57, 0x3, 0x1, 0x0, -1568, 0x4f}
-       case AVMV4RV:
-               return &inst{0x57, 0x3, 0x3, 0x0, -1568, 0x4f}
-       case AVMV8RV:
-               return &inst{0x57, 0x3, 0x7, 0x0, -1568, 0x4f}
        case AVMVSX:
                return &inst{0x57, 0x6, 0x0, 0x0, 1056, 0x21}
        case AVMVVI:
@@ -936,6 +1742,14 @@ func encode(a obj.As) *inst {
                return &inst{0x57, 0x4, 0x0, 0x0, 1504, 0x2f}
        case AVMVXS:
                return &inst{0x57, 0x2, 0x0, 0x0, 1056, 0x21}
+       case AVMV1RV:
+               return &inst{0x57, 0x3, 0x0, 0x0, -1568, 0x4f}
+       case AVMV2RV:
+               return &inst{0x57, 0x3, 0x1, 0x0, -1568, 0x4f}
+       case AVMV4RV:
+               return &inst{0x57, 0x3, 0x3, 0x0, -1568, 0x4f}
+       case AVMV8RV:
+               return &inst{0x57, 0x3, 0x7, 0x0, -1568, 0x4f}
        case AVMXNORMM:
                return &inst{0x57, 0x2, 0x0, 0x0, 2016, 0x3f}
        case AVMXORMM:
```

backends/generators/Go/inst.go
markdryan: Thanks.  I'll look into this and try to figure out what's going on.

backends/generators/Go/inst.go
AFOliveira: yes, I confused myself in the answer here, I was also missing ordering the CSRs. Now it's already done. That bit you sent is actually the only one place in which the ordering is wrong, but since I don't think riscv-opcodes does any ordering on instructions I can not reorder my version to match that. Also, I the scope where that happens in limited to those 4 instructions, which should not be that big of a problem, right?

backends/generators/Go/inst.go
AFOliveira: I think you are fetching from csrs.csv assuming it is all 64-bit, but that is not the case for those registers. Likely, it's a risc-v opcodes problem and they need to be changed from csrs.csv to csrs32.csv.

backends/generators/Go/inst.go
markdryan: Yeah, if it's just 4 instructions it will probably be okay.

AFOliveira:
Is there any dependency for this PR  @dhower-qc? Or can we merge this and then I'll start the rest of riscv-opcodes outputs?

dhower-qc (approved):
Looks good to me. We might want to eventually elevate the python interface into the db data to lib, but let's do that later

PR #530, opened 2025-03-21
 YAML files for the eight 16-bit MOP instructions defined in the Zcmo…

### LINKED ISSUE #514: Add Zcmop extension data
Add Zcmop.yaml file to database.

Shehrozkashif:
This PR implements the **C.MOP.N (Compressed May-Be-Operation) instruction** as part of the Zcmop extension. The instruction follows a **16-bit encoding format**, where bits `[15:12]` are fixed as `"0110"`, bits `[7:0]` are set to `"10000001"`, and bits `[11:8]` encode `(n >> 1)`, determining the specific operation. It maps to various **pseudoinstructions** (`c.mop.1`, `c.mop.3`, `c.mop.5`, etc.) based on the `mop_n_field` value. The instruction is available across all privilege levels (`s`, `u`, `vs`, and `vu`), ensuring broad applicability. By introducing this instruction, the PR enhances **code compression efficiency**, reducing instruction footprint while maintaining execution flexibility.
Closes #514


AFOliveira (changes requested):
@Shehrozkashif Is there any good reason to not encapsulate all this instructions in a single file names c.mop.n.yaml?

Shehrozkashif:
@AFOliveira 
To enhance modularity, clarity, and version control, I have organized each Zcmop instruction into its own YAML file. This approach aligns with the existing structure observed in the arch/inst directory, where instructions are maintained in separate YAML files.

AFOliveira:
> @AFOliveira To enhance modularity, clarity, and version control, I have organized each Zcmop instruction into its own YAML file. This approach aligns with the existing structure observed in the arch/inst directory, where instructions are maintained in separate YAML files.

You are right in that explanation and this is great work! However, this case is slightly special: on MOP - may be operations - like the C.MOPs, I think there is little point to make them separate as of now. If you check [mop.r.n , in this link,](https://github.com/riscv-software-src/riscv-unified-db/blob/main/arch/inst/Zimop/mop.r.n.yaml) you will see that we also opted to avoid duplication and described the individual instructions as pseudo-instructions. This may be better because it aligns with riscv-opcodes(https://github.com/riscv/riscv-opcodes) and will make it easier in this first approach of replacing that repo with UDB. Does this make sense to you, @Shehrozkashif?

Shehrozkashif:
@AFOliveira 
Got it! That makes sense—keeping MOP instructions together as pseudo-instructions helps avoid duplication and aligns with riscv-opcodes. Given the transition to UDB, I see how this approach is more efficient for now. I’ll follow this structure for MOPs while keeping the modular approach for other instructions. Thanks for the clarification!

AFOliveira:
> @AFOliveira 
> Got it! That makes sense—keeping MOP instructions together as pseudo-instructions helps avoid duplication and aligns with riscv-opcodes. Given the transition to UDB, I see how this approach is more efficient for now. I’ll follow this structure for MOPs while keeping the modular approach for other instructions. Thanks for the clarification!


Exactly, thanks!

AFOliveira:
Great work! I think we are almost there!

arch/inst/Zcmop/c.mop.n.yaml
```diff
@@ -0,0 +1,41 @@
...
+kind: instruction
+name: c.mop.n
+long_name: "Zcmop: Compressed May-Be-Operation Instruction"
+description: |
+  No description available.
+definedBy: Zcmop
+assembly: c.mop.n
+encoding:
+  # 16-bit encoding: bits [15:12] are fixed as "0110", bits [11:8] encode the (n >> 1) value,
+  # and bits [7:0] are fixed as "10000001".
```
ThinkOpenly, line 13: Suggested, certainly not mandatory, change here. Some of this text is redundant with the next line, and could be considered fairly obvious as well. I suggest removing this comment. The interesting part is associated with a single field, so a comment could be added there, below.

ThinkOpenly (changes requested):

arch/inst/Zcmop/c.mop.n.yaml
```diff
@@ -0,0 +1,41 @@
...
+  match: 0110----10000001
+  variables:
+    - name: mop_n_field
+      location: 11-8
```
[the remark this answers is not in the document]
ThinkOpenly: If you wanted to add a comment here, you could, but I don't think it's necessary, given all of the context.

arch/inst/Zcmop/c.mop.n.yaml
Shehrozkashif: Thanks for the suggestion! I've removed the redundant comment as you recommended.

ThinkOpenly (commented):
I'm not seeing any changes...?

Shehrozkashif:
@ThinkOpenly 
My bad changes were not commited now I have pushed again

ThinkOpenly:
@dhower-qc , should we look to covert this to the "layout" as used in "arch/csr/Zihpm/hpmcounterN.layout", for example?

ThinkOpenly (approved):

arch/inst/Zcmop/c.mop.n.yaml
```diff
@@ -0,0 +1,41 @@
+# yaml-language-server: $schema=../../../schemas/inst_schema.json
+
+$schema: "inst_schema.json#"
+kind: instruction
+name: c.mop.n
+long_name: "Zcmop: Compressed May-Be-Operation Instruction"
```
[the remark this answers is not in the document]
AFOliveira: @Shehrozkashif

Shehrozkashif:
@AFOliveira 
yes now they are visible to me

AFOliveira:
> @AFOliveira yes now they are visible to me

I forgot to comment other than the individual comments. My bad, sorry!

arch/inst/Zcmop/c.mop.n.yaml
```diff
@@ -0,0 +1,41 @@
...
+description: |
+  No description available.
+definedBy: Zcmop
+assembly: c.mop.n
+encoding:
+  # 16-bit encoding: bits [15:12] are fixed as "0110", bits [11:8] encode the (n >> 1) value,
+  # and bits [7:0] are fixed as "10000001".
+  match: 0110----10000001
```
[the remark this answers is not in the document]
ThinkOpenly: `01100---10000001`

arch/inst/Zcmop/c.mop.n.yaml
```diff
@@ -0,0 +1,41 @@
+# yaml-language-server: $schema=../../../schemas/inst_schema.json
+
+$schema: "inst_schema.json#"
+kind: instruction
+name: c.mop.n
+long_name: "Zcmop: Compressed May-Be-Operation Instruction"
+description: |
```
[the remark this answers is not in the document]
Shehrozkashif: sure ill add it in the description

arch/inst/Zcmop/c.mop.n.yaml
```diff
@@ -0,0 +1,41 @@
...
+  No description available.
+definedBy: Zcmop
+assembly: c.mop.n
```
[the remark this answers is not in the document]
Shehrozkashif: Sure ill remove the quotes to maintain consistency with the rest of the file

arch/inst/Zcmop/c.mop.n.yaml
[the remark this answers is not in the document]
Shehrozkashif: Sure, I'll remove the double quotes to maintain consistency with the rest of the file

arch/inst/Zcmop/c.mop.n.yaml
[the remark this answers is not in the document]
AFOliveira: this is still not fixed @Shehrozkashif

AFOliveira (changes requested):
Still one tiny detail to fo, and after thet its ready to merge

AFOliveira:
@Shehrozkashif One tiny detail to go, can you see it in the review?

Shehrozkashif:
@AFOliveira, no, I am unable to see what actually pointing out

AFOliveira:
> @AFOliveira, no, I am unable to see what actually pointing out

That is weird. I am talking about the assembly field, you may just leave it empty, but remove the instruction name
image

Shehrozkashif:
Now I have got it!

arch/inst/Zcmop/c.mop.n.yaml
```diff
@@ -0,0 +1,39 @@
+# yaml-language-server: $schema=../../../schemas/inst_schema.json
+
+$schema: inst_schema.json#
+kind: instruction
+name: c.mop.n
+long_name: Compressed May-Be-Operation Instruction
+description: "|"
```
AFOliveira, line 7: Can you add the description "C.MOP.n is encoded in the reserved
encoding space corresponding to C.LUI xn, 0. Unlike the MOPs defined in the
Zimop extension, the C.MOP.n instructions are defined to not write any register. Their encoding allows
future extensions to define them to read register x[n]."

AFOliveira (changes requested):

arch/inst/Zcmop/c.mop.n.yaml
Shehrozkashif: sure, ill add it

AFOliveira (approved):
Great job! 

Assembly field could be empty, but I also don't think it's a big problem due to the inital state of RISC-V assembly in this repo.

Shehrozkashif:
Understood! I'll leave the assembly field as is for now, but I'll keep this in mind for future refinements.

AFOliveira:
I think this ready to merge, just waiting your approval

arch/inst/Zcmop/c.mop.n.yaml
```diff
@@ -0,0 +1,39 @@
...
+kind: instruction
+name: c.mop.n
+long_name: Compressed May-Be-Operation Instruction
+description: C.MOP.n is encoded in the reserved encoding space corresponding to C.LUI xn, 0. Unlike the MOPs defined in the Zimop extension, the C.MOP.n instructions are defined to not write any register. Their encoding allows future extensions to define them to read register x[n].
+definedBy: Zcmop
+assembly: "|"
+encoding:
+  match: 01100---10000001
+  variables:
+    - name: mop_n_field
```
dhower-qc, line 13: Can we rename this to just `n`? I just did the same for the uncompressed mops, so that would give us consistency

dhower-qc (changes requested):

arch/inst/Zcmop/c.mop.n.yaml
```diff
@@ -0,0 +1,39 @@
+# yaml-language-server: $schema=../../../schemas/inst_schema.json
+
+$schema: inst_schema.json#
+kind: instruction
+name: c.mop.n
+long_name: Compressed May-Be-Operation Instruction
+description: C.MOP.n is encoded in the reserved encoding space corresponding to C.LUI xn, 0. Unlike the MOPs defined in the Zimop extension, the C.MOP.n instructions are defined to not write any register. Their encoding allows future extensions to define them to read register x[n].
+definedBy: Zcmop
+assembly: "|"
```
ThinkOpenly, line 9: This should be:
```
assembly: ""
```
(right?)

ThinkOpenly (changes requested):

arch/inst/Zcmop/c.mop.n.yaml
Shehrozkashif: sure ill rename it

arch/inst/Zcmop/c.mop.n.yaml
```diff
@@ -0,0 +1,39 @@
...
+      location: 10-8
+      left_shift: 1
+access:
+  s: always
+  u: always
+  vs: always
+  vu: always
+data_independent_timing: false
+pseudoinstructions:
+  - when: (mop_n_field == 0x1)
```
dhower-qc, line 23: `mop_n_field` needs to be renamed to `n` here, too (and below)

dhower-qc (changes requested):

Shehrozkashif:
@dhower-qc, @ThinkOpenly, if you need more refinements to merge this  pr, let me know

arch/inst/Zcmop/c.mop.n.yaml
```diff
@@ -0,0 +1,39 @@
...
+long_name: Compressed May-Be-Operation Instruction
+description: C.MOP.n is encoded in the reserved encoding space corresponding to C.LUI xn, 0. Unlike the MOPs defined in the Zimop extension, the C.MOP.n instructions are defined to not write any register. Their encoding allows future extensions to define them to read register x[n].
+definedBy: Zcmop
+assembly: ""
+encoding:
+  match: 01100---10000001
+  variables:
+    - name: n
+      location: 10-8
+      left_shift: 1
```
dhower-qc, line 15: There is no left shift on this.

arch/inst/Zcmop/c.mop.n.yaml
```diff
@@ -0,0 +1,39 @@
...
+    to: c.mop.7
+  - when: (n == 0x9)
+    to: c.mop.9
+  - when: (n == 0xB)
+    to: c.mop.11
+  - when: (n == 0xD)
+    to: c.mop.13
+  - when: (n == 0xF)
+    to: c.mop.15
+operation(): |
```
dhower-qc, line 39: Should be

```yaml
operation(): "" # does nothing
```

dhower-qc (changes requested):

arch/inst/Zcmop/c.mop.n.yaml
```diff
@@ -0,0 +1,39 @@
+# yaml-language-server: $schema=../../../schemas/inst_schema.json
+
+$schema: inst_schema.json#
+kind: instruction
+name: c.mop.n
+long_name: Compressed May-Be-Operation Instruction
```
ThinkOpenly, line 6: nit: could you remove "Instruction" from the name? It's a bit redundant, since the file is describing an instruction (see "kind" attribute, above). (Also, no other instruction YAML includes that word.)

arch/inst/Zcmop/c.mop.n.yaml
dhower-qc: hmm, but that doesn't jive with the mappings:

image

where:

`n[3:1] == 0  => 1 (not 0)`
`n[3:1] == 1  => 3 (not 2)`
...

Shehrozkashif: Since only n[3:1] is stored in bits 10-8, the left shift is needed to reconstruct the full 4-bit value n[3:0], ensuring n[0] is correctly set to 0

AFOliveira: Amazing how we can always find more corner cases in defining the spec :) 

Should we have a `set` field? Should we allow logic in one of the fields to describe the behavior of the field?

dhower-qc: I suggest we make sure this gets captured by Usman's "binning" exercise, and we can address it there as part of Phase II.

Then we can just remove left_shift for now.

ThinkOpenly: Having the left_shift is accurate, but incomplete. How should it be represented in the absence of the shift?

For correctness now, shall we:
- Leave the shift and add one in `when` cases?
  ```
  when: (n+1 == 0x1)
  to: c.mop.1
  ```
- Abandon the shift (temporarily, I presume) and append a '1' bit in the `when` cases?
  ```
  when: ((n<<1)+1)
  to: c.mop.1
  ```
- Go back to the original `when` cases?
  ```
  when: (n == 0)
  to: c.mop.1
  ```

Should we rename `n`, then, to `n_3_1`?

Shehrozkashif: Keeping the shift and adding 1 when needed seems like the safer approach for now.

Removing the shift and appending 1 in specific cases might oversimplify things.

Reverting to the original when: (n == 0) could work, but I need to ensure it doesn’t break anything.

@dhower-qc, what would you suggest?

ThinkOpenly: Let's open an issue to track it, before we move past and forget.

arch/inst/Zcmop/c.mop.n.yaml
```diff
@@ -0,0 +1,38 @@
...
+    to: c.mop.7
+  - when: (n == 0x9)
+    to: c.mop.9
+  - when: (n == 0xB)
+    to: c.mop.11
+  - when: (n == 0xD)
+    to: c.mop.13
+  - when: (n == 0xF)
+    to: c.mop.15
+operation(): ""
```
AFOliveira, line 38: I believe @dhower-qc  asked for a comment such as `operation(): "" #do nothing `

Shehrozkashif: sure ill add the comment as well

AFOliveira (approved):

AFOliveira:
@Shehrozkashif  you need to run the pre-commit checks locally in order to be able to merge with main

dhower-qc (approved):

AFOliveira:
@ThinkOpenly Just waiting for your approval!

ThinkOpenly:
Is there an issue open regarding the interesting field transformation required by this encoding?
```
n = (field_n[3:1] << 1) | 1
```
Other than that, this is ready to be merged, IMO.

AFOliveira:
@Shehrozkashif do you want to do the honours and open the issue with what you came across here?

Shehrozkashif:
@AFOliveira Sure, happy to! I’ll open an issue shortly outlining the transformation and its context so we can track it properly. Thanks for the nudge!

ThinkOpenly (approved):

PR #532, opened 2025-03-24
Change several CSRs from unspecified base to base: 32 

AFOliveira:
To comply with the privileged spec. This is required for several use cases like Go risc-v support which relies only on 64-bit spec, thus being able to filter rv32 CSRs is mandatory.


AFOliveira:
As per the privileged ISA table - https://github.com/riscv/riscv-isa-manual/blob/main/src/priv-csrs.adoc

dhower-qc (approved):

PR #533, opened 2025-03-24
Separate Zcf and Zcd extensions from Zce 

ayosher:
Zcf and Zcd to be in separate directories and be defined as separate extensions, Zce implies them if F/D extensions defined


AFOliveira:
Just wondering, isn't this something that could go to main?

ayosher:
> Just wondering, isn't this something that could go to main?

I doubt - it rely (conceptually) on changes in cpp_hart branch: instructions set definedBy only Zcf/Zcd, assuming that extension dependencies will be recovered. by engine

PR #524, opened 2025-03-20
Add CSR Definitions for Trigger Module Registers

neverlandiz:
This PR addresses Issue #81 by adding the missing CSR definitions for the following trigger module registers:
- tselect.yaml
- tdata1.yaml
- tdata2.yaml
- tdata3.yaml
- scontext.yaml
- mcontext.yaml
- hcontext.yaml

Issue #519, opened 2025-03-20
Add XLEN length type for CSRs


dhower-qc:
Many CSRs in the debug spec have a length equal to the effective XLEN of the current priv mode. This is a new concept for UDB that needs to be implemented.

This will involve:

- Adding "XLEN" as an option for length in the CSR schema.
- Updating the Ruby CSR object to understand "XLEN".

This issue should also depend on #52 to avoid a name collision.

dhower-qc:
@neverlandiz, can you please respond to this issue so I can assign it to you?

neverlandiz:
sounds good

PR #534, opened 2025-03-25
Added XLEN as a length type for CSRs

neverlandiz:
This PR addresses Issue #519.
Many debug CSRs have a length of XLEN, a new concept that needs to be implemented. The missing implementation was causing errors in PR #524.

- Updated CSR schema to include XLEN as a length type
- Updated the Ruby CSR object to handle XLEN as a valid length type


lib/arch_obj_models/csr.rb
```diff
@@ -185,6 +185,8 @@ def length(cfg_arch, effective_xlen = nil)
         # don't know VSXLEN
         effective_xlen
       end
+    when "XLEN"
+      effective_xlen
```
dhower-qc, line 189: Good find on this one. There are a few more places to look:

- #dynamic_length?
    - In this case, the length is "dynamic" (unknown by a config) if any of the three conditions covered in the existing case statement for that function are true when that mode is supported.
        - For example, the MXLEN condition must always be checked because you must always have M-mode.
        - The SXLEN condition applies if S-mode is possible, which can be checked by `cfg_arch.possible_extensions.map(&:name).include?("S")`
        - The VSXLEN condition applies if VS-mode is possible. It's defined by `H`, so you check it with `cfg_arch.possible_extensions.map(&:name).include?("H")`
- #min_length
- #max_length
- #length_cond32
- #length_cond64
    - For both of the "cond*" functions, you need to check the `priv_mode() == PrivilegeMode::{M,S,VS}`, along with the condition you see in the existing function.
- #length_pretty

schemas/csr_schema.json
```diff
@@ -245,8 +245,8 @@
           "enum": ["M", "S", "U", "VS"]
         },
         "length": {
-          "description": "Length, in bits, of the CSR. Can either be a 32, 64 or MXLEN, SXLEN, VSXLEN to indicate that is is dependent on the effective XLEN for a given mode",
-          "enum": [32, 64, "MXLEN", "SXLEN", "VSXLEN"]
+          "description": "Length, in bits, of the CSR. Can either be a 32, 64 or MXLEN, SXLEN, VSXLEN, XLEN to indicate that is is dependent on the effective XLEN for a given mode",
```
dhower-qc, line 248: This should describe what "XLEN" means, since it is different than the others.

dhower-qc (changes requested):

neverlandiz:
@dhower-qc I've made changes based on your comments, let me know if anything else needs to be added/changed. Thanks!

lib/arch_obj_models/csr.rb
```diff
@@ -124,6 +124,17 @@ def dynamic_length?(cfg_arch)
...
+      # must always have M-mode
+      cfg_arch.mxlen.nil?
+      # SXLEN condition applies if S-mode is possible
+      if cfg_arch.possible_extensions.map(&:name).include?("S")
+        [nil, 3264].include?(cfg_arch.param_values["SXLEN"])
+      end
+      # VSXLEN condition applies if VS-mode is possible
+      if cfg_arch.possible_extensions.map(&:name).include?("H")
+        [nil, 3264].include?(cfg_arch.param_values["VSXLEN"])
+      end
```
dhower-qc, line 137: Lines 129-133 have no effect here. The value of a block (when "XLEN") is it's last statement (if..H). So the value of this case is the value of the last if, which will either be `nil` (if H isn't implemented) or the check on line 136 (if H is implemented).

You want to return the OR of the three checks.

lib/arch_obj_models/csr.rb
```diff
@@ -219,6 +232,8 @@ def max_length(cfg_arch)
       else
         64
       end
+    when "XLEN"
+      64
```
dhower-qc, line 236: This is going to be similar to what you have in `dynamic_length?`; the max length for a given configuration depends on what XLENs are supported in the implemented modes.

lib/arch_obj_models/csr.rb
```diff
@@ -235,6 +250,16 @@ def length_cond32
       "CSR[mstatus].SXL == 0"
     when "VSXLEN"
       "CSR[hstatus].VSXL == 0"
+    when "XLEN"
+      if priv_mode() == PrivilegeMode::M
+        "CSR[misa].MXL == 0"
+      elsif priv_mode() == PrivilegeMode::S
+        "CSR[mstatus].SXL == 0"
+      elsif priv_mode() == PrivilegeMode::VS
+        "CSR[hstatus].VSXL == 0"
```
dhower-qc, line 259: Close. This is returning an IDL statement. We can't check priv_mode() at compile time (when this Ruby code executes) because it changes as software runs on the implementation. So what you want to return is the _string_ `(priv_mode() == PrivilegeMode::M && CSR[misa].MXL == 0) || (priv_mode() == PrivilegeMode::S && ...) || ...`

lib/arch_obj_models/csr.rb
```diff
@@ -249,6 +274,16 @@ def length_cond64
       "CSR[mstatus].SXL == 1"
     when "VSXLEN"
       "CSR[hstatus].VSXL == 1"
+    when "XLEN"
+      if priv_mode() == PrivilegeMode::M
+        "CSR[misa].MXL == 1"
+      elsif priv_mode() == PrivilegeMode::S
+        "CSR[mstatus].SXL == 1"
+      elsif priv_mode() == PrivilegeMode::VS
+        "CSR[hstatus].VSXL == 1"
```
dhower-qc, line 283: same here

lib/arch_obj_models/csr.rb
```diff
@@ -266,6 +301,16 @@ def length_pretty(cfg_arch, effective_xlen=nil)
           "CSR[mstatus].SXL == %%"
         when "VSXLEN"
           "CSR[hstatus].VSXL == %%"
+        when "XLEN"
+          if priv_mode() == PrivilegeMode::M
+            "CSR[misa].MXL == %%"
+          elsif priv_mode() == PrivilegeMode::S
+            "CSR[mstatus].SXL == %%"
+          elsif priv_mode() == PrivilegeMode::VS
+            "CSR[hstatus].VSXL == %%"
```
dhower-qc, line 310: And here

dhower-qc (changes requested):

neverlandiz:
@dhower-qc Thanks for the comments! I've made the changes, but I'm a little unsure if I implemented max_length correctly. Should it be dependent on the privilege mode or supported modes?

lib/arch_obj_models/csr.rb
```diff
@@ -124,6 +124,11 @@ def dynamic_length?(cfg_arch)
     when "VSXLEN"
       # dynamic if either we don't know VSXLEN or VSXLEN is explicitly mutable
       [nil, 3264].include?(cfg_arch.param_values["VSXLEN"])
+    when "XLEN"
+      # must always have M-mode
+      # SXLEN condition applies if S-mode is possible
+      # VSXLEN condition applies if VS-mode is possible
+      (cfg_arch.mxlen.nil?) || (cfg_arch.possible_extensions.map(&:name).include?("S") && [nil, 3264].include?(cfg_arch.param_values["SXLEN"])) || (cfg_arch.possible_extensions.map(&:name).include?("H") && [nil, 3264].include?(cfg_arch.param_values["VSXLEN"]))
```
dhower-qc, line 131: Just a formatting nit here: it's more readable if you put each condition on a separate line. In ruby, you'll need a `\`, like:

```ruby
(cfg_arch.mxlen.nil?) || \
  (cfg_arch.possible_extensions.map(&:name).include?("S") && \
   [nil, 3264].include(cfg_arch.param_values["SXLEN"])) || \ 
  (cfg_arch.possible_extensions.map(&:name).include?("H") && \
   [nil, 3264].include?(cfg_arch.param_values["VSXLEN"]))
```

lib/arch_obj_models/csr.rb
```diff
@@ -219,6 +226,32 @@ def max_length(cfg_arch)
...
+        cfg_arch.mxlen || 64
+      elsif cfg_arch.possible_extensions.map(&:name).include?("S")
+        if cfg_arch.param_values.key?("SXLEN")
+          if cfg_arch.param_values["SXLEN"] == 3264
+            64
+          else
+            cfg_arch.param_values["SXLEN"]
+          end
+        else
+          64
```
dhower-qc, line 240: SXLEN can never be greater than MXLEN, so this should actually be the same as "M" (cfg_arch.mxlen || 64)

lib/arch_obj_models/csr.rb
```diff
@@ -219,6 +226,32 @@ def max_length(cfg_arch)
...
+        end
+      elsif cfg_arch.possible_extensions.map(&:name).include?("H")
+        if cfg_arch.param_values.key?("VSXLEN")
+          if cfg_arch.param_values["VSXLEN"] == 3264
+            64
+          else
+            cfg_arch.param_values["VSXLEN"]
+          end
+        else
+          64
```
dhower-qc, line 250: similar here: VSXLEN can never be greater than MXLEN or SXLEN, so it is the min of those two

lib/arch_obj_models/csr.rb
```diff
@@ -219,6 +226,32 @@ def max_length(cfg_arch)
...
+        end
+      else
+        64
```
dhower-qc, line 253: I don't think we can ever get here, or if you can I'm not sure what it means ;) So let's `raise "unexpected"`

lib/arch_obj_models/csr.rb
```diff
@@ -235,6 +268,8 @@ def length_cond32
       "CSR[mstatus].SXL == 0"
     when "VSXLEN"
       "CSR[hstatus].VSXL == 0"
+    when "XLEN"
+      (priv_mode() == PrivilegeMode::M && "CSR[misa].MXL == 0") || (priv_mode() == PrivilegeMode::S && "CSR[mstatus].SXL == 0") || (priv_mode() == PrivilegeMode::VS && "CSR[hstatus].VSXL == 0")
```
dhower-qc, line 272: This needs to all be a string:

```ruby
"(priv_mode() == PrivilegeMode::M && CSR[misa].MXL == 0) || (priv_mode() == PrivilegeMode::S && CSR[mstatus].SXL == 0) || (priv_mode() == PrivilegeMode::VS && CSR[hstatus].VSXL == 0)"
```

lib/arch_obj_models/csr.rb
```diff
@@ -249,6 +284,8 @@ def length_cond64
       "CSR[mstatus].SXL == 1"
     when "VSXLEN"
       "CSR[hstatus].VSXL == 1"
+    when "XLEN"
+      (priv_mode() == PrivilegeMode::M && "CSR[misa].MXL == 1") || (priv_mode() == PrivilegeMode::S && "CSR[mstatus].SXL == 1") || (priv_mode() == PrivilegeMode::VS && "CSR[hstatus].VSXL == 1")
```
dhower-qc, line 288: string here too

lib/arch_obj_models/csr.rb
```diff
@@ -266,6 +303,8 @@ def length_pretty(cfg_arch, effective_xlen=nil)
           "CSR[mstatus].SXL == %%"
         when "VSXLEN"
           "CSR[hstatus].VSXL == %%"
+        when "XLEN"
+          (priv_mode() == PrivilegeMode::M && "CSR[misa].MXL == %%") || (priv_mode() == PrivilegeMode::S && "CSR[mstatus].SXL == %%") || (priv_mode() == PrivilegeMode::VS && "CSR[hstatus].VSXL == %%")
```
dhower-qc, line 307: and here

dhower-qc (changes requested):

neverlandiz:
@dhower-qc Thanks for the comments, I've added the changes.

dhower-qc (approved):

PR #540, opened 2025-03-25
Ensure that ext req expression satisfiability accounts for ext absence

dhower-qc:
Should fix ext_pdf generation issue with missing instructions.


ayosher (approved):
Worked for me.
Thanks

PR #542, opened 2025-03-25
This commit adds the Zilsd extension, which improves code size and pe…

### LINKED ISSUE #510: Add Zilsd extension data
Zilsd extensions has been recently ratified and at some point needs to be added to UDB.
https://github.com/riscv/riscv-zilsd/releases/tag/v1.0.

This issue is to add the YAML (name, version, etc) for Zilsd. This task does not require IDL (the `operation()` field) to be complete.

### LINKED ISSUE #511: Add Zilsd extension IDL
Zilsd extensions has been recently ratified and at some point needs to be added to UDB.
https://github.com/riscv/riscv-zilsd/releases/tag/v1.0.

This issue is to add IDL for Zilsd instructions.

Shehrozkashif:
…rformance by introducing load/store pair instructions for RV32. It includes 32-bit instructions (ld and sd) for loading and storing doublewords, as well as 16-bit compressed versions (c.ldsp, c.sdsp, c.ld, and c.sd) for more efficient memory access. These instructions work with even/odd register pairs, following the Zdinx extension, and ensure proper alignment for better performance. The extension was officially approved in February 2025 and helps reduce instruction count while making memory operations faster.
Closes #510 
Closes #511

### DIFF
--- a/arch/ext/Zilsd.yaml
+++ b/arch/ext/Zilsd.yaml
@@ -1,0 +1,13 @@
+$schema: "ext_schema.json#"
+kind: extension
+name: Zilsd
+long_name: Load/Store Pair for RV32
+description: |
+  This specification adds load and store instructions using register pairs. It does so by reusing existing instruction encodings which are RV64-only. The specification defines 32-bit encodings.
+  Load and store instructions will use the same definition of even-odd pairs as defined by the Zdinx extension.
+  The extension improves static code density, by replacing two separate load or store instructions with a single one. In addition, it can provide a performance improvement for implementations that can make use of a wider than XLEN memory interface.
+type: unprivileged
+versions:
+  - version: "1.0"
+    state: ratified
+    ratification_date: "2025-02"
--- a/arch/inst/Zilsd/ld.yaml
+++ b/arch/inst/Zilsd/ld.yaml
@@ -1,0 +1,36 @@
+# yaml-language-server: $schema=../../../schemas/inst_schema.json
+
+$schema: "inst_schema.json#"
+kind: instruction
+name: ld
+long_name: Load doubleword to even/odd register pair
+description: |
+  Loads a 64-bit value into registers rd and rd+1. The effective address is obtained by adding
+  register rs1 to the sign-extended 12-bit offset.
+definedBy: Zilsd
+assembly: rd, offset(rs1)
+encoding:
+  match: -----------------011-----0000011
+  variables:
+    - name: rd
+      location: 11-7
+      not: [1, 3, 5, 7, 9, 11, 13, 15, 17, 19, 21, 23, 25, 27, 29, 31]
+    - name: rs1
+      location: 19-15
+    - name: imm
+      location: 31-20
+access:
+  s: always
+  u: always
+  vs: always
+  vu: always
+operation(): |
+  Bits<XLEN> base = X[rs1];
+  Bits<XLEN> offset = $signed(imm);
+  Bits<XLEN> eff_addr = base + offset;
+
+  Bits<64> data = read_memory<64>(eff_addr, $encoding);
+
+  X[rd] = data[31:0];
+  X[rd+1] = data[63:32];
+sail(): "" #not implemented in the sail model yet
--- a/arch/inst/Zilsd/sd.yaml
+++ b/arch/inst/Zilsd/sd.yaml
@@ -1,0 +1,37 @@
+# yaml-language-server: $schema=../../../schemas/inst_schema.json
+
+$schema: "inst_schema.json#"
+kind: instruction
+name: sd
+long_name: Store doubleword from even/odd register pair
+description: |
+  Stores a 64-bit value from registers rs2 and rs2+1. The effective address is obtained by adding
+  register rs1 to the sign-extended 12-bit offset.
+definedBy: Zilsd
+assembly: rs2, offset(rs1)
+encoding:
+  match: -----------------011-----0100011
+  variables:
+    - name: rs1
+      location: 19-15
+    - name: rs2
+      not: [1, 3, 5, 7, 9, 11, 13, 15, 17, 19, 21, 23, 25, 27, 29, 31]
+      location: 24-20
+    - name: imm
+      location: 31-25|11-7
+access:
+  s: always
+  u: always
+  vs: always
+  vu: always
+operation(): |
+  Bits<XLEN> base = X[rs1];
+  Bits<XLEN> offset = $signed(imm);
+  Bits<XLEN> eff_addr = base + offset;
+
+  Bits<32> lower_word = X[rs2];
+  Bits<32> upper_word = X[rs2 + 1];
+  Bits<64> store_data = {upper_word, lower_word};
+
+  write_memory<64>(eff_addr, store_data, $encoding);
+sail(): "" #not implemented in the sail model yet


ayosher:
I believe you cannot have single extension for both Zilsd and Zclsd. It is 2 separate extensions. This extension does not imply Zdinx. It uses register pairs, but not implies Zdinx. The Zclsd requires at least Zca. The Zilsd does not require it.

Shehrozkashif:
Thank you for the clarification. I now understand that Zilsd and Zclsd are separate extensions: Zilsd introduces 32-bit load (ld) and store (sd) instructions for doublewords using even/odd register pairs in RV32, while Zclsd adds compressed 16-bit versions (c.ld, c.sd, c.ldsp, c.sdsp) for more efficient memory access. Additionally, Zclsd depends on the Zca extension, whereas Zilsd operates independently of Zdinx. ​

ayosher:
I think you have to add arch/ext/Zclsd.yaml, I do not see it yet.
As well, need to change the following files:
arch/inst/I/sd.yaml 
arch/inst/I/ld.yaml 
arch/inst/C/c.ld.yaml 
arch/inst/C/c.ldsp.yaml 
arch/inst/C/c.sdsp.yaml 
arch/inst/C/c.sd.yaml

For each of them you should add condition that these instructions for RV32 only should Zilsd/Zclsd respectively, while for RV64 still remain be dependent from I/C.

arch/ext/Zilsd.yaml
```diff
@@ -0,0 +1,23 @@
+$schema: "ext_schema.json#"
+kind: extension
+name: Zilsd
+long_name: Load/Store Pair for RV32 (Zilsd)
+description: |
+  The Zilsd extension adds RV32-only load and store pair instructions with 32-bit encoding.
+
+  **Instructions:**
+  - **ld rd, offset(rs1):** Loads a 64-bit doubleword into an aligned register pair. The lower-numbered (even) register receives bits 31:0 and the higher (odd) register receives bits 63:32.
+  - **sd rs2, offset(rs1):** Stores a 64-bit doubleword from an aligned register pair.
```
ThinkOpenly, line 10: Let's leave the set of instructions to the associated files in `arch/inst/Zilsd`, so delete these lines.

arch/ext/Zilsd.yaml
```diff
@@ -0,0 +1,23 @@
+$schema: "ext_schema.json#"
+kind: extension
+name: Zilsd
+long_name: Load/Store Pair for RV32 (Zilsd)
+description: |
+  The Zilsd extension adds RV32-only load and store pair instructions with 32-bit encoding.
```
ThinkOpenly, line 6: The text in the spec uses plural "encodings", so let's use that here.

arch/ext/Zilsd.yaml
```diff
@@ -0,0 +1,23 @@
...
+
```
ThinkOpenly, line 7: Maybe we should include more of the text from the spec:
> This specification adds load and store instructions using register pairs. It does so by reusing existing instruction encodings which are RV64-only. The specification defines 32-bit encodings.
> 
> Load and store instructions will use the same definition of even-odd pairs as defined by the Zdinx extension.
> 
> The extension improves static code density, by replacing two separate load or store instructions with a single one. In addition, it can provide a performance improvement for implementations that can make use of a wider than XLEN memory interface.

arch/ext/Zilsd.yaml
```diff
@@ -0,0 +1,23 @@
...
+  **Instructions:**
+  - **ld rd, offset(rs1):** Loads a 64-bit doubleword into an aligned register pair. The lower-numbered (even) register receives bits 31:0 and the higher (odd) register receives bits 63:32.
+  - **sd rs2, offset(rs1):** Stores a 64-bit doubleword from an aligned register pair.
+
+  **Alignment and Atomicity:**
+  - If the effective address is a multiple of 8, the access is naturally aligned and will not raise a misaligned trap.
+  - If the address is a multiple of 4 but not 8, each word access is performed atomically; however, a misaligned trap may occur.
+  - For load instructions, if a trap is taken the base register retains its original value, ensuring resumable trap handling.
+  - Implementations that do not perform the doubleword load atomically must delay writeback or sequence register writes to satisfy these requirements.
```
ThinkOpenly, line 16: This text is more associated with the instructions than the extension as a whole, at least arguably. Let's remove this text as well.

arch/ext/Zilsd.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+  This extension is independent and does not imply any other extension.
```
ThinkOpenly, line 18: This is arguably redundant with any "implies" field, or, in this case, the lack thereof. So, let's remove this line, too. :-)

dhower-qc:
> I think you have to add arch/ext/Zclsd.yaml, I do not see it yet. As well, need to change the following files: arch/inst/I/sd.yaml arch/inst/I/ld.yaml arch/inst/C/c.ld.yaml arch/inst/C/c.ldsp.yaml arch/inst/C/c.sdsp.yaml arch/inst/C/c.sd.yaml
> 
> For each of them you should add condition that these instructions for RV32 only should be definedBy Zilsd/Zclsd respectively, while for RV64 still remain be definedBy I/C.

Good point. This is another new concept for the schema. I think we'll have to split `definedBy` by XLEN, similar to how we can split encodings.

arch/ext/Zilsd.yaml
AFOliveira: +1 on this, instructions actually are identified like this one each extension, but it is up to the tooling to generate it, since they already coupled by extensions
image

ThinkOpenly:
> I think you have to add arch/ext/Zclsd.yaml, I do not see it yet.

That could be in a separate PR.

> As well, need to change the following files: arch/inst/I/sd.yaml arch/inst/I/ld.yaml arch/inst/C/c.ld.yaml arch/inst/C/c.ldsp.yaml arch/inst/C/c.sdsp.yaml arch/inst/C/c.sd.yaml
> 
> For each of them you should add condition that these instructions for RV32 only should be definedBy Zilsd/Zclsd respectively, while for RV64 still remain be definedBy I/C.

Hmm, what's the best way to do that? Would it be to remove the "base" definition in those instructions, but then add a condition on `xlen()` and `implemented?(ExtensionName:Zilsd)` and add new code for this extension, else throw an exception?

One possibility for now is to accept this PR to define the extension only, and open a new issue for modifying the IDL. (I don't think the Sail code handles this extension, either, but that's an upstream issue.)

dhower-qc:
> That could be in a separate PR.

Yep.

> Hmm, what's the best way to do that? Would it be to remove the "base" definition in those instructions, but then add a condition on xlen() and implemented?(ExtensionName:Zilsd) and add new code for this extension, else throw an exception?

Assuming we have to modify `definedBy` to understand something different for RV32/RV64, then I think the condition just needs to check `xlen()`; in RV32, Zilsd will have to be implemented to decode the instruction.

> One possibility for now is to accept this PR to define the extension only, and open a new issue for modifying the IDL. (I don't think the Sail code handles this extension, either, but that's an upstream issue.)

I'm ok with that. @Shehrozkashif, do you also want to do the IDL changes?

To recap, I think we need 3 new issues:

- Zclsd extension
- Implement split definedBy
- IDL for Zilsd / Zclsd

Shehrozkashif:
@dhower-qc 
Sure! If you want me to work on it, I'll be happy to do so

arch/ext/Zilsd.yaml
Shehrozkashif: Sure ill include more text

arch/ext/Zilsd.yaml
```diff
@@ -0,0 +1,16 @@
+$schema: "ext_schema.json#"
+kind: extension
+name: Zilsd
+long_name: Load/Store Pair for RV32 (Zilsd)
```
dhower-qc, line 4: We should remove (Zilsd).

dhower-qc (changes requested):

ThinkOpenly (changes requested):
In addition to the inline comments, could you shorted the commit subject line?
```
Subject: [PATCH] Added support for the Zilsd extension, which introduces load/store pair instructions for RV32. These instructions allow efficient 64-bit memory operations using register pairs. Ensured proper alignment rules, trap handling, and atomicity requirements. This improves memory efficiency and reduces instruction count for doubleword operations.

As a general rule, commit subject lines should easily fit on a single line. I know reviewers who prefer 50 characters max.

It should also be in imperative tense.

The single commit here could have a subject of "Add Zilsd extension YAML".

File: arch/ext/Zilsd.yaml (hunk 3 above)
Comment (Shehrozkashif) [reply arch/ext/Zilsd.yaml:8]: Got it! I’ll make sure to use the exact wording from the ISA spec whenever possible and keep any modifications minimal and clearly justified.

File: arch/ext/Zilsd.yaml [hunk 7]
```diff
@@ -0,0 +1,18 @@
...
+long_name: Load/Store pair instructions
+description: |
+  As the access size is 64-bit, accesses are only considered naturally aligned for effective addresses
+  that are a multiple of 8. In this case, these instructions are guaranteed to not raise an address-
+  misaligned exception. Even if naturally aligned, the memory access might not be performed atomically.
+  If the effective address is a multiple of 4, then each word access is required to be performed
+  atomically.
+  To ensure resumable trap handling is possible for the load instructions, the base register must have its
+  original value if a trap is taken. The other register in the pair can have been updated. This affects x2
+  for the stack pointer relative instruction and rs1 otherwise.
```
ThinkOpenly, line 13: This text isn't what I was expecting, as it doesn't describe the extension.

I included the text I suggested in one of my comments:
> This specification adds load and store instructions using register pairs. It does so by reusing existing instruction encodings which are RV64-only. The specification defines 32-bit encodings.
> 
> Load and store instructions will use the same definition of even-odd pairs as defined by the Zdinx extension.
> 
> The extension improves static code density, by replacing two separate load or store instructions with a single one. In addition, it can provide a performance improvement for implementations that can make use of a wider than XLEN memory interface.

(This is from chapter 1 of the extension specification, modified slightly to better suit the context here, since this is only for one of the extensions described in the specification.)

Shehrozkashif: Sure, I'll add the text you've provided from chapter 1

arch/ext/Zilsd.yaml
ThinkOpenly: The change here went from "Load/Store Pair for RV32 (Zilsd)" to "Load/Store pair instructions". The title of the spec is "Load/Store Pair for RV32 (Zilsd &
Zclsd)". Since we are covering exactly one of the included extensions here, it certainly makes sense to remove "(Zilsd)", being redundant. However, the title of the spec should be preserved verbatim. Please use "Load/Store Pair for RV32".

AFOliveira (approved):

arch/inst/Zilsd/ld.yaml
```diff
@@ -0,0 +1,37 @@
...
+description: |
+  Loads a 64-bit value into registers rd and rd+1. The effective address is obtained by adding
+  register rs1 to the sign-extended 12-bit offset.
+definedBy: Zilsd
+assembly: rd, offset(rs1)
+encoding:
+  match: -----------------011-----0000011
+  variables:
+    - name: rd
+      location: 11-7
```
ThinkOpenly, line 16: Thinking more about this, we have the capability to screen out the invalid (odd) values here, so let's go ahead and do that with:
```
  not: [ 1, 3, 5, 7, 9, 11, 13, 15, 17, 19, 21, 23, 25, 27, 29, 31 ]
```

arch/inst/Zilsd/ld.yaml
```diff
@@ -0,0 +1,37 @@
...
+  vs: always
+  vu: always
+operation(): |
+  Bits<XLEN> base = X[rs1];
+  Bits<XLEN> offset = $signed(imm);
+  Bits<XLEN> eff_addr = base + offset;
+
+  Bits<64> data = read_memory<64>(eff_addr, $encoding);
+
+  if ((rd & 1) == 0) {
```
ThinkOpenly, line 33: And if we screen out the values with `not` above, we no longer need this conditional.

arch/inst/Zilsd/sd.yaml
```diff
@@ -0,0 +1,36 @@
...
+  register rs1 to the sign-extended 12-bit offset.
+definedBy: Zilsd
+assembly: rs2, offset(rs1)
+encoding:
+  match: -----------------011-----0100011
+  variables:
+    - name: rs1
+      location: 19-15
+    - name: rs2
+      location: 24-20
```
ThinkOpenly, line 18: Same concept here, so add:
```
  not: [ 1, 3, 5, 7, 9, 11, 13, 15, 17, 19, 21, 23, 25, 27, 29, 31 ]
```

ThinkOpenly (changes requested):
I couldn't figure out your commit structure. I see two commits: add Zilsd.yaml, and a merge commit with lots of content. Would it be possible to have just two commits (or 1) with the 3 new files?

If you are modifying the commits, feel free to reduce your commit message to:
> Add Zilsd extension

The rest of the text is redundant with the content.

(If you have questions, please ask. :-) )

arch/inst/Zilsd/ld.yaml
```diff
@@ -0,0 +1,37 @@
...
+  vu: always
+operation(): |
+  Bits<XLEN> base = X[rs1];
+  Bits<XLEN> offset = $signed(imm);
+  Bits<XLEN> eff_addr = base + offset;
+
+  Bits<64> data = read_memory<64>(eff_addr, $encoding);
+
+    X[rd] = data[31:0];
+    X[rd+1] = data[63:32];
```
ThinkOpenly, line 35: nit: could you align this code with the code above it? (2 spaces less indent.)

ThinkOpenly (approved):

ThinkOpenly:
"This branch is out-of-date with the base branch" :slightly_frowning_face:

Could you rebase and push once more? Sorry for all the trouble.

ThinkOpenly:
> "This branch is out-of-date with the base branch" 🙁
> 
> Could you rebase and push once more? Sorry for all the trouble.

Nevermind. It's mergeable now.

ThinkOpenly:
I must've squashed the commits. Not what I intended. Too late. Must be more careful next time. :slightly_frowning_face:

ThinkOpenly:
Huh. I didn't realize that. All of the merge types are enabled, but the "protect branch" rule for the main branch forces use of the merge queue, which in turn forces squash merges. It sure would be nice to allow the merger to specify which type of merge to use when adding to the merge queue.

ThinkOpenly:
I'm not alone in my thinking: https://github.com/orgs/community/discussions/68723

;-)

AFOliveira:
I agree, that's where I was getting with my issue on merging ways on the other day

ThinkOpenly:
> I agree, that's where I was getting with my issue on merging ways on the other day

Now I understand your earlier comments there. I had thought it was "which is preferred among the possible choices available". Rather, it was "do we really want everything squashed". It should be a "false choice", but I now see that it is not. With the merge queue, there can be only one method.

PR #552, opened 2025-03-26
[WIP] ISA manual integration

dhower-qc:
This branch/PR will serve as the proof-of-concept for integrating riscv-isa-manual with riscv-unified-db


dhower-qc:
@wmat, this is ready.

You can generate the ISA manual with:

```bash
```

interesting issue with this: I was trying to find the commit in UDB corresponding to ISA manual 20240411, but couldn't. I realized that's because the commit never made it into main (which is the branch I pulled with subtree), but instead lives on a dead-end branch [v20240411](https://github.com/riscv/riscv-isa-manual/tree/v20240411). So, for now, I set the SHA for 20240411 to the nearest common ancestor.

wmat:
If I remember correctly there was no specific commit for 20240411 but rather a release was manually generated for it.

wmat (approved):
This all seems reasonable to me.  Do you want to just point @aswaterman at this PR?

wmat:
Another question, will you simply rebase against the riscv-isa-manual daily or on a schedule to pull in changes?  And if yes, does @aswaterman 's workflow change at all?

PR #553, opened 2025-03-27
spike support

AFOliveira:
This need to go after the Golang support, I'll fix the commit history once that is merged.

For the output difference check:
https://www.diffchecker.com/22skd4o8/

There are some differences that relate to how UDB views the data, but I think that is not entirely related to this PR and needs proper discussion on UDB vs riscv-opcodes on those particular instructions.

Also, ordering is messed up, but riscv-opcodes does not enforce any ordering so it is impossible to match the original order


backends/generators/generator.py
```diff
@@ -0,0 +1,422 @@
...
+                # Use address if available, otherwise use indirect_address
+                addr_to_use = address if address is not None else indirect_address
+                if isinstance(addr_to_use, int):
+                    addr_int = addr_to_use
+                else:
+                    addr_int = int(addr_to_use, 0)
+
+                # For BOTH architecture, add suffix to RV32-specific CSRs
+                if target_arch == "BOTH" and base == 32:
+                    csrs[addr_int] = f"{name.upper()}.RV32"
```
ThinkOpenly, line 381: This will eventually end up emitting invalid C code:
```
#define CSR_CYCLEH.RV32 0xc80
```

Preprocessor tokens cannot have ".". Maybe use underscore here?

backends/generators/spike/generate_encoding.py
```diff
@@ -0,0 +1,226 @@
...
+        (0xA, "virtual supervisor ecall"),
+        (0xB, "machine ecall"),
+        (0xC, "fetch page fault"),
+        (0xD, "load page fault"),
+        (0xF, "store page fault"),
+        (0x14, "fetch guest page fault"),
+        (0x15, "load guest page fault"),
+        (0x16, "virtual instruction"),
+        (0x17, "store guest page fault"),
+    ]
```
dhower-qc, line 124: Ideally, we should pull this from UDB; the information is already there, and then this list will stay up to date over time without intervention.

dhower-qc, line 124: Ideally, we should pull this from UDB; the information is already there, and then this list will stay up to date over time without intervention.

dhower-qc (changes requested):

backends/generators/spike/generate_encoding.py
AFOliveira: I had no clue we had this information, I'll do that change

jordancarlin:
This generated header is a generic C header and isn't specific to Spike. It is used in several other places already (eg riscv-arch-test: https://github.com/riscv-non-isa/riscv-arch-test/blob/dev/riscv-test-suite/env/encoding.h and custom tests for the Sail model: https://github.com/riscv/sail-riscv/blob/master/test/first_party/src/common/encoding.h). It probably makes more sense to name the generator for it C or C-header or something along those lines to account for this.

AFOliveira:
> This generated header is a generic C header and isn't specific to Spike. It is used in several other places already (eg riscv-arch-test: https://github.com/riscv-non-isa/riscv-arch-test/blob/dev/riscv-test-suite/env/encoding.h and custom tests for the Sail model: https://github.com/riscv/sail-riscv/blob/master/test/first_party/src/common/encoding.h). It probably makes more sense to name the generator for it C or C-header or something along those lines to account for this.

Oh, cool! Thank you for that information, Jordan! I will of course change it, it is even better this way.

backends/generators/spike/generate_encoding.py
AFOliveira: I can't find this data, where is it?

AFOliveira:
Hi @jordancarlin Are you familiar with those code bases? There are some questions I have, maybe we could set up a call and try to understand how much impact small changes to those files would have on the downstream users

jordancarlin:
Any changes would definitely also need to be run by the Spike community, but I’d be happy to speak to how some of the other projects are using it.

backends/generators/spike/generate_encoding.py
dhower-qc: https://github.com/riscv-software-src/riscv-unified-db/blob/215341e718b575a3039f2dacce1d4a8432825ab7/arch/ext/Sm.yaml#L103-L169

AFOliveira: I already commited the changes, the names don't exactly match riscv-opcodes but I do not think that is a problem. Thanks for the pointer!

AFOliveira:
enconding.out.h is only there fore review purposes, I'll remove it before merging. It would be great to see this merged before the summit

backends/generators/tasks.rake
```diff
@@ -34,4 +32,35 @@ namespace :gen do
...
     sh "#{$root}/.home/.venv/bin/python3 #{$root}/backends/generators/Go/go_generator.py --inst-dir=#{inst_dir} --csr-dir=#{csr_dir} --output=#{output_dir}inst.go"
   end
+
+  desc <<~DESC
+    Generate C encoding header from RISC-V instruction and CSR definitions
+    This is used by Spike, ACTs and the Sail Model
+
+    Options:
+     * CONFIG - Configuration name (defaults to "_")
+     * OUTPUT_DIR - Output directory for generated Spike headers (defaults to "#{$root}/gen/c_header")
```
jordancarlin, line 42:
```suggestion
     * OUTPUT_DIR - Output directory for generated C header (defaults to "#{$root}/gen/c_header")
```

backends/generators/tasks.rake
```diff
@@ -34,4 +32,35 @@ namespace :gen do
...
+    csr_dir = $root / "arch" / "csr"
+
+    # If using a specific config other than the default, use the resolved arch
+    if config_name != "_"
+      inst_dir = $root / "gen" / "resolved_arch" / config_name / "inst"
+      csr_dir = $root / "gen" / "resolved_arch" / config_name / "csr"
+      ext_dir = $root / "gen" / "resolved_arch" / config_name / "ext"
+    end
+
+    # Run the Spike generator script using the same Python environment
```
jordancarlin, line 62:
```suggestion
    # Run the C header generator script using the same Python environment
```

backends/generators/tasks.rake
```diff
@@ -34,4 +32,35 @@ namespace :gen do
...
+    # The script generates encoding.h for Spike simulator
```
jordancarlin, line 63: Maybe something like this instead?
```suggestion
    # The script generates encoding.h for inclusion in C programs
```

backends/generators/c_header/encoding.out.h
```diff
@@ -0,0 +1,4206 @@
...
+#define CSR_MHPMCOUNTER23 0xb17
+#define CSR_MHPMCOUNTER24 0xb18
+#define CSR_MHPMCOUNTER25 0xb19
+#define CSR_MHPMCOUNTER26 0xb1a
+#define CSR_MHPMCOUNTER27 0xb1b
+#define CSR_MHPMCOUNTER28 0xb1c
+#define CSR_MHPMCOUNTER29 0xb1d
+#define CSR_MHPMCOUNTER30 0xb1e
+#define CSR_MHPMCOUNTER31 0xb1f
+#define CSR_MCYCLEH.RV32 0xb80
```
jordancarlin, line 2556: This needs to use an underscore instead of a dot. Alternatively, the version from riscv-opcodes doesn't add any suffix to the CSRs since they don't have overlapping encodings with RV64 versions (unlike the instructions where the suffix is necessary). RV32 only CSRs can be identified by the `H` already anyway.

AFOliveira: Ups, I have an outdated version in the checked out encoding.out.h, this has been fixed

AFOliveira:
Thanks for the suggestions @jordancarlin

ThinkOpenly (approved):

ThinkOpenly:
I now possess the power to merge this. :muscle: I see you addressed Derek's suggestion to extract the interrupt codes from UDB. I see no other issues/comments/suggestions. Absent objection by end-of-my-day (say 6 hours from now), I'll go ahead and merge this.

AFOliveira:
> I now possess the power to merge this. 💪 I see you addressed Derek's suggestion to extract the interrupt codes from UDB. I see no other issues/comments/suggestions. Absent objection by end-of-my-day (say 6 hours from now), I'll go ahead and merge this.

Glad to see you're using your powers for good—merging my PRs! 🤣

dhower-qc (approved):

jordancarlin:
Looks like generating the C header wasn't added to regression. I see that generating the Go file is, and this seems similar, so I assume it should be?

PR #571, opened 2025-03-27
documentation updates

dhower-qc:


dhower-qc:
It may be easier to review this by looking at the branch:

https://github.com/riscv-software-src/riscv-unified-db/tree/readme

AFOliveira:
Should we enforce a READme per feature? I think it goes with what @ayosher proposed, each ./do should have proper documenation of flags and how to use it. That could also go into the contributing guidelines - if you add a new feature, create enough documentation that everyone can use it without having to read your source code

dhower-qc:
Yes, those definitely need to be documented better. If we put the documentation for them in a README file, then we should also make sure that text shows up with `do --desc TASK`. For example,

```ruby
# tasks.rake

desc File.read('path/to/readme')
task :manual
# ...
```

dhower-qc:
I added a README for gen:html_manual so we can see what it would look like.  This is what you see:

```
hu-dhower-lv:readme (4) ./do --desc gen:html_manual
Using singularity environment
Running with 1 job(s)
do gen:html_manual
    Generate an HTML site for one or more versions of the manual (./do --desc for options)
    
    == Usage
    
    Options:
    
      * MANUAL_NAME: The database name (key) of the manual to generate.
      * VERSIONS: A comma-separated list of versions to generate, or "all".
    
    Examples:
    
      ./do gen:html_manual MANUAL_NAME=isa VERSIONS=20191008,20240411
    
    Result:
    
      A static HTML website will be written into gen/manual/MANUAL_NAME/<hash of versions>/html

```

README.adoc
```diff
@@ -1,73 +1,58 @@
...
+* `cfgs`: Configurations used by backends to customize generated outputs
+* `schemas`: Data schemas for the content in `arch`
 
-If it is not installed, either ask your IT admin or:
+Additionally, developers will be interested in the following:
 
- * For Apptainer, see https://apptainer.org/docs/admin/main/installation.html[Apptainer Installation].
- * For Singularity CE, see https://docs.sylabs.io/guides/latest/admin-guide/installation.html[Singularity CE Installation].
+* `backends`: Tools to generate artifacts (documents, simulators, etc.) from UDB data
+* `bin`: Wrapper scripts to run commands within the container enviornment
```
ThinkOpenly, line 48: s/enviornment/environment/

README.adoc
```diff
@@ -87,26 +72,21 @@ You can then run `Dev Containers: Open Folder in Container...` from the Command
 
 You can start a GitHub Codespace for this repository by clicking the "Code" button and selecting "Open with Codespaces".
 
-== Tasks
+=== Tasks
 
-Quick start:
+UnifiedDB uses https://github.com/ruby/rake[Rake], a Ruby-based Makefile alternative, to manage tasks.
+For convienence, running Rake inside the container is encapulated in `do`. For example:
```
ThinkOpenly, line 78: s/convienence/convenience/

ThinkOpenly: and s/encapulated/encapsulated/

dhower-qc: 🦤

AFOliveira:
On the contributing guidelines - I think we should have one guideline that asks authors that whenever a new task is added, all of its flags should be explicit on the Rakefile description.

dhower-qc:
> On the contributing guidelines - I think we should have one guideline that asks authors that whenever a new task is added, all of its flags should be explicit on the Rakefile description.

Maybe that detail should be in a "New backend HOW TO"?

.gitmessage
```diff
@@ -0,0 +1,29 @@
...
+# BREAKING CHANGE: <summary of what breaks>
+#
+# <Breaking change description + migration instructions>
+
+# At the end: Include Co-authored-by if relevant.
+# Co-authored-by: name <user@users.noreply.github.com>
+
+# common title types:
+#   build:    build system
+#   ci:       continous integeration
```
ThinkOpenly, line 17: spell-o: "integration"

dhower-qc: fixed

dhower-qc:
@ThinkOpenly, ok with this? We should probably get this merged and then continue discussions on additions with new PRs.

ThinkOpenly (approved):

Issue #512, opened 2025-03-19
Add Smctr extension data


dhower-qc:
Add YAML (name, versions) for Smctr.

PR #573, opened 2025-03-28
Add Smctr extension and SCTRCLR instruction

syedowaisalishah:
## Description
- Added the **Smctr** extension, which includes Control Transfer Record management in Machine and Supervisor modes.
- Added the **SCTRCLR** instruction to clear the CTR buffer and reset related registers.

Fixes issue #512.

### DIFF
--- a/arch/ext/Smctr.yaml
+++ b/arch/ext/Smctr.yaml
@@ -1,0 +1,78 @@
+---
+# yaml-language-server: $schema=../../schemas/ext_schema.json
+
+$schema: "ext_schema.json#"
+kind: extension
+name: Smctr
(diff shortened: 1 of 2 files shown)


ThinkOpenly:
Unfortunately, this is a duplicate of #554.

#554 should've mentioned #512 somewhere so a reference would be put in #512.

syedowaisalishah:
I understand now that #573 is a duplicate of #554, and I appreciate the clarification. Should I go ahead and close #573?

dhower-qc:
Yes, let's close this one. My bad on not assigning myself to the issue. I will be more diligent about that going forward.

syedowaisalishah:
Thanks for the confirmation! No worries at all. I'll make sure to stay on top of the issue moving forward as well.

PR #574, opened 2025-03-28
Display rv32/rv64 encodings without tabs in HTML view

### LINKED ISSUE #488: ui suggestion: make rv32/rv64 encoding differences more obvious
i regularly accidentally look at the rv32 encoding for an instruction when there's a difference between rv32 and rv64.

personally, i'd like to see the current rv32/rv64 tabs disappear and just always show both encodings.

the obvious alternative would be to default to showing rv64 --- which would be fine for _me_, but cause the microcontroller folks to file the bug asking for it to be changed back, because they'd be having the problem instead now :-)

**Additional context**
slli is an example instruction, where the shift amount is one bit wider for rv64: https://riscv-software-src.github.io/riscv-unified-db/manual/html/isa/isa_20240411/insts/slli.html

(note that there's a similar problem with the "Decode Variables" section too. again, personally i'd like to just see both [given that defaulting to rv64 would just move the problem over to the microcontroller folks rather than fixing it].)

dhower-qc:
closes #488


AFOliveira:
Shall I also PR all the instructions erb templates with the necessary changes to put the numbers in each cell?

ThinkOpenly (approved):
Any before and after screenshots?
Regardless, seems like an extremely simple change, so I'm approving.

PR #575, opened 2025-03-28
Rename Sm#XLEN parameter -> Sm#MXLEN

### LINKED ISSUE #52: Rename XLEN to MXLEN

dhower-qc:
closes #52


ThinkOpenly:
_[Showing my ignorance]_ Can you elaborate a bit on the algorithm for what's changing?

Is it any use of "XLEN" when in M mode?

dhower-qc:
> _[Showing my ignorance]_ Can you elaborate a bit on the algorithm for what's changing?
> 
> Is it any use of "XLEN" when in M mode?

We've been using `XLEN` as the name of the parameter that represents register width in M-mode. It should have been `MXLEN` all along. There is also the concept of `XLEN` (or `xlen()` in IDL) that represents the current register width given your current mode and CSR settings, which is why it's important to get this renamed.

dhower-qc:
@ThinkOpenly, @AFOliveira: I'd like to get this one in soon. The longer it says in the queue, the more likely we are to miss a bug from concurrent work.

Any changes you want to see?

AFOliveira (approved):

arch/csr/Zihpm/hpmcounter10h.yaml
```diff
@@ -7,7 +7,6 @@ kind: csr
 name: hpmcounter10h
 long_name: User-mode Hardware Performance Counter 7, high half
 address: 0xC8A
-base: 32
```
ThinkOpenly, line 10: Wait. These changes shouldn't be here.

ThinkOpenly (changes requested):

arch/csr/Zihpm/hpmcounter10h.yaml
dhower-qc: Good catch, fixed now

ThinkOpenly (approved):

PR #581, opened 2025-03-31
Zihmp extension: H-registers should be there only for RV32

ayosher:


ThinkOpenly:
Could you fix the spelling mistake in your commit subject line? s/Zihmp/Zihpm/

arch/csr/Zihpm/hpmcounter10h.yaml
```diff
@@ -7,7 +7,6 @@ kind: csr
 name: hpmcounter10h
 long_name: User-mode Hardware Performance Counter 7, high half
 address: 0xC8A
-base: 32
```
ThinkOpenly, line 10: Wait. How does removing the restriction for base 32 only make these available for only base 32?

ayosher: You are right, some mess in my workspace. Let me close PR and delete the branch, sorry for confusion.
Thank you for pointing that out.

PR #585, opened 2025-04-01
Add Ruby typechecking with Sorbet and Tapioca

dhower-qc:


arch/csr/mie.yaml
```diff
@@ -8,7 +8,151 @@ address: 0x304
 priv_mode: M
 length: MXLEN
 definedBy: Sm
-description: "mip.yaml#/description"
+description: |
```
AFOliveira, line 11: Is this supposed to be in this PR?

AFOliveira (changes requested):
The changes look good to me, aside from that small CSR which seems misplaced in the PR.

I do feel like the bar for contributing to UDB’s codebase is a bit higher than it needs to be. It seems difficult for folks who haven’t worked on the project before to make even small changes. I think that’s part of why there’s a lot of interest in UDB, but not as many people actively contributing. As long as we're around and maintaining it, things should be fine—but I do wonder if this might cause issues in the long run.

I don’t have strong opinions on this. I’m not familiar with Sorbet yet, but I’ll learn it. The fact that it’s not mandatory right now feels like a good way to ease into it.

arch/csr/mie.yaml
dhower-qc: No, that's unrelated. It was just a bug I caught along the way. I'll pull it out into another PR.

ThinkOpenly (approved):
I'm more of a rube than a Ruby programmer, but doesn't seem too bad, especially if adoption for new code is currently optional.

AFOliveira (approved):

PR #587, opened 2025-04-02
task: sort apt pkgs

kbroch-rivosinc:
In sorting I noticed a difference between the two lists.  Should they be the same?


dhower-qc:
> In sorting I noticed a difference between the two lists. Should they be the same?

Yes, they should be the same. The Dockerfile is missing a few.

PR #588, opened 2025-04-02
Fix mret and mnret instructions 

ayosher:
Add support for Smdbltrp (MDT bit) and MPRV bit


arch/inst/I/mret.yaml
```diff
@@ -14,9 +14,16 @@ access:
 encoding:
   match: "00110000001000000000000001110011"
 operation(): |
-  if (implemented?(ExtensionName::S) && CSR[mstatus].MPP != 2'b11) {
+  if (CSR[mstatus].MPP != 2'b11) {
```
ThinkOpenly, line 17: Why was the `implemented` check removed?

arch/inst/Smrnmi/mnret.yaml
```diff
@@ -3,9 +3,9 @@
 $schema: inst_schema.json#
 kind: instruction
 name: mnret
-long_name: No synopsis available.
+long_name: Machine NMI Return
```
ThinkOpenly, line 6: Where did this text come from?

What I find is that the "mnret" instruction is defined by "Smrnmi" extension for resumable non-maskable interrupts. And these textual references to this particular instruction:
> MNRET, to resume from the RNMI handler.

> MNRET is an M-mode-only instruction [...] to return to the program counter, privilege mode, and virtualization mode of the interrupted context.

... no great references.

M = Machine. Fine.

Given the first reference, and specifically the use of "RNMI", perhaps the "long_name" should be something like "Machine Resume from RNMI handler".

arch/inst/Smrnmi/mnret.yaml
```diff
@@ -3,9 +3,9 @@
...
 description: |
-  No description available.
+  Returns from an NMI in M-mode.
```
ThinkOpenly, line 8: Same question. Given a very strong preference for using normative text, I prefer:
> `MNRET` is an M-mode-only instruction that uses the values in `mnepc` and `mnstatus` to return to the program counter, privilege mode, and virtualization mode of the interrupted context. This instruction also sets `mnstatus.NMIE`. If `MNRET` changes the privilege mode to a mode less privileged than M, it also sets `mstatus.MPRV` to 0. If the Zicfilp extension is implemented, then if `mnstatus.MNPP` holds the value _y_, `MNRET` sets `ELP` to the logical AND of `y`LPE and `mnstatus.MNPELP`.

arch/inst/Smrnmi/mnret.yaml
```diff
@@ -18,9 +18,17 @@ access:
   vu: always
 data_independent_timing: false
 operation(): |
-  if (implemented?(ExtensionName::S) && CSR[mstatus].MPP != 2'b11) {
+  if (CSR[mnstatus].MNPP != 2'b11) {
```
ThinkOpenly, line 21: MPP changed to MNPP. Is that intentional? Why? This needs to be explained in the commit message.

ThinkOpenly (changes requested):
One area of this project we're trying to improve is the commit messages. I'd appreciate it (and it would certainly help me review an area with which I'm not familiar) if the commit message included:
- description of the issue with the current code
- explanation of what was changed, and how that fixes the issue

arch/inst/I/mret.yaml
ayosher: Because spec says: 
"An MRET or SRET instruction is used to return from a trap in M-mode or S-mode respectively. When
executing an xRET instruction, supposing xPP holds the value y, xIE is set to xPIE; the privilege mode
is changed to y; xPIE is set to 1; and xPP is set to the least-privileged supported mode (U if U-mode is
implemented, else M). If y≠M, xRET also sets MPRV=0."
So, since MPP != M (2'b11) we should reset MPRV. Nothing said about need that S mode will be implemented.

arch/inst/Smrnmi/mnret.yaml
ayosher: I will fix

dhower-qc (approved):

ThinkOpenly (approved):

Issue #547, opened 2025-03-26
Add Smstateen/Sstateen extension and CSRs


dhower-qc:
assigned to @neverlandiz (can you please reply to this thread so I can assign you?)

neverlandiz:
sounds good

ThinkOpenly:
Marking complete via #592 .

Issue #557, opened 2025-03-27
Add Smcsrind/Sscsrind YAML


dhower-qc:
Add extension and CSR files (excluding IDL) for Smcsrind/Sscsrind

Need to add:

- MISELECT
- MIREG
- MIREG2
- MIREG3
- MIREG4
- MIREG5
- MIREG6
- SISELECT
- SIREG
- SIREG2
- SIREG3
- SIREG4
- SIREG5
- SIREG6
- VSISELECT
- VSIREG
- VSIREG2
- VSIREG3
- VSIREG4
- VSIREG5
- VSIREG6

syedowaisalishah:
I would like to work on this issue. Can you assign it to me?

syedowaisalishah:
@dhower-qc I’ve added the extension YAML files for Smcsrind and Sscsrind. For the CSR YAML files, could you recommend a structure that I should follow? Additionally, should I create separate folders for MISELECT, SISELECT, and VSISELECT CSRs?

dhower-qc:
> could you recommend a structure that I should follow?

Any YAML file under `arch/csr` will have the general structure. The key thing is to fill in the correct data for the CSR (like address) and then a list of fields (which must be present, even if there is a single 'field' for the entire CSR).

> should I create separate folders for ex

We should create a `arch/csr/Smcsrind` folder and add all the CSR yaml files there.

syedowaisalishah:
@dhower-qc Thanks for the clarification! I’ll add all the Smcsrind CSR YAMLs under arch/csr/Smcsrind and make sure each file includes the correct address and fields, even if it's just one.

PR #592, opened 2025-04-04
Add Smstateen/Ssstateen Extension and CSRs

neverlandiz:
This PR addresses Issue #547 and adds the Smstateen extension, along with the following Smstateen/Ssstateen CSRs
- mstateen0, mstateen1, mstateen2, mstateen3
- mstateen0h, mstateen1h, mstateen2h, mstateen3h
- hstateen0, hstateen1, hstateen2, hstateen3
- hstateen0h, hstateen1h, hstateen2h, hstateen3h
- sstateen0, sstateen1, sstateen2, sstateen3

### DIFF
--- a/arch/csr/hstateen0.yaml
+++ b/arch/csr/hstateen0.yaml
@@ -1,0 +1,100 @@
+# yaml-language-server: $schema=../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: hstateen0
+long_name: Hypervisor State Enable 0 Register
+address: 0x60C
+priv_mode: S
+length: 64
+description: |
+  For each hstateen CSR, bit 63 is defined to control access to the matching sstateen CSR.
+  Bit 63 of hstateen0 controls access to sstateen0.
+
+  With the hypervisor extension, the hstateen CSRs have identical encodings to the mstateen
+  CSRs, except controlling accesses for a virtual machine (from VS and VU modes).
+
+  For every bit in an hstateen CSR that is zero (whether read-only zero or set to zero),
+  the same bit appears as read-only zero in sstateen when accessed in VS-mode.
+
+  A bit in an hstateen CSR cannot be read-only one unless the same bit is read-only one
+  in the matching mstateen CSR.
+
+definedBy:
+  allOf:
+    - H
+    - Smstateen
+    - Ssstateen
+fields:
+  SEO:
+    location: 63
+    description: |
+      The SE0 bit in hstateen0 controls access to the sstateen0 CSR.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  ENVCFG:
+    location: 62
+    description: |
+      The ENVCFG bit in hstateen0 controls access to the senvcfg CSRs.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  CSRIND:
+    location: 60
+    description: |
+      The CSRIND bit in hstateen0 controls access to the siselect and the
+      sireg*, (really vsiselect and vsireg*) CSRs provided by the Sscsrind
+      extensions.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  AIA:
+    location: 59
+    description: |
+      The AIA bit in hstateen0 controls access to all state introduced by
+      the Ssaia extension and is not controlled by either the CSRIND or the
+      IMSIC bits of hstateen0.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  IMSIC:
+    location: 58
+    description: |
+      The IMSIC bit in hstateen0 controls access to the guest IMSIC state,
+      including CSRs stopei (really vstopei), provided by the Ssaia extension.
+
+      Setting the IMSIC bit in hstateen0 to zero prevents a virtual machine
+      from accessing the hart’s IMSIC the same as setting `hstatus.`VGEIN = 0.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  CONTEXT:
+    location: 57
+    description: |
+      The CONTEXT bit in hstateen0 controls access to the scontext CSR provided
+      by the Sdtrig extension.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  JVT:
+    location: 2
+    description: |
+      The JVT bit controls access to the jvt CSR provided by the Zcmt extension.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  FCSR:
+    location: 1
+    description: |
+      The FCSR bit controls access to fcsr for the case when floating-point instructions
+      operate on x registers instead of f registers as specified by the Zfinx and related
+      extensions (Zdinx, etc.). Whenever misa.F = 1, FCSR bit of mstateen0 is read-only
+      zero (and hence read-only zero in hstateen0 and sstateen0 too). For convenience,
+      when the stateen CSRs are implemented and misa.F = 0, then if the FCSR bit of a
+      controlling stateen0 CSR is zero, all floating-point instructions cause an illegal
+      instruction trap (or virtual instruction trap, if relevant), as though they all access
+      fcsr, regardless of whether they really do.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  C:
+    location: 0
+    description: |
+      The C bit controls access to any and all custom state. This bit is not custom state itself.
+      The C bit of these registers is not custom state itself; it is a standard field of a
+      standard CSR, either mstateen0, hstateen0, or sstateen0.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
--- a/arch/csr/hstateen0h.yaml
+++ b/arch/csr/hstateen0h.yaml
@@ -1,0 +1,70 @@
+# yaml-language-server: $schema=../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: hstateen0h
+long_name: Upper 32 bits of Hypervisor State Enable 0 Register
+address: 0x61C
+priv_mode: S
+length: 32
+description: |
+  For RV64 harts, the Smstateen/Ssstateen extension adds four new 64-bit CSRs at machine level: mstateen0 (Machine State Enable 0),
+  mstateen1, mstateen2, and mstateen3. If supervisor mode is implemented, another four CSRs are defined at
+  supervisor level: sstateen0, sstateen1, sstateen2, and sstateen3. And if the hypervisor extension is implemented,
+  another set of CSRs is added: hstateen0, hstateen1, hstateen2, and hstateen3.
+
+  For RV32, the registers listed above are 32-bit, and for the machine-level and hypervisor CSRs there is a corresponding
+  set of high-half CSRs for the upper 32 bits of each register: mstateen0h, mstateen1h, mstateen2h, mstateen3h, hstateen0h,
+  hstateen1h, hstateen2h, and hstateen3h.
+
+definedBy:
+  allOf:
+    - H
+    - Smstateen
+    - Ssstateen
+fields:
+  SEO:
+    location: 31
+    description: |
+      The SE0 bit in hstateen0h controls access to the sstateen0 CSR.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  ENVCFG:
+    location: 30
+    description: |
+      The ENVCFG bit in hstateen0h controls access to the senvcfg CSRs.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  CSRIND:
+    location: 28
+    description: |
+      The CSRIND bit in hstateen0h controls access to the siselect and the
+      sireg*, (really vsiselect and vsireg*) CSRs provided by the Sscsrind
+      extensions.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  AIA:
+    location: 27
+    description: |
+      The AIA bit in hstateen0h controls access to all state introduced by
+      the Ssaia extension and is not controlled by either the CSRIND or the
+      IMSIC bits of hstateen0.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  IMSIC:
+    location: 26
+    description: |
+      The IMSIC bit in hstateen0h controls access to the guest IMSIC state,
+      including CSRs stopei (really vstopei), provided by the Ssaia extension.
+
+      Setting the IMSIC bit in hstateen0h to zero prevents a virtual machine
+      from accessing the hart’s IMSIC the same as setting `hstatus.`VGEIN = 0.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  CONTEXT:
+    location: 25
+    description: |
+      The CONTEXT bit in hstateen0h controls access to the scontext CSR provided
+      by the Sdtrig extension.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
--- a/arch/csr/mstateen0.yaml
+++ b/arch/csr/mstateen0.yaml
@@ -1,0 +1,95 @@
+# yaml-language-server: $schema=../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: mstateen0
+long_name: Machine State Enable 0 Register
+address: 0x30C
+priv_mode: M
+length: 64
+description: |
+  For each mstateen CSR, bit 63 is defined to control access to the matching sstateen and hstateen
+  CSRs. Bit 63 of mstateen0 controls access to sstateen0 and hstateen0.
+
+  On reset, all writable mstateen bits are initialized by the hardware to zeros. If machine-level software
+  changes these values, it is responsible for initializing the corresponding writable bits of the hstateen
+  and sstateen CSRs to zeros too. Software at each privilege level should set its respective stateen CSRs
+  to indicate the state it is prepared to allow less-privileged software to access. For OSes and hypervisors,
+  this usually means the state that the OS or hypervisor is prepared to swap on a context switch, or to
+  manage in some other way.
+definedBy: Smstateen
+fields:
+  SEO:
+    location: 63
+    description: |
+      The SE0 bit in mstateen0 controls access to the hstateen0, hstateen0h, and the sstateen0 CSRs.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  ENVCFG:
+    location: 62
+    description: |
+      The ENVCFG bit in mstateen0 controls access to the henvcfg, henvcfgh, and the senvcfg CSRs.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  CSRIND:
+    location: 60
+    description: |
+      The CSRIND bit in mstateen0 controls access to the siselect, sireg*, vsiselect, and the vsireg*
+      CSRs provided by the Sscsrind extensions.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  AIA:
+    location: 59
+    description: |
+      The AIA bit in mstateen0 controls access to all state introduced by the Ssaia extension and is not
+      controlled by either the CSRIND or the IMSIC bits.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  IMSIC:
+    location: 58
+    description: |
+      The IMSIC bit in mstateen0 controls access to the IMSIC state, including CSRs stopei and vstopei,
+      provided by the Ssaia extension.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  CONTEXT:
+    location: 57
+    description: |
+      The CONTEXT bit in mstateen0 controls access to the scontext and hcontext CSRs provided by the
+      Sdtrig extension.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  P1P13:
+    location: 56
+    description: |
+      The P1P13 bit in mstateen0 controls access to the hedelegh introduced by Privileged Specification
+      Version 1.13.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  JVT:
+    location: 2
+    description: |
+      The JVT bit controls access to the jvt CSR provided by the Zcmt extension.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  FCSR:
+    location: 1
+    description: |
+      The FCSR bit controls access to fcsr for the case when floating-point instructions
+      operate on x registers instead of f registers as specified by the Zfinx and related
+      extensions (Zdinx, etc.). Whenever misa.F = 1, FCSR bit of mstateen0 is read-only
+      zero (and hence read-only zero in hstateen0 and sstateen0 too). For convenience,
+      when the stateen CSRs are implemented and misa.F = 0, then if the FCSR bit of a
+      controlling stateen0 CSR is zero, all floating-point instructions cause an illegal
+      instruction trap (or virtual instruction trap, if relevant), as though they all access
+      fcsr, regardless of whether they really do.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  C:
+    location: 0
+    description: |
+      The C bit controls access to any and all custom state. This bit is not custom state itself. The C bit of
+      these registers is not custom state itself; it is a standard field of a standard CSR, either mstateen0,
+      hstateen0, or sstateen0.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
--- a/arch/csr/sstateen0.yaml
+++ b/arch/csr/sstateen0.yaml
@@ -1,0 +1,71 @@
+# yaml-language-server: $schema=../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: sstateen0
+long_name: Supervisor State Enable 0 Register
+address: 0x10C
+priv_mode: S
+length: 64
+description: |
+  For the supervisor-level sstateen registers, high-half CSRs are not added at this
+  time because it is expected the upper 32 bits of these registers will always be zeros.
+
+  For every bit with a defined purpose in an sstateen CSR, the same bit is defined in the
+  matching mstateen CSR to control access below machine level to the same state.
+
+  Each bit of a stateen CSR controls less-privileged access to an extension’s state,
+  for an extension that was not deemed "worthy" of a full XS field in sstatus like the
+  FS and VS fields for the F and V extensions. The number of registers provided at each
+  level is four because it is believed that 4 * 64 = 256 bits for machine and hypervisor
+  levels, and 4 * 32 = 128 bits for supervisor level, will be adequate for many years to
+  come, perhaps for as long as the RISC-V ISA is in use. The exact number four is an
+  attempted compromise between providing too few bits on the one hand and going overboard
+  with CSRs that will never be used on the other.
+
+  The stateen registers at each level control access to state at all less-privileged levels,
+  but not at its own level.
+
+  When a stateen CSR prevents access to state for a privilege mode, attempting to execute in
+  that privilege mode an instruction that implicitly updates the state without reading it may
+  or may not raise an illegal instruction or virtual instruction exception. Such cases must
+  be disambiguated by being explicitly specified one way or the other.
+  In some cases, the bits of the stateen CSRs will have a dual purpose as enables for the ISA
+  extensions that introduce the controlled state.
+  Each bit of a supervisor-level sstateen CSR controls user-level access (from U-mode or VU-mode)
+  to an extension’s state. The intention is to allocate the bits of sstateen CSRs starting at the
+  least- significant end, bit 0, through to bit 31, and then on to the next-higher-numbered
+  sstateen CSR.
+
+definedBy:
+  allOf:
+    - Smstateen
+    - Ssstateen
+fields:
+  JVT:
+    location: 2
+    description: |
+      The JVT bit controls access to the jvt CSR provided by the Zcmt extension.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  FCSR:
+    location: 1
+    description: |
+      The FCSR bit controls access to fcsr for the case when floating-point instructions
+      operate on x registers instead of f registers as specified by the Zfinx and related
+      extensions (Zdinx, etc.). Whenever misa.F = 1, FCSR bit of mstateen0 is read-only
+      zero (and hence read-only zero in hstateen0 and sstateen0 too). For convenience,
+      when the stateen CSRs are implemented and misa.F = 0, then if the FCSR bit of a
+      controlling stateen0 CSR is zero, all floating-point instructions cause an illegal
+      instruction trap (or virtual instruction trap, if relevant), as though they all access
+      fcsr, regardless of whether they really do.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  C:
+    location: 0
+    description: |
+      The C bit controls access to any and all custom state. This bit is not custom state itself. The C bit of
+      these registers is not custom state itself; it is a standard field of a standard CSR, either mstateen0,
+      hstateen0, or sstateen0.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
--- a/arch/ext/Smstateen.yaml
+++ b/arch/ext/Smstateen.yaml
@@ -1,0 +1,25 @@
+# yaml-language-server: $schema=../../schemas/ext_schema.json
+
+$schema: "ext_schema.json#"
+kind: extension
+name: Smstateen
+long_name: Machine-mode view of the state-enable extension
+description: |
+  Machine-mode view of the state-enable extension. The Smstateen
+  extension specification comprises the mstateen*, sstateen*,
+  and hstateen* CSRs and their functionality.
+
+  NOTE: The Smstateen extension specification is an M-mode extension as
+  it includes M-mode features, but the supervisor-mode visible
+  components of the extension are named as the Ssstateen extension.  Only
+  Ssstateen is mandated in the RVA22S64 profile when the hypervisor
+  extension is implemented.  These registers are not mandated or
+  supported options without the hypervisor extension, as there are no
+  RVA22S64 supported options with relevant state to control in the
+  absence of the hypervisor extension.
+
+type: privileged
+versions:
+  - version: "1.0.0"
+    state: ratified
+    ratification_date: null
--- a/arch/csr/hstateen1.yaml
+++ b/arch/csr/hstateen1.yaml
@@ -1,0 +1,34 @@
+# yaml-language-server: $schema=../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: hstateen1
+long_name: Hypervisor State Enable 1 Register
+address: 0x60D
+priv_mode: S
+length: 64
+description: |
+  For each hstateen CSR, bit 63 is defined to control access to the matching sstateen CSR.
+  Bit 63 of hstateen1 controls access to sstateen1.
+
+  With the hypervisor extension, the hstateen CSRs have identical encodings to the mstateen
+  CSRs, except controlling accesses for a virtual machine (from VS and VU modes).
+
+  For every bit in an hstateen CSR that is zero (whether read-only zero or set to zero),
+  the same bit appears as read-only zero in sstateen when accessed in VS-mode.
+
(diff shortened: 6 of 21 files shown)


arch/csr/hstateen0.yaml
```diff
@@ -0,0 +1,100 @@
...
+  For every bit in an hstateen CSR that is zero (whether read-only zero or set to zero),
+  the same bit appears as read-only zero in sstateen when accessed in VS-mode.
+
+  A bit in an hstateen CSR cannot be read-only one unless the same bit is read-only one
+  in the matching mstateen CSR.
+
+definedBy:
+  allOf:
+    - H
+    - Smstateen
```
AFOliveira, line 26: Is Smstateen really part of this allOf?

"The Smstateen extension specification comprises the mstateen*, sstateen*, and hstateen*
CSRs and their functionality. The Ssstateen extension specification comprises only the sstateen* and
hstateen* CSRs and their functionality."

AFOliveira (changes requested):

arch/csr/hstateen0.yaml
ThinkOpenly: "Smstateen" is the base upon which the others build:
> For RV64 harts, this extension adds four new 64-bit CSRs at machine level: mstateen0 (Machine State Enable 0), mstateen1, mstateen2, and mstateen3.
> 
> If supervisor mode is implemented, another four CSRs are defined at supervisor level: sstateen0, sstateen1, sstateen2, and sstateen3.
> 
> And if the hypervisor extension is implemented, another set of CSRs is added: hstateen0, hstateen1, hstateen2, and hstateen3.

So, you can't have "hstateen0" without "Smstateen" (and "H", and presumably "Sstateen"?).

arch/csr/hstateen0.yaml
```diff
@@ -0,0 +1,100 @@
...
+    location: 2
+    description: |
+      The JVT bit controls access to the jvt CSR provided by the Zcmt extension.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  FCSR:
+    location: 1
+    description: |
+      The FCSR bit controls access to fcsr for the case when floating-point instructions
+      operate on x registers instead of f registers as specified by the Zfinx and related
```
ThinkOpenly, line 84: backtick `f` here.

General rule: if it's fixed-font-bold in the spec, it should be backticked in the description fields.

arch/csr/hstateen0.yaml
```diff
@@ -0,0 +1,100 @@
...
+      controlling stateen0 CSR is zero, all floating-point instructions cause an illegal
+      instruction trap (or virtual instruction trap, if relevant), as though they all access
+      fcsr, regardless of whether they really do.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  C:
+    location: 0
+    description: |
+      The C bit controls access to any and all custom state. This bit is not custom state itself.
+      The C bit of these registers is not custom state itself; it is a standard field of a
```
ThinkOpenly, line 97: Oddly redundant, but that's what the spec has. I guess we leave it for now. Interested in opening an issue against the spec?

arch/csr/hstateen0h.yaml
ThinkOpenly, line 1: This needs a "base: 32", as it is only defined for RV32.

ThinkOpenly (changes requested):
My comments in hstateen0/hstateen0h generally apply to the others as well.

arch/csr/sstateen0.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+  When a `stateen` CSR prevents access to state for a privilege mode, attempting to execute in
+  that privilege mode an instruction that implicitly updates the state without reading it may
+  or may not raise an illegal instruction or virtual instruction exception. Such cases must
+  be disambiguated by being explicitly specified one way or the other.
+  In some cases, the bits of the `stateen` CSRs will have a dual purpose as enables for the ISA
+  extensions that introduce the controlled state.
+  Each bit of a supervisor-level `sstateen` CSR controls user-level access (from U-mode or VU-mode)
+  to an extension’s state. The intention is to allocate the bits of `sstateen` CSRs starting at the
+  least- significant end, bit 0, through to bit 31, and then on to the next-higher-numbered
```
ThinkOpenly, line 37: remove the space in "least-significant", here and in 3 other places

arch/ext/Smstateen.yaml
```diff
@@ -0,0 +1,25 @@
+# yaml-language-server: $schema=../../schemas/ext_schema.json
+
+$schema: "ext_schema.json#"
+kind: extension
+name: Smstateen
+long_name: Machine-mode view of the state-enable extension
+description: |
+  Machine-mode view of the state-enable extension. The Smstateen
+  extension specification comprises the mstateen*, sstateen*,
+  and hstateen* CSRs and their functionality.
```
ThinkOpenly, line 10: a few more backtickies here

arch/csr/hstateen0.yaml
```diff
@@ -0,0 +1,100 @@
...
+  Bit 63 of hstateen0 controls access to sstateen0.
+
+  With the hypervisor extension, the hstateen CSRs have identical encodings to the mstateen
+  CSRs, except controlling accesses for a virtual machine (from VS and VU modes).
+
+  For every bit in an hstateen CSR that is zero (whether read-only zero or set to zero),
+  the same bit appears as read-only zero in sstateen when accessed in VS-mode.
+
+  A bit in an hstateen CSR cannot be read-only one unless the same bit is read-only one
+  in the matching mstateen CSR.
```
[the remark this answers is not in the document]
neverlandiz: Thanks for the feedback! I'm a bit confused about this part. I just added some `stateen` text to `hstateen0.yaml`, is this what you were looking for?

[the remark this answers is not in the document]
ThinkOpenly: How about this (with backticks needing to be added, of course :slightly_smiling_face:):
> Each bit of a stateen CSR controls less-privileged access to an extension’s state, for an extension that was not deemed "worthy" of a full XS field in sstatus like the FS and VS fields for the F and V extensions.
> 
> The stateen registers at each level control access to state at all less-privileged levels, but not at its own level. This is analogous to how the existing counteren CSRs control access to performance counter registers. Just as with the counteren CSRs, when a stateen CSR prevents access to state by less-privileged levels, an attempt in one of those privilege modes to execute an instruction that would read or write the protected state raises an illegal instruction exception, or, if executing in VS or VU mode and the circumstances for a virtual instruction exception apply, raises a virtual instruction exception instead of an illegal instruction exception.
> 
> When this extension is not implemented, all state added by an extension is accessible as defined by that extension.
> 
> When a stateen CSR prevents access to state for a privilege mode, attempting to execute in that privilege mode an instruction that implicitly updates the state without reading it may or may not raise an illegal instruction or virtual instruction exception. Such cases must be disambiguated by being explicitly specified one way or the other.
> 
> In some cases, the bits of the stateen CSRs will have a dual purpose as enables for the ISA extensions that introduce the controlled state.
> 
> With the hypervisor extension, the hstateen CSRs have identical encodings to the mstateen CSRs, except controlling accesses for a virtual machine (from VS and VU modes).
> 
> Each standard-defined bit of a stateen CSR is WARL and may be read-only zero or one, subject to the following conditions.
> 
> Bits in any stateen CSR that are defined to control state that a hart doesn’t implement are read-only zeros for that hart. Likewise, all reserved bits not yet given a defined meaning are also read-only zeros. For every bit in an mstateen CSR that is zero (whether read-only zero or set to zero), the same bit appears as read-only zero in the matching hstateen and sstateen CSRs. For every bit in an hstateen CSR that is zero (whether read-only zero or set to zero), the same bit appears as read-only zero in
sstateen when accessed in VS-mode.
> 
> A bit in a supervisor-level sstateen CSR cannot be read-only one unless the same bit is read-only one in the matching mstateen CSR and, if it exists, in the matching hstateen CSR. A bit in an hstateen CSR cannot be read-only one unless the same bit is read-only one in the matching mstateen CSR.
> 
> On reset, all writable mstateen bits are initialized by the hardware to zeros. If machine-level software changes these values, it is responsible for initializing the corresponding writable bits of the hstateen and sstateen CSRs to zeros too. Software at each privilege level should set its respective stateen CSRs to indicate the state it is prepared to allow less-privileged software to access. For OSes and hypervisors, this usually means the state that the OS or hypervisor is prepared to swap on a context switch, or to manage in some other way.
> 
> For each mstateen CSR, bit 63 is defined to control access to the matching sstateen and hstateen CSRs. That is, bit 63 of mstateen0 controls access to sstateen0 and hstateen0; bit 63 of mstateen1 controls access to sstateen1 and hstateen1; etc. Likewise, bit 63 of each hstateen correspondingly controls access to the matching sstateen CSR.
> 
> A hypervisor may need this control over accesses to the sstateen CSRs if it ever must emulate for a virtual machine an extension that is supposed to be affected by a bit in an sstateen CSR. Even if such emulation is uncommon, it should not be excluded.
Machine-level software needs identical control to be able to emulate the hypervisor extension. That is, machine level needs control over accesses to the supervisor-level sstateen CSRs in order to emulate the hstateen CSRs, which have such control.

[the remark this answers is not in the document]
neverlandiz: Sounds good, thanks! Should I change the description for all the `stateen` CSRs to this?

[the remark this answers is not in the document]
ThinkOpenly: > Should I change the description for all the `stateen` CSRs to this?

That extract is really just for `hstateen` CSRs. I can create similar extracts for `mstateen` and `sstateen`, or you could give it a try. Either way.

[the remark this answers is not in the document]
neverlandiz: I'll give it a try, thanks!

arch/csr/hstateen0h.yaml
```diff
@@ -0,0 +1,71 @@
...
+      from accessing the hart’s IMSIC the same as setting `hstatus.`VGEIN = 0.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  CONTEXT:
+    location: 25
+    description: |
+      The CONTEXT bit in `hstateen0h` controls access to the `scontext` CSR provided
+      by the Sdtrig extension.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
```
dhower-qc, line 71: Since this is an alias, need sw_read():

```yaml
sw_read(): return $bits(CSR[hstateen0])[63:32];
```

dhower-qc (changes requested):
The comments apply to all the CSRs.

dhower-qc (commented):
We also need an arch/ext/Ssstateen.yaml file.

neverlandiz:
I think there is already a Ssstateen.yaml file?

dhower-qc:
> I think there is already a Ssstateen.yaml file?

You're right. Ignore my ignorance please ;)

.pre-commit-config.yaml
```diff
@@ -119,7 +119,7 @@ repos:
     rev: v5.0.2
     hooks:
       - id: reuse-lint-file
-
+
```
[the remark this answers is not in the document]
ThinkOpenly: Remove this file from the PR. It must've snuck in accidentally?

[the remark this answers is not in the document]
neverlandiz: Is there a way to remove a specific file from a PR? I'm having trouble with this.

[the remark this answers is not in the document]
ThinkOpenly: Just create a new commit which undoes any changes. When the PR is merged, it'll be squashed, and any changes to that file will get canceled out in the final commit.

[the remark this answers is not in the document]
neverlandiz: Ok, thanks!

arch/csr/mstateen0.yaml
```diff
@@ -0,0 +1,150 @@
...
+
+definedBy: Smstateen
+fields:
+  SE0:
+    location: 63
+    base: 64
+    description: |
+      The SE0 bit in `mstateen0` controls access to the `hstateen0`, `hstateen0h`, and the `sstateen0` CSRs.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
```
jordancarlin, line 76: The spec calls out
> On reset, all writable `mstateen` bits are initialized by the hardware to zeros.

Should that be encoded into the `reset_value` somehow?

arch/csr/hstateen0.yaml
```diff
@@ -0,0 +1,144 @@
...
+    - Ssstateen
+fields:
+  SE0:
+    location: 63
+    base: 64
+    description: |
+      The SE0 bit in `hstateen0` controls access to the `sstateen0` CSR.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  ENVCFG:
```
jordancarlin, line 74: The {h/s}stateen bits are 0 if the corresponding mstateen bit is 0. It doesn't look like that is captured here right now. I assume this needs a `sw_read()` call to represent this?

dhower-qc: Yep, that's right.

neverlandiz:
Regression tests are failing due to missing Sscsrind (depends on PR #557) and Zfinx extensions.

arch/csr/hstateen0.yaml
ThinkOpenly: Here's what the spec says today:
> The C bit controls access to any and all custom state.
> The C bit of these registers is not custom state itself; it is a
> standard field of a standard CSR, either `mstateen0`, `hstateen0`, or
> `sstateen0`.

Could you use that text here, instead?

ThinkOpenly:
> Regression tests are failing due to missing Sscsrind

Sscsrind is going in as we speak.

> and Zfinx extensions.

I don't even see an issue open for that. :-/

ThinkOpenly:
There are still a couple of unresolved comments about changing a reset_value and adding a `sw_read()`.

As for Zfinx, I'd be tempted to comment it out in the PR, since Zfinx isn't even started, then open an issue for Zfinx to be added, including a note to uncomment the content here.

jordancarlin:
#711 adds the Zfinx (and other Z*inx) extensions.

ThinkOpenly:
> #711 adds the Zfinx (and other Z*inx) extensions.

Ah, so it does. I searched issues and PRs, but unsurprisingly didn't search for Z.*f.*inx.

Still, that PR and this one depend on each other. I suggest proceeding as planned... comment out Zfinx here, and move this one along.

@neverlandiz , if you would, please add a comment there requesting that your commented out content be uncommented via that PR.

arch/csr/hstateen0.yaml
neverlandiz: @jordancarlin Thanks for pointing that out! Could you check if this implementation is correct (for hstateen0)?
```
sw_read(): |
  # for every bit in an mstateen CSR that is zero, the same bit 
  # appears as read-only zero in the matching hstateen CSR

  Bit<64> mstateen0_mask = CSR[mstateen0].csr_value;
  Bit<64> hstateen0_value = csr_value & mstateen_mask;
  return hstateen0_value;
```

neverlandiz:
Sounds good, thanks!

arch/csr/hstateen0.yaml
jordancarlin: I think that seems reasonable. For sstateen it will probably be a little more complicated since it changes depending on if the H extension is enabled.

neverlandiz: Thanks!

ThinkOpenly: Do we need a `sw_write()` here, too, that enforces read-only for corresponding bits that are zero in `mstateen0`?

neverlandiz: Would it make sense to use `type()` instead of `sw_write()` to set the specific field to RO whenever the corresponding field is zero? For example, when CSR[mstateen0].JVT is 0, then CSR[sstateen0].JVT would have type RO.

ThinkOpenly: I understand type() to be an expression that must be knowable at compile time. Since JVT can vary during runtime, I believe it needs to be a dynamic check with each write operation.

ThinkOpenly: I thought to look at the definition of that attribute in the schema to see if what I was saying was (1) accurate, and (2) documented, and nicely it is:
```
        "type()": {
          "type": "string",
          "description": "Function that returns a configuration-dependent type. The return value should be a CsrFieldType enum, and must be compile-time-known."
        },
```

neverlandiz: I see, thanks for the clarification!

arch/csr/hstateen0.yaml
```diff
@@ -0,0 +1,201 @@
...
+    reset_value: UNDEFINED_LEGAL
+    sw_write(csr_value): |
+      if (CSR[mstateen0].C == 1'b0){
+        return 0;
+      }
+sw_read(): |
+  # for every bit in an mstateen CSR that is zero, the same bit
+  # appears as read-only zero in the matching hstateen CSR
+
+  Bits<64> mstateen0_mask = CSR[mstateen0].csr_value;
```
ThinkOpenly, line 199: The CI failures are complaining about this. Looking at other files, when the entire CSR value is accessed, it seems you should use:
```
$bits(CSR[mstateen0])
```
here and elsewhere

arch/csr/sstateen0.yaml
```diff
@@ -77,6 +77,13 @@ fields:
       The JVT bit controls access to the `jvt` CSR provided by the Zcmt extension.
     type: RW
     reset_value: UNDEFINED_LEGAL
+    sw_write(csr_value): |
+      if (CSR[mstateen0].JVT == 1'b0){
+        return 0;
+      }
+      else if (mode() == PrivilegeMode::VS && CSR[hstateen0].JVT == 1'b0) {
+        return 0;
+      }
```
dhower-qc, line 86: This brings up an interesting question. What happens if you write '1' to sstateen0.JVT while mstateen0.JVT is 0, and then change mstateen0.JVT to 1? Should sstateen0.JVT read as 0, 1, or undefined? I don't see any clarity on this in the spec. The answer will dictate whether this gets handled via sw_write() or sw_read().

While we figure that out...

All the stateen bits are WARL, and can thus be configured to be RW, read-only-0, or read-only-1. This will require a parameter for every bit in every mode. For example, for JVT:

arch/ext/Zcmt.yaml
```yaml
parameters:
  MSTATEEN_JVT_TYPE:
    when: Smstateen
    schema:
      enum: [rw, read-only-0, read-only-1]
    description: |
      Behavior of the mstateen0.JVT bit:

        * 'rw': read-write
        * 'read-only-0': read-only, fixed to 0
        * 'read-only-1': read-only, fixed to 1
  HSTATEEN_JVT_TYPE:
    when:
      allOf:
        - H
        - Ssstateen
    schema:
      enum: [rw, read-only-0, read-only-1]
    description: |
      Behavior of the hstateen0.JVT bit:

        * 'rw': read-write
        * 'read-only-0': read-only, fixed to 0
        * 'read-only-1': read-only, fixed to 1
    extra_validation: |
      assert HSTATEEN_JVT_TYPE == 'read-only-0' if MSTATEEN_JVT_TYPE == 'read-only-0'
      assert HSTATEEN_JVT_TYPE == 'read-only-1' if MSTATEEN_JVT_TYPE == 'read-only-1'
  SSTATEEN_JVT_TYPE:
    when:
      allOf:
        - S
        - Sstateen
    schema:
      enum: [rw, read-only-0, read-only-1]
    description: |
      Behavior of the sstateen0.JVT bit:

        * 'rw': read-write
        * 'read-only-0': read-only, fixed to 0
        * 'read-only-1': read-only, fixed to 1
    extra_validation: |
      assert SSTATEEN_JVT_TYPE == 'read-only-0' if MSTATEEN_JVT_TYPE == 'read-only-0'
      assert SSTATEEN_JVT_TYPE == 'read-only-0' if HSTATEEN_JVT_TYPE == 'read-only-0'
      assert SSTATEEN_JVT_TYPE == 'read-only-1' if MSTATEEN_JVT_TYPE == 'read-only-1'
      assert SSTATEEN_JVT_TYPE == 'read-only-1' if HSTATEEN_JVT_TYPE == 'read-only-1'
```

dhower-qc (changes requested):

arch/csr/sstateen0.yaml
neverlandiz: Thanks for the feedback! Could you explain the purpose of this line `when: Smstateen`? When I tried adding it, it resulted in a schema error because `Smstateen` is not a when_condition. I looked at other extension yaml files and noticed that some use `also_defined_in`, but I'm not sure if it applies in this case.

dhower-qc: My bad. After checking the schema, it's looking for an extension _requirement_ (e.g., "~> 1.0"), which is assumed to be a requirement of the extension being defined (Zcmt here). That's not descriptive enough for us in this case, so we'll need a schema change. I will get that in ASAP and point you to it.

dhower-qc: #751 adds the support. You can say:

```yaml
  when:
    name: Smstateen
    version: ~> 1.0
```

neverlandiz: Ok, thanks!

neverlandiz: Will this schema also be added?

```
when:
      allOf:
        - S
        - Sstateen
```

ThinkOpenly:
There was a newer version of _The RISC-V Instruction Set Manual: Volume II: Privileged Archtecture_ last week (20250508). Sorry to do this to you, but could you verify the current content in this PR is up-to-date? @syedowaisalishah mentioned in #653 that, for example, `mstateen0` now has more defined bits (SRMCFG and CTR).

neverlandiz:
Ok, sounds good!

arch/csr/hstateen0.yaml
```diff
@@ -0,0 +1,216 @@
...
+    location: 63
+    base: 64
+    description: |
+      The SE0 bit in `hstateen0` controls access to the `sstateen0` CSR.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+    sw_write(csr_value): |
+      if (CSR[mstateen0].SE0 == 1'b0){
+        return 0;
+      }
```
ThinkOpenly, line 73: You need to return the value that will actually be written to the field:
```
  return csr_value.SE0;
```
This comment applies to all `sw_write` implementations.

arch/ext/Smstateen.yaml
ThinkOpenly, line 1: The parameterization you added for the JVT bits is also needed for the other bits in the `*stateen` registers, mostly here, I think. CSRIND will go in Sscsrind.yaml, IMSIC in Ssaia, etc.

ThinkOpenly (changes requested):

arch/csr/sstateen0.yaml
dhower-qc: yes, that it also supported

arch/csr/hstateen0.yaml
neverlandiz: Thanks for the feedback! I'm a little confused about this part. How would this make `hstateen0.SE0` read-only 0 when `mstateen0.SE0` is 0?

ThinkOpenly: The `return` statement I suggested is in addition to (after) the `if/return 0` (not a replacement for). Is that where your question comes from?

neverlandiz: Oh I see! I thought it was a replacement for `return 0`.

ThinkOpenly: Sorry for the confusion! I could've been more clear.

neverlandiz: No worries!

ThinkOpenly (approved):
just some pretty minor issues. Thanks for sticking with it!

arch/ext/S.yaml
```diff
@@ -333,3 +333,33 @@ params:
...
+        * 'rw': read-write
+        * 'read-only-0': read-only, fixed to 0
+        * 'read-only-1': read-only, fixed to 1
+  HSTATEEN_ENVCFG_TYPE:
+    when:
+      allOf:
+        - name: H
+        - version: ~> 1.0
+        - name: Ssstateen
+        - version: ~> 1.0
```
dhower-qc, line 354: should be:

```
  - name: H
    version: ~> 1.0
  - name: Ssstateen
    version: ~> 1.0
```

arch/ext/Sdtrig.yaml
```diff
@@ -28,3 +28,35 @@ versions:
...
+        * 'rw': read-write
+        * 'read-only-0': read-only, fixed to 0
+        * 'read-only-1': read-only, fixed to 1
+  HSTATEEN_CONTEXT_TYPE:
+    when:
+      allOf:
+        - name: H
+        - version: ~> 1.0
+        - name: Ssstateen
+        - version: ~> 1.0
```
dhower-qc, line 51: should be:

```
- name: H
  version: ~> 1.0
- name: Ssstateen
  version: ~> 1.0
```

arch/ext/Ssaia.yaml
```diff
@@ -14,3 +14,65 @@ versions:
...
+        * 'rw': read-write
+        * 'read-only-0': read-only, fixed to 0
+        * 'read-only-1': read-only, fixed to 1
+  HSTATEEN_AIA_TYPE:
+    when:
+      allOf:
+        - name: H
+        - version: ~> 1.0
+        - name: Ssstateen
+        - version: ~> 1.0
```
dhower-qc, line 37: should be:

```
- name: H
  version: ~> 1.0
- name: Ssstateen
  version: ~> 1.0
```

arch/ext/Ssaia.yaml
```diff
@@ -14,3 +14,65 @@ versions:
...
+        * 'rw': read-write
+        * 'read-only-0': read-only, fixed to 0
+        * 'read-only-1': read-only, fixed to 1
+  HSTATEEN_IMSIC_TYPE:
+    when:
+      allOf:
+        - name: H
+        - version: ~> 1.0
+        - name: Ssstateen
+        - version: ~> 1.0
```
dhower-qc, line 67: same as above

arch/ext/Zcmt.yaml
```diff
@@ -90,3 +90,55 @@ versions:
...
+        * 'rw': read-write
+        * 'read-only-0': read-only, fixed to 0
+        * 'read-only-1': read-only, fixed to 1
+  HSTATEEN_JVT_TYPE:
+    when:
+      allOf:
+        - name: H
+        - version: ~> 1.0
+        - name: Ssstateen
+        - version: ~> 1.0
```
dhower-qc, line 113: same

arch/ext/Zcmt.yaml
```diff
@@ -90,3 +90,55 @@ versions:
...
+    extra_validation: |
+      assert HSTATEEN_JVT_TYPE == 'read-only-0' if MSTATEEN_JVT_TYPE == 'read-only-0'
+      assert HSTATEEN_JVT_TYPE == 'read-only-1' if MSTATEEN_JVT_TYPE == 'read-only-1'
+  SSTATEEN_JVT_TYPE:
+    when:
+      allOf:
+        - name: S
+        - version: ~> 1.0
+        - name: Sstateen
+        - version: ~> 1.0
```
dhower-qc, line 131: similar

dhower-qc (changes requested):

AFOliveira (approved):
Withdrawing my request for changes

arch/ext/S.yaml
```diff
@@ -333,3 +333,33 @@ params:
       When not implemented mstatus.TVM will be read-only-zero.
     schema:
       type: boolean
+  MSTATEEN_ENVCFG_TYPE:
+    when:
+      name: Smstateen
+      version: ~> 1.0
```
ThinkOpenly, line 339: Missed the dash before "name" here:
```suggestion
    when:
￼     - name: Smstateen
￼       version: ~> 1.0
```

arch/ext/Sdtrig.yaml
```diff
@@ -28,3 +28,35 @@ versions:
   - version: "1.0.0"
     state: ratified
     ratification_date: null
+
+params:
+  MSTATEEN_CONTEXT_TYPE:
+    when:
+      name: Smstateen
```
ThinkOpenly, line 35: here, too

arch/ext/Ssaia.yaml
```diff
@@ -14,3 +14,65 @@ versions:
     requires:
       name: S
       version: ">= 1.12"
+
+params:
+  MSTATEEN_AIA_TYPE:
+    when:
+      name: Smstateen
```
ThinkOpenly, line 21: and here :-)

arch/ext/Ssaia.yaml
```diff
@@ -14,3 +14,65 @@ versions:
...
+
+        * 'rw': read-write
+        * 'read-only-0': read-only, fixed to 0
+        * 'read-only-1': read-only, fixed to 1
+    extra_validation: |
+      assert HSTATEEN_AIA_TYPE == 'read-only-0' if MSTATEEN_AIA_TYPE == 'read-only-0'
+      assert HSTATEEN_AIA_TYPE == 'read-only-1' if MSTATEEN_AIA_TYPE == 'read-only-1'
+  MSTATEEN_IMSIC_TYPE:
+    when:
+      name: Smstateen
```
ThinkOpenly, line 51: here

arch/ext/Ssaia.yaml
```diff
@@ -14,3 +14,65 @@ versions:
     requires:
       name: S
```
ThinkOpenly, line 15: this might need one, too?

arch/ext/Sscsrind.yaml
```diff
@@ -42,3 +42,35 @@ versions:
           version: ~> 1.13
         - name: Smcsrind
           version: ~> 1.0
+
+params:
+  MSTATEEN_CSRIND_TYPE:
+    when:
+      name: Smstateen
```
ThinkOpenly, line 49: here

arch/ext/Zcmt.yaml
```diff
@@ -90,3 +90,55 @@ versions:
       #   - [Zcb, "1.0.0"]
       #   - [Zcmp, "1.0.0"]
       #   - [Zcmt, "1.0.0"]
+
+params:
+  MSTATEEN_JVT_TYPE:
+    when:
+      name: Smstateen
```
ThinkOpenly, line 97: here

ThinkOpenly (changes requested):
tests are failing, I believe because of some YAML format problems.

neverlandiz:
I think the tests are failing because PR #751 hasn't been merged yet?

ThinkOpenly:
> I think the tests are failing because PR #751 hasn't been merged yet?

Maybe? I confess to my own confusion about the different representations between:
```
    when:
      name: Smstateen
      version: ~> 1.0
```
and this:
```
    when:
      allOf:
        - name: H
          version: ~> 1.0
        - name: Ssstateen
          version: ~> 1.0
```
(Why does "name" need a dash in one case and not in the other??)

You could rebase on the branch hosting #751 and see. (If and when #751 gets merged, your branch here will be automatically rebased.)

dhower-qc:
> > I think the tests are failing because PR #751 hasn't been merged yet?
> 
> Maybe? I confess to my own confusion about the different representations between:
> 
> ```
>     when:
>       name: Smstateen
>       version: ~> 1.0
> ```
> 
> and this:
> 
> ```
>     when:
>       allOf:
>         - name: H
>           version: ~> 1.0
>         - name: Ssstateen
>           version: ~> 1.0
> ```
> 
> (Why does "name" need a dash in one case and not in the other??)
> 
> You could rebase on the branch hosting #751 and see. (If and when #751 gets merged, your branch here will be automatically rebased.)

This:

```yaml
when:
  name: Smstateen
  version: ~> 1.0
```

Is a nested hash.

This:

```yaml
when:
  allOf:
    - name: H
      version: ~> 1.0
    - name: Ssstateen
      version: ~> 1.0
```

Is a nested hash with an array of hashes. It might be more clear in the equivalent JSON-like syntax:

```yaml
when: { name: Smstateen, version: ~> 1.0 }
```

```yaml
when: {
  allOf: [
    { name: H, version: ~> 1.0] },
    { name: Smstateen, version: ~> 1.0 }
  ]
}
```

The `{name:, version:}` hash is an _Extension Requirement_. `when` can accept arbitrary boolean expressions of Extension Requirements. So the first says you must have Smstateen, version compatible with 1.0. The second says you must have *both* Smstateen, version compatible with 1.0 and H, version compatible with 1.0.

ThinkOpenly:
Not to belabor the point, but my simple brain will now think "the dash means 'I'm an array element!'". I suppose we could unify the appearance a bit more by one-lining each requirement, like:
```
when: { name: Smstateen, version: ~> 1.0 }
when:
  allOf: 
    - { name: H, version: ~> 1.0 }
    - { name: Ssstateen, version: ~> 1.0 }
```
In that way (to me, at least), the dash is more associated with "allOf" than with "name".
Anyway, I understand now.

ThinkOpenly (approved):

arch/csr/hstateen0.yaml
```diff
@@ -0,0 +1,226 @@
...
+definedBy:
+  allOf:
+    - H
+    - Smstateen
+    - Ssstateen
+fields:
+  SE0:
+    long_name: sstateen0 access control
+    location: 63
+    base: 64
```
ThinkOpenly: CI failure:
```
/home/runner/work/riscv-unified-db/riscv-unified-db/lib/idl/ast.rb:298:in `type_error': In file gen/arch/rv32/csr/hstateen0h.yaml (Idl::AstNode::TypeError)
On line 35
In the code:

  34: if (**HERE** >> CSR[mstateen0].SE0 << **HERE** == 1'b0){
  35:   return 0;
  36: }
  

A type error occurred
  CSR[mstateen0].SE0 is not defined in RV32
```

Looks like we need to remove the
```
    base: 64
```
From the fields which are aliased from the `*h` CSRs, and then probably add a check for `xlen() > 32` in `sw_read()` to prevent direct reading in RV32 mode.

arch/ext/Zcmt.yaml
```diff
@@ -90,3 +90,58 @@ versions:
...
+        * 'read-only-1': read-only, fixed to 1
+    extra_validation: |
+      assert HSTATEEN_JVT_TYPE == 'read-only-0' if MSTATEEN_JVT_TYPE == 'read-only-0'
+      assert HSTATEEN_JVT_TYPE == 'read-only-1' if MSTATEEN_JVT_TYPE == 'read-only-1'
+  SSTATEEN_JVT_TYPE:
+    when:
+      allOf:
+        - name: S
+          version: ~> 1.0
+        - name: Sstateen
```
ThinkOpenly:
```suggestion
        - name: Ssstateen
```

ThinkOpenly (changes requested):

arch/csr/hstateen0.yaml
neverlandiz: Thanks for the feedback! I removed the line `base: 64` in the previous commit, and it seemed to fix the error. Should I still add the `xlen() > 32` check?

ThinkOpenly:
You might have run into a bug in the Ruby code...
```
/home/runner/work/riscv-unified-db/riscv-unified-db/lib/cfg_arch.rb:433:in `map': undefined method `version' for Extension#A:Extension (NoMethodError)

        extensions.map(&:version).flatten
                  ^^^^
Did you mean?  versions
```
I'm not familiar with this code, but it looks like it should be "versions". @dhower-qc ?

arch/csr/hstateen0.yaml
ThinkOpenly: > Should I still add the `xlen() > 32` check?

Fair question, and I'm not sure. The value needs to be truncated in RV32 mode, but I'm not sure if that's already handled somehow or not. Let's get the CI tests passing, and see if @dhower-qc can offer an answer. (You may leave the test out for now.)

ThinkOpenly:
Running your branch with the following change got the regress-gen-isa-manual test to pass:
```
diff --git a/lib/cfg_arch.rb b/lib/cfg_arch.rb
index f60ed1121..16de8bf4b 100644
--- a/lib/cfg_arch.rb
+++ b/lib/cfg_arch.rb
@@ -433 +433 @@ class ConfiguredArchitecture < Architecture
-        extensions.map(&:version).flatten
+        extensions.map(&:versions).flatten
```
If you want to include that change with this PR, I would not object.

ThinkOpenly (approved):

dhower-qc (approved):

PR #593, opened 2025-04-04
Add functions for conversion, add check_f_encoding

ShashankVM:
- Add function for converting unsigned 32-bit integer to single-precision float
- Add function to convert signed 32-bit integer to single-precision float
- Add check_f_ok(encoding) in floating point addition and subtraction


dhower-qc (changes requested):
It looks like there is an issue; the only changes are extra whitespace in `fcvt.s.wu.yaml`.

ShashankVM:
Oops, sorry.

PR #598, opened 2025-04-08
schema: add array form of description

dhower-qc:
Adds an array form of "description" fields that enables tagged statements.

In this first cut, statements can be tagged as:

  - normative or non-normative.
  - conditional, using an IDL function to evaluate the condition

### DIFF
--- a/arch/csr/hstatus.yaml
+++ b/arch/csr/hstatus.yaml
@@ -19,7 +19,8 @@ fields:
     location: 33-32
     base: 64
     description:
-      - normative: false
+      - id: csr-hstatus-vsxl-values
+        normative: false
         text: |
           Determines the effective XLEN in VS-mode. Valid values are:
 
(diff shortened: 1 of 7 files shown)


ThinkOpenly (approved):
Are you marking everything as non-normative for now?

arch/csr/hstatus.yaml
```diff
@@ -69,69 +69,74 @@ fields:
...
-      *Virtual Trap SRET*
-
-      When `hstatus.VTSR` is set, executing the `sret` instruction in VS-mode
-      raises a `Virtual Instruction` exception.
-
-      When `hstatus.VTSR` is clear, an `sret` instruction in VS-mode returns control
-      to the mode stored in `vsstatus.SPP`.
+    long_name: Virtual Trap SRET
+    description:
+      - normative: false
```
AFOliveira, line 74: Why do we have two split normative comments instead of just one?

AFOliveira (approved):
I think it's a graet approach, my only question is why to split in two when we only have normative text?

arch/csr/hstatus.yaml
ThinkOpenly: Good question. I wonder if the 2nd stanza is needed at all, as it describes the normal operation of the `sret` instruction, which is of course described with that instruction.

dhower-qc:
> Are you marking everything as non-normative for now?

I marked it all non-normative because the text I changed isn't copy-pasted. My thought was that we'd only apply the normative tag to statements that are an exact match.

arch/csr/hstatus.yaml
AFOliveira: Ok, I see the purpose of splitting, but this adds some overhead to extensions writers, right?

PR #600, opened 2025-04-09
refactor regress actions with anchors

kbroch-rivosinc:


dhower-qc:
nice. I didn't realize workflow files could use anchors.

dhower-qc (approved):

kbroch-rivosinc:
closing as #602 is a better solution (allows for separate files so action can be reused across multiple workflow files)

Issue #569, opened 2025-03-27
Add Sscofpmf CSR


dhower-qc:
Need to add:

- SCOUNTOVF

syedowaisalishah:
@dhower-qc I've started working on this issue. Could you please assign it to me?

PR #607, opened 2025-04-11
Add CSR YAML files for Sscofpmf: scountovf

syedowaisalishah:
Extension CSR YAML files for the `Sscofpmf` extension.

CSR YAMLs for:
- SCOUNTOVF

Closes Add Sscofpmf CSR #569


syedowaisalishah:
> @syedowaisalishah, did you intend to close this?

Yes, I closed it because the regression test failed in this PR. I’ll address the issues and reopen it once everything passes.

arch/csr/Sscofpmf/scountovf.yaml
```diff
@@ -0,0 +1,31 @@
+# yaml-language-server: $schema=../../../schemas/csr_schema.json
+$schema: csr_schema.json#
+kind: csr
+name: scountovf
+long_name: Supervisor Count Overflow (scountovf) Register
```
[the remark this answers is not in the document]
syedowaisalishah: Got it — I’ll remove "(scountovf)" from the long_name.

arch/csr/Sscofpmf/scountovf.yaml
```diff
@@ -0,0 +1,31 @@
...
+address: 0x700
```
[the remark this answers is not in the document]
syedowaisalishah: sorry my mistake ,You're right — thanks for catching that! I’ll update the address to 0xDA0 to match the spec

arch/csr/Sscofpmf/scountovf.yaml
```diff
@@ -0,0 +1,31 @@
...
+priv_mode: S
+length: 32
+definedBy: Sscofpmf
+description: |
+  This extension adds the scountovf CSR, a 32-bit read-only
```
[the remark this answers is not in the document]
syedowaisalishah: sure

arch/csr/Sscofpmf/scountovf.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+  Read access to bit X is subject to the same mcounteren (or mcounteren and hcounteren)
+  CSRs that mediate access to the hpmcounter CSRs by S-mode (or VS-mode).
+  In M-mode, scountovf bit X is always readable. In S/HS-mode, scountovf bit X is readable
+  when mcounteren bit X is set, and otherwise reads as zero. Similarly, in VS mode,
+  scountovf bit X is readable when mcounteren bit X and hcounteren bit X are both set, and otherwise reads as zero
+
+fields:
+  OF_shadow_bits:
+    location: 31
```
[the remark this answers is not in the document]
syedowaisalishah: > So, bits 3-31 are valid here, and location should be:

Thanks for the clarification! I tried using location: 31-3 as suggested, but the regression fails with that format. I’m currently looking into the correct way to express the range — will update once I get it working.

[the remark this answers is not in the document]
syedowaisalishah: > Also, we need a `sw_read()` method for this CSR, given the "Read access" paragraph above. Do you need help with that?

Yes, I’d appreciate some guidance on that. Noted that we need to add a sw_read() method — I’ll look at similar implementations to get started, but any pointers would be very helpful.

syedowaisalishah:
I've made all the suggested changes — just haven’t added the sw_read() yet due to an error. Will follow up on that soon.

arch/csr/Sscofpmf/scountovf.yaml
```diff
@@ -0,0 +1,32 @@
...
+name: scountovf
+long_name: Supervisor Count Overflow
+address: 0xDA0
+priv_mode: S
+length: 32
+definedBy: Sscofpmf
+description: |
+  A 32-bit read-only
+  register that contains shadow copies of the OF bits in the 29 mhpmevent CSRs
+  (mhpmevent3 - mhpmevent31) - where scountovf bit X corresponds to mhpmeventX.
```
ThinkOpenly, line 13: The spec doesn't have this, but it should... could you backtick the CSRs here and below?

`mhpmevent3`, `mhpmevent31`, `scountovf`, `mhpmeventX`, `mcounteren`, `hcounteren`, `hpmcounter`.

arch/csr/Sscofpmf/scountovf.yaml
```diff
@@ -0,0 +1,32 @@
...
+  In M-mode, scountovf bit X is always readable. In S/HS-mode, scountovf bit X is readable
+  when mcounteren bit X is set, and otherwise reads as zero. Similarly, in VS mode,
+  scountovf bit X is readable when mcounteren bit X and hcounteren bit X are both set, and otherwise reads as zero
+
+fields:
+  OF_shadow_bits:
+    location: 31-3
+    type: RO
+    description: |
+      Read-only shadow copies of the OF bits from mhpmevent3 to mhpmevent31.
```
ThinkOpenly, line 30: backticks around `mhpmevent3` and `mhpmevent31`, please.

ThinkOpenly (changes requested):

arch/csr/Sscofpmf/scountovf.yaml
syedowaisalishah: Sure, will do!

arch/csr/Sscofpmf/scountovf.yaml
syedowaisalishah: Got it, will fix!

arch/csr/Sscofpmf/scountovf.yaml
```diff
@@ -0,0 +1,32 @@
...
+  (without needing to make an execution environment call or series of calls ultimately up to M-mode)
+
+  Read access to bit X is subject to the same `mcounteren` (or `mcounteren` and `hcounteren`)
+  CSRs that mediate access to the `hpmcounter` CSRs by S-mode (or VS-mode).
+  In M-mode, `scountovf` bit X is always readable. In S/HS-mode, `scountovf` bit X is readable
+  when `mcounteren` bit X is set, and otherwise reads as zero. Similarly, in VS mode,
+  `scountovf` bit X is readable when `mcounteren` bit X and `hcounteren` bit X are both set, and otherwise reads as zero.
+
+fields:
+  OF_shadow_bits:
```
dhower-qc, line 26: Since this is a bunch of alias bits, we are going to need to split this into parts, like so:

```yaml
fields:
  OF3:
    alias: mhpmevent3.OF
    # ...
  OF4:
    alias: mhpmevent4.OF
    # ...
```

You can either do this by hand, your own script that doesn't get checked in, or using a `.layout` file like `arch/csr/Zicntr/mcountinhibit.layout`. Any way is fine with me.

arch/csr/Sscofpmf/scountovf.yaml
```diff
@@ -0,0 +1,32 @@
...
+    location: 31-3
+    type: RO
+    description: |
+      Read-only shadow copies of the OF bits from `mhpmevent3` to `mhpmevent31`.
+    reset_value: UNDEFINED_LEGAL
+sw_read(): |
```
dhower-qc, line 32: This needs to return the combined value. It will look like:

```yaml
sw_read(): |
  Bits<32> mask;
  if (priv_mode() == PrivilegedMode::VS) {
    # in VS-mode, scountovf.OFX access is determined by mcounteren/hcounteren
    mask = $bits(CSR[mcounteren]) & $bits(CSR[hcounteren]);
  } else {
    # in M-mode and S-mode, scountovf.OFX access is determined by mcounteren/scounteren
    mask = $bits(CSR[mcounteren]) & $bits(CSR[scounteren]);
  }
  Bits<32> value =
    (CSR[mhpmevent3].OF << 3) |
    (CSR[mhpmevent4].OF << 4) |
    # ...
    (CSR[mhpmevent30].OF << 30) |
    (CSR[mhpmevent31].OF << 31);
  return value & mask;
```

dhower-qc (changes requested):

arch/csr/Sscofpmf/scountovf.yaml
syedowaisalishah: Thanks for the clarification. I’ve updated the sw_read() method as described, applying the masking logic with mcounteren, scounteren, and hcounteren. However, I’m encountering an error and am working on resolving it.

dhower-qc: if you need help with the error, post it here

syedowaisalishah: Hi @dhower-qc ,
I was able to resolve the error I previously encountered—thanks again for your help.
Now, I'm running the smoke test, but it seems to get stuck at the following point and doesn't proceed further:

Resolving arch: 100%|####################| 1651/1651 [00:00<00:00, 23581.00it/s]  
[INFO] Resolved architecture files written to gen/resolved_arch/rv64  
Parsing IDL code for RV32...done  
Type checking IDL code for rv32...  
[INFO] 2025-04-30 20:34:24: Allocating more SymbolTables                       
Instructions: |================================================================|  
CSRs: |=====================                                                   |

arch/csr/Sscofpmf/scountovf.yaml
syedowaisalishah: Thanks for the suggestion, @dhower-qc 
I’m doing both:

> Explicitly defining each field (OF3 to OF31) in the YAML file, as shown in your example.
> Using the .layout file for the bit-level mappings.

Let me know if this approach looks good or if any adjustments are needed!

arch/csr/Sscofpmf/scountovf.yaml
syedowaisalishah: The smoke test does run, but it takes quite a bit of time to proceed when running in VS Code

ThinkOpenly:
> You can either do this by hand, your own script that doesn't get checked in, or using a `.layout` file like `arch/csr/Zicntr/mcountinhibit.layout`. Any way is fine with me.

Derek was suggesting here to either explicitly describe all of the bits OF3-OF31, _or_ parameterize their description using a layout like:
```
  <%- (3..31).each do |of_num| -%>
  OF<%= of_num %>:
    location: <%= of_num %>
[...]
```

Since you've already done the former (describing each of them), you do not need the ".layout" file.

(For reference, if you did choose to use a "layout" file, you'd need to have it explicitly handled in the `Rakefile` similarly to `mcountinhibit`. That has me wondering if we should make that handling more automatic... a thought for another day, perhaps.)

arch/csr/Sscofpmf/scountovf.layout
dhower-qc, line 1: Looks like this was just a file you used locally. If that's true, and it's not tied into the build anywhere, then let's not check it in.

arch/csr/Sscofpmf/scountovf.yaml
```diff
@@ -0,0 +1,269 @@
...
+  CSRs that mediate access to the `hpmcounter` CSRs by S-mode (or VS-mode).
+  In M-mode, `scountovf` bit X is always readable. In S/HS-mode, `scountovf` bit X is readable
+  when `mcounteren` bit X is set, and otherwise reads as zero. Similarly, in VS mode,
+  `scountovf` bit X is readable when `mcounteren` bit X and `hcounteren` bit X are both set, and otherwise reads as zero.
+
+fields:
+  OF3:
+    alias: mhpmevent3.OF
+    location: 3
+    type: RO
```
dhower-qc, line 29: All the types are RO-H (Read Only, with Hardware Update)

arch/csr/Sscofpmf/scountovf.layout
syedowaisalishah: @ThinkOpenly @dhower-qc Got it—thanks for the clarification! I'm now trying to set this up using a .layout file and adding the necessary handling in the Rakefile, similar to how it's done for mcountinhibit. This seems like a more maintainable approach for a large range like OF3–OF31. Appreciate the guidance!

syedowaisalishah:
@ThinkOpenly , @dhower-qc Added .layout file for OF3–OF31 YAML generation and updated the Rakefile accordingly (similar to mcountinhibit). Verified functionality with rake gen:arch.

arch/csr/Sscofpmf/scountovf.layout
```diff
@@ -0,0 +1,84 @@
...
+    (CSR[mhpmevent22].OF << 22) |
+    (CSR[mhpmevent23].OF << 23) |
+    (CSR[mhpmevent24].OF << 24) |
+    (CSR[mhpmevent25].OF << 25) |
+    (CSR[mhpmevent26].OF << 26) |
+    (CSR[mhpmevent27].OF << 27) |
+    (CSR[mhpmevent28].OF << 28) |
+    (CSR[mhpmevent29].OF << 29) |
+    (CSR[mhpmevent30].OF << 30) |
+    (CSR[mhpmevent31].OF << 31);
```
ThinkOpenly, line 82: Since you are now an expert on parameterization, could you convert this, too?

arch/csr/Sscofpmf/scountovf.yaml
ThinkOpenly, line 1: If you are correctly generating this content now, you no longer need to include this file in your PR.

arch/csr/Sscofpmf/scountovf.layout
syedowaisalishah: Absolutely—I'll go ahead and parameterize this section as well. Thanks for the nudge!

dhower-qc (changes requested):
Great, almost there. When using layout files, we also check in the generated files so that they are accessible via browsing GitHub. So, if you can add scountovf.yaml to the PR, then I think this looks good.

arch/csr/Sscofpmf/scountovf.layout
```diff
@@ -0,0 +1,59 @@
...
+      alias: mhpmevent<%= of_num %>.OF
+      location: <%= of_num %>
+      description: |
+        [when="mhpmevent<%= of_num %>.EN == true"]
+        Shadow copy of mhpmevent<%= of_num %> overflow (OF) bit.
+
+        [when="mhpmevent<%= of_num %>.EN == false"]
+        This field is read-only zero because the event is not enabled.
+      type(): |
+        return mhpmevent<%= of_num %>.EN ? CsrFieldType::RO : CsrFieldType::RO-H;
```
dhower-qc, line 38: This (and others) is actually:

```
return CSR[mhpmevent<%= of_num %>].EN ? # ...
```

That's why the regressions are failing.

syedowaisalishah: Thanks @dhower-qc — I updated the type() and reset_value() lines in scountovf.layout as you suggested, using `CSR[mhpmevent<%= of_num %>].EN` instead of the bare `mhpmevent<%= of_num %>.EN`
However, I’m encountering issues:

When I run regressions locally in VS Code, the test appears to hang during type-checking. It stalls here with no further output:
`Type checking IDL code for rv32...
[INFO] 2025-05-06 13:43:32: Allocating more SymbolTables
Instructions: |========================================================|
CSRs: |=====================                                           |
`
When I pushed the changes, 4 tests failed in CI.

dhower-qc (changes requested):
The regressions are failing because it can't find `CSR[mhpmeventX].OF`. That's because those fields don't exist yet...they are added by this extension!

We'll need to modify mhpmeventN.layout to add the fields [OF, MINH, SINH, UINH, VSINH, VUINH] to the event CSRs.

A few things to note: the OF and MINH fields are `definedBy` Sscofpmf. The SINH field is `definedBy: allOf: [S, Sscofpmf]`. Similar for the remaining *INH fields.

arch/csr/Sscofpmf/scountovf.yaml
ThinkOpenly: file is still present. Could you remove it from your PR?
(Create a new commit that removes it.)

ThinkOpenly:
> The regressions are failing because it can't find `CSR[mhpmeventX].OF`. That's because those fields don't exist yet...they are added by this extension!

Why would the regression tests fail for this PR because of something added by this PR? (Maybe I don't understand.)

I wonder if it could be because we're still defining everything twice, until `arch/csr/Sscofpmf/scountovf.yaml` is removed?

syedowaisalishah:
> The regressions are failing because it can't find `CSR[mhpmeventX].OF`. That's because those fields don't exist yet...they are added by this extension!
> 
> We'll need to modify mhpmeventN.layout to add the fields [OF, MINH, SINH, UINH, VSINH, VUINH] to the event CSRs.
> 
> A few things to note: the OF and MINH fields are `definedBy` Sscofpmf. The SINH field is `definedBy: allOf: [S, Sscofpmf]`. Similar for the remaining *INH fields.

I’ve updated mhpmeventN.layout to include the new fields (OF, MINH, SINH, UINH, VSINH, VUINH) as described, and regenerated the mhpmevent3–31 YAML files accordingly. However, the regressions are still failing, and when I run regress in VSCode, it hangs without producing any output.

arch/csr/Sscofpmf/scountovf.layout
```diff
@@ -0,0 +1,59 @@
...
+fields:
+  <%- (3..31).each do |of_num| -%>
+    OF<%= of_num %>:
+      alias: mhpmevent<%= of_num %>.OF
+      location: <%= of_num %>
+      description: |
+        [when="mhpmevent<%= of_num %>.EN == true"]
+        Shadow copy of mhpmevent<%= of_num %> overflow (OF) bit.
+
+        [when="mhpmevent<%= of_num %>.EN == false"]
```
ThinkOpenly, line 35: Should all of these references to `CSR[mhpmevent...].EN`, here and below, instead be
```
HPM_COUNTER_EN[<%= of_num %>]
```

arch/csr/Sscofpmf/scountovf.layout
```diff
@@ -29,17 +29,18 @@ fields:
...
+        [when="HPM_COUNTER_EN[<%= of_num %>] == true"]
         Shadow copy of mhpmevent<%= of_num %> overflow (OF) bit.
 
-        [when="mhpmevent<%= of_num %>.EN == false"]
-        This field is read-only zero because the event is not enabled.
+        [when="HPM_COUNTER_EN[<%= of_num %>] == false"]
+        This field is read-only zero because the counter is not enabled.
       type(): |
-        return CSR[mhpmevent<%= of_num %>].EN ? CsrFieldType::RO : CsrFieldType::RO-H;
+        return HPM_COUNTER_EN[<%= of_num %>] ? CsrFieldType::RO : CsrFieldType::RO-H;
```
ThinkOpenly, line 38: Valid `CsrFieldType` enums can be found in `arch/isa/globals.isa`:
```
enum CsrFieldType {
  RO   0
  ROH  1
  RW   2
  RWR  3
  RWH  4
  RWRH 5
}
```
So, here you should use `CsrFieldType::ROH`.

syedowaisalishah: Thanks for pointing that out! You’re right — I’ll update RO-H to ROH as suggested.

dhower-qc (approved):
Thanks for the hard work with this one!

ThinkOpenly (approved):

ThinkOpenly:
A few of the CI tests are failing, seemingly running forever? Are you seeing this, too? (I seem to see it locally as well.)

syedowaisalishah:
@ThinkOpenly ,
Yes, I’m seeing the same issue. When I run the tests locally, it hangs during the IDL parsing step for rv32:
```
[INFO] 2025-06-02 22:26:02: Allocating more SymbolTables
Instructions: |===============================================================|
```
It gets stuck here and doesn’t move forward — looks like it’s the same hang we're seeing in CI.
Let me know if you’d like me to help debug or gather more logs.

ThinkOpenly:
After a brief investigation, it looks like commit 74b114745 introduced the CI failures.

ThinkOpenly:
> After a brief investigation, it looks like commit [74b1147](https://github.com/riscv-software-src/riscv-unified-db/commit/74b11474523a4ca9a08537aa9673d35e7a99a6f1) introduced the CI failures.

The only odd things that stand out are:
- There are some funny characters:
  ```
  +  (`mhpmevent3` - `mhpmevent31`) <E2><80><94> where `scountovf` bit X corresponds to `mhpmeventX`.
  ```
- There are a bunch of "when" statements:
  ```
  +      [when="mhpmevent3.EN == true"]
  ```

You might try deleting one or both of them to see if they are causing the issue.

syedowaisalishah:
Thanks for the investigation, @ThinkOpenly .
I’ll try removing the special characters (<E2><80><94>) and the [when=...] statements you pointed out — and then rerun the tests to see if that resolves the CI hang.

I’ll report back shortly with what I find.

syedowaisalishah:
@ThinkOpenly , I also noticed the same behavior during debugging—commenting out that one extra line seemed to prevent the failure. However, after removing the unexpected characters and cleaning up the problematic “when” statements, the test no longer fails even with all lines included.

Should I now commit the full code with all those comment lines and push?

ThinkOpenly:
> However, after removing the unexpected characters and cleaning up the problematic “when” statements, the test no longer fails even with all lines included.
> 
> Should I now commit the full code with all those comment lines and push?

It looks like you pushed those changes, correct? The CI tests are still failing.

Could you try the following:
```
  Bits<32> value = 0;
  value = value | (CSR[mhpmevent3].OF << 3);
  value = value | (CSR[mhpmevent4].OF << 4);
  value = value | (CSR[mhpmevent5].OF << 5);
  value = value | (CSR[mhpmevent6].OF << 6);
  value = value | (CSR[mhpmevent7].OF << 7);
  value = value | (CSR[mhpmevent8].OF << 8);
  value = value | (CSR[mhpmevent9].OF << 9);
  value = value | (CSR[mhpmevent10].OF << 10);
  value = value | (CSR[mhpmevent11].OF << 11);
  value = value | (CSR[mhpmevent12].OF << 12);
  value = value | (CSR[mhpmevent13].OF << 13);
  value = value | (CSR[mhpmevent14].OF << 14);
  value = value | (CSR[mhpmevent15].OF << 15);
  value = value | (CSR[mhpmevent16].OF << 16);
  value = value | (CSR[mhpmevent17].OF << 17);
  value = value | (CSR[mhpmevent18].OF << 18);
  value = value | (CSR[mhpmevent19].OF << 19);
  value = value | (CSR[mhpmevent20].OF << 20);
  value = value | (CSR[mhpmevent21].OF << 21);
  value = value | (CSR[mhpmevent22].OF << 22);
  value = value | (CSR[mhpmevent23].OF << 23);
  value = value | (CSR[mhpmevent24].OF << 24);
  value = value | (CSR[mhpmevent25].OF << 25);
  value = value | (CSR[mhpmevent26].OF << 26);
  value = value | (CSR[mhpmevent27].OF << 27);
  value = value | (CSR[mhpmevent28].OF << 28);
  value = value | (CSR[mhpmevent29].OF << 29);
  value = value | (CSR[mhpmevent30].OF << 30);
  value = value | (CSR[mhpmevent31].OF << 31);
```

syedowaisalishah:
Yes, I tried this approach and all the tests are now passing.

ThinkOpenly:
> Yes, I tried this approach and all the tests are now passing

Awesome!

Could you open an issue about the IDL compiler being unable to handle that original code?

ThinkOpenly (changes requested):
Almost forgot: could you put back those "when" statements. The documentation is incorrect without them.

ThinkOpenly (approved):

Issue #604, opened 2025-04-10
vscode Devcontainer not working: tries to rebuild container and fails at `apt install`


james-ball-qualcomm:
After this commit to change the order to be alphabetical, it triggered a re-build of my container in vscode and it no longer worked. I had to revert the file to get it working again.

https://github.com/riscv-software-src/riscv-unified-db/commit/f716423476a0d25f8eceb2524793b9a84379cda5

ThinkOpenly:
> no longer worked

Do you have any error messages? Any further information?

james-ball-qualcomm:
BTW, how do you test the changes you made to Dockerfile? I'm using vscode on Windows and it opens a container using docker.

dhower-qc:
@james-ball-qualcomm, this smells like a Qualcomm IT problem, not a UDB problem.

Closing until we have a way to reproduce it.

kbroch-rivosinc:
I am also seeing an  issue with using devcontainers in vscode currently  but I don't think the problem is related to the reordering of apt pkgs.

Here's what I see on main right now when I ask vscode to "Dev Container: Open folder in container ..." .  It starts rebuilding and then:

```
1330.7 Err:5 http://ports.ubuntu.com/ubuntu-ports noble-updates/main arm64 libpython3.12-minimal arm64 3.12.3-1ubuntu0.4
1330.7   404  Not Found [IP: 185.125.190.36 80]
1331.4 Err:6 http://ports.ubuntu.com/ubuntu-ports noble-updates/main arm64 libexpat1 arm64 2.6.1-2ubuntu0.2
1331.4   404  Not Found [IP: 185.125.190.36 80]
1332.1 Err:7 http://ports.ubuntu.com/ubuntu-ports noble-updates/main arm64 python3.12-minimal arm64 3.12.3-1ubuntu0.4
1332.1   404  Not Found [IP: 185.125.190.36 80]
1332.8 Err:11 http://ports.ubuntu.com/ubuntu-ports noble-updates/main arm64 tzdata all 2024b-0ubuntu0.24.04
1332.8   404  Not Found [IP: 185.125.190.36 80]
1333.6 Err:15 http://ports.ubuntu.com/ubuntu-ports noble-updates/main arm64 libpython3.12-stdlib arm64 3.12.3-1ubuntu0.4
1333.6   404  Not Found [IP: 185.125.190.36 80]
[...]

PR #608, opened 2025-04-11
chore: add --fix-missing so devcontainer can build

kbroch-rivosinc:
relates to #604


AFOliveira:
If you do ./do clobber and then re-build does it still not work?

Issue #570, opened 2025-03-27
Add remaining debug CSRs


dhower-qc:
need to add:

- DCSR
- DPC

neverlandiz:
Hello @dhower-qc, could you assign this issue to me? Thanks

Issue #215, opened 2024-10-30
Add Sdext (External debug arch)


dhower-qc:
**Describe the data**

Need to add Sdext to the database. Also need IDL with "builtin" interface to interact with an external debugger model.

ThinkOpenly:
Some of this is done (PR #350), some of this is pending merge (#614). Some is going to be pretty challenging:

> Also need IDL with "builtin" interface to interact with an external debugger model.

Not saying "no", but there might be lower-hanging fruit which would be a better "good first issue". (And we might want to remove that tag from this issue, frankly.)

ThinkOpenly:
> Are there any easier tasks or ‘good first issues’ you recommend for someone new to the project?

Perhaps #223?

ThinkOpenly:
The work here should also include defining the `dret` instruction.

shivaansh0610-LUFFY:
Hi @ThinkOpenly and @dhower-qc ,

I saw that parts of this have already been covered by #350 and #614. I'd still be interested in working on the remaining pieces if it's available.

Do you think this issue is still open for contribution?

Thanks!

ThinkOpenly:
@shivaansh0610-LUFFY, given that I just assigned #214 to you, let's pick one to focus on... your choice.

PR #614, opened 2025-04-14
Add DCSR and DPC CSRs

neverlandiz:
This PR addresses Issue #570 and adds the remaining missing debug CSRs
- DCSR
- DPC

### DIFF
--- a/arch/csr/dcsr.yaml
+++ b/arch/csr/dcsr.yaml
@@ -1,0 +1,218 @@
+# yaml-language-server: $schema=../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: dcsr
+long_name: Debug Control and Status Register
+address: 0x7B0
+priv_mode: M
+length: 32
+description: |
+  Upon entry into Debug Mode, v and prv are updated with the privilege level the hart was previously in,
+  and cause is updated with the reason for Debug Mode entry. Other than these fields and nmip, the
+  other fields of dcsr are only writable by the external debugger.
+
+  Priority of reasons for entering Debug Mode from highest to lowest is shown below.
+  5:: resethaltreq
+  6:: halt group
+  3:: haltreq
+  2:: trigger
+  1:: ebreak
+  4:: step
+
+  For compatibility with old versions of this spec, resethaltreq and
+  haltreq are allowed to be at different positions than shown as long as:
+  1. resethaltreq is higher priority than haltreq
+  2. the relative order of the other four causes is maintained
+
+  This CSR is read/write.
+
+definedBy: Sdext
+fields:
+  DEBUGVER:
+    location: 31-28
+    type: RO
+    description: |
+      0 (none):: There is no debug support.
+      4 (1.0):: Debug support exists as it is described in this document.
+      15 (custom):: There is debug support, but it does not conform to any available version of this spec.
+    reset_value: UNDEFINED_LEGAL
+  EXTCAUSE:
+    location: 26-24
+    type: RO
+    description: |
+      When cause is 7, this optional field contains the value of a more specific halt reason than "other."
+      Otherwise it contains 0.
+
+      0 (critical error):: The hart entered a critical error state, as defined in the Smdbltrp extension.
+
+      All other values are reserved for future versions of this spec, or for use by other RISC-V extensions.
+    reset_value: 0
+  CETRIG:
+    location: 19
+    type: RW
+    description: |
+      This bit is part of Smdbltrp and only exists when that extension is implemented.
+      0 (disabled):: A hart in a critical error state does not enter
+      Debug Mode but instead asserts the critical-error signal to
+      the platform.
+      1 (enabled):: A hart in a critical error state enters Debug
+      Mode instead of asserting the critical-error signal to the
+      platform. Upon such entry into Debug Mode, the cause
+      field is set to 7, and the extcause field is set to 0, indicating
+      a critical error triggered the Debug Mode entry. This cause
+      has the highest priority among all reasons for entering
+      Debug Mode. Resuming from Debug Mode following an
+      entry from the critical error state returns the hart to the
+      critical error state.
+
+      When cetrig is 1, resuming from Debug Mode following an entry due to a critical
+      error will result in an immediate re-entry into Debug Mode due to the critical error.
+      The debugger may resume with cetrig set to 0 to allow the platform defined actions on
+      critical-error signal to occur. Other possible actions include initiating a hart or platform
+      reset using the Debug Module reset control.
+    reset_value: 0
+  PELP:
+    location: 18
+    type: RW
+    description: |
+      This bit is part of Zicfilp and only exists when that extension is implemented.
+      0 (NO_LP_EXPECTED):: No landing pad instruction expected.
+      1 (LP_EXPECTED):: A landing pad instruction is expected.
+    reset_value: 0
+  EBREAKVS:
+    location: 17
+    type: RW
+    description: |
+      0 (exception):: ebreak instructions in VS-mode behave as described in the Privileged Spec.
+      1 (debug mode):: ebreak instructions in VS-mode enter Debug Mode.
+      This bit is hardwired to 0 if the hart does not support virtualization mode.
+    reset_value: 0
+  EBREAKVU:
+    location: 16
+    type: RW
+    description: |
+      0 (exception):: ebreak instructions in VU-mode behave as described in the Privileged Spec.
+      1 (debug mode):: ebreak instructions in VU-mode enter Debug Mode.
+      This bit is hardwired to 0 if the hart does not support virtualization mode.
+    reset_value: 0
+  EBREAKM:
+    location: 15
+    type: RW
+    description: |
+      0 (exception):: ebreak instructions in M-mode behave as described in the Privileged Spec.
+      1 (debug mode):: ebreak instructions in M-mode enter Debug Mode.
+    reset_value: 0
+  EBREAKS:
+    location: 13
+    type: RW
+    description: |
+      0 (exception):: ebreak instructions in S-mode behave as described in the Privileged Spec.
+      1 (debug mode):: ebreak instructions in S-mode enter Debug Mode.
+      This bit is hardwired to 0 if the hart does not support S-mode.
+    reset_value: 0
+  EBREAKU:
+    location: 12
+    type: RW
+    description: |
+      0 (exception):: ebreak instructions in U-mode behave as described in the Privileged Spec.
+      1 (debug mode):: ebreak instructions in U-mode enter Debug Mode.
+      This bit is hardwired to 0 if the hart does not support U-mode.
+    reset_value: 0
+  STEPIE:
+    location: 11
+    type: RW
+    description: |
+      0 (interrupts disabled):: Interrupts (including NMI) are disabled during single stepping with step set.
+      This value should be supported.
+      1 (interrupts enabled):: Interrupts (including NMI) are enabled during single stepping with step set.
+      Implementations may hard wire this bit to 0. In that case interrupt behavior can be emulated by the
+      debugger. The debugger must not change the value of this bit while the hart is running.
+    reset_value: 0
+  STOPCOUNT:
+    location: 10
+    type: RW
+    description: |
+      0 (normal):: Increment counters as usual.
+      1 (freeze):: Don’t increment any hart-local counters while in Debug Mode or on ebreak instructions
+      that cause entry into Debug Mode. These counters include the instret CSR. On single-hart cores cycle
+      should be stopped, but on multi-hart cores it must keep incrementing.
+      An implementation may hardwire this bit to 0 or 1.
+    reset_value: UNDEFINED_LEGAL
+  STOPTIME:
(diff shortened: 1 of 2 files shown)


neverlandiz:
For core debug registers like DCSR and DPC, it’s stated that “these registers are only accessible from Debug Mode” (Section 4.9, debug specs). What would be the correct priv_mode for those CSRs? Debug mode isn’t one of the options in the schema (only M, S, U, VS).

ThinkOpenly:
> For core debug registers like DCSR and DPC, it’s stated that “these registers are only accessible from Debug Mode” (Section 4.9, debug specs). What would be the correct priv_mode for those CSRs? Debug mode isn’t one of the options in the schema (only M, S, U, VS).

Hmm. The Priv spec 1.3 "Debug Mode" says:
> Debug mode (D-mode) can be considered an additional privilege mode, with even more access than M-mode.

That sounds like a full Mode (capital M). It probably needs to be treated as such, adding a new mode in `schemas/csr_schema.json`, and an entry in `arch/isa/globals.isa`, but there don't appear to be any CSR bits to identify "D-mode" (I couldn't find any). So, which new enum value to choose is an open question. Maybe @dhower-qc has an idea?

Some additional information: `arch/isa/globals.isa` has:
```
# encoded as defined in the privilege spec
enum PrivilegeMode {
  M  0b011
  S  0b001
  HS 0b001 # alias for S when H extension is used
  U  0b000
  VS 0b101
  VU 0b100
}
```
These _3_ bits appear to be the concatenation of MPV and MPP, or at least they line up that way in the Priv spec, 18.4.1 "Machine Status Registers (mstatus and mstatush)", table 32:
image

It _may_ be that you could just pick a random new enum value (prepend a 4th bit, like `0b1000`?), but I don't know for sure. (You could certainly try!)

dhower-qc:
Yea, we do need to add D-mode in globals.isa, and yes, there are cases where we are relying on the encoding of PrivilegeMode being the table Paul added above. For example:

https://github.com/riscv-software-src/riscv-unified-db/blob/00f1540ddcf2bee1cdc88ac4f448f00edadb9451/arch/isa/globals.isa#L628-L634

As such, we should encode D as 0b1011 since it is M (0b0011) with extra permission.

arch/isa/globals.isa
```diff
@@ -40,6 +40,7 @@ enum PrivilegeMode {
   U  0b000
   VS 0b101
   VU 0b100
+  D  0b1011
```
ThinkOpenly, line 43: I'm debating whether all of the constants here should be made 4 bits wide. Let's go with "yes".

Also, I think it would help if the comment above this definition were expanded:
```
# encoding defined by concatenation of { D, MPV, MPP }:
# - D  : debug mode  (1b)
# - MPV: mstatus.MPV (1b)
# - MPP: mstatus.MPP (2b)
```

arch/csr/dcsr.yaml
```diff
@@ -0,0 +1,229 @@
...
+    type: RW
+    description: |
+      0 (disabled):: mprv in mstatus is ignored in Debug Mode.
+      1 (enabled):: mprv in mstatus takes effect in Debug Mode.
+      Implementing this bit is optional. It may be tied to either 0 or 1.
+    reset_value: UNDEFINED_LEGAL
+    sw_write(csr_value): |
+      if (!MPRVEN_IMPLEMENTED) {
+        unimplemented_csr($encoding);
+      }
```
ThinkOpenly, line 201: Add return of the value to be written:
```
return csr_value.MPRVEN;
```

ThinkOpenly:
One of the regression tests is failing:
```
/home/runner/work/riscv-unified-db/riscv-unified-db/lib/arch_obj_models/csr.rb:215:in `length': Unexpected length field for dpc (RuntimeError)
```
Might need a hint from @dhower-qc here, but adding DXLEN support might also include:
1. Adding a value for it in `cfgs/example_rv64_with_overlay.yaml`
2. Updating `arch/csr/schema.adoc`
3. Updating `arch/README.adoc`
4. Maybe updating `lib/arch_obj_models/csr.rb` (This is the spot which might be causing the failure.)

ThinkOpenly:
> My vote is to drop DXLEN, and get clarification from the debug spec that DXLEN is always identical to MXLEN.

Per Paul Donahue (Ventana, author of the Debug spec):
> I believe that DXLEN is always the same as MXLEN.  I thought that we fixed the definition when mxl became read-only but it looks like that was overlooked.

neverlandiz:
Thanks for the feedback! I'll remove the DXLEN code and use MXLEN for `dpc` instead.

dhower-qc:
> 
> Per Paul Donahue (Ventana, author of the Debug spec):
> 
> > I believe that DXLEN is always the same as MXLEN.  I thought that we fixed the definition when mxl became read-only but it looks like that was overlooked.

Excellent. That's going save us a few headaches.

arch/ext/Sdext.yaml
```diff
@@ -11,3 +11,15 @@ versions:
   - version: "1.0.0"
     state: ratified
     ratification_date: null
+params:
+  DCSR_MPRVEN_TYPE:
+    schema:
+      enum: [tied-to-0, tied-to-1, rw]
```
ThinkOpenly: I think this fixes the CI failures:
```suggestion
    schema:
      type: string
      enum: [tied-to-0, tied-to-1, rw]
```

ThinkOpenly:
@neverlandiz, if you are willing to make a few more changes:
- ratification date in `arch/ext/Sdext.yaml` could be set to 2025-02
- all of STEPIE, STOPCOUNT, and STOPTIME need new configuration parameters (in arch/ext/Sdext.yaml) because they can be hardwared to 0, 1, or read-write, just like MPRVEN.
- they also need their own `type()` and `reset_value` attributes.

If you are sufficiently tired of this PR and want to see it merged and move on with your life, we can merge this and open an issue, or we can take on our own commits here.

What's your preference? (No worries either way, honestly.)

neverlandiz:
@ThinkOpenly I'd be happy to make the changes! I plan on finishing the PRs that I have opened.

arch/csr/dcsr.yaml
```diff
@@ -127,34 +127,70 @@ fields:
...
       1 (interrupts enabled):: Interrupts (including NMI) are enabled during single stepping with step set.
       Implementations may hard wire this bit to 0. In that case interrupt behavior can be emulated by the
       debugger. The debugger must not change the value of this bit while the hart is running.
-    reset_value: 0
+    type(): |
+      if (DCSR_STEPIE_TYPE == "tied-to-0" || DCSR_STEPIE_TYPE == "tied-to-1") {
+        return CsrFieldType::RO;
+      } else if (DCSR_STEPIE_TYPE == "rw") {
+        return CsrFieldType::RW;
+      }
```
ThinkOpenly:
```suggestion
      return CsrFieldType::RW;
```
Personal preference, but I like to see that functions always return _something_. I know we have all the cases covered, but that's not obvious from just looking at this code...you'd have to go look at the definition of `DSCR_STEPIE_TYPE`.

arch/csr/dcsr.yaml
```diff
@@ -127,34 +127,70 @@ fields:
...
+    reset_value(): |
+      if (DCSR_STEPIE_TYPE == "tied-to-0") {
+        return 0;
+      } else if (DCSR_STEPIE_TYPE == "tied-to-1") {
+        return 1;
+      } else if (DCSR_STEPIE_TYPE == "rw") {
+        return UNDEFINED_LEGAL;
+      }
```
ThinkOpenly:
```suggestion
      return UNDEFINED_LEGAL;
```

arch/csr/dcsr.yaml
```diff
@@ -127,34 +127,70 @@ fields:
...
       that cause entry into Debug Mode. These counters include the instret CSR. On single-hart cores cycle
       should be stopped, but on multi-hart cores it must keep incrementing.
       An implementation may hardwire this bit to 0 or 1.
-    reset_value: UNDEFINED_LEGAL
+    type(): |
+      if (DCSR_STOPCOUNT_TYPE == "tied-to-0" || DCSR_STOPCOUNT_TYPE == "tied-to-1") {
+        return CsrFieldType::RO;
+      } else if (DCSR_STOPCOUNT_TYPE == "rw") {
+        return CsrFieldType::RW;
+      }
```
ThinkOpenly:
```suggestion
      return CsrFieldType::RW;
```

arch/csr/dcsr.yaml
```diff
@@ -127,34 +127,70 @@ fields:
...
+    reset_value(): |
+      if (DCSR_STOPCOUNT_TYPE == "tied-to-0") {
+        return 0;
+      } else if (DCSR_STOPCOUNT_TYPE == "tied-to-1") {
+        return 1;
+      } else if (DCSR_STOPCOUNT_TYPE == "rw") {
+        return UNDEFINED_LEGAL;
+      }
```
ThinkOpenly:
```suggestion
      return UNDEFINED_LEGAL;
```

arch/csr/dcsr.yaml
```diff
@@ -127,34 +127,70 @@ fields:
...
       time will reflect the latest value of mtime again.
       While all harts have stoptime=1 and are in Debug Mode, mtime is allowed to stop incrementing.
       An implementation may hardwire this bit to 0 or 1.
-    reset_value: UNDEFINED_LEGAL
+    type(): |
+      if (DCSR_STOPTIME_TYPE == "tied-to-0" || DCSR_STOPTIME_TYPE == "tied-to-1") {
+        return CsrFieldType::RO;
+      } else if (DCSR_STOPTIME_TYPE == "rw") {
+        return CsrFieldType::RW;
+      }
```
ThinkOpenly:
```suggestion
        return CsrFieldType::RW;
```

arch/csr/dcsr.yaml
```diff
@@ -127,34 +127,70 @@ fields:
...
+    reset_value(): |
+      if (DCSR_STOPTIME_TYPE == "tied-to-0") {
+        return 0;
+      } else if (DCSR_STOPTIME_TYPE == "tied-to-1") {
+        return 1;
+      } else if (DCSR_STOPTIME_TYPE == "rw") {
+        return UNDEFINED_LEGAL;
+      }
```
ThinkOpenly:
```suggestion
      return UNDEFINED_LEGAL;
```

arch/csr/dcsr.yaml
dhower-qc: FYI, there is also an `assert` function available, so you can also do:

```
if () {
} else {
  assert(DCSR_STEPIE_TYPE == "rw", "Unhandled DCSR_STEPIE_TYPE value");
  return CsrFieldType::RW;
}
```

arch/csr/dcsr.yaml
```diff
@@ -127,34 +127,70 @@ fields:
...
       debugger. The debugger must not change the value of this bit while the hart is running.
-    reset_value: 0
+    type(): |
+      if (DCSR_STEPIE_TYPE == "tied-to-0" || DCSR_STEPIE_TYPE == "tied-to-1") {
+        return CsrFieldType::RO;
+      } else if (DCSR_STEPIE_TYPE == "rw") {
+        return CsrFieldType::RW;
+      }
+    reset_value(): |
+      if (DCSR_STEPIE_TYPE == "tied-to-0") {
```
dhower-qc: Let's change "tied-to-*" to "read-only-*". That's how the spec refers to the situation.

dhower-qc (changes requested):

arch/csr/dcsr.yaml
ThinkOpenly: I see you added return statements after the "else" clause, but I was actually suggesting to _replace_ the "else" clause entirely.

So, with Derek's suggestion in mind, this becomes:
```
if () {
}
assert(DCSR_STEPIE_TYPE == "rw", "Unhandled DCSR_STEPIE_TYPE value");
return CsrFieldType::RW;
```
(move the "return" outside of the conditional control flow, so it's very obvious that this function will always return something.)

neverlandiz: Oh I see! Thanks for the feedback!

tools/ruby-gems/udb/lib/udb/cfg_arch.rb
```diff
@@ -481,6 +483,7 @@ def initialize(name, config, arch_path)
         name: @name,
         csrs:
       )
+
```
ThinkOpenly: Doesn't look like the CI pre-commit test likes this blank line. (I can't explain why. ??)

neverlandiz: I think the pre-commit tests are also failing due to missing copyright and license information. Looking at the other files, I think it's those two lines?
```
# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
# SPDX-License-Identifier: BSD-3-Clause-Clear
```

ThinkOpenly (approved):

ThinkOpenly:
@dhower-qc can we merge this? We got a request to work on #215, but this PR covers a lot of that work.

dhower-qc (approved):

PR #615, opened 2025-04-15
Add long_name and description to AES instructions

sudo-abdullah:
This PR is to add the missing `long_name` and `description`  in some of the YAML files as @AFOliveira shared.

### DIFF
--- a/arch/inst/Zk/aes32dsi.yaml
+++ b/arch/inst/Zk/aes32dsi.yaml
@@ -3,7 +3,7 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: aes32dsi
-long_name: AES final round decryption instruction for RV32.
+long_name: AES final round decryption instruction for RV32
 description: |
  This instruction sources a single byte from `rs2` according to `bs`. To this it applies the inverse AES
  SBox operation, and XOR's the result with `rs1`. This instruction must always be implemented such
--- a/arch/inst/Zk/aes32dsi.yaml
+++ b/arch/inst/Zk/aes32dsi.yaml
@@ -5,9 +5,9 @@ kind: instruction
 name: aes32dsi
 long_name: AES final round decryption instruction for RV32
 description: |
- This instruction sources a single byte from `rs2` according to `bs`. To this it applies the inverse AES
- SBox operation, and XOR's the result with `rs1`. This instruction must always be implemented such
- that its execution latency does not depend on the data being operated on.
+  This instruction sources a single byte from `rs2` according to `bs`. To this it applies the inverse AES
+  SBox operation, and XOR's the result with `rs1`. This instruction must always be implemented such
+  that its execution latency does not depend on the data being operated on.
 
 definedBy:
   anyOf: [Zk, Zkn, Zknd]
--- a/arch/inst/Zk/aes32dsi.yaml
+++ b/arch/inst/Zk/aes32dsi.yaml
@@ -3,9 +3,12 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: aes32dsi
-long_name: No synopsis available.
+long_name:  AES final round decryption instruction for RV32.
 description: |
-  No description available.
+  Sources a single byte from rs2 according to bs. To this it applies the inverse AES
+  SBox operation, and XOR’s the result with rs1. Must always be implemented such
+  that its execution latency does not depend on the data being operated on.
+
 definedBy:
   anyOf: [Zk, Zkn, Zknd]
 base: 32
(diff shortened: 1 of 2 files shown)


arch/inst/Zk/aes32dsi.yaml
```diff
@@ -3,9 +3,12 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: aes32dsi
-long_name: No synopsis available.
+long_name:  AES final round decryption instruction for RV32.
 description: |
-  No description available.
+  Sources a single byte from rs2 according to bs. To this it applies the inverse AES
+  SBox operation, and XOR’s the result with rs1. Must always be implemented such
```
ThinkOpenly, line 9: Let's just keep all of the text from the spec here, from "This instruction sources..." to "...data being operated on."
As in `aes32dsmi`, below, put backquotes around `rs1`, `rs2`, `bs`.
Also, it's subtle, but use a single quote `'` instead of whatever is there now (`’`).

arch/inst/Zk/aes32dsi.yaml
```diff
@@ -3,9 +3,12 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: aes32dsi
-long_name: No synopsis available.
+long_name:  AES final round decryption instruction for RV32.
```
ThinkOpenly, line 6: nit: remove one of the two spaces before "AES".

arch/inst/Zk/aes32dsmi.yaml
```diff
@@ -3,9 +3,11 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: aes32dsmi
-long_name: No synopsis available.
+long_name: AES middle round decryption instruction for RV32.
 description: |
-  No description available.
+  Sources a single byte from `rs2` according to `bs`. To this, it applies the inverse AES
+  SBox operation, and a partial inverse MixColumn, before XOR’ing the result with rs1. Must always be implemented such that its execution latency does not depend on the
```
ThinkOpenly, line 9: Same request here for replacing `’` with `'`.
Also, let's just keep all of the text from the spec here, from "This instruction sources..." to "...data being operated on."

sudo-abdullah (commented):
@ThinkOpenly Please have a look again.

arch/inst/Zk/aes32dsi.yaml
```diff
@@ -3,9 +3,12 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: aes32dsi
-long_name: No synopsis available.
+long_name: AES final round decryption instruction for RV32.
```
ThinkOpenly, line 6: nit: delete the period `.`
(I should've caught this in the first review.)

arch/inst/Zk/aes32dsmi.yaml
```diff
@@ -3,9 +3,9 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: aes32dsmi
-long_name: No synopsis available.
+long_name: AES middle round decryption instruction for RV32.
```
ThinkOpenly, line 6: same

arch/inst/Zk/aes32dsmi.yaml
```diff
@@ -3,9 +3,9 @@
...
 description: |
-  No description available.
+  This instruction sources a single byte from `rs2` according to `bs`. To this it applies the inverse AES SBox operation, and a partial inverse MixColumn, before XOR'ing the result with `rs1`. This instruction must always be implemented such that its execution latency does not depend on the data being operated on.
```
ThinkOpenly, line 8: nit, optional: since I've asked for changes anyway, could you break this line up into shorter lines? The breaks in the spec are fine:
> This instruction sources a single byte from `rs2` according to `bs`. To this it applies the inverse AES
> SBox operation, and a partial inverse MixColumn, before XOR'ing the result with `rs1`. This
> instruction must always be implemented such that its execution latency does not depend on the
> data being operated on.

ThinkOpenly (changes requested):

ThinkOpenly (approved):

AFOliveira:
Hi @sudo-abdullah. Me and @ThinkOpenly have already approved this PR, so you should not add more information to it. If you prefer, you can create a separate PR and put it as a "Draft PR" (I may help with this) and then add the new information you are adding.

AFOliveira:
@sudo-abdullah you gotta run pre commit checks locally in order to pass this test remotely, can you do that?

dhower-qc (approved):

Issue #625, opened 2025-04-16
typo: Endianness should have two n's


kbroch-rivosinc:
Currently have these instances with only one: 

```
kbroch@kbroch-mac.local:~/rvi/non-repo-repos/riscv-unified-db on  main via  v22.14.0 via 🐍 v3.13.3 via 💎 v3.2.3 took 44s
❯ rg -i endianess
cert_flow.txt
81:  M_MODE_ENDIANESS: little

cfgs/config_validation.rb
81:require_param :S_MODE_ENDIANESS  if ext?(:S)
82:require_param :U_MODE_ENDIANESS  if ext?(:U)
83:require_param :VU_MODE_ENDIANESS if ext?(:H)
84:require_param :VS_MODE_ENDIANESS if ext?(:H)

cfgs/mc100-32-full-example.yaml
28:  M_MODE_ENDIANESS: little

cfgs/MC100-32.yaml
22:  M_MODE_ENDIANESS: little

arch/certificate_model/MockCertificateModel.yaml
135:      M_MODE_ENDIANESS:

cfgs/example_rv64_with_overlay.yaml
426:  M_MODE_ENDIANESS: little
433:  S_MODE_ENDIANESS: little
440:  U_MODE_ENDIANESS: little
447:  VU_MODE_ENDIANESS: little
454:  VS_MODE_ENDIANESS: little

backends/cpp_hart_gen/renode/UdbCpu.cs
34:            Machine machine, Endianess endianness = Endianess.LittleEndian,

arch/certificate_model/MC100-32.yaml
136:      M_MODE_ENDIANESS:

backends/cpp_hart_gen/cpp/test/test_decode.cpp
52:  M_MODE_ENDIANESS: little

docs/ruby/Idl/SymbolTable.html
[...]

PR #627, opened 2025-04-16
task: fix typo and put 2 N's in endianness

kbroch-rivosinc:
relates to #625


dhower-qc (approved):

james-ball-qualcomm:
I just cloned the main branch and tried to build the AC100-CRD artifact and got an error because not all ENDIANESS were changed to ENDIANNESS.

image

 The AC100-CRD artifact isn't in test:regress but it is in test:nightly. I see those haven't been running because they take too long. This is probably related to the performance bug accidentally introduced by Derek's cpp_hart merge.

image

PR #630, opened 2025-04-17
fix: misa[M,S,U,V] wrong offset bits

Kingfish404:
https://github.com/riscv/riscv-isa-manual/blob/main/src/machine.adoc

### DIFF
--- a/arch/csr/misa.yaml
+++ b/arch/csr/misa.yaml
@@ -134,7 +134,7 @@ fields:
     definedBy: I
     reset_value: 1
   M:
-    location: 13
+    location: 12
     description: |
       Indicates support for the `M` (integer multiply/divide) extension.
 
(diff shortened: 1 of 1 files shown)


arch/csr/misa.yaml
```diff
@@ -146,7 +146,7 @@ fields:
       return implemented?(ExtensionName::M) ? 1 : 0;
     definedBy: M
   S:
-    location: 19
+    location: 17
```
AFOliveira, line 149: Isn't this 18?
image

AFOliveira (changes requested):

arch/csr/misa.yaml
Kingfish404: Fix it again

AFOliveira: Great! Thanks!

AFOliveira (approved):

ThinkOpenly (approved):

dhower-qc (approved):

Issue #633, opened 2025-04-17
deploy pages workflow fails after merge of CSC fork into main


james-ball-qualcomm:
https://github.com/riscv-software-src/riscv-unified-db/actions/runs/14523448583/job/40749522289

Error message is: /home/runner/work/riscv-unified-db/riscv-unified-db/backends/cfg_html_doc/adoc_gen.rake:131:in `block (2 levels) in <top (required)>': No config named generic_rv64 (RuntimeError)

Happens while running "Build html documentation for generic_rv64"

james-ball-qualcomm:
Hold on. I think I see some massive redundancy. I just realized that the comment you have in pages.yml that says:

```
      - name: Create deploy dir
        run: /bin/bash lib/deploy.sh

Is much more than just creating the deploy dir. You are invoking the deploy.sh script which builds lots of artifacts and puts them in the _site directory which is redundant with all the stuff I added to the pages.yml:

![Image](https://github.com/user-attachments/assets/651c55db-42ec-4381-9a59-859da83f7c49)

james-ball-qualcomm:
So, I think I just need to remove all the extra stuff I added to pages.yml since deploy.sh handles it.

PR #635, opened 2025-04-17
Fix for #633 - pages.yml duplicating artifact building in deploy.sh

james-ball-qualcomm:
and was out-of-date and still using "generic_rv64" which has been renamed.

Let's see after I create this PR if the deploy pages action now works.


james-ball-qualcomm:
Ran the deploy.sh script on my laptop and it passed. I recommend you approve this PR and we merge it in.

I've attached the output log from the deploy.sh as proof that it now passes.
[deploy.log](https://github.com/user-attachments/files/19812814/deploy.log)

james-ball-qualcomm:
Thanks Afonso for the tip about the "act" package. I don't seem to have the "act" command in my path on WSL but maybe I can add it. Before I do that, Derek, is this what you recommend? In the meantime, I'm going to research if there is a way to get GitHub to run the pages.yaml action like it does when a PR is merged. Afonso, do you know about this?

james-ball-qualcomm:
Hold on. I see I can re-run the pages.yaml workflow on my branch with the fix. I'll give that a shot.

image

james-ball-qualcomm:
Well, that did look. Seems like I need some kind of permission on my branch that I don't have. Should we try to add permission to this branch? If so, can I do this or do I need Derek?

image

dhower-qc:
> Well, that didn't look good. Seems like I need some kind of permission on my branch that I don't have. Should we try to add permission to this branch? If so, can I do this or do I need Derek?

The deploy action publishes artifacts on the Pages site. It can, and should, only run from main.

dhower-qc (approved):

PR #637, opened 2025-04-19
correct: Remove periods from instruction long_names

ThinkOpenly:
Remove extraneous periods at the end of instruction "long_name" fields.

### DIFF
--- a/arch/inst/C/c.ebreak.yaml
+++ b/arch/inst/C/c.ebreak.yaml
@@ -3,7 +3,7 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: c.ebreak
-long_name: Breakpoint exception.
+long_name: Breakpoint exception
 description: |
   The C.EBREAK instruction is used by debuggers to cause control to be transferred back to
   a debugging environment. Unless overridden by an external debug environment,
--- a/arch/inst/C/c.jalr.yaml
+++ b/arch/inst/C/c.jalr.yaml
@@ -3,7 +3,7 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: c.jalr
-long_name: Jump and Link Register.
+long_name: Jump and Link Register
 description: |
   C.JALR (jump and link register) performs the same operation as C.JR, but additionally writes the address of the instruction following the jump (pc+2) to the link register, x1.
   C.JALR expands to jalr x1, 0(rs1).
--- a/arch/inst/F/fclass.s.yaml
+++ b/arch/inst/F/fclass.s.yaml
@@ -3,7 +3,7 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: fclass.s
-long_name: Single-precision floating-point classify.
+long_name: Single-precision floating-point classify
 description: |
   The `fclass.s` instruction examines the value in floating-point register
   _fs1_ and writes to integer register _rd_ a 10-bit mask that indicates
--- a/arch/inst/F/fcvt.w.s.yaml
+++ b/arch/inst/F/fcvt.w.s.yaml
@@ -3,7 +3,7 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: fcvt.w.s
-long_name: Convert single-precision float to integer word to signed 32-bit integer.
+long_name: Convert single-precision float to integer word to signed 32-bit integer
 description: |
   Converts a floating-point number in floating-point register _fs1_ to a signed 32-bit integer indicates
   integer register _rd_.
--- a/arch/inst/Smrnmi/mnret.yaml
+++ b/arch/inst/Smrnmi/mnret.yaml
@@ -3,7 +3,7 @@
 $schema: inst_schema.json#
 kind: instruction
 name: mnret
-long_name: Machine mode resume from the RNMI or Double Trap handler.
+long_name: Machine mode resume from the RNMI or Double Trap handler
 description: |
   MNRET is an M-mode-only instruction that uses the values in mnepc and mnstatus to return to the
   program counter, privilege mode, and virtualization mode of the interrupted context. This instruction
--- a/arch/inst/Zcmp/cm.popret.yaml
+++ b/arch/inst/Zcmp/cm.popret.yaml
@@ -3,7 +3,7 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: cm.popret
-long_name: Destroy function call stack frame and return to `ra`.
+long_name: Destroy function call stack frame and return to `ra`
 description: |
   Destroy stack frame: load `ra` and 0 to 12 saved registers from the stack frame, deallocate the stack frame, return to `ra`.
   This instruction pops (loads) the registers in `reg_list` from stack memory, and then adjusts the stack pointer by `stack_adj` and then return to `ra`.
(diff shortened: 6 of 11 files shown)


ThinkOpenly:
This should probably be made into a pre-commit hook, if I can figure out how to create that. (Lots of other things we could check like this, too.)

AFOliveira:
Easier than that, I think we may change the json schema regex to not allow an expression finished with a .

ThinkOpenly:
> Easier than that, I think we may change the json schema regex to not allow an expression finished with a "."

Would that be by adding:
```
        "long_name": {
          "type": "string",
          "description": "Descriptive name for the CSR"
          "pattern": "[^\.]$"
        },
```
?

AFOliveira:
Almost that, but in JSON, you need to write the pattern with two backslashes: `"pattern": "[^\\.]$"`
This is because the regex needs a backslash to escape the period (\.) and then, JSON itself requires backslashes to be escaped with another backslash

So while the actual regex is [^\.]$, when written in a JSON string it must be [^\\.]$ or the JSON parser will interpret it incorrectly.

kbroch-rivosinc:
with this: https://regex101.com/r/DDVrJ7/1

```diff
diff --git a/schemas/inst_schema.json b/schemas/inst_schema.json
index d6fbdf22..df12355c 100644
--- a/schemas/inst_schema.json
+++ b/schemas/inst_schema.json
@@ -180,7 +180,8 @@
         },
         "long_name": {
           "type": "string",
-          "description": "Short description of the instruction"
+          "description": "Short description of the instruction",
+          "pattern": ".*[^.]$"
         },
         "description": {
           "type": "string",
```

The check-jsonschema will find these.  The problem is there's about 900+ "No synopsis available." ones as well.

```
❯ pre-commit run --all-files check-jsonschema | head
Validate files with jsonschema...........................................Failed
- hook id: check-jsonschema
- exit code: 1

Schema validation errors were encountered.
  arch/inst/V/vsuxseg7ei32.v.yaml::$.long_name: 'No synopsis available.' does not match '.*[^.]$'
  arch/inst/V/vfrdiv.vf.yaml::$.long_name: 'No synopsis available.' does not match '.*[^.]$'
  arch/inst/V/vfmacc.vf.yaml::$.long_name: 'No synopsis available.' does not match '.*[^.]$'
  arch/inst/V/vwmaccu.vx.yaml::$.long_name: 'No synopsis available.' does not match '.*[^.]$'
  arch/inst/V/vloxseg5ei64.v.yaml::$.long_name: 'No synopsis available.' does not match '.*[^.]$'
```

Also for this rule, Shouldn't it really be "no punctuation at the end of the string"?

ThinkOpenly:
> with this [...] The check-jsonschema will find these.
Yay!

> The problem is there's about 900+ "No synopsis available." ones as well.
Boo! Forgot about those. Need to clean those up first.

> Also for this rule, Shouldn't it really be "no punctuation at the end of the string"?

Not sure what context you're referring to. Can you associate a message with a pattern?

kbroch-rivosinc:
> Not sure what context you're referring to. Can you associate a message with a pattern?

Same context (although if this PR is only meant to solve periods at the end then I guess I'm expanding the context of the PR) ,  just saying in the future, we probably don't want commas, exclamation points, etc as well as periods at the end of this string (no matter how unlikely someone is to put those there): 

```diff
diff --git a/schemas/inst_schema.json b/schemas/inst_schema.json
index d6fbdf22..4f0c5c05 100644
--- a/schemas/inst_schema.json
+++ b/schemas/inst_schema.json
@@ -180,7 +180,8 @@
         },
         "long_name": {
           "type": "string",
-          "description": "Short description of the instruction"
+          "description": "Short description of the instruction",
+          "pattern": ".*[^.|!|,]$"
         },
         "description": {
           "type": "string",
```

It's not all punctuation, as there's some strings with parens.

And to expand the context more(probably worthy of another GH issue)  there's long_name in other schemas: 

```
kbroch@kbroch-mac.local:~/rvi/non-repo-repos/riscv-unified-db on  main [!] via  v22.14.0 via 🐍 v3.13.3 via 💎 v3.2.3 took 5m26s
❯ rg long_name schemas
schemas/proc_cert_class_schema.json
5:  "required": ["$schema", "kind", "name", "long_name", "introduction"],
21:    "long_name": {

schemas/inst_schema.json
158:        "long_name",
181:        "long_name": {

schemas/proc_cert_model_schema.json
5:  "required": ["$schema", "kind", "name", "long_name", "base"],
69:    "long_name": {

schemas/ext_schema.json
78:        "long_name",
94:        "long_name": {

schemas/csr_schema.json
18:        "long_name": {
181:        "long_name",
210:        "long_name": {
```

ThinkOpenly:
> Also for this rule, Shouldn't it really be "no punctuation at the end of the string"?

I see. You generalized to "punctuation" instead of "periods". Understood now.

> in the future, we probably don't want commas, exclamation points, etc as well as periods at the end of this string (no matter how unlikely someone is to put those there):

Ha-ha! Surprise! It's not really that bad, if you ignore letters, numbers, close-parens, quotation marks:
```
$ grep -H 'long_name:.*[^a-z)0-9A-Z\"]$' $(find . -name '*.yaml')
./arch/inst/Zcmp/cm.popret.yaml:long_name: Destroy function call stack frame and return to `ra`
./arch/inst/Zcmp/cm.popretz.yaml:long_name: Destroy function call stack frame, move zero to `a0` and return to `ra`

./arch_overlay/example/ext/Xcustom.yaml:long_name: A new custom extension!
./arch_overlay/qc_iu/inst/Xqccmp/qc.cm.popret.yaml:long_name: Destroy function call stack frame and return to `ra`.
./arch_overlay/qc_iu/inst/Xqccmp/qc.cm.popretz.yaml:long_name: Destroy function call stack frame, move zero to `a0` and return to `ra`.
```

The backquotes could probably be removed or replaced with single ticks `'` or something other than backticks, or just leave them alone.

The exclamation point could go, of course.

> 
> ```diff
> +          "pattern": ".*[^.|!|,]$"
> ```

The syntax you used is different than I expected, and different than what Afonso suggested as well. Is there no backslash escaping required?

Also, do you need those vertical bar separators?

kbroch-rivosinc:
> The syntax you used is different than I expected, and different than what Afonso suggested as well. Is there no backslash escaping required?

Works for me, I see pre-commit hook WAI and also vscode flagging problems while editing.
 
EMCAscript version of regex: https://regex101.com/r/Mhceju/1

Not sure what the final list of disallowed punctuation should be but IMO `.*[^.!,(]$` is a good start

> 
> Also, do you need those vertical bar separators?

You're right, vertical bars are unnecessary

AFOliveira:
Funny enough this is what is breaking the pre-commit test 

`JSON Schema Validation Error for ext/Xcustom.yaml:
'A new custom extension!' does not match '.*[^.!,]$'
`

AFOliveira (approved):

kbroch-rivosinc:
> Funny enough this is what is breaking the pre-commit test
> 
> `JSON Schema Validation Error for ext/Xcustom.yaml: 'A new custom extension!' does not match '.*[^.!,]$' `

yeah the check-jsonschema pre-commit hook doesn't look at arch_overlay dir currently.  I'll put a PR in for that.

dhower-qc (approved):

AFOliveira:
Merged with main because it was stuck on cpp-hart test

Issue #224, opened 2024-10-30
Add Zawrs (Wait on reservation set)


dhower-qc:
**Describe the data**

Add spec + IDL

PR #640, opened 2025-04-21
Add spec for Wait-on-Reservation-Set Instructions

sudo-abdullah:
Spec for extension Zawrs instructions (#224)
- `wrs.nto`
- `wrs.sto`


arch/inst/Zk/aes32dsi.yaml
```diff
@@ -1,30 +0,0 @@
-# yaml-language-server: $schema=../../../schemas/inst_schema.json
```
AFOliveira, line 1: Was this intended?

arch/inst/Zk/aes32dsmi.yaml
```diff
@@ -1,30 +0,0 @@
-# yaml-language-server: $schema=../../../schemas/inst_schema.json
```
AFOliveira, line 1: Was this intended?

arch/inst/Zawrs/wrs.sto.yaml
```diff
@@ -3,9 +3,18 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: wrs.sto
-long_name: No synopsis available.
+long_name: Wait-on-Reservation-Set with short timeout
 description: |
-  No description available.
+  wrs.sto` instruction causes the hart to temporarily stall execution in a low-power
+  state as long as the reservation set is valid and no pending interrupts, even if disabled, are observed.
+  `wrs.sto` is to mitigate the wasteful looping incase of polling on memory locations.
```
AFOliveira, line 10: Is this line in spec?

arch/inst/Zawrs/wrs.sto.yaml
```diff
@@ -3,9 +3,18 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: wrs.sto
-long_name: No synopsis available.
+long_name: Wait-on-Reservation-Set with short timeout
 description: |
-  No description available.
+  wrs.sto` instruction causes the hart to temporarily stall execution in a low-power
```
AFOliveira, line 8: wrs.sto\`  should likely be  \`wrs.sto\`

arch/inst/Zawrs/wrs.nto.yaml
```diff
@@ -3,9 +3,17 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: wrs.nto
-long_name: No synopsis available.
+long_name: Wait-on-Reservation-Set with No timeout
 description: |
-  No description available.
+  `wrs.nto` instruction causes the hart to temporarily stall execution in a low-power
+  state as long as the reservation set is valid and no pending interrupts, even if disabled, are observed.
+  `wrs.nto` is to mitigate the wasteful looping incase of polling on memory locations.
```
AFOliveira, line 10: Here we maybe could follow exactly what the spect states, right?

> To mitigate the wasteful looping in such usages, a WRS.NTO (WRS-with-no-timeout) instruction is provided. Instead of polling for a store to a specific memory location, software registers a reservation set that includes all the bytes of the memory location using the LR instruction. Then a subsequent WRS.NTO instruction would cause the hart to temporarily stall execution in a low-power state until a store occurs to the reservation set or an interrupt is observed.

AFOliveira (changes requested):

arch/inst/Zawrs/wrs.sto.yaml
```diff
@@ -3,9 +3,17 @@
...
-  No description available.
+  `wrs.sto` instruction causes the hart to temporarily stall execution in a low-power
+  state as long as the reservation set is valid and no pending interrupts, even if disabled, are observed.
+  Sometimes the program waiting on a memory update may also need to carry out a task at a future time
+  or otherwise place an upper bound on the wait. To support such use cases a second instruction
+  `wrs.sto` is provided that works like `wrs.nto` but bounds the stall duration
+  to an implementation-define short timeout such that the stall is terminated on the timeout if no other
+  conditions have occurred to terminate the stall. The program using this instruction may then
+  determine if its deadline has been reached. `wrs.sto` instruction follows the rules of the WFI
+  instruction for resuming execution on a pending interrupt.
```
ThinkOpenly, line 16: > Instead of polling for a store to a specific memory location, software registers a reservation set that includes all the bytes of the memory location using the LR instruction.
> 
> A subsequent WRS.STO instruction would cause the hart to temporarily stall execution in a low-power state until a store occurs to the reservation set or an interrupt is observed.
> 
> Sometimes the program waiting on a memory update may also need to carry out a task at a future time or otherwise place an upper bound on the wait. To support such use cases, WRS.STO bounds the stall duration to an implementation-define short timeout such that the stall is terminated on the timeout if no other conditions have occurred to terminate the stall. The program using this instruction may then determine if its deadline has been reached.
> 
> WRS.STO causes the hart to temporarily stall execution in a low-power state as long as the reservation set is valid and no pending interrupts, even if disabled, are observed. For WRS.STO the stall duration is bounded by an implementation defined short timeout. These instructions are not supported in a constrained LR/SC loop.
> 
> Hart execution may be stalled while the following conditions are all satisfied:
> a. The reservation set is valid
> b. If WRS.STO, a "short" duration since start of stall has not elapsed
> c. No pending interrupt is observed (see the rules below)
> 
> While stalled, an implementation is permitted to occasionally terminate the stall and complete execution for any reason.
> 
> WRS.STO follows the rules of the WFI instruction for resuming execution on a pending interrupt.
> 
> Since WRS.STO can complete execution for reasons other than stores to the reservation set, software will likely need a means of looping until the required stores have occurred.
> 
> [Note]
> The duration of a WRS.STO instruction’s timeout may vary significantly within and among implementations. In typical implementations this duration should be roughly in the range of 10 to 100 times an on-chip cache miss latency or a cacheless access to main memory.

ThinkOpenly (changes requested):

arch/inst/Zawrs/wrs.sto.yaml
```diff
@@ -3,9 +3,17 @@
...
 definedBy: Zawrs
 assembly: wrs_sto
```
ThinkOpenly, line 18: This also needs to change:
```
assembly: ""
```

sudo-abdullah: @ThinkOpenly I've updated as you asked. Let me know if it requires more.

arch/inst/Zawrs/wrs.nto.yaml
```diff
@@ -3,11 +3,33 @@
...
+  This instruction is not supported in a constrained LR/SC loop.
+  While stalled, an implementation is permitted to occasionally terminate the stall and complete
+  execution for any reason. `wrs.nto` follows the rules of the WFI instruction for resuming execution
+  on a pending interrupt. When the TW (Timeout Wait) bit in `mstatus` is set and `wrs.nto` is executed
+  in any privilege mode otherthan M mode, and it does not complete within an implementation-specific
+  bounded time limit, the `wrs.nto` instruction will cause an illegal instruction exception.
+  When executing in VS or VU mode, if the VTW bit is set in `hstatus`, the TW bit in `mstatus` is clear,
+  and the `wrs.nto` does not complete within an implementation-specific bounded time limit,
+  the `wrs.nto` instruction will cause a virtual instruction exception.
+  [Note]
```
ThinkOpenly, line 22: I think there needs to be a blank line before the "[Note]" line, everywhere.

arch/inst/Zawrs/wrs.nto.yaml
```diff
@@ -3,11 +3,33 @@
...
-long_name: No synopsis available.
+long_name: Wait-on-Reservation-Set-with-No-Timeout
 description: |
-  No description available.
+  To mitigate the wasteful looping in such usages, a `wrs.nto` (WRS-with-no-timeout) instruction is provided.
+  Instead of polling for a store to a specific memory location, software registers a reservation set that
+  includes all the bytes of the memory location using the LR instruction. Then a subsequent `wrs.nto`
+  instruction would cause the hart to temporarily stall execution in a low-power state until a store
+  occurs to the reservation set or an interrupt is observed.
+  This instruction is not supported in a constrained LR/SC loop.
```
ThinkOpenly, line 13: add a blank line between paragraphs, everywhere (or they get merged)

ThinkOpenly (changes requested):

arch/inst/Zawrs/wrs.nto.yaml
```diff
@@ -10,18 +10,22 @@ description: |
...
+
   This instruction is not supported in a constrained LR/SC loop.
   While stalled, an implementation is permitted to occasionally terminate the stall and complete
-  execution for any reason. `wrs.nto` follows the rules of the WFI instruction for resuming execution
+  execution for any reason.
+  `wrs.nto` follows the rules of the WFI instruction for resuming execution
   on a pending interrupt. When the TW (Timeout Wait) bit in `mstatus` is set and `wrs.nto` is executed
   in any privilege mode otherthan M mode, and it does not complete within an implementation-specific
   bounded time limit, the `wrs.nto` instruction will cause an illegal instruction exception.
   When executing in VS or VU mode, if the VTW bit is set in `hstatus`, the TW bit in `mstatus` is clear,
```
ThinkOpenly, line 21: add blank lines here, too:
- after "reason."
- after "pending interrupt."
- after "exception."
... all to match the paragraph structure in the spec.

arch/inst/Zawrs/wrs.nto.yaml
dhower-qc: Bigger discussion: should we start using the "array form" of description to track provenance?

arch/inst/Zawrs/wrs.sto.yaml
```diff
@@ -3,11 +3,37 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: wrs.sto
-long_name: No synopsis available.
+long_name: Wait-on-Reservation-Set-with-Short-Timeout
```
dhower-qc, line 6: Why use dashes?

ThinkOpenly: That's what the spec uses.
> Chapter 15. "Zawrs" Extension for Wait-on-Reservation-Set instructions

> WRS.NTO (WRS-with-no-timeout)

> WRS.STO (WRS-with-short-timeout)

ThinkOpenly:
> should we start using the "array form" of description to track provenance?

Probably, although some training and/or documentation is needed.

arch/inst/Zawrs/wrs.nto.yaml
dhower-qc: Pulled the trigger. Please update the descriptions to the [structure description](https://github.com/riscv-software-src/riscv-unified-db/blob/main/doc/prose_schema.adoc) format.

dhower-qc (approved):

AFOliveira (approved):

ThinkOpenly (approved):

Issue #642, opened 2025-04-21
check-jsonschema pre-commit hook doesn't validate arch_overlay dir files


kbroch-rivosinc:
Currently only arch dir yaml files are being compared to check-jsonschema pre-commit hook
arch_overlay has other yaml files that should but those problems are flagged to CI ruby code here: https://github.com/riscv-software-src/riscv-unified-db/actions/runs/14578986964/job/40891298108?pr=637#step:4:78

kbroch-rivosinc:
I'll put a PR in for this.

kbroch-rivosinc:
Actually this might not work if the overlay is only meant to override some part of an existing arch yaml files like marchid then trying to compare it against the jsonschema files will result in problems like this: 

```
Validate files with jsonschema...........................................Failed
- hook id: check-jsonschema
- exit code: 1

ok -- validation done
ok -- validation done
ok -- validation done
ok -- validation done
Schema validation errors were encountered.
  arch_overlay/example/csr/marchid.yaml::$: '$schema' is a required property
  arch_overlay/example/csr/marchid.yaml::$: 'kind' is a required property
  arch_overlay/example/csr/marchid.yaml::$: 'name' is a required property
  arch_overlay/example/csr/marchid.yaml::$: 'long_name' is a required property
  arch_overlay/example/csr/marchid.yaml::$: 'length' is a required property
  arch_overlay/example/csr/marchid.yaml::$: 'description' is a required property
  arch_overlay/example/csr/marchid.yaml::$: 'priv_mode' is a required property
  arch_overlay/example/csr/marchid.yaml::$: 'definedBy' is a required property
  arch_overlay/example/csr/marchid.yaml::$.fields.Architecture: {'description': 'Custom overlay description'} is not valid under any of the given schemas
  Underlying errors caused this.

  Best Match:
    $.fields.Architecture: 'location' is a required property

  2 other errors were produced. Use '--verbose' to see all errors.
  arch_overlay/example/csr/marchid.yaml::$.fields.Architecture: {'description': 'Custom overlay description'} is not valid under any of the given schemas
  Underlying errors caused this.

  Best Match:
    $.fields.Architecture: 'type' is a required property

  1 other errors were produced. Use '--verbose' to see all errors.
  arch_overlay/example/csr/marchid.yaml::$.fields.Architecture: {'description': 'Custom overlay description'} is not valid under any of the given schemas
  Underlying errors caused this.

  Best Match:
[...]

PR #644, opened 2025-04-21
check-jsonschema pre-commit setup for arch_overlay dir

kbroch-rivosinc:
relates to #642

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,15 +34,15 @@ repos:
     hooks:
       - id: check-jsonschema
         alias: check-jsonschema-inst
-        files: ^arch/inst/.*\.(yaml|yml)$
+        files: ^(arch|arch_overlay/(example|qc_ui))/inst/.*\.(yaml|yml)$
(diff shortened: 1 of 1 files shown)


AFOliveira (commented):
Is there a way we can run this in the whole `arch_overlay ` - instead of just those two custom parts- so that if anyone adds a custom overlay, they also get to have this pre-commit check?

dhower-qc:
> Is there a way we can run this in the whole arch_overlay  - instead of just those two custom parts- so that if anyone adds a custom overlay, they also get to have this pre-commit check?

I was wondering the same

kbroch-rivosinc:
> > Is there a way we can run this in the whole arch_overlay  - instead of just those two custom parts- so that if anyone adds a custom overlay, they also get to have this pre-commit check?
> 
> I was wondering the same

As discussed here: https://github.com/riscv-software-src/riscv-unified-db/issues/642#issuecomment-2831343349
Give the arch_overlay can contain just yaml partials of existing defs, these would fail json schema validation when the json schema `required` fields that weren't being present were checked.

I close this PR.

Issue #646, opened 2025-04-22
Antora HTML Template has useless buttons


AFOliveira:
I realized there are three dead buttons on our html template. Is this something that antora forces us to use, or is it modifiable? I think we do not want those lying there forever.

![Image](https://github.com/user-attachments/assets/d99d4384-9ee3-40d5-b8b7-1655d2fb925f)

AFOliveira:
cc @dhower-qc @ddrcode 

Maybe this is something @ddrcode knows better than all of us :)

kbroch-rivosinc:
Here's how this is happening currently: 

This https://github.com/riscv-software-src/riscv-unified-db/blob/main/backends/manual/templates/playbook.yml.erb#L23

```
ui:
  bundle:
    url: https://gitlab.com/antora/antora-ui-default/-/jobs/artifacts/HEAD/raw/build/ui-bundle.zip?job=bundle-stable
```

pulls in this: 

```html
❯ cat partials/header-content.hbs
<header class="header">
  <nav class="navbar">
    <div class="navbar-brand">
      <a class="navbar-item" href="{{{or site.url siteRootPath}}}">{{site.title}}</a>
      {{#if env.SITE_SEARCH_PROVIDER}}
      <div class="navbar-item search hide-for-print">
        <div id="search-field" class="field">
          <input id="search-input" type="text" placeholder="Search the docs"{{#if page.home}} autofocus{{/if}}>
        </div>
      </div>
      {{/if}}
      <button class="navbar-burger" aria-controls="topbar-nav" aria-expanded="false" aria-label="Toggle main menu">
        <span></span>
        <span></span>
        <span></span>
      </button>
    </div>
    <div id="topbar-nav" class="navbar-menu">
      <div class="navbar-end">
        <a class="navbar-item" href="#">Home</a>
        <div class="navbar-item has-dropdown is-hoverable">
          <a class="navbar-link" href="#">Products</a>
          <div class="navbar-dropdown">
            <a class="navbar-item" href="#">Product A</a>
            <a class="navbar-item" href="#">Product B</a>
            <a class="navbar-item" href="#">Product C</a>
          </div>
        </div>
        <div class="navbar-item has-dropdown is-hoverable">
          <a class="navbar-link" href="#">Services</a>
          <div class="navbar-dropdown">
            <a class="navbar-item" href="#">Service A</a>
            <a class="navbar-item" href="#">Service B</a>
[...]

PR #651, opened 2025-04-23
data: remove bad top nav links from antora site

kbroch-rivosinc:
relates to #646
Default antora bundle zip pulling in template links remove those by overriding in playbook

looks like this now: 

image

In the future we could probably put useful links in the top nav, but for now this just removes the useless ones.

### DIFF
--- a/backends/manual/templates/playbook.yml.erb
+++ b/backends/manual/templates/playbook.yml.erb
@@ -107,3 +107,19 @@ ui:
       .edit-this-page {
          display: none !important;
       }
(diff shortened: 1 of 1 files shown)


AFOliveira (approved):

ThinkOpenly (approved):
Certainly not my area of expertise, but I like the results. :-)

dhower-qc (approved):
Good enough for now. @wmat, do you have plans for anything more comprehensive?

PR #656, opened 2025-04-23
fix: broken refs in generated html

### LINKED ISSUE #641: ISA manual and example_cfg artifacts aren't creating links correctly
I just happened to notice this in the ISA manual and example_cfg artifacts. This %%UDB_DOC_LINK% ... format is supposed to be converted to Antora links by calling AntoraUtils::resolve_links. Doesn't seem like it is happening.

Image

james-ball-qualcomm:
fixes #641


lib/idl/passes/gen_adoc.rb
```diff
@@ -314,6 +313,7 @@ def gen_adoc(indent = 0, indent_spaces: 2)
         if @cfg_arch.csr(csr_text).nil?
           "#{' '*indent}#{csr_text}"
         else
+          # Needs fix. See issue https://github.com/riscv-software-src/riscv-unified-db/issues/641
           "#{' '*indent}%%LINK%csr;#{idx};#{csr_text}%%"
```
dhower-qc, line 317: Why is this one being treated differently?

james-ball-qualcomm: There is no anchor for a CSR that uses its index number (just its name).

ThinkOpenly:
I'm trying to make sense of the PR title, because that's going to be the commit message.
What does "633 deploy pages fails build html gen64" mean?

lib/idl/passes/gen_adoc.rb
dhower-qc: I see. Looking at this, there seems to be some bit rot. You'll never wind up on line 317 because `csr_text` is going to be the string "CSR[#{idx}]".

It should be:

```ruby
if @cfg_arch.csr(idx).nil?
  "#{' '*indent}#{csr_text}"
else
  "#{' '*indent}%%LINK%csr;#{idx};#{csr_text}%%"
end
```

And in this case, you do have the name (it's `idx`), so you should be able to use your function.

james-ball-qualcomm: Committed recommended code in this PR. Will associate issue #641 with this PR.

dhower-qc:
PR title should follow Conventional Commits, like:

```
fix: broken refs in generated html
```

And the why, along with issue links, should be in the PR description

```
fixes #641
```

lib/idl/passes/gen_adoc.rb
```diff
@@ -311,7 +310,7 @@ def gen_adoc(indent = 0, indent_spaces: 2)
         # we don't have the symtab to map this to a csr name
         "#{' '*indent}#{csr_text}"
       else
-        if @cfg_arch.csr(csr_text).nil?
+        if @cfg_arch.csr(idx).nil?
           "#{' '*indent}#{csr_text}"
         else
           "#{' '*indent}%%LINK%csr;#{idx};#{csr_text}%%"
```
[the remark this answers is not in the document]
james-ball-qualcomm: Yes, just merged that in.

dhower-qc (approved):

Issue #662, opened 2025-04-25
add dependabot to check for submodule updates and provide PR


kbroch-rivosinc:
similar to what was done here: https://github.com/riscv-admin/docs-sig/issues/63

AFOliveira:
This should be solved and closed, right?

PR #665, opened 2025-04-25
chore: add dependabot cfg to create PR when git submodules could be u…

kbroch-rivosinc:
…pdated

relates to #662

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+---
+# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem
+version: 2
+updates:
+  - package-ecosystem: gitsubmodule
+    directory: /
+    schedule:
+      interval: daily


dhower-qc (approved):
cool

AFOliveira:
I know I arrived late to this PR, but didn't we always want to have the riscv-isa-manual at the 20240411 version since somethings were being fetch from there?

kbroch-rivosinc:
> I know I arrived late to this PR, but didn't we always want to have the riscv-isa-manual at the 20240411 version since somethings were being fetch from there?

This will only suggest PRs, so if that's the case for isa-manual then this one https://github.com/riscv-software-src/riscv-unified-db/pull/671 could simply be closed.  If it keeps suggesting the same PRs that we don't want than I can remove it from dependabot.

AFOliveira:
> > I know I arrived late to this PR, but didn't we always want to have the riscv-isa-manual at the 20240411 version since somethings were being fetch from there?
> 
> This will only suggest PRs, so if that's the case for isa-manual then this one #671 could simply be closed. If it keeps suggesting the same PRs that we don't want than I can remove it from dependabot.

That's why I did my question, I thought maybe we could ignore the riscv-isa-manual submodule. However, since @dhower-qc has accepted, I also wanted to figure if we are still pulling information from there, or if my comments are outdated.

ThinkOpenly:
There's a lot I don't understand, but is the submodule parked at a tag, or are we updating the submodule because the main branch has changes?

Do we need to open an issue to discuss?

dhower-qc:
> I know I arrived late to this PR, but didn't we always want to have the riscv-isa-manual at the 20240411 version since somethings were being fetch from there?

1. The 20240411 release is stored as a commit in the history of riscv-isa-manual. We can update commits from riscv-isa-manual without changing what we are pointing to.
2. The riscv-isa-manual is currently maintained as a git _subtree_, not a git _submodule_, so this shouldn't affect it anyway.

That said, it'd be nice if we could also automate keeping the subtree up to date.

Issue #655, opened 2025-04-23
Add sub-types to existing instruction formats


AFOliveira:
### Summary  
Track progress on introducing **sub-types** for our R-, I-, B-, S-, U-type instruction formats to improve clarity, tooling and testing.

### Motivation  
- **Mixed encodings** in each format (e.g. immediates vs. hard-coded bits)  
- **Ambiguous docs** some instructions don't match any type
- **Coarse tests** that can’t target specific bit-patterns

### Proposal  
1. Add a `subtype` field to our instruction schema.  
2. Example sub-types:  
   - `R.aqrl` — acquire release Rtype
3. Annotate existing instructions.  
4. Create test suite to validate schema in UDB 

### Next Steps  
- Create full list in a readable format to ask for reviews

AFOliveira:
@Unique-Usman can we discuss this topic here from now on? Please comment this so I can assign it to you

Unique-Usman:
Hi @AFOliveira, Got stuck in college activities. I am back now. I suggest we use this Github thread instead of excel. People Can also see it and easily drop comment here ? What do you think ?

AFOliveira:
> Hi [@AFOliveira](https://github.com/AFOliveira), Got stuck in college activities. I am back now. I suggest we use this Github thread instead of excel. People Can also see it and easily drop comment here ? What do you think ?

That's fine, but how? 1000+ lines is not readable as you posted, so maybe you will need to format it in github, which may be a lot of work.

Unique-Usman:
I think we should use Divide and Conquer approach, I think we should go extension by extension as some of the instruction subtypes are similar across extension. what do you think?

AFOliveira:
Problem is, that won't be easy even for people who may come just to review. Also, the same sub-type may exist in several extensions, but I think it's your call, just try to make it readable, and I'll try to review and ask people to review it

Unique-Usman:
Yeah, Subtypes exist in different extensions actually. Maybe we can have it in excel. The question is, how do we make it synchronized. For excel, if we have everything inside the excel file, people can come randomly and review. It will be add to know what has been reviewed and what has not been ? If we have in a thread, we can go extension by extension. Have discussion on it, then go to the next extension. If someone comes they can view what has been discussed and proposed. And they can dropped their own idea. Will this be easy to do in excel ?

Unique-Usman:
This is the sample of what I am for example starting with a Extension

# A - Extension

[...]

PR #686, opened 2025-04-29
feat(schema): add new instruction schema for types/subtypes

### LINKED ISSUE #676: Add type to instruction schema
Add type to instructions json schema. I'll make it non-mandatory for now, but to allow the UDB data shift to start happening. This is the type of change that will break downstream tools, keep ignoring it for now, since we are just trying to get established and this type of change needs to happen?

AFOliveira:
### Add instruction “type” schema & link sub‐types (closes #676)

- **inst_schema.json**:  
  - Introduce `type` object under `encoding` with a `$ref` to external `inst_type/...#.yaml` files  
- **inst_type_schema.json**:  
  - Import of encoding schema for defining RV32/RV64 encodings for instruction types  
- Lays groundwork for sub‐types in R/I/B/S/U formats as discussed in #655


schemas/inst_type_schema.json
```diff
@@ -0,0 +1,38 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Instruction Type Schema",
+  "description": "Schema for instruction type definitions",
+
+  "$defs": {
+    "encoding": {
+      "$ref": "inst_schema.json#/$defs/encoding"
```
dhower-qc, line 8: let's move the encoding schema to schema_defs.json...that's where we are keeping things shared by multiple schemas.

dhower-qc (changes requested):
Do we also need sub_type?

AFOliveira:
> Do we also need sub_type?

I though you commented against sub_type in https://github.com/riscv-software-src/riscv-unified-db/issues/655#issuecomment-2831379168. However, now I see @ThinkOpenly proposed `format `and `sub-type`. However, with the URI path of `/type/sub-type`, isn't it redundant to have both?

schemas/inst_schema.json
```diff
@@ -190,6 +203,25 @@
           "$ref": "schema_defs.json#/$defs/requires_entry",
           "description": "Extension(s) that defines the instruction"
         },
+        "type": {
```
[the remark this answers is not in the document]
AFOliveira: Then type wouldn't be any reference, but just a string? (i.e. "I")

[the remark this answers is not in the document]
ThinkOpenly: I think it would be a reference to generic definitions of the top-level instruction formats.

schemas/inst_type_schema.json
AFOliveira: Done

schemas/inst_schema.json
```diff
@@ -16,6 +16,19 @@
         }
       ]
     },
+    "type": {
+      "type": "object",
+      "properties": {
+        "$ref": {
+          "type": "string",
+          "format": "uri-reference",
+          "pattern": "^inst_type/[A-Z]/[a-zA-Z0-9_]+\\.yaml#.*$",
```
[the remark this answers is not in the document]
AFOliveira: This is the last thing to do, do we really want it to start with inst_type? where will this files stay within UDB?

schemas/inst_schema.json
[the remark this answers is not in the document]
dhower-qc: Do we need to explicitly mention "type", given that subtype presumably points to its parent?

ThinkOpenly (dismissed):
I suggest we move this along. Seems like an awesome start.

ThinkOpenly:
> I suggest we move this along. Seems like an awesome start.

After "field" moves from "inst_schema" to "schema_defs", of course.

schemas/inst_schema.json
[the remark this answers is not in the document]
ThinkOpenly: I may have resolved this too early. It would be convenient to have the top-level type here. One-stop shopping for anyone looking at the instruction, rather than having to look at every subtype to find the type _and_ make sure they all agree. We could add validation that everything matches, too.

There currently isn't a pointer from subtype to parent. That's probably a good add, and offers the ability to do additional validation.

AFOliveira:
I know I am delaying @Unique-Usman work, I'm just super busy, I'll try to either fix this today or monday

schemas/inst_schema.json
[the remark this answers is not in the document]
AFOliveira: It seems to me that having a parent in the sub-type if we did not describe the main type in the instructions, otherwise how can it not be redundant information?

dhower-qc:
@AFOliveira, if you OK with I can finish this one up.

schemas/inst_schema.json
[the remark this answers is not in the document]
ThinkOpenly: > It seems to me that having a parent in the sub-type if we did not describe the main type in the instructions, otherwise how can it not be redundant information?

I'm not following. I propose having some indication of the top-level type in the instruction and in each subtype. (They do not necessarily need to be the same thing, although they could be.) In this way, both "full information" at hand to facilitate validation, rather than needing an additional level of indirection in order to determine the top-level format.

Is that redundant? Yes, but it allows each to be standalone (for some definition of that term).

If you remove the top-level type from the instruction, you'd need to look at every subtype (usually one, but sometimes 2 [RV32, RV64]) to find the top-level type, and then make sure they all agree.

If you remove the top-level type from the subtype, you'd need to go find all of the instructions that use the subtype to find the top-level type, and then make sure they all agree.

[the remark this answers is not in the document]
AFOliveira: But isn't what you are proposing precisely the opposite of what was decided on the `definedBy` field? In which we removed the B extension that implied `Zbx` extensions? I notice that the relationship between extensions and types is not the same, but they close enough to me in this terms. I don't have any strong opinion on this, it just seems counter-intuitive and against the previous line of thought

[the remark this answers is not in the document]
ThinkOpenly: I see `definedBy` differently. That attribute a fairly static description of precisely which extension(s) actually _define_ the instruction, rather than just include it -- which would be somewhat dynamic. It would be an ongoing maintenance effort (nightmare?) as new umbrella extensions are added to go through the entire extensions hierarchy and add that new extension to the `definedBy` for newly included instructions. I was trying to avoid that error-prone update requirement.

For types/subtypes, that is absolutely static. So, the convenience/cost ratio of any redundancy is pretty high. I described the pros/cons above.

Is your preference to only define the top-level type in subtypes, and not in the instruction itself?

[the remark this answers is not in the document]
AFOliveira: I agree with that distinction between `definedBy` and `type` — the comparison makes total sense. I also don’t think the choice is hard to reverse, so let’s go with your approach for now, and if it makes more sense to switch later, that’s totally fine.

dhower-qc:
I still need to update the Ruby classes, but I pushed the schema updates for review.

arch/inst/Zba/add.uw.yaml
```diff
@@ -10,15 +10,10 @@ description: |
...
-    - name: rs2
-      location: 24-20
-    - name: rs1
-      location: 19-15
-    - name: rd
-      location: 11-7
+format:
+  type: { "$ref": "inst_type/R.yaml#" }
+  subtype: { "$ref": "inst_subtype/R/R-x.yaml#" }
+  encoding: 0000100----------000-----0111011
```
ThinkOpenly, line 16: Do we want to keep this? This might be a longer discussion.
Is there a way to parameterize the type and/or subtype such that the "opcodes" that are known to the instruction can be associated with the opcode fields in the type/subtype, rather than by extraction from this "encoding" bit string? Put differently, can we put here "this is an R-type, R-x-subtype instruction where funct7=0b0000100, funct3=0b000, opcode=0b0111011"?

And eventually, perhaps: "funct7=ADD.UW, funct3=ADD.UW, opcode=OP-32", or something like that, to match the spec?

schemas/inst_var_type_schema.json
```diff
@@ -0,0 +1,44 @@
...
+        }
+      },
+      "then": {
+        "properties": {
+          "register_file": {
+            "enum": ["X", "F", "V"]
+          },
+          "direction": {
+            "enum": ["src", "dst", "srcdst"]
+          }
```
ThinkOpenly, line 39: subjective, but I propose:
```
"access": {
  "enum": ["R", "W", "RW"]
}
```

arch/inst/Zba/add.uw.yaml
dhower-qc: That's a good point. Now that we are enumerating all the opcodes, we can assign them individually. One question is whether we want to go the `$inherits` route or keep it as a `$ref` for the subtype. `$inherits` is probably a better human description, and lets us use JSON Schema validation on the resolved file. The downside is that we are only using `$inherits` right now for profiles, so this would make that custom-to-UDB operator pervasive, which some might not like.

arch/inst_type/R.yaml
```diff
@@ -0,0 +1,6 @@
+# yaml-language-server: $schema=../../schemas/inst_type_schema.json#
+
+$schema: inst_type_schema.json#
+kind: instruction_type
+name: R
+description: R-type instructions usually have two source registers and one destination registers
```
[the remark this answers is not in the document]
dhower-qc: I can add that. Note, though, that generally speaking we can't expect all instructions to validate against their type (e.g., fence)

arch/inst/Zba/add.uw.yaml
dhower-qc: here's what it could look like with `$inherits`:

```yaml
format:
  $inherits: "inst_subtype/R/R-x.yaml#"
  opcodes:
    funct7:
      display_name: ADD.UW
      value: 0b0000100
    funct3:
      display_name: ADD.UW
      value: 0b000
    opcode:
      display_name: OP-32
      value: 0b0111011
```

ThinkOpenly: Significantly better.

However, for at least the "opcode" values, these are shared among potentially many instructions. Should we / can we create a database for them? I guess that would be realized with another `$ref`?
```
opcode: { "$ref": "opcodes/OP-32.yaml#" }
```

arch/inst_type/R.yaml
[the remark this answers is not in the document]
dhower-qc: ah, that we can do

arch/inst_opcode/funct3.yaml
```diff
@@ -0,0 +1,8 @@
+# yaml-language-server: $schema=../../schemas/inst_opcode_schema.json#
+
+$schema: inst_opcode_schema.json#
+kind: instruction_opcode
+name: funct3
+
+data:
+  location: 14-12
```
ThinkOpenly, line 8: Unfortuntely, "funct3" is not always at this location. It depends on the format. For example, CI, CSS, CIW, CL, CS all have a "funct3" at 15-13.

dhower-qc: Would you prefer we just drop the indirection or call it something like R-funct3?

ThinkOpenly: What do you think about something like this?

- `arch/inst_type/R.yaml` (unchanged)
- `arch/inst_subtype/R-x.yaml`:
  ```yaml
  data:
    type: { "$ref": "inst_type/R.yaml#" }
    subtype: { "$ref": "inst_subtype/R/R-x.yaml#" }
    $inherits:
      - inst_type/R.yaml#/opcodes
      - inst_var/xs2.yaml#/data
      - inst_var/xs1.yaml#/data
      - inst_var/xd.yaml#/data
  ```
- `arch/inst/Zba/add.uw.yaml`:
  ```yaml
  format:
    $inherits: inst_subtype/R/R-x.yaml#
    opcodes:
      funct7:
        display_name: ADD.UW
        value: 0b0000100
      funct3:
        display_name: ADD.UW
        value: 0b000
      opcode: { $inherits: inst_opcode/OP-32.yaml#/data }
  ```
- arch/inst_var/xs2.yaml:
  ```yaml
  data:
    $inherits: inst_type/R.yaml#/variables/rs2
    type: { "$ref": "inst_var_type/x_src_reg.yaml#" }
  ```
etc.

So, R-x subtype inherits opcodes from the base type, R.
And `xs2` inherits attributes (just "location") from the base type R's `rs2`.
(I removed the top-level "inherits: OP-32" from "add.uw.yaml". Was that needed?)

dhower-qc: this is implemented mostly as you suggest now. a few minor changes to make it all work.

ThinkOpenly: I don't think this file should exist after the recent changes, right?

dhower-qc: gone now

dhower-qc (approved):

PR #687, opened 2025-04-29
fix: skip fence.tso on llvm-test

AFOliveira:
since compiler and ISA view of the instruction differ

Fixes #460

### DIFF
--- a/ext/auto-inst/test_parsing.py
+++ b/ext/auto-inst/test_parsing.py
@@ -122,6 +122,7 @@ class TestInstructionEncoding:
             or instr_name == "c.nop"
             or instr_name == "fcvtmod.w.d"
             or instr_name == "fence"
+            or instr_name == "fence.tso"
         ):
             pytest.skip(
                 f"Instruction {instr_name} is a corner case and implementation should not follow ISA spec"


dhower-qc (approved):

AFOliveira:
@ThinkOpenly ok, I'll add it

ThinkOpenly:
You're gonna love me...

None of this is in the contributing guidelines, but I wonder if it should be added:
- The commit subject line is so long, it wraps:
  image
  I suggest a message stating only the WHAT :wink::
  > fix: skip fence.tso on llvm-test

  The WHY is in the commit message body.

- The line length in the commit body is 344 characters (!). I suggest wrapping like code at something like 80.

If I'm crazy and the world has moved on from caring about the width of text lines, let me know.

dhower-qc:
> You're gonna love me...
> 
> None of this is in the contributing guidelines, but I wonder if it should be added:
> 
> * The commit subject line is so long, it wraps:
>   image
>   I suggest a message stating only the WHAT 😉:
>   > fix: skip fence.tso on llvm-test
>   
>   
>   The WHY is in the commit message body.
> * The line length in the commit body is 344 characters (!). I suggest wrapping like code at something like 80.
> 
> If I'm crazy and the world has moved on from caring about the width of text lines, let me know.

You're right, it should be added. The [commit template](https://github.com/riscv-software-src/riscv-unified-db/blob/main/.gitmessage) mentions 80 lines, but not the contributing guide.

PR #695, opened 2025-04-29
feat: bump docs-resources

dhower-qc:


ThinkOpenly (approved):

dhower-qc:
I see now that dependabot does this...

PR #708, opened 2025-04-30
chore: get instruction appendix ready for release

dhower-qc:
This does several things:

* fix(build): make instruction appendix dependent on docs-resources
* fix: remove stray '<' in instruction appendix pages
* refactor: use cfg_arch.possible_instructions to get appendix list
    * Also removes InstructionIndex class (redundant now)
* ci: only generate asciidoc for instruction appendix regress

I'll also get this added to the deployments, and add a regression to check a "golden" output.


AFOliveira (commented):
LGTM so far. Will you also add the assembly field to the erb template or do you prefer me to do so in a separate PR?

AFOliveira:
@dhower-qc is it fine with you if I add a commit with a Assembly field being reliant on an environment variable?

dhower-qc:
> @dhower-qc is it fine with you if I add a commit with a Assembly field being reliant on an environment variable?

What would that look like?

AFOliveira:
My idea was to add a couple lines to the erb template that prints the Assembly field if "ASSEMBLY=1".

dhower-qc:
ahhh, I see. I thought you meant a variable that changes the string.

I have no problem with ASSEMBLY=1

lib/deploy.sh
```diff
@@ -73,61 +73,65 @@ deploy_mkdir $DEPLOY_DIR/manual
...
-deploy_do "gen:isa_explorer_browser_inst"
 deploy_log "Create isa_explorer_browser_csr"
-deploy_do "gen:isa_explorer_browser_csr"
+
+parallel :::                                          \
+  "./do gen:isa_explorer_browser_csr"                 \
+  "./do gen:isa_explorer_browser_ext"                 \
+  "./do gen:isa_explorer_browser_inst"                \
+  "./do gen:isa_explorer_spreadsheet"                 \
+  "./do gen:html_manual MANUAL_NAME=isa VERSIONS=all" \
```
ThinkOpenly, line 91: I presume this doesn't reintroduce the problem fixed by commit 34f9a10?

AFOliveira:
> ahhh, I see. I thought you meant a variable that changes the string.
> 
> I have no problem with ASSEMBLY=1

I'll either PR against your branch or main then. Depending on how fast this gets merged

lib/deploy.sh
```diff
@@ -73,61 +73,65 @@ deploy_mkdir $DEPLOY_DIR/manual
...
 deploy_log "Create _site/isa_explorer"
 deploy_mkdir $DEPLOY_DIR/isa_explorer
 deploy_log "Create isa_explorer_browser_ext"
-deploy_do "gen:isa_explorer_browser_ext"
 deploy_log "Create isa_explorer_browser_inst"
-deploy_do "gen:isa_explorer_browser_inst"
 deploy_log "Create isa_explorer_browser_csr"
-deploy_do "gen:isa_explorer_browser_csr"
+
+parallel :::                                          \
```
ThinkOpenly, line 86: I'm far from a Ruby expert, but the syntax here is very different from the examples I've found, that tend to look like:
```
data = [1, 2, 3, 4, 5]
results = Parallel.map(data) do |number|
  number * 2
end
```

dhower-qc: That's because it's not Ruby 😄 

This is a shell script, and that is calling GNU Parallel

ThinkOpenly: > That's because it's not Ruby 😄

I stand corrected. I've never seen that `:::` separator before.

> This is a shell script, and that is calling GNU Parallel

This must be different than the `parallel` I have on my Ubuntu 24.04 system (in package "moreutils"). That's not confusing at all, eh? Looks like there's a "parallel" package for Ubuntu as well. And they conflict, both installing /usr/bin/parallel. But, there's a workaround implemented https://superuser.com/questions/917577/how-can-i-install-gnu-parallel-alongside-moreutils-on-ubuntu-debian/1253492#1253492. I've now learned way too much about this command. ;-)

lib/deploy.sh
ThinkOpenly: Checks out (with a mock-up `do` that just echos it's arguments:
```
$ parallel ::: \
"./do gen:isa_explorer_spreadsheet"                 \
"./do gen:html_manual MANUAL_NAME=isa VERSIONS=all" \
"./do gen:proc_ctp_pdf[MockProcessor]"
do ...
gen:isa_explorer_spreadsheet
do ...
gen:html_manual
MANUAL_NAME=isa
VERSIONS=all
do ...
gen:proc_ctp_pdf[MockProcessor]
```

ThinkOpenly (approved):

dhower-qc:
Cool. this already found a bug in the code it merged from main.

james-ball-qualcomm:
How are the stdout/stderr from each "parallel" do command captured? Do they all get mixed into the same console? If so, how do you debug the output?

PR #725, opened 2025-05-05
Fix mstatus fs/vs field descriptions

jordancarlin:
The existing description said 
> This parameter only has an effect when both S and F mode are disabled.

This is incorrect. The field has an effect when F is disabled and S is enabled. We could update the description, but the preceding line makes this sentence redundant anyway, so this just removes the line instead.

The same issue applies to the "vs" field.

### DIFF
--- a/arch/ext/S.yaml
+++ b/arch/ext/S.yaml
@@ -284,19 +284,19 @@ params:
   MSTATUS_FS_WRITABLE:
     description: |
       When `S` is enabled but `F` is not, mstatus.FS is optionally writable.
-
-      This parameter only has an effect when both S and F mode are disabled.
     schema:
       type: boolean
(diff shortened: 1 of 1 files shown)


ThinkOpenly (approved):

dhower-qc (changes requested):
While we are at it, the extra_validation should also check that mstatus.FS is read-only if neither F nor S is implemented. Something like:

```yaml
  extra_validation: |
    assert MSTATUS_FS_WRITEABLE == true  if ext?(:F)
    assert MSTATUS_FS_WRITEABLE == false if (!ext?(:S) && !ext?(:F))
```

Same for VS.

jordancarlin:
Updated to add the new `extra_validation`.

dhower-qc (approved):

PR #736, opened 2025-05-06
feat(idl): add widening operators in IDL

dhower-qc:
Adds four widening `Bits<N>` operators in IDL:

* `` `<< `` : widening shift left. If the shift amount is a constant, the result is `Bits<N + shamt>`. Otherwise, the result type is `Bits<Infinite>`.
* `` `* ``: widening multiply. The result type when multiplying `Bits<N>` * `Bits<M>` is `Bits<N + M>`
* `` `+ ``: widening add. The result type when adding `Bits<N>` + `Bits<M>` is `Bits<max(M, N) + 1>`
* `` `- ``: widening sub. The result type when subtracting `Bits<N> + Bits<M>` is `Bits<max(M, N) + 1>`.


ayosher:
I would expect any multiply to be widening multiply

ayosher (approved):

backends/cpp_hart_gen/lib/gen_cpp.rb
```diff
@@ -599,16 +599,20 @@ class BinaryExpressionAst
...
+      elsif op == "`<<"
         if rhs.constexpr?(symtab)
           # use template form of shift
           "#{' '*indent}(#{lhs.gen_cpp(symtab, 0, indent_spaces:)}.template sll<#{rhs.value(symtab)}>())"
-        elsif rhs.type(symtab).const?
+        else
           # use widening shift
           "#{' '*indent}(#{lhs.gen_cpp(symtab, 0, indent_spaces:)}.widening_sll(#{rhs.gen_cpp(symtab, 0, indent_spaces:)}))"
-        else
-        "#{' '*indent}(#{lhs.gen_cpp(symtab, 0, indent_spaces:)} << #{rhs.gen_cpp(symtab, 0, indent_spaces:)})"
```
ThinkOpenly, line 610: Trying to understand the changes here...

Previously, there was only "<<", and if the RHS was constant expression, it used "template form of shift" (to `Bits<M+shamt>`?), else if a constant type, widening (to `Bits<infinite>`?), else non-widening shift "<<"?

Now "<<" is gone, replaced by "`<<"?
And for "`<<", if RHS is constant expression, use "template form" (to `Bits<M+shamt>`?) else use widening (to `Bits<infinite>`?)

ayosher: I guess nothing is gone, just for IDL code writers is another option to code using new available operator.

dhower-qc: P.S. I've updated the IDL documentation to reflect this now.

Issue #731, opened 2025-05-06
Still have 404 errors on artifacts web page built by lib/deploy.sh (Deploy Pages GitHub action)


james-ball-qualcomm:
The nightly regression passed for the 1st time in 3 weeks last night. I was hoping all the 404 artifact link errors on https://riscv-software-src.github.io/riscv-unified-db/index.html would have been fixed. However, the following links still are 404:

- All profiles
- ISA Explorer Excel (https://riscv-software-src.github.io/riscv-unified-db/isa_explorer.xlsx)
- AC100 and AC200 CRDs
- IDL Language doc (https://riscv-software-src.github.io/riscv-unified-db/ruby/idl/index.html)
- Ruby Interface doc (https://riscv-software-src.github.io/riscv-unified-db/ruby/arch_def/index.html)

AFOliveira:
I have not taken a look at this, but may it be because our deployments are broken and the web view you linked does not represent the most up-to-date version of UDB?

james-ball-qualcomm:
Here's a link to last night's regression:
   https://github.com/riscv-software-src/riscv-unified-db/actions/runs/14850973611

I'll start looking through the logs since I've improved the display of information and checking of exit status in PRs now in main. Maybe some of you can do the same in case you spot something useful.

The only failure I see is the "Check Latest Commit" annotation when you can find from the link above. Who knows about this? It says:

![Image](https://github.com/user-attachments/assets/ec01bd05-c911-42fa-85f1-fc542b01f0cd)

james-ball-qualcomm:
Hold on, just found this clue at the bottom of the log output:
 This step has been truncated due to its large size. Download the full logs from the menu once the workflow run has completed.

james-ball-qualcomm:
[logs_38132817721.zip](https://github.com/user-attachments/files/20062377/logs_38132817721.zip)

Attached logs.

james-ball-qualcomm:
Just realized I'm looking at the test:nightly results when I should be looking at the Deploy Pages action since that is what runs the lib/deploy.sh script and creates the artifacts on https://riscv-software-src.github.io/riscv-unified-db/index.html

james-ball-qualcomm:
The Deploy Pages doesn't seem happy even though it built find in my branch 597-isa-explorer-backend-missing-extension-dependencies yesterday:

![Image](https://github.com/user-attachments/assets/4461fbd9-7278-4ce1-ba1f-3ea7a5cfe535)

[...]

PR #668, opened 2025-04-25
Utilize Ruby Logger class to created timestamped output to find performance issues.

james-ball-qualcomm:

PR #737, opened 2025-05-06
fix: too much quoting in `lib/deploy.sh`

ThinkOpenly:
Commit 4ddb999 added some nice logging, but also added a lot of quotes. A few too many, in fact, as this:
```
```
became:
```
deploy_do "gen:html_manual MANUAL_NAME=isa VERSIONS=all"
```

The quotes there turn "command argument argument" into a single command string where the spaces are part of the command itself.

Remove the quotes so that the arguments are again treated as arguments.

### DIFF
--- a/lib/deploy.sh
+++ b/lib/deploy.sh
@@ -93,7 +93,7 @@ deploy_log "Copy isa_explorer_spreadsheet"
 deploy_cp_recursive gen/isa_explorer/spreadsheet $DEPLOY_DIR/isa_explorer
 
 deploy_log "Build manual"
-deploy_do "gen:html_manual MANUAL_NAME=isa VERSIONS=all"
+deploy_do gen:html_manual MANUAL_NAME=isa VERSIONS=all
 deploy_log "Copy manual html"
 deploy_cp_recursive gen/manual/isa/top/all/html $DEPLOY_DIR/manual
 deploy_log "Build html documentation for example_rv64_with_overlay"


ThinkOpenly:
Fixes #731 (hopefully)

ThinkOpenly:
Fix to PR #668

dhower-qc (approved):

AFOliveira:
and yet again we have a green deployment :) great!

james-ball-qualcomm:
Paul, thanks so much for finding and fixing the bug I inadvertently introduced.

james-ball-qualcomm:
Instead of changing:
    deploy_do "gen:html_manual MANUAL_NAME=isa VERSIONS=all"
to
    deploy_do gen:html_manual MANUAL_NAME=isa VERSIONS=all

How about we change it to:
    deploy_do "gen:html_manual" "MANUAL_NAME=isa" "VERSIONS=all"

Whatever we decide, seems like we should make all the calls to deploy_do consistent. For example, if we decide some new or existing target needs environment variables set like gen:html_manual does, if we don't have all the calls to deploy_do consistent then we risk having the same hard-to-find bug show up again down the road.

So, I vote for putting quotes around each space-separated argument to deploy_do. That should be the safest solution.

james-ball-qualcomm:
I just noticed I have the same failure when running lib/deploy.sh locally. I'm going to make improvements to this script to make it clearer what happened so it will be hard to miss failures like this again. That work is happening in issue #731.

james-ball-qualcomm:
Actually, I already have double quotes around each argument to "do_deploy". So, no need for a discussion about this.

PR #740, opened 2025-05-07
feat(llvm-test): upload riscv.json to github

AFOliveira:
riscv.json is llvm-tablegen for risc-v.
Uploading it to github actions, allows users to download it, instead of building it locally.

### DIFF
--- a/.github/workflows/regress.yml
+++ b/.github/workflows/regress.yml
@@ -134,7 +134,6 @@ jobs:
           rm -rf ext/llvm-project
           git submodule sync --recursive
           git submodule update --init --recursive --depth=1 ext/llvm-project
-
       - name: Check for required directories and files
         if: ${{ steps.cache-riscv.outputs.cache-hit != 'true' }}
         run: |
@@ -158,6 +157,11 @@ jobs:
             -o ext/llvm-project/riscv.json
       - name: Show riscv.json output
         run: ls -l ext/llvm-project/riscv.json
+      - name: Upload RISC-V JSON as Artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: riscv-json
+          path: ext/llvm-project/riscv.json
   regress-gen-go:
     runs-on: ubuntu-latest
     env:


ThinkOpenly:
> riscv.json is llvm-tablegen for risc-v. Uploading it to github actions, allows users to download it, instead of building it locally.

Do we provide a link to this file somewhere?

AFOliveira:
> > riscv.json is llvm-tablegen for risc-v. Uploading it to github actions, allows users to download it, instead of building it locally.
> 
> Do we provide a link to this file somewhere?

Yes, it is embedded in the action. From what I could grasp, this is standard for the upload and did not document this part.

See
image

ThinkOpenly:
> > > riscv.json is llvm-tablegen for risc-v. Uploading it to github actions, allows users to download it, instead of building it locally.
> > 
> > Do we provide a link to this file somewhere?
> 
> Yes, it is embedded in the action.

I'm confused. Are we expecting users to download this file from the action artifacts?

AFOliveira:
> I'm confused. Are we expecting users to download this file from the action artifacts?

It's not about expecting users to download it. This artifact is generated by compiling LLVM, which is a time-consuming process and requires a reasonably capable machine. It's mainly useful for those who want to compare against LLVM locally or fix errors that appear in the CI (like in my case). Downloading it from the action artifacts is just a convenience for UDB developers to avoid the need to build LLVM locally. Therefore, it can be linked in other way, but I don't think it should have a static location, as it's not an output but rather a debugging aid for developers.

.github/workflows/regress.yml
```diff
@@ -158,6 +157,11 @@ jobs:
             -o ext/llvm-project/riscv.json
       - name: Show riscv.json output
         run: ls -l ext/llvm-project/riscv.json
+      - name: Upload RISC-V JSON as Artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: riscv-json
+          path: ext/llvm-project/riscv.json
```
dhower-qc, line 164: did you mean to upload this to a submodule path?

And do we want to upload this for every regression run, or is this better suited as a deployment action?

AFOliveira: `path` here is the file location of what you want to upload as the artifact.

This doesn't seem deployment-suited since it is a llvm-artifact and not a UDB-one. I don't see any problem with the upload, since it's really lightweight.  I can add a flag to only update this when it has a different commit SHA, so it will not upload the same content twice, but it will make it substantially harder to find the specific regression test that has this file available to download.

dhower-qc (approved):

PR #749, opened 2025-05-09
feat(CI): add c_header generation to regression test

AFOliveira:
Normal regression test for the new ouput.

Missing on previous PR and noticed by @jordancarlin.

### DIFF
--- a/.github/workflows/regress.yml
+++ b/.github/workflows/regress.yml
@@ -169,6 +169,17 @@ jobs:
         uses: ./.github/actions/singularity-setup
       - name: Generate Go code
         run: ./do gen:go
+  regress-gen-c-header:
+    runs-on: ubuntu-latest
+    env:
+      SINGULARITY: 1
+    steps:
+      - name: Clone Github Repo Action
+        uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


dhower-qc (approved):
We should also add this to deployment

AFOliveira:
> We should also add this to deployment

I have no idea on how to do that, do we want to add it do the pages? Should it go on deploy.sh?

ThinkOpenly (approved):

dhower-qc:
> > We should also add this to deployment
> 
> I have no idea on how to do that, do we want to add it do the pages? Should it go on deploy.sh?

Exactly. It needs to get built in deploy.sh, and then linked on the HTML page that deploy.sh generates (at the bottom of deploy.sh).

AFOliveira:
> > > We should also add this to deployment
> > 
> > 
> > I have no idea on how to do that, do we want to add it do the pages? Should it go on deploy.sh?
> 
> Exactly. It needs to get built in deploy.sh, and then linked on the HTML page that deploy.sh generates (at the bottom of deploy.sh).

I'll do it and commit that in a separate PR

PR #750, opened 2025-05-09
feat(dev): have vscode recommend extensions for the project

dhower-qc:


ThinkOpenly:
Are there some details about what this is?

dhower-qc:
It just gets VS Code to prompt you if you want to install the recommended extensions for the project. This is the same set of extensions you get if you run in the devcontainer using VS Code.

See https://code.visualstudio.com/docs/configure/extensions/extension-marketplace#_workspace-recommended-extensions for details

ThinkOpenly (approved):

PR #751, opened 2025-05-09
feat(schema): convert param 'when' to full extension req cond

dhower-qc:
Needed to represent cross-extension conditions, rather than assuming
a parameter would only be dependent on the version of the defining
extension.

### DIFF
--- a/arch/ext/Sm.yaml
+++ b/arch/ext/Sm.yaml
@@ -404,6 +404,7 @@ params:
       minimum: 0
       maximum: 0xFFFFFFFFFFFFFFFF
     when:
+      name: Sm
       version: ">= 1.12.0"
   PMA_GRANULARITY:
     description: |
--- a/backends/portfolio/tasks.rake
+++ b/backends/portfolio/tasks.rake
@@ -7,11 +7,7 @@ require_relative "#{$lib}/idl/passes/gen_adoc"
 
 # @return [Architecture]
 def pf_create_arch
-  # Ensure that unconfigured resolved architecture called "_" exists.
-  Rake::Task["#{$root}/.stamps/resolve-_.stamp"].invoke
-
-  # Create architecture object using the unconfigured resolved architecture called "_".
-  Architecture.new($root / "gen" / "resolved_arch" / "_")
+  cfg_arch_for("_")
 end
 
 # @param portfolio_grp_with_arch [PortfolioGroup] Contains one or more Portfolio objects that have an arch (not a cfg_arch).
--- a/lib/arch_obj_models/parameter.rb
+++ b/lib/arch_obj_models/parameter.rb
@@ -6,8 +6,7 @@ require_relative "../version"
 
 # A parameter (AKA option, AKA implementation-defined value) supported by an extension
 class Parameter
-  # @return [Architecture] The defining architecture
-  attr_reader :arch
+  extend T::Sig
 
   # @return [String] Parameter name
   attr_reader :name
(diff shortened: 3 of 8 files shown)


arch/ext/Sm.yaml
```diff
@@ -404,6 +404,7 @@ params:
       minimum: 0
       maximum: 0xFFFFFFFFFFFFFFFF
     when:
+      name: Sm
```
ThinkOpenly, line 407: Is this fixing a bug?

backends/portfolio/tasks.rake
```diff
@@ -7,11 +7,7 @@ require_relative "#{$lib}/idl/passes/gen_adoc"
 
 # @return [Architecture]
 def pf_create_arch
-  # Ensure that unconfigured resolved architecture called "_" exists.
-  Rake::Task["#{$root}/.stamps/resolve-_.stamp"].invoke
-
-  # Create architecture object using the unconfigured resolved architecture called "_".
-  Architecture.new($root / "gen" / "resolved_arch" / "_")
+  cfg_arch_for("_")
```
ThinkOpenly, line 10: Is this related?

lib/arch_obj_models/parameter.rb
```diff
@@ -46,12 +45,13 @@ def default
...
   # @param name [String]
   # @param data [Hash<String, Object]
+  sig { params(ext: Extension, name: String, data: T::Hash[String, Object]).void }
   def initialize(ext, name, data)
     raise ArgumentError, "Expecting Extension but got #{ext.class}" unless ext.is_a?(Extension)
     raise ArgumentError, "Expecting String but got #{name.class}" unless name.is_a?(String)
     raise ArgumentError, "Expecting Hash but got #{data.class}" unless data.is_a?(Hash)
 
-    @arch = ext.arch
+    @cfg_arch = ext.cfg_arch
```
ThinkOpenly, line 54: unrelated?

lib/config.rb
```diff
@@ -319,10 +319,6 @@ def initialize(portfolio_grp)
 
     @portfolio_grp = portfolio_grp
 
-    portfolio_grp.portfolios.each do |portfolio|
-      raise "Portfolio #{portfolio.name} shouldn't have a non-nil cfg_arch member" if portfolio.cfg_arch?
-      raise "Portfolio #{portfolio.name} shouldn't have a an arch member of type ConfiguredArchitecture" if portfolio.arch.is_a?(ConfiguredArchitecture)
-    end
```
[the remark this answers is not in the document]
dhower-qc: Yes, indirectly. @james-ball-qualcomm was assuming that you always start from a "bare" architecture, but that's not valid. So those assertions need to go away.

The issue was exposed by this patch because you need a configured architecture (with an extension list) to know if a parameter should exists now.

lib/arch_obj_models/parameter.rb
dhower-qc: ditto

lib/arch_obj_models/req_expression.rb
```diff
@@ -572,6 +572,30 @@ def satisfied_by?(&block)
...
+  # returns true if the list of extension requirements *can* satisfy the condition.
+  #
+  # note that it is possible that the condition might not be satisfied for all possible
+  # extension versions implied by ext_req_list. For example, this condition:
+  #
+  #  { "name": "A", "version": ">= 1.2" }
+  #
+  # Will be "satisfied by":
+  #
+  #  { "name": "A, "version": ">= 1.1" }
```
ThinkOpenly, line 584: missing a closing quote after `A`

PR #757, opened 2025-05-09
feat: add resolved_arch tarball to deploy

dhower-qc:


lib/deploy.sh
```diff
@@ -151,7 +153,19 @@ cat <<- EOF > $DEPLOY_DIR/index.html
...
     <ul>
-      <li><a href="$PAGES_URL/resolved_arch/index.yaml">index.yaml</a> Database index, as array of relative paths from $PAGES_URL/resolved_arch</li>
+      <li>
+        <a href="$PAGES_URL/resolved_arch/index.yaml">index.yaml</a>
+        Database index, as array of relative paths from $PAGES_URL/resolved_arch
+        <ul>
+          <li>
+            For example, you can find <a href="$PAGES_URL/resolved_arch/ext/Sm.yaml">Sm.yaml</a> at $PAGES_URL/resolved_arch/ext/Sm.yaml
+          </li>
+        <ul>
```
AFOliveira, line 163: Not super sure, but isn't this \ul?

lib/deploy.sh
```diff
@@ -75,6 +75,8 @@ deploy_mkdir $DEPLOY_DIR/htmls
 
 deploy_log "Resolve / Create Index for base architecture"
 deploy_do "gen:resolved_arch"
+tar czf resolved_arch.tar.gz gen/resolved_arch/_
+mv resolved_arch.tar.gz $DEPLOY_DIR/
```
ThinkOpenly, line 79: One line?
```
tar czf $DEPLOY_DIR/resolved_arch.tar.gz gen/resolved_arch/_
```

lib/deploy.sh
ThinkOpenly: close. :-)
```
</ul>
```

ThinkOpenly (changes requested):
Who's the expected consumer of this artifact? Is this a possible solution to providing "data" without "tools"?

ThinkOpenly:
> > Who's the expected consumer of this artifact? Is this a possible solution to providing "data" without "tools"?
> 
> Exactly. This would provide the _resolved_ data only for anyone that just wants to see that.

I wonder if we should make that content into an occasional "release" artifact. I think you can drop whatever binaries/files you want into an arbitrary release. I see you can also mark it "Set as a pre-release", so it would be more clearly "not validated data, use as your own risk" (hopefully).

(This is not an "instead of" proposal, but "in addition to".)

AFOliveira (approved):
LGTM!

ThinkOpenly (approved):

Issue #650, opened 2025-04-22
Validate proper indentation


ThinkOpenly:
**Is your feature request related to a problem? Please describe.**
Improper indentation should be detected before merging.

**Describe the solution you'd like**
I'd like to see tests added that validate indentation within data files and the fields therein:
- YAML (may already be done?)
  - IDL (not checked... would a SystemVerilog formatter work?)
- JSON (not checked... I'm sure there are JSON formatters out there)

kbroch-rivosinc:
in general, YAML and JSON are formatted by prettier pre-commit https://github.com/riscv-software-src/riscv-unified-db/blob/main/.pre-commit-config.yaml#L29

default config: 

```
kbroch@kbroch-mac.local:~/rvi/non-repo-repos/riscv-unified-db on  main via  v22.14.0 via 🐍 v3.13.3 via 💎 v3.2.3
❯ cat .prettierrc
{}
```

some files being ignored: 

```
kbroch@kbroch-mac.local:~/rvi/non-repo-repos/riscv-unified-db on  main via  v22.14.0 via 🐍 v3.13.3 via 💎 v3.2.3
❯ cat .prettierignore
gen
.home
.singularity
node_modules
.asciidoctor
.bundle
.git
.github
.vscode
docs/ruby
ext
```

AFOliveira:
@kbroch-rivosinc does IDL gets indented with this? Maybe that would be a big effort, but would it be possible to extend prettier to do it?  I think most of the indentation @ThinkOpenly is refering to is IDL related.

dhower-qc:
There isn't a formatter for IDL. Looking briefly at what it would take, I see that the [Ruby plugin](https://github.com/prettier/plugin-ruby/blob/main/src/server.rb) for prettier actually just wraps a call to a formatter implemented in Ruby via javascript. That seems like the easiest path here, too. If we add a formatted output function to the existing IDL syntax tree, we can just call it from javascript.

FYI, there is a `to_idl` function on the IDL AST, but it prints everything in minimal format. We'd just need to add a similar function that adds nice whitespace.

syedowaisalishah:
@dhower-qc @ThinkOpenly
Hi! I'd like to work on the IDL formatting part of this issue by adding a pretty-printer to the IDL AST in Ruby and integrating it with pre-commit hooks. Please let me know if anyone is already working on this or if you have any suggestions or advice before I begin. Thanks!

[...]

PR #772, opened 2025-05-15
Add basic IDL pretty-printer with AST indentation (Option 2 from #650)

syedowaisalishah:
This PR follows Option 2 from issue #650 to implement a simple scope-based pretty-printer for IDL using the existing AST.

- Added `pretty_idl(indent = 0)` methods to AST classes in `lib/idl/ast.rb`
- Created CLI tool at `tools/format_idl.rb` to format `.idl` files via AST

issue [#650](https://github.com/riscv-software-src/riscv-unified-db/issues/650l)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -96,7 +96,6 @@ repos:
     hooks:
       - id: pyupgrade
         stages: [pre-commit]
-
   - repo: local
     hooks:
       - id: shellcheck
@@ -105,6 +104,15 @@ repos:
         language: system
         entry: shellcheck
         args: ["--severity=error"]
+
+  - repo: local
+    hooks:
+      - id: format-idl
+        name: Format IDL files
+        entry: tools/format_idl.rb
+        language: system
+        types: [file]
+        files: \.idl$
 
   - repo: https://github.com/scop/pre-commit-shfmt
     rev: v3.11.0-1
--- a/backends/manual/templates/instruction.adoc.erb
+++ b/backends/manual/templates/instruction.adoc.erb
@@ -5,18 +5,16 @@
 
 *<%= inst.long_name %>*
 
-== Assembly format
-
-`<%= inst.name %> <%= inst.assembly.gsub('x', 'r') %>`
-
-== Synopsis
+<%= inst.fix_entities(inst.description) %>
 
 <%- if inst.data_independent_timing? -%>
 [IMPORTANT]
 This instruction must have data-independent timing when extension `Zkt` is enabled.
 <%- end -%>
 
-<%= inst.fix_entities(inst.description) %>
+== Assembly format
+
+`<%= inst.name %> <%= inst.assembly.gsub('x', 'r') %>`
 
 == Decode Variables
 
--- a/lib/idl/ast.rb
+++ b/lib/idl/ast.rb
@@ -402,6 +402,10 @@ module Idl
     # @!macro to_idl
     # @abstract
     def to_idl = raise NotImplementedError, "#{self.class.name} must implement to_idl"
+    def pretty_idl(indent = 0)
+      ('  ' * indent) + to_idl + "\n"
+    end
+
 
     def inspect = self.class.name.to_s
   end
@@ -667,6 +671,11 @@ module Idl
 
     # @!macro to_idl
     def to_idl = name
+    def pretty_idl(indent = 0)
+      indent_str = '  ' * indent
+      "#{indent_str}#{type} #{name};\n"
+    end
+
   end
 
   class GlobalWithInitializationSyntaxNode < Treetop::Runtime::SyntaxNode
@@ -727,6 +736,11 @@ module Idl
     def to_idl
       var_decl_with_init.to_idl
     end
+    def pretty_idl(indent = 0)
+      indent_str = '  ' * indent
+      "#{indent_str}#{type} #{name};\n"
+    end
+
   end
 
   class GlobalSyntaxNode < Treetop::Runtime::SyntaxNode
@@ -880,7 +894,13 @@ module Idl
     end
 
     def to_idl = "$array_size(#{expression.to_idl})"
+    def pretty_idl(indent = 0)
+      indent_str = '  ' * indent
+      "#{indent_str}#{type} #{name};\n"
+    end
+
   end
+
 
 
   class EnumSizeSyntaxNode < Treetop::Runtime::SyntaxNode
@@ -916,6 +936,11 @@ module Idl
     end
 
     def to_idl = "$enum_size(#{enum_class.to_idl})"
+    def pretty_idl(indent = 0)
+      indent_str = '  ' * indent
+      "#{indent_str}#{type} #{name};\n"
+    end
+
   end
 
   class EnumElementSizeSyntaxNode < Treetop::Runtime::SyntaxNode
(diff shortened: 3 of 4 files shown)


syedowaisalishah:
Hi @dhower-qc, @kbroch-rivosinc, and @ThinkOpenly,
I’ve implemented a basic pretty-printer based on the AST, following Option 2 from issue [#650](https://github.com/riscv-software-src/riscv-unified-db/issues/650). The goal is to ensure proper indentation of .idl files using the existing AST structure.

**Approach**
I added pretty_idl(indent = 0) methods to all relevant AST classes and created a CLI tool (tools/format_idl.rb) that formats .idl files using the AST. This avoids tracking comments/whitespace, staying consistent with Option 2’s simpler scope-based formatting path.

**Blocker**
I'm currently stuck because I couldn't find how .idl files are parsed into AST nodes — no parser.rb, IDL.parse, or similar entry point is present in the repo.

**Question**
Could you confirm:
Is my approach consistent with what you envisioned for Option 2?
Is there an existing parser function or method I can use to parse .idl files into AST nodes?
If not, what would be the recommended next step?

Once that’s clear, I’ll wire up the formatting tool fully.
Thanks!

kbroch-rivosinc:
From the pre-commit edits it looks like you are only formatting `*.idl` files of which there are only:

```
❯ fd \.idl$
arch/isa/fetch.idl
arch/isa/interrupts.idl
arch/isa/util.idl
arch/isa/fp.idl
arch/isa/builtin_functions.idl
```


Is there any consideration to format the IDL in the yaml files like this: https://github.com/riscv-software-src/riscv-unified-db/blob/main/arch/inst/I/ecall.yaml#L32

kbroch-rivosinc:
> I'm a bit outside my comfort zone, but how would we expect it to work? Would there be a new pre-commit hook, defined similarly to the hook defined here, but operating on YAML files? It would load the YAML files, look for fields that contain YAML? (Maybe it needs to know which fields to look for, because they are just defined as being "string" type.) Then, it formats the string as IDL, and... reports if there are differences between the original string and the formatted string? (It can't magically fix the YAML, can it?)

That's how I was thinking it would work.  The potentially tricky part is that there's effectively two formatters running on the IDL code in the yaml file and if they don't agree problems occur. I think prettier will leave the IDL string alone as long as:
* whitespace before first non-whitespace char is maintained as prettier wants it
*  line length isn't exceeded

dhower-qc:
Definitely on the right track for the vision.

For the parser:

The grammar is defined in https://github.com/riscv-software-src/riscv-unified-db/blob/main/lib/idl/idl.treetop. The AST gets constructed from that. For example:

https://github.com/riscv-software-src/riscv-unified-db/blob/808077a49727f05e75d3e8150256f757eda5eef8/lib/idl/idl.treetop#L49-L53

When parsing, a match for rule `bitfield_definition` will construct an `Idl::BitfieldDefinitionSyntaxNode`:

https://github.com/riscv-software-src/riscv-unified-db/blob/808077a49727f05e75d3e8150256f757eda5eef8/lib/idl/ast.rb#L1292-L1300

Which exists just to define a `to_ast` method that returns `BitfieldDefinitionAst`.

You can invoke this all using the functions in https://github.com/riscv-software-src/riscv-unified-db/blob/main/lib/idl.rb.

For example, to get the AST for an entire `.idl` file:

```ruby
ast = Idl::Compiler.compile_file(path)
```

For the following, you can use a `ConfiguredArchitecture` object to get the AST of YAML-defined IDL:

```ruby
cfg_arch = cfg_arch_for('_')  # for pretty printing, you probably want the 'unconfigured' config '_'
```

To get the AST for an instruction `operation()`:

```ruby
ast = cfg_arch.instruction('ld').operation_ast  # internally, calls Idl::Compiler.compile_inst_operation
```

For a CSR `sw_read()`:

```ruby
ast = cfg_arch.csr('mstatus').sw_read_ast
```

For a CSR field `type()`:

```ruby
ast = cfg_arch.csr('mstatus').field('SD').type_ast
```

Similar for `reset_value_ast`, `sw_write_ast`.

dhower-qc:
> From the pre-commit edits it looks like you are only formatting `*.idl` files of which there are only:
> 
> ```
> ❯ fd \.idl$
> arch/isa/fetch.idl
> arch/isa/interrupts.idl
> arch/isa/util.idl
> arch/isa/fp.idl
> arch/isa/builtin_functions.idl
> ```
> 
> Is there any consideration to format the IDL in the yaml files like this: https://github.com/riscv-software-src/riscv-unified-db/blob/main/arch/inst/I/ecall.yaml#L32

`arch/isa/globals.isa` is also an IDL file. For better or worse, it has a different extension to mark that it is the special top-level file for compilation (you'll see the other files get included from it).

syedowaisalishah:
Thanks @dhower-qc  That clears things up. I’ll update my formatter to use Idl::Compiler.compile_file(path) for parsing and will explore YAML support via ConfiguredArchitecture next. Appreciate the help!

syedowaisalishah:
@dhower-qc,
Thanks again for your guidance earlier!
I’ve completed implementing the AST-based pretty-printer using the simpler indentation-based approach (Option 2).
Could you please review the implementation to confirm if it aligns with the intended direction?
Also, what’s the best way to verify that the formatting logic is working correctly?
Thanks!

dhower-qc:
Great. Now we need to figure out how to retain comments and possibly intentional whitespace (like blank lines between statements). Currently, comments and whitespace are both parsed but then thrown away (don't make it into the AST). 

Warning, this isn't going to be simple.

The first step is to create AST nodes for comments and whitespace. As we add comments/intentional whitespace in the AST, we have to make decisions about what they are attached to in the tree.

Take statements as an example. There are at least four scenarios to account for:

```
# this is a pre-comment for the statement
X[xd] = X[xs1] + X[xs2];
```

```
# this is a free-floating comment

X[xd] = X[xs1] + X[xs2];
```

```
X[xd] = X[xs1] + X[xs2]; # this is a post-comment for the statement
```

```
X[xd] = X[xs1] + X[xs2];

# this is a free-floating comment
```

Intuitively, we probably want the pre-comment and post-comment attached to the StatementAst node. But where do the free-floating comments go? When we see statements, they are usually (always?) part of a list of statements in the AST (e.g., an IfBody points to a list of statements). Now that list probably becomes a combined list of statements and comments (and intentional whitespace).

That's just an example; we have deal with this anywhere there is whitespace. Consider:

```
if (A # comment for A
    || B# comment for B
    # random comment
 # another random comment
    && C) # comment for if?
{
}
```

All of this is going to require changing the grammar structure and the Ast node classes.

Take-away: formatters aren't easy, even for 'simple' languages :)

There may be a better way to approach this that I'm not thinking about. I'm also not opposed to changing the language a bit if we think of a way to make things easier (random example; maybe it's easier to attach a post-comment if they all have to start with `#<` instead of just `#`; or maybe we find there is a verilog formatter out there that would work with some small changes)

syedowaisalishah:
> @syedowaisalishah, let me know if you want to keep this open. There is now a new effort to do pretty printing through a holistic UDB tool

Thanks for the update, @dhower-qc. Since I’m currently in a full-time role and in my probation period, I have limited bandwidth. If the new UDB tool will handle pretty printing, I’m okay with closing this. Otherwise, feel free to decide what works best.

Issue #773, opened 2025-05-16
Profile releases show options multiple times


james-ball-qualcomm:
Problem with Porfolio.uses_optional_types? algorithm.

PR #774, opened 2025-05-16
fix: RVI20 incorrectly shows optional extension multiple times

james-ball-qualcomm:
fixes: #773


lib/arch_obj_models/portfolio.rb
```diff
@@ -665,11 +665,11 @@ def uses_optional_types?
 
     @uses_optional_types = false
 
-    # Iterate through different kinds of optional using the "object" version (not the string version).
-    Presence.optional_types_obj.each do |optional_type_obj|
-      # See if any extension reqs have this type of optional.
-      unless in_scope_ext_reqs(optional_type_obj).empty?
-        @uses_optional_types = true
+    in_scope_ext_reqs(Presence.optional)&.each do |ext_req|
+      unless ext_req.presence.nil?
```
dhower-qc, line 669: Is this check needed? Doesn't `in_scope_ext_reqs(Presence.optional)` only return ext_reqs with a presence (of optional)?

james-ball-qualcomm: Good catch

dhower-qc (approved):

PR #775, opened 2025-05-16
Xqccmp: Left shift `spimm` in pop instructions

AntonJohansson:
Shifting `spimm` left by 4 matches the behaviour of cm.push/cm.pop, and moreover since qc.cm.push shifts by 4 the stack address is not calculated correctly upon popping. E.g.

    qc.cm.push      {ra}, -64
    qc.cm.pop       {ra}, 64

will not recover the stack address in addition to loading incorrect values for all saved registers.

### DIFF
--- a/arch_overlay/qc_iu/inst/Xqccmp/qc.cm.pop.yaml
+++ b/arch_overlay/qc_iu/inst/Xqccmp/qc.cm.pop.yaml
@@ -24,6 +24,7 @@ encoding:
       not: [0, 1, 2, 3]
     - name: spimm
       location: 3-2
+      left_shift: 4
 access:
   s: always
   u: always
--- a/arch_overlay/qc_iu/inst/Xqccmp/qc.cm.popret.yaml
+++ b/arch_overlay/qc_iu/inst/Xqccmp/qc.cm.popret.yaml
@@ -24,6 +24,7 @@ encoding:
       not: [0, 1, 2, 3]
     - name: spimm
       location: 3-2
+      left_shift: 4
 access:
   s: always
   u: always
--- a/arch_overlay/qc_iu/inst/Xqccmp/qc.cm.popretz.yaml
+++ b/arch_overlay/qc_iu/inst/Xqccmp/qc.cm.popretz.yaml
@@ -24,6 +24,7 @@ encoding:
       not: [0, 1, 2, 3]
     - name: spimm
       location: 3-2
+      left_shift: 4
 access:
   s: always
   u: always
--- a/arch_overlay/qc_iu/inst/Xqccmp/qc.cm.pushfp.yaml
+++ b/arch_overlay/qc_iu/inst/Xqccmp/qc.cm.pushfp.yaml
@@ -26,6 +26,7 @@ encoding:
       not: [0, 1, 2, 3, 4]
     - name: spimm
       location: 3-2
+      left_shift: 4
 access:
   s: always
   u: always


AFOliveira (commented):
Adding @james-ball-qualcomm @ayosher since they are both also involved in qc's extension

arch_overlay/qc_iu/inst/Xqccmp/qc.cm.pop.yaml
```diff
@@ -24,6 +24,7 @@ encoding:
       not: [0, 1, 2, 3]
     - name: spimm
       location: 3-2
+      left_shift: 4
```
dhower-qc, line 27: Currently, this field is only informative. To actually get the shift, you need to implement it in the `operation()`:

```
  XReg virtual_address_new_sp = virtual_address_sp + stack_aligned_adj + (spimm `<< 4);
```

(Note, this uses widening left shift, prefixed with a backtick)

This will apply to all four files.

ayosher: Folks,
we cannot have both left_shift (informative) and widening left shift in operation. It will mislead people when reading PDF. I believe we need support correctly left_shift inthe infrastructure.
Regarding the widening operations - currently it looks weird in PDF.

dhower-qc: Yes, the plan is to remove the redundancy. It's tied up in a larger schema change for instructions, though, so will take a few weeks at least.

dhower-qc (approved):

arch_overlay/qc_iu/inst/Xqccmp/qc.cm.pop.yaml
AntonJohansson: I see, I've addressed this and also added a `<< 4 to `qc.cm.push` which only specified `left_shift: 4`.

Always assumed `left_shift:`,`not:` etc. were to be performed during decoding so that `operation()` always has the guarantee that `spimm` is shifted by 4 to recover the value encoded in the instruction.

Also, if `left_shift` is decorative then I assume so is `not`, meaning we would have to add
```
if (rlist < 4) {
    raise(...)
}
```
and similar to a lot of instructions, which is not too nice.

dhower-qc: Quite confusingly, `not` is applied and does not need to explicitly checked

dhower-qc:
@ayosher, excellent point. I'll try to accelerate this change.

AntonJohansson:
Currently we have `left_shift: 4` and `` `<< 4`` for the instructions touched here, but all other instructions only specify `left_shift`. So ignoring `left_shift` is not viable.

IMO, anything involved in producing a correct disassembly should be part of variable and not `op()`, so I think either duplicating `left_shift` and `` `<<`` everywhere, or dropping `` `<<`` is a good idea.

If the concern is `left_shift` not being expressive enough, what about something like
```yaml
  variables:
    - name: imm
      location: 29-25
      decode_op: imm `<< 2
```
this way you could also deal with fields such as `r1s`
```yaml
  variables:
    - name: r1s
      location: 9-7
      decode_op: (r1s[2:1]>0) ? {1,0,r1s[2:0]} : {0,1,r1s[2:0]}
```
just an idea.

ayosher:
@Derek ***@***.***>,

Can we add support for left_shift, so it will be enough and not require additional things? Once this and XLEN-wide constants will be in, we can fix all Xqci/Xqccmp instructions to remove all widening currently was added.

Albert
________________________________

Issue #562, opened 2025-03-27
Add Zcmt CSR


dhower-qc:
Need to add:

- JVT

neverlandiz:
Hello, I'd like to work on this issue if no one else is working on it. Thanks!

PR #782, opened 2025-05-19
Add JVT CSR

neverlandiz:
This PR closes Issue #562 and adds the missing Zcmt CSR (jvt.yaml).


arch/csr/jvt.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+      `jvt.base` is a virtual address, whenever virtual memory is enabled.
+
+      The memory pointed to by `jvt.base` is treated as instruction memory for the purpose of executing table
+      jump instructions, implying execute access permission.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+  MODE:
+    location_rv64: 5-0
+    location_rv32: 5-0
```
dhower-qc, line 40: When the location is the same, just use `location`:

```yaml
MODE:
  location: 5-0
```

dhower-qc (changes requested):

arch/csr/jvt.yaml
ThinkOpenly, line 1: I think we need a `sw_read` and `sw_write`, based on:
> If the Smstateen extension is implemented, then bit 2 in mstateen0, sstateen0, and hstateen0 is implemented. If bit 2 of a controlling stateen0 CSR is zero, then access to the jvt CSR [...] results in an illegal-instruction trap (or, if appropriate, a virtual-instruction trap).

arch/csr/jvt.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+      The memory pointed to by `jvt.base` is treated as instruction memory for the purpose of executing table
+      jump instructions, implying execute access permission.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+    sw_write(csr_value): |
+      if (implemented?(ExtensionName::Smstateen) && (CSR[mstateen0].JVT == 1'b0 || CSR[sstateen0].JVT == 1'b0 || CSR[hstateen0].JVT == 1'b0)) {
+        return ILLEGAL_WLRL;
+      }
+      return csr_value.BASE;
```
dhower-qc, line 42: Three considerations:

- Each of mstateen, sstateen, and hstateen need to be considered separately since they control access in different modes.
- The effect of the stateen.JVT bit is that the entire JVT csr appears to be unimplemented.
- BASE is WARL, so we need parameters on the Zcmt extension. Something like

```yaml
  JVT_BASE_TYPE:
    schema:
      enum: [mask, custom]
    description: |
      Type of the jvt.BASE CSR field. One of:

        * mask: jvt.BASE contains one or more implemented bits, which are indicated by JVT_BASE_MASK.
        * custom: Custom behavior. Will cause hart to enter 'unpredictable' state on a write to jvt.BASE.
  JVT_BASE_MASK:
    schema:
      type: integer
      minimum: 64
      maximum: 0x7ffffffffffffff # 58 bits
    # required_if() isn't part of the schema, but maybe it should be?
    # required_if(): return JVT_BASE_TYPE == "mask";
    description: |
      Mask representing the implemented bits of jvt.BASE. Includes the implicitly-zero bits of jvt.BASE, so JVT_BASE_MASK[5:0] must always be 0.
    extra_validation: |
      assert JVT_BASE_MASK < 0xffffffc0 if MXLEN == 32
```

sw_write will be something like

```
  if ((mode() != PrivMode::M) && implemented?(ExtensionName::Smstateen)) {
    if (CSR[mstateen0].JVT == 1'b0) {
      unimplemented_csr();
    }
  }
  else if ((mode() == PrivMode::U) && implemented?(ExtensionName::Ssstaten)) {
    if (CSR[sstateen0].JVT == 1'b0) {
      unimplemented_csr();
    }
  }
  else if ((mode() == PrivMode::VS) && implemented?(ExtensionName::Ssstaten)) {
    if (CSR[hstateen0].JVT == 1'b0) {
      unimplemented_csr();
    }
  }
  else if ((mode() == PrivMode::VU) && implemented?(ExtensionName::Ssstateen)) {
    if ((CSR[sstateen0].JVT == 1'b0) || (CSR[hstateen0].JVT == 1'b0)) {
      unimplemented_csr();
    }
  }
  else {
    if (JVT_BASE_TYPE == "custom") {
      unpredictable("jvt.BASE has custom behavior");
    } else {
      return csr_value.BASE & (JVT_BASE_MASK >> 6);
    }
    return csr_value.BASE == 0 ? 0 : UNDEFINED_LEGAL_DETERMINISTIC;
  }
```

arch/csr/jvt.yaml
```diff
@@ -0,0 +1,64 @@
...
+      The value in the BASE field must always be aligned on a 64-byte boundary. Note that the CSR contains only
+      bits XLEN-1 through 6 of the address base. When computing jump-table accesses, the lower six bits of base
+      are filled with zeroes to obtain an XLEN-bit jump-table base address `jvt.base` that is always aligned on a
+      64-byte boundary.
+
+      `jvt.base` is a virtual address, whenever virtual memory is enabled.
+
+      The memory pointed to by `jvt.base` is treated as instruction memory for the purpose of executing table
+      jump instructions, implying execute access permission.
+    type: RW
```
dhower-qc, line 36: This can be RW or RO. We need a parameter on Zcmt to know which. From the spec:

> If [Section 29.10](https://github.com/riscv-software-src/riscv-unified-db/pull/782/files#Zcmt) is implemented then jvt must also be implemented, but can contain a read-only value.

arch/csr/jvt.yaml
```diff
@@ -0,0 +1,64 @@
...
+        return ILLEGAL_WLRL;
+      }
+      return csr_value.BASE;
+  MODE:
+    location: 5-0
+    description: |
+      `jvt.mode` is a WARL field, so can only be programmed to modes which are implemented. Therefore the
+      discovery mechanism is to attempt to program different modes and read back the values to see which
+      are available. Jump table mode must be implemented.
+    type: RW
```
dhower-qc, line 49: MODE is actually RO according the spec:

> jvt.mode is a WARL field, so can only be programmed to modes which are implemented. 

MODE can currently only take one value (0), so the field is Read-Only.

arch/csr/jvt.yaml
```diff
@@ -0,0 +1,64 @@
...
+    reset_value: UNDEFINED_LEGAL
```
dhower-qc, line 50: 0, until another mode is defined.

arch/csr/jvt.yaml
```diff
@@ -0,0 +1,64 @@
...
+    sw_write(csr_value): |
+      if (implemented?(ExtensionName::Smstateen) && (CSR[mstateen0].JVT == 1'b0 || CSR[sstateen0].JVT == 1'b0 || CSR[hstateen0].JVT == 1'b0)) {
+        return ILLEGAL_WLRL;
+      }
+      return csr_value.MODE;
```
dhower-qc, line 55: Not needed.

arch/csr/jvt.yaml
```diff
@@ -0,0 +1,64 @@
...
+sw_read(): |
+  # If the Smstateen extension is implemented, then bit 2 in `mstateen0`, `sstateen0`, and `hstateen0` is
+  # implemented. If bit 2 of a controlling `stateen0` CSR is zero, then access to the `jvt` CSR and execution
+  # of a `cm.jalt` or `cm.jt` instruction by a lower privilege level results in an illegal-instruction trap (or, if
+  # appropriate, a virtual-instruction trap).
+
+  if (implemented?(ExtensionName::Smstateen) && (CSR[mstateen0].JVT == 1'b0 || CSR[sstateen0].JVT == 1'b0 || CSR[hstateen0].JVT == 1'b0)) {
+    raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+  }
```
dhower-qc, line 64: similar to the example for BASE above

dhower-qc (changes requested):

arch/csr/jvt.yaml
neverlandiz: I think the spec refers to section 28.10, which is Zcmt. Does that mean it's RO when Zcmt is implemented? Would it be accurate to use 
```
type(): |
      if (implemented?(ExtensionName::Zcmt)) {
        return CsrFieldType::RO;
      } 
      else {
        return CsrFieldType::RW;
      }
```
Thanks!

ThinkOpenly: > Does that mean it's RO when Zcmt is implemented? Would it be accurate to use

The text that Derek quoted is important, saying "if Zcmt is implemented, JVT is implemented, but _can_ contain a read-only value". So, as Derek mentioned a new parameter needs to be added to `arch/ext/Zcmt.yaml` to indicate whether JVT is read-only. Does that make sense?

neverlandiz: Yes, that makes sense. Thank you so much!

arch/csr/jvt.yaml
```diff
@@ -0,0 +1,106 @@
...
+        }
+        return csr_value.BASE == 0 ? 0 : UNDEFINED_LEGAL_DETERMINISTIC;
+      }
+  MODE:
+    location: 5-0
+    description: |
+      `jvt.mode` is a WARL field, so can only be programmed to modes which are implemented. Therefore the
+      discovery mechanism is to attempt to program different modes and read back the values to see which
+      are available. Jump table mode must be implemented.
+    type: RO
```
ThinkOpenly: This should be the same `type()` as the BASE field, above, since it is the entire `jvt` CSR that is covered by "If [Zcmt] is implemented, then `jvt` must also be implemented, but can contain a read-only value`.

arch/csr/jvt.yaml
```diff
@@ -0,0 +1,106 @@
...
+    reset_value: 0
+sw_read(): |
+  # If the Smstateen extension is implemented, then bit 2 in `mstateen0`, `sstateen0`, and `hstateen0` is
+  # implemented. If bit 2 of a controlling `stateen0` CSR is zero, then access to the `jvt` CSR and execution
+  # of a `cm.jalt` or `cm.jt` instruction by a lower privilege level results in an illegal-instruction trap (or, if
+  # appropriate, a virtual-instruction trap).
```
dhower-qc: This means that if the current execution mode is VS or VU, then the exception should be `ExceptionCode::VirtualInstruction`. That's going to require changes to lines 89, 99, and 104

arch/csr/jvt.yaml
```diff
@@ -0,0 +1,112 @@
...
+    type(): |
+      if (JVT_READ_ONLY) {
+        return CsrFieldType::RO;
+      }
+      else {
+        return CsrFieldType::RW;
+      }
+    reset_value: UNDEFINED_LEGAL
+    sw_write(csr_value): |
+      if ((mode() != PrivMode::M) && implemented?(ExtensionName::Smstateen)) {
```
ThinkOpenly: CI is tripping over this.
```suggestion
      if ((mode() != PrivilegeMode::M) && implemented?(ExtensionName::Smstateen)) {
```
make this change everywhere for "PrivMode".

arch/csr/jvt.yaml
ThinkOpenly: This still needs attention. The MODE field has only one defined valid value, 0.

I differ from Derek's suggestion that it should be read-only because it only has one legal value. I think it is legal to write the valid value of 0 to this field. So, for MODE, I see:
- needs to respect JVT_READ_ONLY
- is writable (when JVT_READ_ONLY is false)
- can only accept a value of 0

Something like this, perhaps? (Untested.)
```
    sw_write(csr_value): |
￼       if (JVT_READ_ONLY || (csr_value != 0)) {
          ￼ raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
        }
        return 0;
```

ThinkOpenly: Are you able to address this last comment?

neverlandiz: Sorry about the delay, the changes have been pushed now.

ThinkOpenly (approved):

Issue #727, opened 2025-05-05
The How-to/recipes section in top level README.adoc are broken


AFOliveira:
**Describe the bug**
The how-to/recipes hyperlinks in the readme.md don't take anywhere inside the repo.
![Image](https://github.com/user-attachments/assets/a51d29e9-48ab-4e2c-acd1-fb7704a9a8f3)

That may be because we don't yet have any links to link to, but IMO at least placeholders should be there

kbroch-rivosinc:
I'll put a PR in for this.  link to README: https://github.com/riscv-software-src/riscv-unified-db/blob/main/README.adoc

PR #783, opened 2025-05-19
chore: fix broken links in README

kbroch-rivosinc:
couple xrefs had nothing to point to and others
need adoc on end to point to correct file
relates to #727


AFOliveira (approved):
Thanks for the help!

ThinkOpenly (approved):

Issue #788, opened 2025-05-20
Missing Fields in `misa` Register Definition


AFOliveira:
**Observed:**

The current definition of the `misa` register in our system includes the following fields:
`A`, `B`, `C`, `D`, `F`, `G`, `H`, `I`, `M`, `S`, `U`, `V`

**Expected :**

According to the RISC-V Privileged Specification (Volume II), the `misa` register should define bits for a broader range of extensions. The table below (from the spec) illustrates the standard bit assignments:

![Standard MISA bit assignments](https://github.com/user-attachments/assets/34ad321c-a5a3-497a-b1a6-50b8ef6e927a)

**Gap Analysis:**

Comparing our current implementation against the specification, the following fields appear to be missing from our `misa` definition:
`E`, `J`, `K`, `L`, `N`, `O`, `P`, `Q`, `R`, `T`, `W`, `X`, `Y`, `Z`

**Next steps:**

We should review and update our `misa` register definition to include these standard fields

dhower-qc:
Most of those are intentionally omitted since they don't correspond to anything and are read-only-0. (unspecified fields are implicitly read-only-0). The exceptions are:

- E (needs E in UDB)
- Q (needs Q in UDB)

X should never be in the standard definition, and can be added in an overlay.

AFOliveira:
Q is in UDB, right?

dhower-qc:
ah, so it is

AFOliveira:
Fixed by #801

PR #801, opened 2025-05-22
feat: add misa Q bit, add MUTABLE_MISA_Q parameter to Q extension

BKhristenko:
This pull request introduces the following changes:

### MISA Q Bit
- Added a new field `Q` in `arch/csr/misa.yaml` to indicate support for the Q (quad precision float) extension.

### MUTABLE_MISA_Q Parameter
- Introduced a new parameter `MUTABLE_MISA_Q` in `arch/ext/Q.yaml`.

### Additional Notes
- Changes to `F` and `D` bits might be required. Looking forward to getting some feedback on this.

**Related Issue:**
- Issue #788

### DIFF
--- a/arch/csr/misa.yaml
+++ b/arch/csr/misa.yaml
@@ -169,6 +169,28 @@ fields:
           .. Execute every in-scope multiply extension instruction
           . check
           .. Check that every multiply extension instruction throws illegal instruction exception
+  Q:
+    location: 16
+    description: |
+      Indicates support for the `Q` (quad precision float) extension.
+
+      [when,"MUTABLE_MISA_Q == true"]
+      --
+      Writing 0 to this field will cause all quad-precision floating point instructions to raise an `IllegalInstruction` exception.
+      --
+    type(): |
+      return MUTABLE_MISA_Q ? CsrFieldType::RW : CsrFieldType::RO;
+    reset_value: 1
+    definedBy: Q
+    sw_write(csr_value): |
+      if ((csr_value.F == 0 || csr_value.D == 0) && csr_value.Q == 1) {
+        return UNDEFINED_LEGAL_DETERMINISTIC;
+      }
+
+      # fall-through; write the intended value
+      return csr_value.Q;
+    legal?(csr_value): |
+      return !(csr_value.Q == 1 && csr_value.D == 0);
   S:
     location: 18
     description: |
--- a/arch/ext/Q.yaml
+++ b/arch/ext/Q.yaml
@@ -18,3 +18,9 @@ versions:
     state: ratified
     ratification_date: null
     requires: D
+params:
+  MUTABLE_MISA_Q:
+    description: |
+      Indicates whether or not the `Q` extension can be disabled with the `misa.Q` bit.
+    schema:
+      type: boolean


arch/csr/misa.yaml
```diff
@@ -169,6 +169,29 @@ fields:
...
+    definedBy: Q
+    sw_write(csr_value): |
+      if (csr_value.Q == 0) {
+        return UNDEFINED_LEGAL_DETERMINISTIC;
+      }
+
+      # fall-through; write the intended value
+      return csr_value.Q;
+    legal?(csr_value): |
+      return !(csr_value.Q == 0);
```
ThinkOpenly, line 194: I don't think this is correct. My reading is that this is saying you can't disable Q. I think what you want here is:
```
return !(csr_value.Q == 1 && csr_value.D == 0);
```
...you can't enable Q with D disabled.

arch/csr/misa.yaml
```diff
@@ -169,6 +169,29 @@ fields:
...
+      --
+      Writing 0 to this field will cause all quad-precision floating point instructions to raise an `IllegalInstruction` exception.
+      --
+    type(): |
+      return (implemented?(ExtensionName::Q) && MUTABLE_MISA_Q) ? CsrFieldType::RW : CsrFieldType::RO;
+    reset_value(): |
+      return implemented?(ExtensionName::Q) ? 1 : 0;
+    definedBy: Q
+    sw_write(csr_value): |
+      if (csr_value.Q == 0) {
```
ThinkOpenly, line 187: I think you want:
```
if (csr_value.D == 0 && csr_value.Q == 1) {
```

ThinkOpenly (changes requested):
Thanks for the PR! A couple of things to change...

arch/csr/misa.yaml
AFOliveira: Should we add this to D as well?

arch/csr/misa.yaml
```diff
@@ -169,6 +169,29 @@ fields:
...
+      --
+      Writing 0 to this field will cause all quad-precision floating point instructions to raise an `IllegalInstruction` exception.
+      --
+    type(): |
+      return (implemented?(ExtensionName::Q) && MUTABLE_MISA_Q) ? CsrFieldType::RW : CsrFieldType::RO;
+    reset_value(): |
+      return implemented?(ExtensionName::Q) ? 1 : 0;
+    definedBy: Q
+    sw_write(csr_value): |
+      if (csr_value.D == 0 && csr_value.Q == 1) {
```
dhower-qc, line 187: The transitive requirement of "D implies F" won't be checked here, so you'll want to make it:

```
if ((csr_value.F == 0 || csr.value.D == 0) && csr_value.Q == 1) {
```

dhower-qc (changes requested):
Thanks for contributing!

ThinkOpenly:
> Question [...] if D and F should reflect the introduction of Q? Per example, by being mutually exclusive

I'm unsure about the question.

F stands alone.
D depends on F, and they can coexist.
Q depends on D, and they can all coexist.

AFOliveira:
> > Question [...] if D and F should reflect the introduction of Q? Per example, by being mutually exclusive
> 
> I'm unsure about the question.
> 
> F stands alone. D depends on F, and they can coexist. Q depends on D, and they can all coexist.

Nevermind, my understanding of this was wrong (it's starting to pretty common these days :))

arch/csr/misa.yaml
```diff
@@ -169,6 +169,29 @@ fields:
...
+      Indicates support for the `Q` (quad precision float) extension.
+
+      [when,"MUTABLE_MISA_Q == true"]
+      --
+      Writing 0 to this field will cause all quad-precision floating point instructions to raise an `IllegalInstruction` exception.
+      --
+    type(): |
+      return (implemented?(ExtensionName::Q) && MUTABLE_MISA_Q) ? CsrFieldType::RW : CsrFieldType::RO;
+    reset_value(): |
+      return implemented?(ExtensionName::Q) ? 1 : 0;
```
ThinkOpenly, line 184: Out of curiosity, @dhower-qc, do we need to check for `implemented?` here, given that the field is `definedBy: Q`? (I see this is how it is done for other fields, so the implementation here is following those examples.) If Q is not implemented, does this not just become a read-only zero field?

ThinkOpenly (approved):

arch/csr/misa.yaml
dhower-qc: Good catch Paul -- this doesn't need to check implemented?(Q). If Q isn't implemented, then the field disappears. So, in this case, we just need:

```
type(): |
  return MUTABLE_MISA_Q ? CsrFieldType::RW : CsrFieldType::RO;
reset_value: 1
```

arch/csr/misa.yaml
```diff
@@ -169,6 +169,29 @@ fields:
...
+      Indicates support for the `Q` (quad precision float) extension.
+
+      [when,"MUTABLE_MISA_Q == true"]
+      --
+      Writing 0 to this field will cause all quad-precision floating point instructions to raise an `IllegalInstruction` exception.
+      --
+    type(): |
+      return MUTABLE_MISA_Q ? CsrFieldType::RW : CsrFieldType::RO;
+    reset_value(): |
+      return 1;
```
dhower-qc, line 184: This isn't wrong, but to maintain consistency with the rest of the files, let's use the non-function form:

```yaml
reset_value: 1
```

dhower-qc (changes requested):

dhower-qc (approved):

PR #803, opened 2025-05-22
fix(data): Fix remaining old register names

ThinkOpenly:
- `rd` to `xd`
- `rs1` to `xs1`
- `rs2` to `xs2`

Fix a few cases where a `$` appeared before the register name.

Add a few missing "long_names".

### DIFF
--- a/arch/inst/C/c.add.yaml
+++ b/arch/inst/C/c.add.yaml
@@ -5,20 +5,20 @@ kind: instruction
 name: c.add
 long_name: Add
 description: |
-  Add the value in rs2 to rd, and store the result in rd.
-  C.ADD expands into `add rd, rd, rs2`.
+  Add the value in xs2 to xd, and store the result in xd.
+  C.ADD expands into `add xd, xd, xs2`.
 definedBy:
   anyOf:
     - C
     - Zca
-assembly: xd, rs2
+assembly: xd, xs2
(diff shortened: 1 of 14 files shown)


ThinkOpenly:
At least for me, this PR seems to be stuck processing the 2nd commit. I'm not sure how to unwedge it?

dhower-qc:
Looks like it ran now? The failing one needs the golden instruction appendix updated. We should probably make it easier to do that...

ThinkOpenly:
> Looks like it ran now? The failing one needs the golden instruction appendix updated. We should probably make it easier to do that...

How many commits do you see here? I only see one. Still.

dhower-qc:
> > Looks like it ran now? The failing one needs the golden instruction appendix updated. We should probably make it easier to do that...
> 
> How many commits do you see here? I only see one. Still.

I only see one. If you "push" locally, git says it's up to date?

ThinkOpenly:
> I only see one. If you "push" locally, git says it's up to date?

Yep. And at the top of this screen, next to the title, I see this spinner:
image

dhower-qc:
oh, I see that now. based on this:

https://stackoverflow.com/questions/77238329/github-shows-processing-updates-on-my-pull-request-after-every-push

pushing an empty commit might get it to wake up

PR #806, opened 2025-05-22
fix(data): floating-point loads/stores several issues

ThinkOpenly:
- make sure the loads load to a float register
- make sure the stores store from a float register
- update the corresponding field names, description, IDL
- make sure the assembly syntax are correct
- add a few missing "long_name"
- remove an extraneous `$`

### DIFF
--- a/arch/inst/D/fld.yaml
+++ b/arch/inst/D/fld.yaml
@@ -3,19 +3,19 @@
 $schema: inst_schema.json#
 kind: instruction
 name: fld
-long_name: No synopsis available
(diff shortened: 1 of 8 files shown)


AFOliveira (approved):
Thanks for stepping in and helping on these! I don't have enough bandwidth to solve all this problems as of now. All seems correct, but I think you will also need to update the regression test to get this through CI. Should be as simple as replacing the golden adoc with yours.

dhower-qc (approved):

PR #808, opened 2025-05-23
fix(data): fix remaining old register names

ThinkOpenly:
- `rd` to `xd`
- `rs1` to `xs1`
- `rs2` to `xs2`

Fix a few cases where a `$` appeared before the register name.

Add a few missing "long_names".


ThinkOpenly:
> #803 got in a really weird state...

#803 is still in an odd state, but this one seems fine. (closing #803, marking this ready for review.)

PR #809, opened 2025-05-23
fix(data): correct assembly in a few places

ThinkOpenly:
- `c.nop` has no operands
- For CSR instructions, `imm` should be `csr` to better match Sail
- Correct operand order and names

### DIFF
--- a/arch/inst/Zicsr/csrrw.yaml
+++ b/arch/inst/Zicsr/csrrw.yaml
@@ -13,7 +13,7 @@ description: |
   If `xd=x0`, then the instruction shall not read the CSR and shall not
   cause any of the side effects that might occur on a CSR read.
 definedBy: Zicsr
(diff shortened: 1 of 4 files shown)


arch/inst/Zicsr/csrrw.yaml
```diff
@@ -13,7 +13,7 @@ description: |
   If `xd=x0`, then the instruction shall not read the CSR and shall not
   cause any of the side effects that might occur on a CSR read.
 definedBy: Zicsr
-assembly: xd, imm, xs1
+assembly: xd, csr, imm
```
jordancarlin, line 16: `csrrw` writes a register value to the desired CSR, not an immediate value.
```suggestion
assembly: xd, csr, xs1
```

dhower-qc (approved):

PR #810, opened 2025-05-23
Add spec to FP Computational and Conversion Instructions

sudo-abdullah:
This PR adds missing long_name and description to some FP Computational and remaining FP Conversion Instructions.

### DIFF
--- a/arch/inst/D/fadd.d.yaml
+++ b/arch/inst/D/fadd.d.yaml
@@ -4,12 +4,12 @@ $schema: inst_schema.json#
 kind: instruction
 name: fadd.d
 long_name: Floating-point Add Double-precision
-description: |
-  -id: inst-fadd.d-behaviour
-  -normative: false
-  -text: |
-    `fadd.d` is analogous to `fadd.s` and performs double-precision floating-point addition between
-    `xs1` and `xs2` and writes the final result to `xd`.
+description:
+  - id: inst-fadd.d-behaviour
+    normative: false
+    text: |
+      `fadd.d` is analogous to `fadd.s` and performs double-precision floating-point addition between
+      `xs1` and `xs2` and writes the final result to `xd`.
 definedBy: D
 assembly: xd, xs1, xs2, rm
 encoding:
--- a/arch/inst/D/fcvt.d.l.yaml
+++ b/arch/inst/D/fcvt.d.l.yaml
@@ -4,12 +4,12 @@ $schema: inst_schema.json#
 kind: instruction
 name: fcvt.d.l
 long_name: Floating-point Convert Long to Double-precision
-description: |
-  -id: inst-fcvt.d.l-behaviour
-  -normative: false
-  -text: |
-     `fcvt.d.l` converts a 64-bit signed integer, in integer register `xs1` into a double-precision
-     floating-point number in floating-point register `fd`.
+description:
+  - id: inst-fcvt.d.l-behaviour
+    normative: false
+    text: |
+      `fcvt.d.l` converts a 64-bit signed integer, in integer register `xs1` into a double-precision
+      floating-point number in floating-point register `fd`.
 definedBy: D
 assembly: xd, xs1, rm
 encoding:
--- a/arch/inst/D/fcvt.d.lu.yaml
+++ b/arch/inst/D/fcvt.d.lu.yaml
@@ -4,12 +4,12 @@ $schema: inst_schema.json#
 kind: instruction
 name: fcvt.d.lu
 long_name: Floating-point Convert Unsigned Long to Double-precision
-description: |
-  -id: inst-fcvt.d.lu-behaviour
-  -normative: false
-  -text: |
-     `fcvt.d.lu` converts to or from a 64-bit unsigned integer, `xs1` into a double-precision
-     floating-point number in floating-point register `fd`.
+description:
+  - id: inst-fcvt.d.lu-behaviour
+    normative: false
+    text: |
+      `fcvt.d.lu` converts to or from a 64-bit unsigned integer, `xs1` into a double-precision
+      floating-point number in floating-point register `fd`.
 definedBy: D
 assembly: xd, xs1, rm
 encoding:
--- a/arch/inst/D/fcvt.d.s.yaml
+++ b/arch/inst/D/fcvt.d.s.yaml
@@ -4,14 +4,14 @@ $schema: inst_schema.json#
 kind: instruction
 name: fcvt.d.s
 long_name: Floating-point Convert Single-precision to Double-precision
-description: |
-  -id: inst-fcvt.d.s-behaviour
-  -normative: false
-  -text: |
-    The single-precision to double-precision conversion instruction, `fcvt.d.s` is encoded in the OP-FP
-    major opcode space and both the source and destination are floating-point registers. The `xs2` field
-    encodes the datatype of the source, and the `fmt` field encodes the datatype of the destination.
-    `fcvt.d.s` will never round.
+description:
+  - id: inst-fcvt.d.s-behaviour
+    normative: false
+    text: |
+      The single-precision to double-precision conversion instruction, `fcvt.d.s` is encoded in the OP-FP
+      major opcode space and both the source and destination are floating-point registers. The `xs2` field
+      encodes the datatype of the source, and the `fmt` field encodes the datatype of the destination.
+      `fcvt.d.s` will never round.
 definedBy: D
 assembly: xd, xs1, rm
 encoding:
--- a/arch/inst/D/fcvt.d.w.yaml
+++ b/arch/inst/D/fcvt.d.w.yaml
@@ -4,13 +4,13 @@ $schema: inst_schema.json#
 kind: instruction
 name: fcvt.d.w
 long_name: Floating-point Convert Word to Double-precision
-description: |
-  -id: inst-fcvt.d.w-behaviour
-  -normative: false
-  -text: |
-    `fcvt.d.w` converts a 32-bit signed integer, in integer register `xs1` into a double-precision
-    floating-point number in floating-point register `fd`.
-    Note `fcvt.d.w` always produces an exact result and is unaffected by rounding mode.
+description:
+  - id: inst-fcvt.d.w-behaviour
+    normative: false
+    text: |
+      `fcvt.d.w` converts a 32-bit signed integer, in integer register `xs1` into a double-precision
+      floating-point number in floating-point register `fd`.
+      Note `fcvt.d.w` always produces an exact result and is unaffected by rounding mode.
 definedBy: D
 assembly: xd, xs1, rm
 encoding:
--- a/arch/inst/D/fcvt.d.wu.yaml
+++ b/arch/inst/D/fcvt.d.wu.yaml
@@ -4,13 +4,13 @@ $schema: inst_schema.json#
 kind: instruction
 name: fcvt.d.wu
 long_name: Floating-point Convert Unsigned Word to Double-precision
-description: |
-  -id: inst-fcvt.d.wu-behaviour
-  -normative: false
-  -text: |
-    `fcvt.d.wu` converts a 32-bit unsigned integer in integer register `fs1` into a double-precision
-    floating-point number in floating-point register `fd`.
-    Note `fcvt.d.wu` always produces an exact result and is unaffected by rounding mode.
+description:
+  - id: inst-fcvt.d.wu-behaviour
+    normative: false
+    text: |
+      `fcvt.d.wu` converts a 32-bit unsigned integer in integer register `fs1` into a double-precision
+      floating-point number in floating-point register `fd`.
+      Note `fcvt.d.wu` always produces an exact result and is unaffected by rounding mode.
 definedBy: D
 assembly: xd, xs1, rm
 encoding:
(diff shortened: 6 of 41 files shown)


arch/inst/Zfh/fcvt.h.lu.yaml
```diff
@@ -3,9 +3,12 @@
 $schema: inst_schema.json#
 kind: instruction
 name: fcvt.h.lu
-long_name: No synopsis available
+long_name: Half-Precision Floating-Point 64-bit Unsigned Integer Conversion Instruction
```
AFOliveira, line 6: Shouldn't this be "Half-Precision Floating-Point to Long Unsigned Integer Conversion"?

Saying instruction in the instruction name seems clearly a no-go.

About the 64-bit vs Long, I think it's better to define long, but other reviewrs may differ opinion.

arch/inst/Zfh/fcvt.h.lu.yaml
```diff
@@ -3,9 +3,12 @@
...
 description: |
-  No description available.
+  -id: inst-fcvt.h.lu-behaviour
+  -normative: false
```
AFOliveira, line 9: I don't think this is a requirement to enter, but since you specified, I believe all of this is normative text.

BTW, Normative text specifies the binding rules and requirements that implementations must follow (using terms like “MUST,” “SHALL,” or “SHOULD”), whereas non-normative text is purely informational—providing background, examples, rationale, or guidance without imposing any mandatory obligations on implementers. Normative clauses form the “contract” of a specification, and violating them means non-compliance, while non-normative sections merely aid understanding and carry no compliance weight.

arch/inst/Zfh/fcvt.h.l.yaml
```diff
@@ -3,9 +3,12 @@
 $schema: inst_schema.json#
 kind: instruction
 name: fcvt.h.l
-long_name: No synopsis available
+long_name: Half-Precision Floating-Point 64-bit Signed Integer Conversion Instruction
```
AFOliveira, line 6: same

arch/inst/Q/fcvt.d.q.yaml
```diff
@@ -3,9 +3,12 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: fcvt.d.q
-long_name: No synopsis available
+long_name: Quad to Double Precision Conversion
```
AFOliveira, line 6: missing "-"

arch/inst/D/fcvt.d.wu.yaml
```diff
@@ -3,9 +3,15 @@
 $schema: inst_schema.json#
 kind: instruction
 name: fcvt.d.wu
-long_name: No synopsis available
+long_name: Double-Precision to 32-bit Unsigned Integer Conversion Instruction
```
AFOliveira, line 6: same

AFOliveira (changes requested):
Great work! 

Just a couple things to go:
Saying instruction in the instruction name seems clearly a no-go. Please remove from all the instructions long-names.

About the 64-bit vs Long, I think it's better to define long, but other reviewers may differ opinion.

arch/inst/D/fcvt.d.w.yaml
```diff
@@ -3,9 +3,14 @@
 $schema: inst_schema.json#
 kind: instruction
 name: fcvt.d.w
-long_name: No synopsis available
+long_name: Double-Precision to 32-bit Signed Integer Conversion
```
ThinkOpenly, line 6: Floating-point convert Word to Double-precision

ThinkOpenly (changes requested):
To my understanding, you have a lot of the "from" and "to" types reversed. I've suggested new text that reads better left-to-right with the mnemonic string and reverses those types.

sudo-abdullah:
> Aren't all the descriptions normative?

As per @ThinkOpenly's guidance :
Only descriptions that are exact matches to normative spec text should be marked as `normative: true`.
Paraphrased or simplified descriptions, even if accurate, should not be marked normative.

ThinkOpenly (changes requested):

arch/inst/D/fcvt.d.s.yaml
```diff
@@ -3,9 +3,15 @@
 $schema: inst_schema.json#
 kind: instruction
 name: fcvt.d.s
-long_name: No synopsis available
+long_name: Floating-point Convert Double-precision to Single-precision
```
ThinkOpenly: This should be:
> Floating-point Convert Single-precision to Double-precision

arch/inst/D/fcvt.d.s.yaml
```diff
@@ -3,9 +3,15 @@
...
 description: |
-  No description available.
+  -id: inst-fcvt.d.s-behaviour
+  -normative: false
+  -text: |
+    The double-precision to single-precision conversion instruction, `fcvt.d.s` is encoded in the OP-FP
```
ThinkOpenly: This needs to be reversed, too:
> The single-precision to double-precision conversion instruction

arch/inst/D/fcvt.d.l.yaml
```diff
@@ -3,9 +3,13 @@
...
 name: fcvt.d.l
-long_name: No synopsis available
+long_name: Floating-point Convert Long to Double-precision
 description: |
-  No description available.
+  -id: inst-fcvt.d.l-behaviour
+  -normative: false
+  -text: |
+     `fcvt.d.l` converts a 64-bit signed integer, in integer register `xs1` into a double-precision
+     floating-point number in floating-point register `xd`.
```
ThinkOpenly: I did mess this up. The target register is floating-point, so change `xd` to `fd`. Sorry.

arch/inst/D/fcvt.d.lu.yaml
```diff
@@ -3,9 +3,13 @@
...
 name: fcvt.d.lu
-long_name: No synopsis available
+long_name: Floating-point Convert Unsigned Long to Double-precision
 description: |
-  No description available.
+  -id: inst-fcvt.d.lu-behaviour
+  -normative: false
+  -text: |
+     `fcvt.d.lu` converts to or from a 64-bit unsigned integer, `xs1` into a double-precision
+     floating-point number in floating-point register `xd`.
```
ThinkOpenly: Here, too: change `xd` to `fd`.

arch/inst/D/fcvt.d.w.yaml
```diff
@@ -3,9 +3,14 @@
...
 name: fcvt.d.w
-long_name: No synopsis available
+long_name: Floating-point Convert Word to Double-precision
 description: |
-  No description available.
+  -id: inst-fcvt.d.w-behaviour
+  -normative: false
+  -text: |
+    `fcvt.d.w` converts a 32-bit signed integer, in integer register `xs1` into a double-precision
+    floating-point number in floating-point register `xd`.
```
ThinkOpenly: Change `xd` to `fd`.

arch/inst/D/fcvt.d.wu.yaml
```diff
@@ -3,9 +3,15 @@
...
-long_name: No synopsis available
+long_name: Floating-point Convert Unsigned Word to Double-precision
 description: |
-  No description available.
+  -id: inst-fcvt.d.wu-behaviour
+  -normative: false
+  -text: |
+    This variant of the `fcvt` instruction converts to or from unsignedn integer values.
+    `fcvt.d.wu` converts a 32-bit unsigned integer in integer register `xs1` into a double-precision
+    floating-point number in floating-point register `xd`.
```
ThinkOpenly: `fd`

dhower-qc:
> (I'm content at this point to let correct text get merged, absent glaring omissions. Verbatim spec content is nice, but not required, and also difficult to contextualize when extracted. My criteria is "good enough". The text battles come later.)

Agreed; I'm assuming there will be a deep pass of all text down the road, so it's not critical at this point.

arch/inst/D/fcvt.l.d.yaml
```diff
@@ -3,9 +3,13 @@
...
 kind: instruction
 name: fcvt.l.d
-long_name: No synopsis available
+long_name: Floating-point Convert Double-precision to Long
 description: |
-  No description available.
+  -id: inst-fcvt.l.d-behaviour
+  -normative: false
+  -text: |
+    `fcvt.l.d` converts a double-precision floating-point number in floating-point register `xs1`
```
ThinkOpenly: change `xs1` to `fs1`

arch/inst/D/fcvt.wu.d.yaml
```diff
@@ -3,9 +3,13 @@
 $schema: inst_schema.json#
 kind: instruction
 name: fcvt.wu.d
-long_name: No synopsis available
+long_name: Floating-point Convert Double-precision to Word
```
ThinkOpenly:
```suggestion
long_name: Floating-point Convert Double-precision to Unsigned Word
```

arch/inst/D/fcvt.d.l.yaml
sudo-abdullah: > I did mess this up. The target register is floating-point, so change `xd` to `fd`. Sorry.

But the assembly says otherwise:
> assembly: xd, xs1, rm

ThinkOpenly: > > I did mess this up. The target register is floating-point, so change `xd` to `fd`. Sorry.
> 
> But the assembly says otherwise:
> 
> > assembly: xd, xs1, rm

Yeah, the assembly is also incorrect, but that's being fixed in PR #822.

arch/inst/D/fcvt.l.d.yaml
sudo-abdullah: All of 'xs1' and 'xs2' must be replaced by their FP counterparts everywhere, right?

ThinkOpenly: Since you mentioned `xs2`, and there is no `xs2` in this particular file, I presume your question is more global to the PR. If so, then the answer is "no".

Which register names are used depends on context. An instruction like `fcvt.X.Y` converts a value of Y type in source register to a different type X in destination register. An instruction like `fadd.X` operates on three registers of type X. The types and registers are:
| Type short name | Type | Register prefix |
|--|--|--|
| w(u) | 32-bit (unsigned) integer | x |
| l(u) | 64-bit (unsigned) integer | x |
| s | 32-bit floating-point | f |
| d | 64-bit floating-point | f |
| q | 128-bit floating-point | f |

arch/inst/Q/fcvt.q.l.yaml
```diff
@@ -3,9 +3,12 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: fcvt.q.l
-long_name: No synopsis available
+long_name: Floating-point Convert Long integer to Quad-precision
```
ThinkOpenly:
```suggestion
long_name: Floating-point Convert Long to Quad-precision
```

arch/inst/Zfh/fcvt.h.wu.yaml
```diff
@@ -3,9 +3,12 @@
 $schema: inst_schema.json#
 kind: instruction
 name: fcvt.h.wu
-long_name: No synopsis available
+long_name: Floating-point Convert unsigned Word to Half-precision
```
ThinkOpenly:
```suggestion
long_name: Floating-point Convert Unsigned Word to Half-precision
```

ThinkOpenly:
For that one CI test that is failing, you need to regenerate the "golden reference" instruction appendix:
```
$ ./do gen:instruction_appendix
$ mv gen/instructions_appendix/all_instructions.adoc backends/instructions_appendix/all_instructions.golden.adoc
$ git add backends/instructions_appendix/all_instructions.golden.adoc
```
Then commit as you see fit and push the changes here.

sudo-abdullah:
> For that one CI test that is failing, you need to regenerate the "golden reference" instruction appendix:
> 
> ```
> $ ./do gen:instruction_appendix
> $ mv gen/instructions_appendix/all_instructions.adoc backends/instructions_appendix/all_instructions.golden.adoc
> $ git add backends/instructions_appendix/all_instructions.golden.adoc
> ```
> 
> Then commit as you see fit and push the changes here.

I tried it, but it didn't generate "golden reference" in the directory as mentioned in move command

AFOliveira:
What did you try and what did not generate the golden reference?

sudo-abdullah:
I tried to generate golden reference as Mr. Paul has mentioned and it went smooth but upon running move command, I noticed that it was not in the directory `gen/instructions_appendix/` but `all_instructions.golden.adoc` was already in `backends/instructions_appendix/`.

AFOliveira:
The objective is to move the generated adoc from the instructions appendix and update the already existing `all_instructions.golden.adoc`. Try first just building the appendix and locating the adoc. It should be under `gen/instructions_appendix`. If it is there, you gotta move it to the other file and replace it.

arch/inst/D/fadd.d.yaml
```diff
@@ -3,9 +3,13 @@
 $schema: inst_schema.json#
 kind: instruction
 name: fadd.d
-long_name: No synopsis available
+long_name: Floating-point Add Double-precision
 description: |
```
dhower-qc: All of these descriptions need to be turned into arrays. To do so, remove the `|` and use a _single_ dash:

```yaml
description:
  - id: inst-fadd.d-behaviour
    normative: false
    text: |
      `fadd.d` is analogous to `fadd.s` and performs double-precision floating-point addition between
      `xs1` and `xs2` and writes the final result to `xd`.
```

ThinkOpenly (approved):

PR #838, opened 2025-06-10
fix(data): vsetvli/vsetivli: correct assembly and fields

ThinkOpenly:
`vsetvli` and `vsetivli` accept the _components_ of VTYPEI as
individual parameters, and they are encoded in individual fields.
    
...with a few additional `0` bits as a result, per the
definition of vtype.

### DIFF
--- a/arch/inst/V/vsetivli.yaml
+++ b/arch/inst/V/vsetivli.yaml
@@ -7,12 +7,18 @@ long_name: No synopsis available
 description: |
   No description available.
 definedBy: V
-assembly: xd, uimm, vtypei
+assembly: xd, uimm, sew, lmul, ta, ma
 encoding:
-  match: 11---------------111-----1010111
+  match: 1100-------------111-----1010111
   variables:
-    - name: vtypei
-      location: 29-20
+    - name: ma
+      location: 27
+    - name: ta
+      location: 26
+    - name: sew
+      location: 25-23
+    - name: lmul
+      location: 22-20
     - name: uimm
       location: 19-15
     - name: xd
--- a/arch/inst/V/vsetvli.yaml
+++ b/arch/inst/V/vsetvli.yaml
@@ -7,12 +7,18 @@ long_name: No synopsis available
 description: |
   No description available.
 definedBy: V
-assembly: xd, xs1, vtypei
+assembly: xd, xs1, sew, lmul, ta, ma
 encoding:
-  match: 0----------------111-----1010111
+  match: 0000-------------111-----1010111
   variables:
-    - name: xs1
-      location: 30-20
+    - name: ma
+      location: 27
+    - name: ta
+      location: 26
+    - name: sew
+      location: 25-23
+    - name: lmul
+      location: 22-20
     - name: xs1
       location: 19-15
     - name: xd
--- a/backends/instructions_appendix/all_instructions.golden.adoc
+++ b/backends/instructions_appendix/all_instructions.golden.adoc
@@ -1,6 +1,6 @@
 = Instruction Appendix
 :doctype: book
-:wavedrom: /workspaces/riscv-unified-db/node_modules/.bin/wavedrom-cli
+:wavedrom: /workspace/riscv-unified-db/node_modules/.bin/wavedrom-cli
 // Now the document header is complete and the wavedrom attribute is active.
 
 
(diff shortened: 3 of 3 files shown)


jordancarlin:
Some additional discussion over in this Sail PR: https://github.com/riscv/sail-riscv/pull/1010.

For valid values, the 4-component approach is used, but expressing them as a single immediate is also accepted by the assemblers and is the only way to express reserved values (which in this case does not mean that the instruction itself is reserved).

ThinkOpenly:
For this sequence of instructions:
```
        vsetvli t0, a0, 192
        vsetvli t0, a0, e8, m1, ta, ma
        vsetvli t0, a0, e8, ta, ma
```
GCC emits the same encoding for each.
LLVM (clang 20 2b9fc0f1c82e514f8b46c4c7d282edecb0346089) does not respect the optional "lmul" operand:
```
$ clang -c vsetvli.s
vsetvli.s:21:25: error: operand must be e[8|16|32|64],m[1|2|4|8|f2|f4|f8],[ta|tu],[ma|mu]
        vsetvli t0, a0, e8, ta, ma
                        ^
```

lenary:
What's going on here is that LLVM has a custom parser for the vtype information: https://github.com/llvm/llvm-project/blob/b3873e8aa41be17b6664c918fd6ba30307361d08/llvm/lib/Target/RISCV/AsmParser/RISCVAsmParser.cpp#L2295 

I haven't looked closely at how `parseVTypeToken` works, but you have an example that GCC accepts and LLVM does not, and you have a reason you think LLVM is wrong, so it would be good to file that against LLVM: http://github.com/llvm/llvm-project/issues (link back to this in case someone wants to see where you found the issue) 

I would expect that LLVM will want to continue to bundle all the vsetvli operands into a single immediate, but I could be wrong, so this might be somewhere where we need a little gap between the "asmstring" in LLVM (which is using a more complex parser underneath), and the unified db information, when doing the cross-verification.

AFOliveira:
Thankfully, @lenary arrived to the issue earlier than me and clarified much better than what I could have :)

I see two ways of going forward here:

If @ThinkOpenly opens that issue against LLVM:
 - I can skip this test in the llvm tests for now
 - We wait to see if that gets fixed

I don't quite know how fast LLVM is in fixing this if they agree with our view here, but I can really just fix this in couple minutes if you want me to

ThinkOpenly:
> What's going on here is that LLVM has a custom parser for the vtype information: https://github.com/llvm/llvm-project/blob/b3873e8aa41be17b6664c918fd6ba30307361d08/llvm/lib/Target/RISCV/AsmParser/RISCVAsmParser.cpp#L2295

I see. Is there anything in the TableGen JSON that indicates that this instruction has a custom parser, or is weird in some way? (I didn't see anything obvious.) I'm just looking for a programmatic way to separate these two instructions from the herd.

> file that against LLVM: [http://github.com/llvm/llvm-project/issues](https://github.com/llvm/llvm-project/issues) (link back to this in case someone wants to see where you found the issue)

https://github.com/llvm/llvm-project/issues/143744

> I would expect that LLVM will want to continue to bundle all the vsetvli operands into a single immediate, but I could be wrong, so this might be somewhere where we need a little gap between the "asmstring" in LLVM (which is using a more complex parser underneath), and the unified db information, when doing the cross-verification.

UDB is going to need to accommodate all 3 different syntax possibilities somehow. There is no way to do that at present.

Validating those three against the TableGen JSON is perhaps not possible.

topperc:
The intent of the vector spec is that the LMUL is not optional. It was explicitly removed here https://github.com/riscv/riscv-isa-manual/commit/75e857da51db157814f67e465c7262ae0a74c277

topperc:
> The intent of the vector spec is that the LMUL is not optional. Text saying it was optional was explicitly removed here [riscv/riscv-isa-manual@75e857d](https://github.com/riscv/riscv-isa-manual/commit/75e857da51db157814f67e465c7262ae0a74c277)

Unfortunately, this removal happened years after ratification. So I guess we might need to support it be compatible with the ratified spec prior to this change?

ThinkOpenly:
> If @ThinkOpenly opens that issue against LLVM:
> 
> * I can skip this test in the llvm tests for now
> * We wait to see if that gets fixed
> 
> I don't quite know how fast LLVM is in fixing this if they agree with our view here, but I can really just skip those test for now in a couple minutes if you want me to

I don't see that a fix coming from LLVMland that would help, given our current use of the JSON.

I think Sam's awesomely detailed overview of how to interpret the JSON could be used to intelligently skip this class of cases in general:
1. look up the instruction as is already done
2. walk through the "InOperandList":"args", for each "def"...
3. look up each "def" value at the top level, I think
4. then look up its "ParserMatchClass":"def" (this is the "anonymous" thing") also at top level
5. see if that has a "ParserMethod" attribute. If so, skip.

PR #840, opened 2025-06-11
feat: mock-up of rules-based B extension

dhower-qc:


dhower-qc:
This effort is being spearheaded in the CSC now, this proof of concept can close

PR #854, opened 2025-06-20
andn: use the new instruction schema

Unique-Usman:
for types/subtypes. This commit change B/andn
instruction to use the new schema. It is also part of effort in (#655).


Unique-Usman:
@dhower-qc @ThinkOpenly - Take a look a this.

ThinkOpenly (approved):

PR #855, opened 2025-06-20
fix: paths for CTP, isa_explorer, and deploy

dhower-qc:
Should fix the deployment

### DIFF
--- a/backends/isa_explorer/isa_explorer.rb
+++ b/backends/isa_explorer/isa_explorer.rb
@@ -1,8 +1,12 @@
+
+# typed: true
 # frozen_string_literal: true
-#
+
 # Generate
 
+require "sorbet-runtime"
 require "write_xlsx"
+
 require "udb/architecture"
 
 # @param presence [String] Can be nil
@@ -85,9 +89,8 @@ end
 
 # @param arch [Architecture] The entire RISC-V architecture
 # @return [Hash<String,Array<String>] Instruction table data
+sig { params(arch: Udb::Architecture).returns(T::Hash[String, T::Array[String]]) }
 def arch2inst_table(arch)
-  raise ArgumentError, "arch is a #{arch.class} class but needs to be Architecture" unless arch.is_a?(Architecture)
-
   sorted_profile_releases = get_sorted_profile_releases(arch)
 
   inst_table = {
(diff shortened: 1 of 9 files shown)


ThinkOpenly:
> Should fix the deployment

Maybe when it passes CI? ;-)

dhower-qc:
As an aside: the coverage check isn't running. I think I need Rafael (with riscv-software-src admin privilege) to click a button for us. Looking into it...

backends/isa_explorer/isa_explorer.rb
```diff
@@ -85,9 +89,8 @@ def arch2ext_table(arch)
 
 # @param arch [Architecture] The entire RISC-V architecture
 # @return [Hash<String,Array<String>] Instruction table data
+sig { params(arch: Udb::Architecture).returns(T::Hash[String, T::Array[String]]) }
 def arch2inst_table(arch)
-  raise ArgumentError, "arch is a #{arch.class} class but needs to be Architecture" unless arch.is_a?(Architecture)
```
james-ball-qualcomm, line 89: Does this is_a?(Architecture) need to be Udb::Architecture? Alternatively, is there a way to tell Ruby that references to Architecture should automatically use Udb?

backends/isa_explorer/isa_explorer.rb
```diff
@@ -134,9 +137,8 @@ def arch2inst_table(arch)
 
 # @param arch [Architecture] The entire RISC-V architecture
 # @return [Hash<String,Array<String>] CSR table data
+sig { params(arch: Udb::Architecture).returns(T::Hash[String, T::Array[String]]) }
 def arch2csr_table(arch)
-  raise ArgumentError, "arch is a #{arch.class} class but needs to be Architecture" unless arch.is_a?(Architecture)
```
james-ball-qualcomm, line 138: Ditto

backends/isa_explorer/tasks.rake
```diff
@@ -81,7 +81,7 @@ file "#{GEN_XLSX}" => [
     __FILE__,
     src_pnames
 ].flatten do |t|
-    arch = create_arch
+    arch = $resolver.cfg_arch_for("_")
```
james-ball-qualcomm, line 84: Glad to see the $resolver. Seems to remove hardwired paths.

james-ball-qualcomm (commented):
Looks good to me!

backends/isa_explorer/isa_explorer.rb
dhower-qc: This is a removed line; it's not needed because Sorbet checks it. In fact, if you leave it in Sorbet type checking fails because it knows that's an unreachable statement.

Re: Architecture vs. Udb::Architecture; yes, it would need to be Udb::Architecture now. If you want to avoid that, you need to wrap the isa explorer functions in the Udb module.

backends/isa_explorer/tasks.rake
dhower-qc: yes, should be easier to work with now.

backends/isa_explorer/isa_explorer.rb
ThinkOpenly: This PR is already merged, but I've already submitted a PR #856 to address these issues. (Reviews welcome!)

PR #866, opened 2025-06-23
865 Improve ISA Explorer extension dependency columns - Implied By changed to Implies

james-ball-qualcomm:
Made changes with Derek's help

image

### DIFF
--- a/spec/std/isa/ext/F.yaml
+++ b/spec/std/isa/ext/F.yaml
@@ -14,6 +14,7 @@ versions:
     ratification_date: 2019-12
     changes:
       - Define NaN-boxing scheme, changed definition of FMAX and FMIN
+    requires: Zicsr
 description: |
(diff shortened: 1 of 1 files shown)


james-ball-qualcomm:
I think you are asking me to fix F.yaml and add a dependency on Zicsr. Is that correct? If so, I'll do it but it really is a separate issue.

Here's the current F.yaml:
```
$schema: "ext_schema.json#"
kind: extension
name: F
type: unprivileged
long_name: Single-precision floating-point
versions:
  - version: "2.2.0"
    state: ratified
    ratification_date: 2019-12
    changes:
      - Define NaN-boxing scheme, changed definition of FMAX and FMIN

dhower-qc:
> Added "requires: Zicsr" to F.yaml. Here's the ISA explorer output:
> 
> image

Excellent, thanks.

PR #871, opened 2025-06-25
Xqci: Fix unintended overflows

AntonJohansson:
This commit fixes two kinds of bugs:

1. Xreg width = width_minus1 + 1, where width_minus1 is a Bits<N> type with N smaller than 32. Since the C literal "1" is 1 bit in size it will be extended to N bits and the N-bit addition might overflow. The range of width is then [1,31] instead of [1,32] as intended.

2. (32'b1 << width)-1, when translated to C++ results in undefined behaviour if width == 32.

Bug 1. is fixed by using a widening add and 2. by swapping 32'b1 -> 33'b1. An alternative for 2. is computing the mask via {32{1}} >> (32-width).

Is `32'b1 << N` where `N >= 32` well defined in the IDL? If so 2. is not a problem.

What is the rationale for having C-style literals be represented by as few bits as possible, whereas Verilog-style literals are MXLEN when no explicit size is given? Why not have both be MXLEN in size? I feel this will lead to over-use of widening operations or a lot of unintended bugs when using C literals.

### DIFF
--- a/spec/custom/isa/qc_iu/inst/Xqci/qc.c.extu.yaml
+++ b/spec/custom/isa/qc_iu/inst/Xqci/qc.c.extu.yaml
@@ -32,5 +32,5 @@ access:
   vs: always
   vu: always
 operation(): |
-  XReg width = width_minus1 + 1;
-  X[rd] = (X[rd] >> 0) & ((32'b1 << width) - 1);
+  XReg width = width_minus1 `+ 1;
+  X[rd] = (X[rd] >> 0) & ((33'b1 << width) - 1);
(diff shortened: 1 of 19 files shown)


dhower-qc:
> Is 32'b1 << N where N >= 32 well defined in the IDL? If so 2. is not a problem.

It is well-defined, and should always give you zero. Looking at the code, I see that is true in the Ruby evaluation but may not be true in the C++ evaluation. I will correct the C++.

> What is the rationale for having C-style literals be represented by as few bits as possible, whereas Verilog-style literals are MXLEN when no explicit size is given?

Fair question. To be honest, I don't remember the reasoning. Which probably means it wasn't a very good reason.

I'm happy to make this change; I don't think it will be intrusive to any of our existing code.

@ThinkOpenly, any opinions?

AntonJohansson:
If the decision is to make C-style literals MXLEN-sized, I'll can go ahead and close this!:)

PR #872, opened 2025-06-25
fix(data): fix field names to be closer to operands

ThinkOpenly:
- `dret`, `mnret`, and `sctrclr` take no operands.
- ~~`sspush` and `sspopchk` are each single mnemonics that take a very
  restricted set of operand values.
  Also include a bit of documentation.~~
- `vaeskf1.vi`, `vaeskf2.vi`, `vsm3c.vi`, `vsm4k.vi`, `vwsll.vi`:
  renamed field to match operand.

### DIFF
--- a/backends/instructions_appendix/all_instructions.golden.adoc
+++ b/backends/instructions_appendix/all_instructions.golden.adoc
@@ -1,6 +1,6 @@
 = Instruction Appendix
 :doctype: book
-:wavedrom: /workspaces/riscv-unified-db/node_modules/.bin/wavedrom-cli
+:wavedrom: /workspace/riscv-unified-db/node_modules/.bin/wavedrom-cli
 // Now the document header is complete and the wavedrom attribute is active.
 
 
@@ -7145,9 +7145,6 @@ Included in::
 
 Synopsis::
 No synopsis available
-
-Assembly::
-dret dret
 
 Encoding::
 [wavedrom, ,svg,subs='attributes',width="100%"]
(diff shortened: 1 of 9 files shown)


ThinkOpenly:
@dhower-qc , any clue why the CI tests are failing? They fail without the updated instruction appendix commit as well, so just the first commit is a problem, but there's not a ton there... some deleted and created files causing issues?

BTW, the "prettier" pre-commit hook made me make the list of unacceptable values for `sspush`/`sspopchk` uglier. (I had them on a single line, but it didn't like that.)

dhower-qc:
The error message is a big cryptic because it manifests as a Sorbet type error, but it's caused by the fact a $ref wasn't resolved. Presumably, that happens because you removed a file that was the target of a ref.

dhower-qc:
Back to your changes, I suspect you are seeing this because the instructions you are changing are "hints" of may-be-ops. So there is going to be a reference in their definition that needs to be updated.

ThinkOpenly:
I've removed the `sspopchk` and `sspush` changes. I think these need some additional changes in how the decode tree is constructed, because they both could map nominally to the same "match" string:
```
sspopchk: 110011011100-----100000001110011
sspush  : 1100111-----00000100000001110011
```
...if the `not` constraint is ignored, as the first string's variable must be 1 or 5, not 0 (`00000`) and the second string's variable must be 1 or 5, not 28 (`11100`).

I'll open an issue.

@dhower-qc, could you review this PR?

PR #873, opened 2025-06-26
perf(ruby): make merged instruction appendix a direct target

ThinkOpenly:
Currently, producing the "merged instruction" .adoc file also builds the dependent PDF file. This can take some time.

Since updating the "golden" reference instruction appendix .adoc file only requires building a new .adoc file, and this operation is relatively common, make that its own target. This builds much faster.

Also, remove references to the "ASSEMBLY" variable, support for which was removed in #867.

### DIFF
--- a/backends/instructions_appendix/tasks.rake
+++ b/backends/instructions_appendix/tasks.rake
@@ -56,28 +56,39 @@ namespace :gen do
   task instruction_appendix_adoc: MERGED_INSTRUCTIONS_FILE.to_s
 
   desc <<~DESC
-    Generate the instruction appendix (merged .adoc and PDF)
+    Generate the instruction appendix (.adoc)
 
-    By default this will produce the “merged instructions” AsciiDoc file and
-    then render it to PDF.
-
-    Environment flags:
-
-     * ASSEMBLY - set to `1` to include an “Assembly” line (instruction mnemonic + operands)
-                  before the Encoding section for each instruction.
+    This will produce the “merged instructions” AsciiDoc file.
 
     Examples:
 
-     # Just regenerate AsciiDoc + PDF:
-     $ do gen:instruction_appendix
+     $ do gen:instruction_appendix_adoc
 
-     # Include assembly templates in the docs:
-     $ do gen:instruction_appendix ASSEMBLY=1
+  DESC
+  task :instruction_appendix_adoc do
+    # Generate the merged instructions adoc.
+    Rake::Task[MERGED_INSTRUCTIONS_FILE.to_s].invoke
+    puts "SUCCESS: Instruction appendix adoc generated at '#{MERGED_INSTRUCTIONS_FILE}'"
+  end
+end
+
+namespace :gen do
+  task instruction_appendix: MERGED_INSTRUCTIONS_FILE.to_s
+
+  desc <<~DESC
+    Generate the instruction appendix (merged .adoc and PDF)
+
+    This will produce the “merged instructions” AsciiDoc file and
+    then render it to PDF.
+
+    Examples:
+
+     $ do gen:instruction_appendix
 
   DESC
   task :instruction_appendix do
     # Generate the merged instructions adoc.
-    Rake::Task[MERGED_INSTRUCTIONS_FILE.to_s].invoke
+    Rake::Task["gen:instruction_appendix_adoc"].invoke
     # Then generate the PDF.
     Rake::Task[MERGED_INSTRUCTIONS_PDF.to_s].invoke
     puts "SUCCESS: Instruction appendix generated at '#{MERGED_INSTRUCTIONS_FILE}' and PDF at '#{MERGED_INSTRUCTIONS_PDF}'"


backends/instructions_appendix/tasks.rake
```diff
@@ -56,28 +56,39 @@ namespace :gen do
   task instruction_appendix_adoc: MERGED_INSTRUCTIONS_FILE.to_s
```
dhower-qc: I'm a little confused: On line 56, the `gen:instruction_appendix_adoc` is already defined (and it's what the GitHub regression is calling). It's not documented, though. Maybe that's what you are trying to do? If so, just remove line 56.

dhower-qc (changes requested):

dhower-qc:
Since you are in here, any chance you want to add a `chore:update_golden_iappendix` task?

backends/instructions_appendix/tasks.rake
ThinkOpenly: I'm sure I'm showing my ignorance, but maybe I haven't quite grok'd the structure of the task/desc blocks, and how they are glued together.

My goal is splitting the `gen:instruction_appendix` task into two tasks, one to generate just the `.adoc`, and one to generate both `.adoc` and PDF.

Given the original structure:
```
namespace :gen do
  task instruction_appendix_adoc: MERGED_INSTRUCTIONS_FILE.to_s
  desc <<~DESC
[...]
  DESC
  task :instruction_appendix do
[...]
  end
end
```
What's the purpose of the first `task` line?

AFOliveira: I believe what @dhower-qc is pointing out is that UDB already supports the behavior you’re looking for. It doesn’t appear in the top-level tasks simply because it has no description—only items with a description are promoted to appear in ./do --tasks

backends/instructions_appendix/tasks.rake
```diff
@@ -56,28 +56,39 @@ namespace :gen do
...
+    Rake::Task[MERGED_INSTRUCTIONS_FILE.to_s].invoke
+    puts "SUCCESS: Instruction appendix adoc generated at '#{MERGED_INSTRUCTIONS_FILE}'"
+  end
+end
 
-    Environment flags:
+namespace :gen do
+  task instruction_appendix: MERGED_INSTRUCTIONS_FILE.to_s
 
-     * ASSEMBLY - set to `1` to include an “Assembly” line (instruction mnemonic + operands)
```
AFOliveira: Why are we removing assembly? I think the functionality of generating the instruction appendix with or without appendix is still useful, or not really?

AFOliveira (commented):
In general if your sole purpose is to get a two-step way of generating ,adoc and PDF spearately, IMO this PR can be simplified to only add a descriptions to line 56 -  
 `task instruction_appendix_adoc: MERGED_INSTRUCTIONS_FILE.to_s`
 
 It would look something like this:
 
```
 desc "Generate or update the instruction appendix AsciiDoc file"
 task instruction_appendix_adoc: MERGED_INSTRUCTIONS_FILE.to_s
```

backends/instructions_appendix/tasks.rake
ThinkOpenly: OH! OK. Then let me take a simpler approach. @dhower-qc said "remove line 56" and @AFOliveira said ~ 'add a description'. I presume the latter is preferred?

I'll be back with an updated and much simpler PR using the latter approach.

backends/instructions_appendix/tasks.rake
ThinkOpenly: I'm thinking not (obviously, because I removed it unconditionally), but happy to be convinced otherwise. I just thought we had made it optional because it was often broken in UDB. Now that it's pretty solid, I'd expect there to be assembly syntax in any description of an instruction.

backends/instructions_appendix/tasks.rake
AFOliveira: That makes sense, but in that case you will also have to change instructions.adoc.erb line 13. 

See, the thing you removed is only the explanation of what happens under the hood in 

```
<%- if ENV['ASSEMBLY']=="1" && inst.assembly.to_s.strip != "" -%>
Assembly::
<%= inst.fix_entities("#{inst.name.downcase} #{inst.assembly}") %>

<%- end -%>
```

backends/instructions_appendix/tasks.rake
AFOliveira: That makes sense, but in that case you will also have to change instructions.adoc.erb line 13. 

See, the thing you removed is only the explanation of what happens under the hood in 

```
<%- if ENV['ASSEMBLY']=="1" && inst.assembly.to_s.strip != "" -%>
Assembly::
<%= inst.fix_entities("#{inst.name.downcase} #{inst.assembly}") %>

<%- end -%>
```

backends/instructions_appendix/tasks.rake
dhower-qc: I think both are correct ;) If you remove line 56 and keep your addition, you are left with a described task.

backends/instructions_appendix/tasks.rake
ThinkOpenly: > That makes sense, but in that case you will also have to change instructions.adoc.erb line 13.

This was already done in #867, but I failed to remove the description in that same PR.

AFOliveira: Missed that, thanks for explaining!

PR #874, opened 2025-06-27
fix(CI): update get_yaml_directory to return spec/std/isa/inst

k4neka:
With the latest folder structure updates, the previous logic for resolving the YAML and JSON paths was outdated, causing LLVM-related tests to always be skipped. This commit updates the test to use the correct paths, ensuring that the tests run as intended and are no longer erroneously skipped.

### DIFF
--- a/ext/auto-inst/parsing.py
+++ b/ext/auto-inst/parsing.py
@@ -49,7 +49,7 @@ def get_json_path():
 
 
 def get_yaml_directory():
-    return "arch/inst/"
+    return "spec/std/isa/inst"
 
 
 def load_inherited_variable(var_path, repo_dir):


AFOliveira (approved):
LGTM! Thanks!

ThinkOpenly:
Should we change this, too:
https://github.com/riscv-software-src/riscv-unified-db/blob/56cda536878376f1194b85f116c1e43317639bc3/ext/auto-inst/test_parsing.py#L26-L28
...to yell a bit louder if the directory isn't found? Maybe an assert?
```
        _repo_dir = get_yaml_directory()
        assert(os.path.exists(_repo_dir))
```
Looking at `test_parsing.py`, there are a lot of places where we "skip" out... I wonder if more of those should be `assert` statements?

PR #880, opened 2025-07-02
feat(ci): add GitHub merge workflow

dhower-qc:
This adds a workflow that should run when a PR is added to the merge queue. The workflow has all the same jobs that run to deploy pages. This is step 1 of a full fix; step two is to deploy using the same GitHub actions instead of the deploy.sh script.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -1,0 +1,217 @@
+name: Merge Queue Tests
+permissions:
+  contents: read
+  pull-requests: write
+on:
+  merge_group:
+    types: [checks_requested]
+  workflow_dispatch:
+jobs:
+  build-reuse-manifest:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+      - name: run reuse
+        run: ./bin/reuse spdx
+  build-udb-api-doc:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/merge.yml
```diff
@@ -0,0 +1,217 @@
...
+permissions:
+  contents: read
+  pull-requests: write
+on:
+  merge_group:
+    types: [checks_requested]
+  workflow_dispatch:
+jobs:
+  build-reuse-manifest:
+    runs-on: ubutnut-latest
```
AFOliveira: typo

dhower-qc: thanks, fixed

lenary (commented):
LGTM.

.github/workflows/merge.yml
```diff
@@ -0,0 +1,217 @@
...
+permissions:
+  contents: read
+  pull-requests: write
+on:
+  merge_group:
+    types: [checks_requested]
+  workflow_dispatch:
+jobs:
+  build-reuse-manifest:
+    runs-on: ubutnu-latest
```
ThinkOpenly: I'm actually finding the misspellings of "ubuntu" quite entertaining. "ubutnut" has to be the best so far. ;-)

dhower-qc: :) fixed.

ThinkOpenly (approved):

PR #883, opened 2025-07-05
inst_type/I: Add I-type instruction schema

Unique-Usman:
for types/subtypes. This commit add the schema
for the I-type. It is also part of effort in (#655).

### DIFF
--- a/spec/std/isa/inst_type/I.yaml
+++ b/spec/std/isa/inst_type/I.yaml
@@ -1,0 +1,22 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../schemas/inst_type_schema.json#
+
+$schema: inst_type_schema.json#
+kind: instruction_type
+name: I
+description: I-type instructions usually have one source register, an immediate value, and one destination register
+length: 32
+opcodes:
+  funct3:
+    location: 14-12
+  opcode:
+    location: 6-0
+variables:
+  imm:
+    location: 31-20
+  rs1:
+    location: 19-15
+  rd:
+    location: 11-7
--- a/backends/instructions_appendix/all_instructions.golden.ado
+++ b/backends/instructions_appendix/all_instructions.golden.ado
@@ -1,0 +1,46367 @@
+= Instruction Appendix
+:doctype: book
+:wavedrom: /home/uniqueusman/riscv-unified-db/node_modules/.bin/wavedrom-cli
+// Now the document header is complete and the wavedrom attribute is active.
+
+
+[#udb:doc:inst:add]
+== add
+
+Synopsis::
+Integer add
+
+Assembly::
+add xd, xs1, xs2
+
+Encoding::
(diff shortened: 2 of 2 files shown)


spec/std/isa/inst_type/I.yaml
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
ThinkOpenly: You are fine to grant the copyright to this file to Qualcomm, but you can also put your name here, since you created the content.
```
# Copyright (c) 2025 Usman Akinyemi
```
is probably fine. (Although I am not a lawyer and cannot give legal advice...)

ThinkOpenly (approved):

ThinkOpenly:
One of the CI tests is failing, seemingly complaining about the end of file in the golden reference. Not sure what's going on there.

Also, I notice that you are doing work on your "main" branch. I suggest, at least in the future, to do new work on a new topic-specific branch. It makes keeping the trees in sync much easier, I find, because your tree doesn't get ahead of the upstream tree with unmerged changes. Just my advice.

spec/std/isa/inst_type/I.yaml
Unique-Usman: I think I should put mine then if it is fine

ThinkOpenly: Go for the glory when you can. :-)

Unique-Usman:
> One of the CI tests is failing, seemingly complaining about the end of file in the golden reference. Not sure what's going on there.
Yeah, it seems so. I will check

> 
> Also, I notice that you are doing work on your "main" branch. I suggest, at least in the future, to do new work on a new topic-specific branch. It makes keeping the trees in sync much easier, I find, because your tree doesn't get ahead of the upstream tree with unmerged changes. Just my advice.

Noted, I made the commit already before I had noticed that I was on the main branch. Thank you.

Unique-Usman:
> Though it's not going to hurt, you don't actually need the change to all_instructions.golden.adoc since nothing (except your local path, which is ignored) changed.
> 
> Let's remove that change, and then it looks good.

Thanks for the review, is there any rule of thumb on when to update the all_instructions.golden.adoc and when not to ? 

Thank you.

ThinkOpenly:
> Thanks for the review, is there any rule of thumb on when to update the all_instructions.golden.adoc and when not to ?

The (current) contents of the instruction appendix are primarily from `spec/std/isa/inst`, although your work is starting to expand that a bit. It can't hurt to generate it and look at the diff.

That there was a diff at all makes me wonder if we can generate the appendix without the local path, or maybe strip it off as part of the generation so that there is no diff unless there needs to be?

Unique-Usman:
> Great that you keep on with this work!

Yeah, Thanks, it might not fast, but, I will definitely keep on with it.

spec/std/isa/inst_type/I.yaml
dhower-qc: absolutely. I started transitioning to per-file headers (was part of the big file move PR) for exactly this purpose: the right people should get credit!!

PR #892, opened 2025-07-11
feat(schema): Add schema for I-x-x subtype

Unique-Usman:
for types/subtypes. This commit add the schema
for the I-x-x subtype. This commit also add schema
for immediate variable and variable type. It is also
part of effort in (#655).


spec/std/isa/inst_var/imm.yaml
dhower-qc, line 1: We are going to have a lot of different kinds of immediates, so we probably want to give this one a more descriptive name (like i-imm12)

dhower-qc (commented):
Other than the nit about name, this looks good to me.

spec/std/isa/inst_subtype/I/I-x-x.yaml
```diff
@@ -0,0 +1,21 @@
...
+name: I-x-x
+
+data:
+  type: { "$ref": "inst_type/I.yaml#" }
+  subtype: { "$ref": "inst_subtype/I/I-x-x.yaml#" }
+  opcodes:
+    $inherits: inst_type/I.yaml#/opcodes
+  variables:
+    func11:
+      $inherits: inst_var/i-imm12.yaml#/data
```
ThinkOpenly: Bike-shedding, perhaps...
Since the `i` in `i-imm12` references the instruction type, `I`, perhaps it should be capitalized to match: `I-imm12`.
Further, if we don't expect other lengths for an I immediate, perhaps it should be `I-imm`? (Or, even if we do expect other lengths, this 12-bit length will surely dominate and the others would be exceptions to the rule and can be named as such if and when the need arises.)

Unique-Usman: I think there are other length also. Since adding the `i-imm12` make it more descriptive, that might be a good reason to also leave it like that.

ThinkOpenly: The only obvious exception is when that 12-bit "imm" field is broken up into 2 fields for shift instructions, and those will have to be new variables (and opcodes!) anyway, with names like "shamt" instead of "imm".

Unique-Usman: Sounds good. Let go with `I-imm,` should it be I-imm or `i-imm` ?

ThinkOpenly: `I-imm`, please.

spec/std/isa/inst_subtype/I/I-x-x.yaml
```diff
@@ -0,0 +1,21 @@
...
+kind: instruction_subtype
+name: I-x-x
+
+data:
+  type: { "$ref": "inst_type/I.yaml#" }
+  subtype: { "$ref": "inst_subtype/I/I-x-x.yaml#" }
+  opcodes:
+    $inherits: inst_type/I.yaml#/opcodes
+  variables:
+    func11:
```
ThinkOpenly: Sorry, I should've caught this before now. This is a 12-bit field, so "func11" is confusing. However, in the spec, it's pretty consistently referred to as "imm", so let's use that name here.

ThinkOpenly (changes requested):

PR #904, opened 2025-07-15
fix(gen): support subtype-based encodings in C header generator

### LINKED ISSUE #893: Generated C header missing instructions with new subtype schema
**Describe the bug**
All instructions that have been migrated to use the new type/subtype encoding schema are skipped by `generator.py`, which is used to produce both the c_header and golang outputs. The resulting files are missing both the `MATCH` and `MASK` clauses for these instructions.

**To Reproduce**
1. Run `./do gen:c_header`

```
⬢ [podman] ❯ ./do gen:c_header
Using devcontainer environment
Running with 1 job(s)
/workspaces/riscv-unified-db/.home/.venv/bin/python3 /workspaces/riscv-unified-db/backends/generators/c_header/generate_encoding.py --inst-dir=/workspaces/riscv-unified-db/gen/resolved_spec/_/inst --csr-dir=/workspaces/riscv-unified-db/gen/resolved_spec/_/csr --ext-dir=/workspaces/riscv-unified-db/gen/resolved_spec/_/ext --output=/workspaces/riscv-unified-db/gen/c_header/encoding.out.h --include-all
INFO:: Loading instructions from /workspaces/riscv-unified-db/gen/resolved_spec/_/inst
INFO:: Searching for instruction files in /workspaces/riscv-unified-db/gen/resolved_spec/_/inst for target architecture BOTH
ERROR:: Missing 'encoding' field in instruction orn in /workspaces/riscv-unified-db/gen/resolved_spec/_/inst/B/orn.yaml
ERROR:: Missing 'encoding' field in instruction andn in /workspaces/riscv-unified-db/gen/resolved_spec/_/inst/B/andn.yaml
ERROR:: Missing 'encoding' field in instruction ror in /workspaces/riscv-unified-db/gen/resolved_spec/_/inst/B/ror.yaml
ERROR:: Missing 'encoding' field in instruction clmulh in /workspaces/riscv-unified-db/gen/resolved_spec/_/inst/B/clmulh.yaml
ERROR:: Missing 'encoding' field in instruction xnor in /workspaces/riscv-unified-db/gen/resolved_spec/_/inst/B/xnor.yaml
ERROR:: Missing 'encoding' field in instruction rorw in /workspaces/riscv-unified-db/gen/resolved_spec/_/inst/B/rorw.yaml
ERROR:: Missing 'encoding' field in instruction rolw in /workspaces/riscv-unified-db/gen/resolved_spec/_/inst/B/rolw.yaml
ERROR:: Missing 'encoding' field in instruction clmul in /workspaces/riscv-unified-db/gen/resolved_spec/_/inst/B/clmul.yaml
ERROR:: Missing 'encoding' field in instruction rol in /workspaces/riscv-unified-db/gen/resolved_spec/_/inst/B/rol.yaml
ERROR:: Missing 'encoding' field in instruction add.uw in /workspaces/riscv-unified-db/gen/resolved_spec/_/inst/Zba/add.uw.yaml
INFO:: Found 1150 instruction definitions in 1150 files
INFO:: Filtered out 10 instructions due to encoding issues
```

**Additional context**
`generator.py` is looking specifically for an `encoding` field, which no longer exists in these instructions. It needs a new way to retrieve the encoding.

There should probably be golden reference files for these artifacts (similar to what is already done for the instruction appendix) to prevent inadvertent issues like these. Or at least a check in CI for error messages.

blazethunderstorm:
Fixes #893

Updated the generator.py to correctly detect instruction encodings when using the new type/subtype schema. Previously, it looked only for the old encoding field, so some instructions were skipped. Now it checks subtype fields and extracts encoding info from there.

image
image


blazethunderstorm:
@ThinkOpenly would do that

dhower-qc:
Thanks for the contribution! Like @ThinkOpenly said, you'll need to build up what used to be explicit in 'match'. Basically, start with an instruction-length string of '-'s and then replace any position that is occupied by an opcode value.

So if you have:

```yaml
# andn.yaml
format:
  funct7:
    display_name: ANDN
    location: 31-25
    value: 0b0100000
  funct3:
    display_name: ANDN
    location: 14-12
    value: 0b111
  opcode:
    display_name: OP
    location: 6-0
    value: 0b0110011
```

You want to wind up with the string:

`0100000----------111-----0110011`

blazethunderstorm:
@dhower-qc thanks for help would make the changes as req

backends/generators/generator.py
```diff
@@ -177,12 +177,40 @@ def load_instructions(
...
+                    for field, info in opcodes.items():
+                        if not isinstance(info, dict):
+                            continue  
+
+                        value = info.get("value")
+                        location = info.get("location")
+                        if value is None or location is None:
+                            continue
+
+                        if isinstance(location, str) and "-" in location:
```
ThinkOpenly: You need to handle when `location` does not have `-`. They all do today, but this is not guaranteed. In this case, the field is a single bit.

backends/generators/generator.py
```diff
@@ -177,12 +177,40 @@ def load_instructions(
...
+                            val_int = int(value, 0) if isinstance(value, str) else value
+                            val_bin = bin(val_int)[2:].zfill(width)
+
+                            match_bits[31 - hi : 32 - lo + 1] = list(val_bin)
+                    match_str = "".join(match_bits)
+                    instr_dict[name] = {"match": match_str}
+                except Exception as e:
+                    logging.error(f"Failed to construct match string for {name} in {path}: {e}")
+                    encoding_filtered += 1
+                    continue
```
ThinkOpenly: Do we need this? This would have the effect of logged, but otherwise ignored errors. We may actually want a crash here.

backends/generators/generator.py
```diff
@@ -177,12 +177,40 @@ def load_instructions(
...
+
+                continue
```
ThinkOpenly: does this have any effect?

backends/generators/generator.py
```diff
@@ -255,14 +283,30 @@ def load_instructions(
                 encoding_filtered += 1
                 continue
 
-            match_str = encoding_to_use.get("match")
+            match_str = encoding_to_use.get("match") if encoding_to_use else None
+            mask_str = encoding_to_use.get("mask") if encoding_to_use else None
```
ThinkOpenly: Do we need to do any work with "mask" here? I thought it was calculated elsewhere based on the "match" string.

blazethunderstorm:
@ThinkOpenly would think about this and inform you and make changes

backends/generators/generator.py
```diff
@@ -5,7 +5,7 @@
 import pprint
 
 pp = pprint.PrettyPrinter(indent=2)
-logging.basicConfig(level=logging.INFO, format="%(levelname)s:: %(message)s")
+logging.basicConfig(level=logging.DEBUG, format="%(levelname)s:: %(message)s")
```
AFOliveira: not-related

AFOliveira (changes requested):

backends/generators/generator.py
blazethunderstorm: Would make changes

AFOliveira: Thanks for the effort!

blazethunderstorm: sry i wanted to see the logs of whats happening so i wrote debug and forgot to change it back would do that

blazethunderstorm:
@AFOliveira i have made changes asked by @ThinkOpenly

backends/generators/generator.py
AFOliveira: @blazethunderstorm can you see this review? It also seems randomly placed to me

AFOliveira (commented):
@blazethunderstorm not to this one, I thin k

AFOliveira:
@blazethunderstorm You still have not addressed @ThinkOpenly comments, can you please take a look at them?

blazethunderstorm:
Alright would do that sry

backends/generators/generator.py
blazethunderstorm: i have fixed this now pls review

backends/generators/generator.py
blazethunderstorm: You're right, we don't use mask here directly. It's just fetched in case something later needs it. I can remove it if we’re sure it’s not used, but keeping it for now seems safe.

backends/generators/generator.py
blazethunderstorm: fixed pls review

blazethunderstorm:
@ThinkOpenly @AFOliveira pls review

backends/generators/generator.py
ThinkOpenly: Remove references to `mask` here. It could be confusing to compute it here, not use it here, then compute it again somewhere else.

blazethunderstorm: Should I complete remove it then @ThinkOpenly

AFOliveira: What @ThinkOpenly means is that you should remove it from this part as it's unused

AFOliveira:
> @AFOliveira sry to ask but what loose end like I have to fix

@ThinkOpenly comment.

backends/generators/generator.py
blazethunderstorm: Ohk would do that

backends/generators/generator.py
ThinkOpenly: The `continue` statement is still there, but looking again it does have a purpose: the code above handles the "format" case, and the code below handles the "encoding" case. These are mutually exclusive, so OK.

Nit: could you remove the extra blank line _above_ the `continue` statement?

blazethunderstorm: Yeah sure

backends/generators/generator.py
blazethunderstorm: i have fixed this pls  review

backends/generators/generator.py
ThinkOpenly: still there :-)

ThinkOpenly:
CI is failing, complaining about the extra blank line and three other formatting issues.

blazethunderstorm:
Would fix this

blazethunderstorm:
> CI is failing, complaining about the extra blank line and three other formatting issues.
i have fixed this pls review

backends/generators/generator.py
```diff
@@ -177,12 +177,56 @@ def load_instructions(
...
+                opcodes = format_data.get("opcodes", {})
+
+                if not opcodes:
+                    logging.error(
+                        f"Missing both 'encoding' and 'format.opcodes' for {name} in {path}"
+                    )
+                    encoding_filtered += 1
+                    continue
+
+                match_bits = ["-" for _ in range(32)]
```
ThinkOpenly: Does this work for compressed (extension "C" and friends) instructions? For example:https://github.com/riscv-software-src/riscv-unified-db/blob/292b34f289cc0620ba99089a840467cf5d8c5671/spec/std/isa/inst/C/c.addi.yaml#L21

backends/generators/generator.py
```diff
@@ -177,12 +177,56 @@ def load_instructions(
...
+                                else:
+                                    hi = lo = int(location)
+
+                                width = hi - lo + 1
+                                val_int = (
+                                    int(value, 0) if isinstance(value, str) else value
+                                )
+                                val_bin = bin(val_int)[2:].zfill(width)
+
+                                match_bits[31 - hi : 32 - lo + 1] = list(val_bin)
```
ThinkOpenly: Also needs to accommodate compressed instructions (16 bit encoding)

blazethunderstorm:
@ThinkOpenly pls review

backends/generators/generator.py
blazethunderstorm: fixed pls review

backends/generators/generator.py
```diff
@@ -187,7 +187,22 @@ def load_instructions(
                     encoding_filtered += 1
                     continue
 
-                match_bits = ["-" for _ in range(32)]
+                bit_length = 32
+                if "length" in data:
```
ThinkOpenly, line 191: Is this ever true?

backends/generators/generator.py
```diff
@@ -187,7 +187,22 @@ def load_instructions(
...
+                    try:
+                        bit_length = int(data["length"])
+                    except ValueError:
+                        pass
+                elif "C" in enabled_extensions:
+                    if any(
+                        loc.get("location", "").startswith("0..15")
```
ThinkOpenly, line 198: Is this ever true?

backends/generators/generator.py
```diff
@@ -211,7 +226,10 @@ def load_instructions(
                                 )
                                 val_bin = bin(val_int)[2:].zfill(width)
 
-                                match_bits[31 - hi : 32 - lo + 1] = list(val_bin)
+                                inst_width = 16 if encoding.get("compressed", False) else 32
```
ThinkOpenly, line 229: Is `encoding.get("compressed"...` ever true?

backends/generators/generator.py
blazethunderstorm: No i tried its not working sry would do it properly and would make the requested changes

AFOliveira:
> > Looks good to me, please fix the commit history!
> 
> What do you seek, @AFOliveira , given that the merge queue will squash everything?


Nevermind, I'm just not used to the merge queue squash.

ThinkOpenly:
Superceded by #1051

PR #905, opened 2025-07-16
feat(csr): inherit attrs from aliases

### LINKED ISSUE #781: Field aliases should inherit attributes
There are many cases where a CSR field is simply an alias of field in another CSR (search for "alias" in the CSR definitions). Currently, the alias field simply serves as a pointer from one bit to another, and all other attributes associated with the field must be duplicated for the alias. For example, the MBE field in https://github.com/riscv-software-src/riscv-unified-db/blob/808077a49727f05e75d3e8150256f757eda5eef8/arch/csr/mstatus.yaml#L88-L108

and it's alias in https://github.com/riscv-software-src/riscv-unified-db/blob/808077a49727f05e75d3e8150256f757eda5eef8/arch/csr/mstatush.yaml#L52-L58

Fields which are not location-specific ("type()", "reset_value()") are duplicated, and "description" is only a reference. Since these fields can be non-trivial, as seen above, duplication can be error-prone.

Fields which are not location-specific (or perhaps fields which are not overridden by the alias definition) should be inherited from the aliased field.

blazethunderstorm:
Closes #781

Made CSR field aliases automatically inherit missing attributes like description, type, and reset value from the original field. This avoids duplication and keeps alias definitions cleaner and less error-prone.

### DIFF
--- a/backends/generators/generator.py
+++ b/backends/generators/generator.py
@@ -103,6 +103,38 @@ def parse_extension_requirements(extensions_spec):
     # that have an unrecognized format rather than excluding them
     return lambda exts: True
 
+def resolve_field_alias(field_def, alias_path, csr_root, current_name):
+    try:
+        alias_csr_name, alias_field_name = alias_path.split(".")
+        alias_file_path = None
+
+        for root, _, files in os.walk(csr_root):
+            for file in files:
+                if file == f"{alias_csr_name}.yaml":
+                    alias_file_path = os.path.join(root, file)
+                    break
+            if alias_file_path:
+                break
+
+        if not alias_file_path or not os.path.exists(alias_file_path):
+            logging.warning(f"[CSR] Alias file not found for {alias_path} in {current_name}")
+            return
+
+        with open(alias_file_path, encoding="utf-8") as alias_file:
+            alias_data = yaml.safe_load(alias_file)
+            alias_fields = alias_data.get("fields", {})
+
+            alias_field_def = alias_fields.get(alias_field_name)
+            if not isinstance(alias_field_def, dict):
+                logging.warning(f"[CSR] Alias field {alias_field_name} not found or invalid in {alias_file_path}")
+                return
+
+            for attr_key, attr_value in alias_field_def.items():
+                if attr_key not in ("location", "alias") and attr_key not in field_def:
+                    field_def[attr_key] = attr_value
+
+    except Exception as e:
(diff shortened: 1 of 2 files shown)


blazethunderstorm:
@dhower-qc i am using alias only but like removing other things is giving error what to do for that

blazethunderstorm:
@dhower-qc i have resolved the issue raising pr now

blazethunderstorm:
image

blazethunderstorm:
@dhower-qc @ThinkOpenly pls review

tools/ruby-gems/udb/python/yaml_resolver.py
```diff
@@ -407,6 +407,45 @@ def _resolve(obj, obj_path, obj_file_path, doc_obj, arch_root, do_checks):
...
+                                        for attr_key, attr_value in alias_field_def.items():
+                                            if (attr_key not in ("location", "alias") and 
+                                                attr_key not in field_def):
+                                                if not ((attr_key == "type" and "type()" in field_def) or
+                                                    (attr_key == "type()" and "type" in field_def) or
+                                                    (attr_key == "reset_value" and "reset_value()" in field_def) or
+                                                    (attr_key == "reset_value()" and "reset_value" in field_def)):
+                                                    field_def[attr_key] = deepcopy(attr_value)
+                                                
+                        except Exception as e:
```
dhower-qc, line 447: What would cause an exception in this code?

tools/ruby-gems/udb/python/yaml_resolver.py
```diff
@@ -407,6 +407,45 @@ def _resolve(obj, obj_path, obj_file_path, doc_obj, arch_root, do_checks):
 
         return final_obj
     else:
+        if isinstance(obj, dict) and "fields" in obj:
```
dhower-qc, line 410: Is this just checking that the alias points to something valid?

If so, we typically do that via Ruby in the object "validate" functions.

backends/generators/generator.py
```diff
@@ -103,9 +103,41 @@ def check_alternative(alt, exts):
     # that have an unrecognized format rather than excluding them
     return lambda exts: True
 
+def resolve_field_alias(field_def, alias_path, csr_root, current_name):
```
dhower-qc, line 106: Is this just a check? The function doesn't seem to return anything.

If it is just a check, we can probably remove it. I think there is an implicit assumption here that the database you are working with is consistent (and if we are missing a check, it should be added with the other consistency checks in Ruby instead of here).

Issue #827, opened 2025-06-02
Improve Zclsd Constraint Visibility Outside of operation() Function


Shehrozkashif:
In [PR #577](https://github.com/your-repo-link/pull/577), the check for the Zclsd extension currently resides within the operation() function. However, this design hides the constraint from downstream tools like disassemblers that do not evaluate operation() to determine instruction validity.

As suggested in the discussion, it would be better to lift this constraint to the instruction definition level for better visibility and consistency. Unfortunately, the current schema does not support this kind of conditional inclusion, since definedBy is limited to simple extension-based checks.

Proposal:

Consider extending the schema to support conditional availability of instructions, potentially using a structure like:
```
definedBy:
  anyOf:
    - { name: C, when: xlen() > 32 }
    - { name: Zca, when: xlen() > 32 }
    - { name: Zclsd, when: xlen() == 32 }
```
Alternatively, introduce a new availableIf field or similar mechanism to handle such cases.

dhower-qc:
I think I'd prefer if we moved the XLEN up:

```yaml
definedBy:
  RV32:
    name: Zclsd
  RV64:
    anyOf:
      - name: C
      - name: Zca
```

That's more in line with how we've handled differences between RV32/64 elsewhere (like encodings)

PR #891, opened 2025-07-11
[WIP] feat: consolidate all YAML conditionals

dhower-qc:
WIP

PR #907, opened 2025-07-16
FIXES ISSUE #827: Improve Zclsd Constraint Visibility Outside of operation() Function

omsuneri:
this PR resolves issue #827 by moving the `Zclsd` constraint from the `operation()` function to the instruction-level `definedBy` field in `c.ld.yaml`, improving visibility for downstream tools such as disassemblers that do not evaluate `operation()` logic

why not used the suggested schema format?

```yaml
definedBy:
  RV32:
    name: Zclsd
  RV64:
    anyOf:
      - name: C
      - name: Zca
````

This format is not valid under the current `inst_schema.json` constraints. Specifically, `definedBy` only supports `string`, `anyOf`, `allOf`, `oneOf`, or `not` constructs based on `requires_entry`, and does not allow architecture-specific keys like `RV32` or `RV64` So this approach had to be adapted within the allowed schema

### DIFF
--- a/spec/std/isa/inst/C/c.ld.yaml
+++ b/spec/std/isa/inst/C/c.ld.yaml
@@ -18,6 +18,7 @@ definedBy:
     - C
     - Zca
     - Zclsd
+base: 32
 assembly: xd, imm(xs1)
 encoding:
   RV32:


dhower-qc:
Thanks! This is a great suggestion. So good, in fact, that it is being enabled by #891 😄.

Once 891 is in main, we can make the change.

PR #908, opened 2025-07-16
fix: cfgs with custom overlays; was broken by refactoring

dhower-qc:
- Fixes the generation of an extension PDF when a custom overlay is involved
- Fixes some minor IDL issues in Xqci that were revealed by recent compiler improvements
- Adds the Xqci document generation to regression tests

### DIFF
--- a/.github/workflows/regress.yml
+++ b/.github/workflows/regress.yml
@@ -201,3 +201,14 @@ jobs:
         uses: ./.github/actions/singularity-setup
       - name: Run cpp unit tests
         run: ./do test:cpp_hart CONFIG=rv64 JOBS=4
+  regress-xqci-doc:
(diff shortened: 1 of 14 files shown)


.github/workflows/regress.yml
```diff
@@ -201,3 +201,14 @@ jobs:
         uses: ./.github/actions/singularity-setup
       - name: Run cpp unit tests
         run: ./do test:cpp_hart CONFIG=rv64 JOBS=4
+  regress-xqci-doc:
```
ThinkOpenly, line 204: It would be really nice to make this generic, so that _all_ custom extensions are tested, and there are no config-specific or extension-specific references here. This is probably not trivial to implement, though.

PR #909, opened 2025-07-16
Merge new command "list csrs" CLI

adingank-qualcomm:
Added `test_list_csrs` with an assertion for equality of the number of CSRs reported with [the number of yaml files under `csr` directory](https://github.com/riscv-software-src/riscv-unified-db/blob/d0e69a4e02b33e0ea082a6fbc83f6f6e38480031/tools/ruby-gems/udb/test/test_cli.rb#L77).


adingank-qualcomm:
... and the new test failed with expected result of 0! :-) 
I suspect that the smoke regression not run from the root and hence `spec/std/isa/csr` is not found (unlike my local runs). 
Have to learn the right process for testing (even for this baby step PR)!

tools/ruby-gems/udb/test/test_cli.rb
```diff
@@ -62,4 +62,27 @@ def test_disasm
...
+    # the number of yaml files under csr directory.
+    num_csr_yaml_files = `find spec/std/isa/csr/ -name '*.yaml' | wc -l`.to_i
+    puts num_csr_yaml_files
+    assert_equal num_csr_yaml_files, out.split.length
+    # The puts above is reported when the individual test is run in the VSCode
+    # terminal from the root of the riscv-unified-db repo with:
+    # ./bin/ruby -Itest tools/ruby-gems/udb/test/test_cli.rb --name test_list_csrs
+
+    # Don't hard code the number of reported CSRs!
+    # assert_equal 371, out.split.length
```
dhower-qc: Let's remove these comments.

The bit about how to run is useful, but would be better at the top of the file.

dhower-qc (changes requested):

dhower-qc:
> ... and the new test failed with expected result of 0! :-) I suspect that the smoke regression not run from the root and hence `spec/std/isa/csr` is not found (unlike my local runs). Have to learn the right process for testing (even for this baby step PR)!

You can use the resolver to get the root path: `Udb::Resolver.repo_root`

tools/ruby-gems/udb/test/test_cli.rb
adingank-qualcomm: Removed comments

adingank-qualcomm:
> > ... and the new test failed with expected result of 0! :-) I suspect that the smoke regression not run from the root and hence `spec/std/isa/csr` is not found (unlike my local runs). Have to learn the right process for testing (even for this baby step PR)!
> 
> You can use the resolver to get the root path: `Udb::Resolver.repo_root`

Thanks! Resolver class didn't work but `Udb.repo_root` did!

tools/ruby-gems/udb/lib/udb/cli.rb
```diff
@@ -245,6 +245,37 @@ def parameters
...
+            Pathname.new("#{options[:config_dir]}/#{options[:config]}.yaml")
+          else
+            raise ArgumentError, "Cannot find config: #{options[:config]}"
+          end
+
+        resolver =
+          Udb::Resolver.new(
+            std_path_override: Pathname.new(options[:arch]),
+            gen_path_override: Pathname.new(options[:gen]),
+            custom_path_override: Pathname.new(options[:arch_overlay])
```
ThinkOpenly: totally a nit, so feel free to ignore me.

The order of these is input-output-input_overlay, which is unexpected to me
```suggestion
            std_path_override: Pathname.new(options[:arch]),
            custom_path_override: Pathname.new(options[:arch_overlay],
            gen_path_override: Pathname.new(options[:gen]))
```

PR #913, opened 2025-07-20
Add Zkr extension and seed CSR for physical entropy

### LINKED ISSUE #561: Add Zkr CSR
Need to add:

- SEED

blazethunderstorm:
Added a new extension called Zkr and a special register called seed.
This register gives software access to real random bits from the hardware, which can be used for security and cryptography.
The register is read-only and only gives out up to 16 bits of random data at a time.

Closes #561

### DIFF
--- a/spec/std/isa/csr/seed.yaml
+++ b/spec/std/isa/csr/seed.yaml
@@ -18,6 +18,9 @@ description: |
   The `seed` CSR exposes physical entropy directly to software. This register is useful for cryptographic
   applications and random number generation. Reads may have side effects depending on the implementation.
 
+sw_read: |
+  return read_entropy();
+
 fields:
   seed_value:
     location_rv32: 31-0
@@ -27,7 +30,4 @@ fields:
       Entropy bits returned from the physical entropy source.
       This field is WARL and read-only. On each read, the value may be consumed
       or updated depending on platform behavior (see `ZKR_READ_SIDE_EFFECTS`).
-    reset_value: UNDEFINED_LEGAL  
-    sw_read: |
-      return read_entropy();
-      
+    reset_value: UNDEFINED_LEGAL
--- a/spec/std/isa/csr/seed.yaml
+++ b/spec/std/isa/csr/seed.yaml
@@ -1,4 +1,4 @@
-# Copyright (c) Qualcomm Technologies, Inc.
+# Copyright (c) Anirudh Narang
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 
 # yaml-language-server: $schema=../../schemas/csr_schema.json
@@ -12,6 +12,7 @@ type: unprivileged
 priv_mode: U
 length: XLEN
 definedBy: Zkr
+
 description:
   - id: csr-seed-purpose
     normative: true
@@ -25,23 +26,34 @@ description:
       The entropy value returned by `seed` may change between reads. Some implementations may repeat values,
       return zero if entropy is not available, or exhibit other implementation-defined behavior.
       The returned value must never cause traps or undefined instruction behavior.
-      Only bits [15:0] are guaranteed to contain valid entropy.
+      Only bits [15:0] are guaranteed to contain valid entropy. Bits [31:30] may indicate entropy status.
 
 fields:
   SEED:
     location_rv64: 63-0
     location_rv32: 31-0
     description: |
-      Contains the physical entropy bits. Bits [15:0] are valid entropy; upper bits are zero or undefined.
-      This field is read-only. Reading it may consume entropy.
-    type(): return CsrFieldType::RO;
-    reset_value: UNDEFINED_LEGAL
-    sw_write(_): raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+      Contains the physical entropy bits. Bits [15:0] are valid entropy. Bits [31:30] may indicate entropy availability state.
+      Upper bits are reserved or undefined. This field may be read/write with hardware update (RW-H), but writes are ignored.
+    type: RW-H
+    reset_value: UNDEFINED
+    sw_write(_): |
+      # Check permission via mseccfg.USEED or mseccfg.SSEED (implementation-defined)
+      if (!has_seed_write_permission(mode())) {
+        raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+      }
+      # Ignore all written values (read-only from software perspective)
 
 sw_read(): |
-  if (!implemented?(ExtensionName::Zkr)) {
+  # Check access permission via mseccfg.USEED or mseccfg.SSEED
+  if (!has_seed_read_permission(mode())) {
     raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
   }
 
-  # Provide only the entropy bits in [15:0], rest are zero/undefined
-  return entropy_device.read() & ((1 << ZKR_SEED_WIDTH) - 1);
+  # Optionally: disallow CSRRS/CSRRC reads
+  if (is_read_modify_write_encoding($encoding)) {
+    raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+  }
+
+  # Return entropy value; get_random() is assumed to be a defined IDL function
+  return get_random() & ((1 << ZKR_SEED_WIDTH) - 1);
--- a/spec/std/isa/csr/seed.yaml
+++ b/spec/std/isa/csr/seed.yaml
@@ -22,9 +22,11 @@ fields:
   seed_value:
     location_rv32: 31-0
     location_rv64: 63-0
-    type: RO
+    type: RO-H
     description: |
       Entropy bits returned from the physical entropy source.
       This field is WARL and read-only. On each read, the value may be consumed
       or updated depending on platform behavior (see `ZKR_READ_SIDE_EFFECTS`).
-    reset_value: UNDEFINED_LEGAL
+    reset_value: UNDEFINED_LEGAL  
+    sw_read: |
+      return read_entropy();
--- a/spec/std/isa/csr/seed.yaml
+++ b/spec/std/isa/csr/seed.yaml
@@ -6,54 +6,25 @@
 $schema: "csr_schema.json#"
 kind: csr
 name: seed
-long_name: Physical Entropy Seed Register
+long_name: Entropy Seed Register
 address: 0x015
-type: unprivileged
 priv_mode: U
-length: XLEN
-definedBy: Zkr
-
-description:
-  - id: csr-seed-purpose
-    normative: true
-    text: |
-      The `seed` register is an XLEN-bit WARL read-only CSR that provides up to 16 bits of physical entropy.
-      Software can read this CSR to obtain entropy for cryptographic use. If no entropy is available, the value
-      may return 0 or an undefined pattern. Reading from `seed` may have side effects, depending on implementation.
-  - id: csr-seed-behavior
-    normative: false
-    text: |
-      The entropy value returned by `seed` may change between reads. Some implementations may repeat values,
-      return zero if entropy is not available, or exhibit other implementation-defined behavior.
-      The returned value must never cause traps or undefined instruction behavior.
-      Only bits [15:0] are guaranteed to contain valid entropy. Bits [31:30] may indicate entropy status.
+length: 64
+writable: false
+definedBy:
+  name: Zkr
+  version: ">=1.0.0"
+description: |
+  The `seed` CSR exposes physical entropy directly to software. This register is useful for cryptographic
+  applications and random number generation. Reads may have side effects depending on the implementation.
 
 fields:
-  SEED:
-    location_rv64: 63-0
+  seed_value:
     location_rv32: 31-0
+    location_rv64: 63-0
+    type: RO
     description: |
-      Contains the physical entropy bits. Bits [15:0] are valid entropy. Bits [31:30] may indicate entropy availability state.
-      Upper bits are reserved or undefined. This field may be read/write with hardware update (RW-H), but writes are ignored.
-    type: RW-H
-    reset_value: UNDEFINED
-    sw_write(_): |
-      # Check permission via mseccfg.USEED or mseccfg.SSEED (implementation-defined)
-      if (!has_seed_write_permission(mode())) {
-        raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
-      }
-      # Ignore all written values (read-only from software perspective)
-
-sw_read(): |
-  # Check access permission via mseccfg.USEED or mseccfg.SSEED
-  if (!has_seed_read_permission(mode())) {
-    raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
-  }
-
-  # Optionally: disallow CSRRS/CSRRC reads
-  if (is_read_modify_write_encoding($encoding)) {
-    raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
-  }
-
-  # Return entropy value; get_random() is assumed to be a defined IDL function
-  return get_random() & ((1 << ZKR_SEED_WIDTH) - 1);
+      Entropy bits returned from the physical entropy source.
+      This field is WARL and read-only. On each read, the value may be consumed
+      or updated depending on platform behavior (see `ZKR_READ_SIDE_EFFECTS`).
+    reset_value: UNDEFINED_LEGAL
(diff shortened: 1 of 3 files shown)


spec/std/isa/ext/Zkr.yaml
```diff
@@ -1,17 +1,69 @@
-# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# Copyright (c) Qualcomm Technologies, Inc.
```
dhower-qc: This shouldn't be copyrighted to Qualcomm. It should be you personally or your company, depending on if you are acting independently and/or your company's policy.

spec/std/isa/ext/Zkr.yaml
```diff
@@ -1,17 +1,69 @@
...
+  It is a read-only, WARL register visible to unprivileged software.
+
 type: unprivileged
+
 versions:
   - version: "1.0.0"
     state: ratified
     ratification_date: null
+
+csrs:
```
dhower-qc: CSRs are not defined in the extension YAML file. They are defined in their own file under spec/std/isa/csr(/Zkr).

spec/std/isa/csr/seed.yaml
```diff
@@ -0,0 +1,47 @@
...
+      Only bits [15:0] are guaranteed to contain valid entropy.
+
+fields:
+  SEED:
+    location_rv64: 63-0
+    location_rv32: 31-0
+    description: |
+      Contains the physical entropy bits. Bits [15:0] are valid entropy; upper bits are zero or undefined.
+      This field is read-only. Reading it may consume entropy.
+    type(): return CsrFieldType::RO;
```
dhower-qc: This field is a bit strange. Technically, it is read-write, though all write values should always be ignored. 
 
- When the type is not configuration-dependent, we prefer the `type:` alternative
- This is "Read/Write with Hardware Update", so the type should be:

```yaml
  type: RW-H
```

spec/std/isa/csr/seed.yaml
```diff
@@ -0,0 +1,47 @@
...
+    reset_value: UNDEFINED_LEGAL
+    sw_write(_): raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
```
dhower-qc: This needs to do several things;

- Check if the hart has permission to access the register (via mseccfg.SSEED and mseccfg.USEED)
- raise an IllegalInstruction -or- VirtualInstruction exception if needed
- do nothing otherwise (all write values are ignored)

spec/std/isa/csr/seed.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+sw_read(): |
+  if (!implemented?(ExtensionName::Zkr)) {
```
dhower-qc: You don't need to check if Zkr is implemented, that is captured by `definedBy`.

Like sw_write, this needs to check for access permission and raise the appropriate exception if needed.

This is strange CSR in that it is supposed to raise an exception if it is being accessed by a read-only CSRRS/CSRRC/CSRRSI/CSRRCI. That needs to be checked here, too.

spec/std/isa/csr/seed.yaml
```diff
@@ -0,0 +1,47 @@
...
+    raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+  }
+
+  # Provide only the entropy bits in [15:0], rest are zero/undefined
+  return entropy_device.read() & ((1 << ZKR_SEED_WIDTH) - 1);
```
dhower-qc: what is entropy_device?

This likely needs an IDL builtin function (like get_random()) added to builtin_functions.idl.

Also, bits 31:30 aren't zero/undefined; they indicate the entropy state.

dhower-qc (changes requested):

blazethunderstorm:
@dhower-qc would do requested changes

blazethunderstorm:
@dhower-qc pls review the pr

spec/std/isa/csr/seed.yaml
```diff
@@ -0,0 +1,59 @@
...
+    location_rv64: 63-0
+    location_rv32: 31-0
+    description: |
+      Contains the physical entropy bits. Bits [15:0] are valid entropy. Bits [31:30] may indicate entropy availability state.
+      Upper bits are reserved or undefined. This field may be read/write with hardware update (RW-H), but writes are ignored.
+    type: RW-H
+    reset_value: UNDEFINED
+    sw_write(_): |
+      # Check permission via mseccfg.USEED or mseccfg.SSEED (implementation-defined)
+      if (!has_seed_write_permission(mode())) {
```
dhower-qc: This function needs to be implemented somewhere. Take a look at `spec/std/isa/isa/globals.isa`

spec/std/isa/csr/seed.yaml
```diff
@@ -0,0 +1,59 @@
...
+        raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+      }
+      # Ignore all written values (read-only from software perspective)
+
+sw_read(): |
+  # Check access permission via mseccfg.USEED or mseccfg.SSEED
+  if (!has_seed_read_permission(mode())) {
```
dhower-qc: same; needs an implementation

spec/std/isa/csr/seed.yaml
```diff
@@ -0,0 +1,59 @@
...
+    raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+  }
+
+  # Optionally: disallow CSRRS/CSRRC reads
+  if (is_read_modify_write_encoding($encoding)) {
+    raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+  }
+
+  # Return entropy value; get_random() is assumed to be a defined IDL function
+  return get_random() & ((1 << ZKR_SEED_WIDTH) - 1);
```
dhower-qc: same; get_random() needs an implementation. In this case, get_random is probably a builtin function whose implementation needs to be provided by whatever is using IDL (see `spec/std/isa/isa/builtin_functions.idl`

dhower-qc (changes requested):
You'll need to make sure that this passes before the PR can merge:

```bash
./do test:smoke
```

That will, among other things, parse and type check the IDL code.

blazethunderstorm:
@dhower-qc pls review

blazethunderstorm:
@dhower-qc @ThinkOpenly @AFOliveira pls review the pr

spec/std/isa/csr/seed.yaml
```diff
@@ -0,0 +1,30 @@
...
+  version: ">=1.0.0"
+description: |
+  The `seed` CSR exposes physical entropy directly to software. This register is useful for cryptographic
+  applications and random number generation. Reads may have side effects depending on the implementation.
+
+fields:
+  seed_value:
+    location_rv32: 31-0
+    location_rv64: 63-0
+    type: RO
```
dhower-qc: This is "RO-H", which means read only with hardware update

spec/std/isa/csr/seed.yaml
```diff
@@ -0,0 +1,30 @@
...
+    description: |
+      Entropy bits returned from the physical entropy source.
+      This field is WARL and read-only. On each read, the value may be consumed
+      or updated depending on platform behavior (see `ZKR_READ_SIDE_EFFECTS`).
+    reset_value: UNDEFINED_LEGAL
```
dhower-qc, line 33: To model this, we'll also need a sw_read function. Since the value read is completely implementation-dependent, it should just call a builtin function. See spec/std/isa/isa/builtin_functions.idl

spec/std/isa/ext/Zkr.yaml
```diff
@@ -1,17 +1,35 @@
...
 # yaml-language-server: $schema=../../schemas/ext_schema.json
 
 $schema: "ext_schema.json#"
 kind: extension
 name: Zkr
-long_name: Entropy Source
-description: |
-  Defines the `seed` CSR.
-  This CSR provides up to 16 physical entropy bits that can be used to seed cryptographic random bit generators.
+long_name: Entropy Source Extension
```
dhower-qc: Remove the word extension here; it's implied

spec/std/isa/ext/Zkr.yaml
```diff
@@ -1,17 +1,35 @@
...
+      minimum: 1
+      maximum: 16
+    description: |
+      Defines the number of valid entropy bits returned in the `seed` CSR. Must be ≤ 16.
+
+  ZKR_READ_SIDE_EFFECTS:
+    schema:
+      type: boolean
+    description: |
+      If true, reading the `seed` CSR consumes entropy and may affect future read values (e.g., advancing a ring buffer).
```
dhower-qc, line 35: This is an interesting one. I agree this is a parameter in the spec, but I'm not sure what we can learn from it, especially since the side effect consequence could be anything.

@ThinkOpenly, @allenjbaum, thoughts?

dhower-qc (changes requested):

spec/std/isa/csr/seed.yaml
blazethunderstorm: Yeah sry would change that

spec/std/isa/csr/seed.yaml
blazethunderstorm: Would do the req changes

spec/std/isa/ext/Zkr.yaml
blazethunderstorm: Sure

spec/std/isa/ext/Zkr.yaml
ThinkOpenly: Hmm. Maybe just define it and move on. It at least forces those that define configurations to be aware. (Unless and until we make a "default" value for it, I guess.)

blazethunderstorm: so  should i just keep it @ThinkOpenly @dhower-qc

spec/std/isa/ext/Zkr.yaml
blazethunderstorm: done @dhower-qc

spec/std/isa/csr/seed.yaml
blazethunderstorm: done @dhower-qc

spec/std/isa/csr/seed.yaml
blazethunderstorm: done @dhower-qc

blazethunderstorm:
@dhower-qc @ThinkOpenly pls review

spec/std/isa/ext/Zkr.yaml
ThinkOpenly: I say keep it. I'd rather have an unused parameter than nothing.

blazethunderstorm: Ohk got it

blazethunderstorm:
Would fix the failing test

spec/std/isa/isa/builtin_functions.idl
```diff
@@ -637,3 +637,11 @@ builtin function atomic_read_modify_write_64 {
     it's assumed the RMW is OK to proceed.
   }
 }
+
+builtin function read_entropy {
+  returns Bits<64>
```
dhower-qc: should be Bits<ZKR_SEED_WIDTH>

spec/std/isa/csr/seed.yaml
```diff
@@ -0,0 +1,33 @@
...
+  seed_value:
+    location_rv32: 31-0
+    location_rv64: 63-0
+    type: RO-H
+    description: |
+      Entropy bits returned from the physical entropy source.
+      This field is WARL and read-only. On each read, the value may be consumed
+      or updated depending on platform behavior (see `ZKR_READ_SIDE_EFFECTS`).
+    reset_value: UNDEFINED_LEGAL  
+    sw_read: |
```
dhower-qc: sw_read() is a property of the CSR, not the fields. It should look like:

```yaml
sw_read(): |
  return read_entropy();
```

dhower-qc (changes requested):

spec/std/isa/isa/builtin_functions.idl
blazethunderstorm: done made req changes

spec/std/isa/csr/seed.yaml
blazethunderstorm: done

spec/std/isa/csr/seed.yaml
```diff
@@ -0,0 +1,33 @@
...
+length: 64
+writable: false
+definedBy:
+  name: Zkr
+  version: ">=1.0.0"
+description: |
+  The `seed` CSR exposes physical entropy directly to software. This register is useful for cryptographic
+  applications and random number generation. Reads may have side effects depending on the implementation.
+
+sw_read: |
```
ThinkOpenly:
```suggestion
sw_read(): |
```

spec/std/isa/csr/seed.yaml
ThinkOpenly: This and the other comments still need to be addressed. They all refer to creating implementations of the new functions you are calling, very similar to the "read_entropy" function you already added, to my understanding.

spec/std/isa/ext/Zkr.yaml
```diff
@@ -1,17 +1,35 @@
...
 # yaml-language-server: $schema=../../../schemas/ext_schema.json
 
 $schema: "ext_schema.json#"
 kind: extension
 name: Zkr
-long_name: Entropy Source
-description: |
-  Defines the `seed` CSR.
-  This CSR provides up to 16 physical entropy bits that can be used to seed cryptographic random bit generators.
+long_name: Entropy Source
```
ThinkOpenly: CI is failing in pre-commit due to a trailing space:

```suggestion
long_name: Entropy Source
```

spec/std/isa/csr/seed.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+# yaml-language-server: $schema=../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: seed
+long_name: Entropy Seed Register
+address: 0x015
+priv_mode: U
+length: 64
```
ThinkOpenly, line 12:
```suggestion
length: 32
```

spec/std/isa/csr/seed.yaml
```diff
@@ -0,0 +1,33 @@
...
+writable: false
```
ThinkOpenly, line 13: The register is technically writable, although:
> The write value (in rs1 or uimm) must be ignored by implementations. The
purpose of the write is to signal polling and flushing.
```suggestion
writable: true
```
So, you'll need `sw_write()` method for each field.

henrikg-qc:
> @henrikg-qc, this PR needs an ISS implementation for a new builtin (stub) routine. Can you please take a look?

I can definitively help, but I want to synch with @blazethunderstorm as he responded "Sure"

@blazethunderstorm, if you are not working on the implementation of the builtin function I can help. let me know

spec/std/isa/csr/seed.yaml
```diff
@@ -0,0 +1,33 @@
...
+  name: Zkr
+  version: ">=1.0.0"
+description: |
+  The `seed` CSR exposes physical entropy directly to software. This register is useful for cryptographic
+  applications and random number generation. Reads may have side effects depending on the implementation.
+
+sw_read(): |
+  return read_entropy();
+
+fields:
```
ThinkOpenly, line 24: The fields need to match the definition in the spec:
- OPST: 31:30
- custom: 23:16
- entropy: 15:0

read/"write" access is managed through `mseccfg.sseed` and `mseccfg.useed`, depending on mode. These restrictions need to be realized in `sw_read()/sw_write()` methods.

ThinkOpenly (changes requested):
A description of the SEED CSR can be found in the RISC-V ISA unprivileged spec, "The seed CSR".

spec/std/isa/ext/Zkr.yaml
```diff
@@ -7,11 +7,29 @@ $schema: "ext_schema.json#"
...
     ratification_date: null
+
+params:
+  ZKR_SEED_WIDTH:
+    schema:
+      type: integer
+      minimum: 1
+      maximum: 16
+    description: |
+      Defines the number of valid entropy bits returned in the `seed` CSR. Must be ≤ 16.
```
dhower-qc, line 29: I don't think this is a parameter; the spec states that the seed is always 16 bits

dhower-qc (changes requested):

ThinkOpenly:
Superseded by #2236 .

PR #914, opened 2025-07-21
B/rev8: use the new instruction schema

Unique-Usman:
for types/subtypes. This commit change B/rev8
instruction to use the new schema. It is also part of effort in (#655).

### DIFF
--- a/spec/std/isa/inst/B/rev8.yaml
+++ b/spec/std/isa/inst/B/rev8.yaml
@@ -20,21 +20,31 @@ description: |
 definedBy:
   anyOf: [Zbb, Zbkb]
 assembly: xd, xs1
-encoding:
+format:
   RV32:
-    match: 011010011000-----101-----0010011
+    $inherits:
+      - inst_subtype/I/I-x-x.yaml#/data
+    opcodes:
+      funct3:
+        display_name: REV8
+        value: 0b101
+      opcode: { $inherits: inst_opcode/OP-IMM.yaml#/data }
     variables:
-      - name: xs1
-        location: 19-15
-      - name: xd
-        location: 11-7
+      imm:
+        $inherits: inst_var/I-imm.yaml#/data
+        value: 0b011010011000
   RV64:
-    match: 011010111000-----101-----0010011
+    $inherits:
+      - inst_subtype/I/I-x-x.yaml#/data
+    opcodes:
+      funct3:
+        display_name: REV8
+        value: 0b101
+      opcode: { $inherits: inst_opcode/OP-IMM.yaml#/data }
     variables:
-      - name: xs1
-        location: 19-15
-      - name: xd
-        location: 11-7
+      imm:
+        $inherits: inst_var/I-imm.yaml#/data
+        value: 0b011010111000
 access:
   s: always
   u: always
--- a/spec/std/isa/inst_opcode/OP-IMM.yaml
+++ b/spec/std/isa/inst_opcode/OP-IMM.yaml
@@ -1,0 +1,12 @@
+# Copyright (c) 2025 Usman Akinyemi
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../schemas/inst_opcode_schema.json#
+
+$schema: inst_opcode_schema.json#
+kind: instruction_opcode
+name: OP-IMM
+
+data:
+  display_name: OP-IMM
+  value: 0b0010011


Unique-Usman:
@ThinkOpenly and @dhower-qc. kindly help go through this, I am not able to get the schema to match the instruction.

spec/std/isa/inst/B/rev8.yaml
```diff
@@ -20,21 +20,31 @@ description: |
 definedBy:
   anyOf: [Zbb, Zbkb]
 assembly: xd, xs1
-encoding:
+format:
   RV32:
```
ThinkOpenly, line 24: The (new) "format" stanza does not support "RV32"/"RV64" like "encoding" does. @dhower-qc was that intentional?

dhower-qc: No, that wasn't intentional. It should be added back in.

spec/schemas/inst_schema.json
```diff
@@ -430,6 +430,20 @@
           "description": "Path to the source file. Used by downstream tooling; not expected to be found in handwritten files",
           "type": "string"
         },
+        "all_bases_format": {
```
ThinkOpenly: @dhower-qc This is a new attribute, "all_bases_format", distinct from "format" (below), which is what Unique has been using thus far. What that intentional? :-) Do you want Unique to use this attribute instead of "format" when RV32/RV64 is required?

dhower-qc: sorry, sometimes I'm far too reliant on GH checks to do my testing ;) Fix coming

spec/std/isa/inst/B/rev8.yaml
```diff
@@ -20,21 +20,31 @@ description: |
...
+        value: 0b101
+      opcode: { $inherits: inst_opcode/OP-IMM.yaml#/data }
     variables:
-      - name: xs1
-        location: 19-15
-      - name: xd
-        location: 11-7
+      imm:
+        $inherits: inst_var/I-imm.yaml#/data
+        value: 0b011010011000
```
dhower-qc: We can't do this; all variables have to come from the subtype. That means that rev8 isn't an I-x-x instruction, and we'll need a new subtype.

Unique-Usman: The I-x-x subtype has the imm variable though and it was specifically created for the rev8 instruction. Reference -> https://github.com/riscv-software-src/riscv-unified-db/issues/655#issuecomment-2893176019

Maybe we need to make adjustment to the I-x-x ?

dhower-qc: ah, ok. Then, yes, we need to change I-x-x. `imm` isn't a variable for brev8; it's an opcode.  And, as such, it should probably have a different name (like funct12)

ThinkOpenly: What Derek is saying is that "imm" is actually a _variable_ in the "I-x-x" subtype, but in this REV8 instruction, it's a constant -- an opcode, and needs to be handled as such.

Unique-Usman: @dhower-qc and @ThinkOpenly,

I wanted to update the I-x-x sub_type but, I noticed the I-type opcodes section do not have funct12 since the imm is part of the vairables. I type -> https://github.com/riscv-software-src/riscv-unified-db/blob/main/spec/std/isa/inst_type/I.yaml

How do I reference the func12 from I-x-x to the opcodes in I type ?

Thanks

ThinkOpenly: _Hmm_.

What I think you need to do is create a new subtype, say "I-x-x-op", that inherits the opcodes from I, but also adds a new opcode, something like:
```
  opcodes:
    $inherits: inst_type/I.yaml#/opcodes
    funct12:
      location: 31-20
```
(And then don't include the "imm" variable, of course.) See if that works. If not, you may need to inherit the two opcodes individually:
```
  opcodes:
    funct12:
      location: 31-20
    funct3:
      $inherits: inst_type/I.yaml#/opcodes/funct3
    opcode:
      $inherits: inst_type/I.yaml#/opcodes/opcode
```
Hopefully, one of those approaches works.

ThinkOpenly:
> Unique-Usman force-pushed
> Unique-Usman requested review from ThinkOpenly and dhower-qc 16 hours ago

The disadvantage to force-push is it's difficult for reviewers to see what's changed since the last review. Since we're using the merge queue, the PR gets squash-merged, so all intermediate history is lost (for good or bad). Feel free to just add new commits while the PR is under review.

Can you describe what you changed?

I see the CI tests are failing, although the error messages are not terribly helpful to me.

Looking at things with fresh eyes, I see that the new subtype, I-x-x-op, is self-referential (as is the original subtype, R-x):
```
  subtype: { "$ref": "inst_subtype/R/R-x.yaml#" }
```
(Not sure if that causes problems or not, but now looks odd to me.)

I also see that you chose this implementation:
```
  opcodes:
    funct12:
      location: 31-20
    funct3:
      $inherits: inst_type/I.yaml#/opcodes/funct3
    opcode:
      $inherits: inst_type/I.yaml#/opcodes/opcode
  variables:
    xs1:
      $inherits: inst_var/xs1.yaml#/data
    xd:
      $inherits: inst_var/xd.yaml#/data
```
Did that "work"? Did you try the first suggestion? Just curious.

dhower-qc:
> Looking at things with fresh eyes, I see that the new subtype, I-x-x-op, is self-referential (as is the original subtype, R-x):
> 
> ```
>   subtype: { "$ref": "inst_subtype/R/R-x.yaml#" }
> ```
> 
> (Not sure if that causes problems or not, but now looks odd to me.)

That's intentional. Odd, yes. Maybe there is a better way, but I put it in there so that when instructions $inherit a subtype, the subtype shows up as a key in the resolved YAML.

```yaml
format:
  # without the self-reference, you'd have to peek at $parent_of in the resolved files to find your subtype
  $inherits:
    - inst_subtype/R/R-x.yaml#/data
```

dhower-qc:
> > ```
> > SON Schema Validation Error for inst_type/I.yaml:
> > Additional properties are not allowed ('$parent_of' was unexpected)
> > Validating arch:  68%|#############5      | 1161/1717 [00:04<00:02, 253.28it/s]
> > ```
> 
> 
> I looked into this a bit, but I'm getting confused in the schema definitions. When to use `$ref` versus `$inherits`. The use of `$inherits` here results in a `$parent_of` attribute being added to the generated/resolved subtypes and instructions, and the respective schema do not currently allow that.

**Re: When to use `$ref` versus `$inherits`.**

$refs are pointers. They point to something that is immutable.

$inherits is polymorphism. It pulls in a copy of the target, which can then be modified.

So, $ref if you will never change the pointed-to object, and $inherits if you will.

**Re: The use of `$inherits` here results in a `$parent_of` attribute being added to the generated/resolved subtypes and instructions, and the respective schema do not currently allow that.**

Correct. We can't use $inherits anywhere without the schema knowing that it is a possibility. The resolver is generic, and will happily inherit everything, but the schema check is going to fail. Maybe there is way to tell schema checkers to ignore $parent_of/$child_of? In that case, we wouldn't have to explicitly identify all the places it might be used.

spec/std/isa/inst/B/rev8.yaml
```diff
@@ -20,21 +20,29 @@ description: |
...
-    variables:
-      - name: xs1
-        location: 19-15
-      - name: xd
-        location: 11-7
+    $inherits:
+      - inst_subtype/I/I-x-x.yaml#/data
+    opcodes:
+      funct12:
+        display_name: RV32
```
dhower-qc: I think in most of the existing diagrams in the ISA manual, this funct12 field would have the same 'REV8' name as funct3.

spec/std/isa/inst/B/rev8.yaml
```diff
@@ -20,21 +20,29 @@ description: |
...
-    variables:
-      - name: xs1
-        location: 19-15
-      - name: xd
-        location: 11-7
+    $inherits:
+      - inst_subtype/I/I-x-x.yaml#/data
+    opcodes:
+      funct12:
+        display_name: RV64
```
dhower-qc: Same here; I think it should be "REV8"

spec/std/isa/inst_subtype/I/I-x-x-op.yaml
```diff
@@ -0,0 +1,24 @@
...
+data:
+  type: { "$ref": "inst_type/I.yaml#" }
+  subtype: { "$ref": "inst_subtype/I/I-x-x-op.yaml#" }
+  opcodes:
+    funct12:
+      location: 31-20
+    funct3:
+      $inherits: inst_type/I.yaml#/opcodes/funct3
+    opcode:
+      $inherits: inst_type/I.yaml#/opcodes/opcode
```
dhower-qc, line 19: Opcodes come from the type, so this is expected to look like:

```yaml
opcodes:
  $inherits: inst_type/I.yaml#/opcodes
```

In this case, that doesn't give what we want since 'funct12' is an 'imm' variable in I.

I think (?) that means that I-x-x-op is actually a subtype of a new type, not of I.

@ThinkOpenly, thoughts?

ThinkOpenly: The idea was for I-x-x-op to be a subtype of I, inherit the 2 opcodes that I has (funct3, opcode), and add it's own (funct12). Is that possible?

dhower-qc: There are a few assumptions built in that you inherit all the type's opcode locations (e.g., in the CLI tool to create an instruction). It can of course be changed, but first we need to decide:

What does it mean to be of an "Instruction Type"? I struggle to see much benefit of the type buckets if we let everything change.

PR #918, opened 2025-07-22
read-only-zero case when CSR disabled - misa.yaml file 

### LINKED ISSUE #576: MISA can be read-only-0
UDB does not account for the fact that the `misa` CSR can be read-only-0. This has two consequences:

- The misa.yaml file needs updated to reflect the option.
- All uses of `CSR[misa].FIELD` in IDL need attention, since the value cannot be relied upon.

To fix this, we'll need a new parameter in `Sm` and the above issues addressed.

jevinjojo:
## Issue 

- Fixes #576

## Proposed Changes

- All CSR fields always return 0 when misa CSR = false
- writes Ignored when CSR disabled


jevinjojo:
@ThinkOpenly @dhower-qc Please take a look

spec/std/isa/csr/misa.yaml
```diff
@@ -20,7 +24,7 @@ fields:
     description: XLEN in M-mode.
     type: RO
     reset_value(): |
-      return (MXLEN == 32) ? 2'b01 : 2'b10;
+      return (!MISA_CSR_IMPLEMENTED) ? 0 : (MXLEN == 32) ? 2'b01 : 2'b10;
```
dhower-qc: I would suggest either adding parenthesis or converting this to `if (MISA_CSR_IMPLEMENTED) { .. } else { ... }` for clarity.

(applies many times in this PR)

spec/std/isa/csr/misa.yaml
```diff
@@ -261,4 +302,4 @@ cert_test_procedures:
       . Loop
       .. Turn off each present bit invidually and try affected behaviors
       . Check
-      .. Fail unless turning off bit disables extension as expected
+      .. Fail unless turning off bit disables extension as expected
```
dhower-qc: You'll need a newline at the end for regressions to pass

dhower-qc (changes requested):
Thanks for the contribution! Just a few small nits

dhower-qc:
Partially addresses #576 ; still leaves checks of CSR[misa].extension without checking for read-only-0 status.

jevinjojo:
> Partially addresses #576 ; still leaves checks of CSR[misa].extension without checking for read-only-0 status.

@dhower-qc @ThinkOpenly Shall I go ahead with misa field update, so this can complete the remaining work

spec/std/isa/csr/misa.yaml
jevinjojo: should we also convert those existing ternary conditions to if-else blocks.

jevinjojo: should we also convert those existing ternary conditions to if-else blocks.

ThinkOpenly: > should we also convert those existing ternary conditions to if-else blocks.

leave the existing ternaries, please.

ThinkOpenly:
> > Partially addresses #576 ; still leaves checks of CSR[misa].extension without checking for read-only-0 status.
> 
> @dhower-qc @ThinkOpenly Shall I go ahead with misa field update, so this can complete the remaining work

```
$ grep 'CSR\[misa\]' $(find spec/std/isa -type f) | wc -l
558
```

@jevinjojo, you mean you don't want to change all 558 occurences where `misa` is read? :wink:

@dhower-qc, I presume the tedious fix would be to change all of them from:
```
CSR[misa].X == value
```
to something like:
```
(!MISA_CSR_IMPLEMENTED) || (CSR[misa].X == value)
```
Basically, let everything "pass" if `misa` is read-only-0, right?

Do we want a more maintainable implementation? I can't think of a great one offhand.

jevinjojo:
@ThinkOpenly @dhower-qc No 😄, I was referring the phase 3 from our issue breakdown

jevinjojo:
@ThinkOpenly @dhower-qc added the guards to CSR[misa] reference in the core specs and CSR yaml files, also updated the ruby code-gen template. take a look

spec/std/isa/csr/cycle.yaml
```diff
@@ -45,7 +45,7 @@ sw_read(): |
       raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
     }
   } else if (mode() == PrivilegeMode::U) {
-    if (CSR[misa].S == 1'b1) {
+    if ((!MISA_CSR_IMPLEMENTED) || ((!MISA_CSR_IMPLEMENTED) || (CSR[misa].S == 1'b1))) {
```
ThinkOpenly: This seems redundant. Can it be simplified to:
```suggestion
    if (!MISA_CSR_IMPLEMENTED || CSR[misa].S == 1'b1) {
```
Here and elsewhere, of course.

spec/std/isa/csr/misa.yaml
```diff
@@ -29,9 +37,17 @@ fields:
       [when,"MUTABLE_MISA_A == true"]
       Writing 0 to this field will cause all atomic instructions to raise an `IllegalInstruction` exception.
     type(): |
-      return (implemented?(ExtensionName::A) && MUTABLE_MISA_A) ? CsrFieldType::RW : CsrFieldType::RO;
+      if (MISA_CSR_IMPLEMENTED) {
+        return (implemented?(ExtensionName::A) && MUTABLE_MISA_A) ? CsrFieldType::RW : CsrFieldType::RO;
+      } else {
+        return CsrFieldType::RO;
+      }
```
ThinkOpenly: Could you use the slightly simpler implementation below, here and elsewhere?
```suggestion
      if (MISA_CSR_IMPLEMENTED && implemented?(ExtensionName::A) && MUTABLE_MISA_A) {
        return CsrFieldType::RW;
      } else {
        return CsrFieldType::RO;
      }
```

spec/std/isa/csr/misa.yaml
```diff
@@ -182,18 +267,35 @@ fields:
...
+        return MUTABLE_MISA_Q ? CsrFieldType::RW : CsrFieldType::RO;
+      } else {
+        return CsrFieldType::RO;
+      }
+    reset_value(): |
+      if (MISA_CSR_IMPLEMENTED) {
+        return 1;
+      } else {
+        return 0;
+      }
```
ThinkOpenly: This is reasonably replaced with a ternary.
```suggestion
      return (MISA_CSR_IMPLEMENTED) ? 1 : 0;
```

spec/std/isa/csr/misa.yaml
```diff
@@ -214,38 +324,60 @@ fields:
...
+      (CSR[misa].H << 7) |
+      ((CSR[misa].A & CSR[misa].M & CSR[misa].F & CSR[misa].D) << 6) | # 'G'
+      (CSR[misa].F << 5) |
+      (CSR[misa].D << 3) |
+      (CSR[misa].C << 2) |
+      (CSR[misa].B << 1) |
+      CSR[misa].A);
+  } else {
+    return 0;
+  }
```
ThinkOpenly: There's no need to change the indentation here if you insert a test/return at the beginning:
```suggestion
  if (!MISA_CSR_IMPLEMENTED) {
    return 0;
  }
  return (
    (CSR[misa].MXL << (xlen() - 2)) |
    (CSR[misa].V << 21) |
    (CSR[misa].U << 20) |
    (CSR[misa].S << 18) |
    (CSR[misa].M << 12) |
    (CSR[misa].I << 8) |
    (CSR[misa].H << 7) |
    ((CSR[misa].A & CSR[misa].M & CSR[misa].F & CSR[misa].D) << 6) | # 'G'
    (CSR[misa].F << 5) |
    (CSR[misa].D << 3) |
    (CSR[misa].C << 2) |
    (CSR[misa].B << 1) |
    CSR[misa].A);
```

spec/std/isa/ext/Sm.yaml
```diff
@@ -455,6 +455,15 @@ params:
...
       type: boolean
+      default: true
+  MISA_READ_ONLY_ZERO:
+    description: |
+      When true, indicates the `misa` CSR is hardwired to read-only zero.
+      When false, the `misa` CSR behaves as a standard WARL register.
+      Only applicable when MISA_CSR_IMPLEMENTED == false.
+    schema:
+      type: boolean
+      default: false
```
ThinkOpenly: I don't think this is needed, as it is redundant with MISA_CSR_IMPLEMENTED.

tools/ruby-gems/udb/lib/udb/obj/csr.rb
```diff
@@ -325,13 +325,13 @@ def max_length
   def length_cond32
     case @data["length"]
     when "MXLEN"
-      "CSR[misa].MXL == 0"
+      "(!MISA_CSR_IMPLEMENTED) || (CSR[misa].MXL == 0)"
```
ThinkOpenly: If `misa` is not implemented, it is read-only-0. Since this is just testing against zero, you need not test to see if it's implemented.

Same comment applies anywhere the test is with zero (equal or not equal).

ThinkOpenly (changes requested):
I think this is very close to complete. Thank you for your efforts.

I suggest a few ways to make the changes simpler.

spec/std/isa/csr/cycle.yaml
dhower-qc: Almost. We can't just give a 'pass' when misa is read-only-0. In that case, we have to check 'implemented?(ExtensionName::NAME)'. When misa is not read-only-0, having the bit set implies the extension is implemented (and has not been dynamically disabled). When misa is read-only-0, the misa bit does not give any information on the extension.

ThinkOpenly: Ah. Understood. If `misa` is not implemented, don't just avoid it, grab the setting from the configuration. So, for this instance:
```suggestion
    if ((!MISA_CSR_IMPLEMENTED && implemented?(ExtensionName::S) || ((!MISA_CSR_IMPLEMENTED) || (CSR[misa].S == 1'b1))) {
```

jevinjojo:
@ThinkOpenly @dhower-qc I think it's almost done :)

spec/std/isa/csr/misa.yaml
```diff
@@ -86,38 +126,62 @@ fields:
...
+        return implemented?(ExtensionName::F) ? 1 : 0;
+      } else {
+        return 0;
+      }
     definedBy: F
     sw_write(csr_value): |
-      if (csr_value.F == 0 && csr_value.D == 1) {
-        return UNDEFINED_LEGAL_DETERMINISTIC;
-      }
+      if (MISA_CSR_IMPLEMENTED) {
```
ThinkOpenly: @dhower-qc, is this test needed? In the `!MISA_CSR_IMPLEMENTED` case, all fields are marked read-only, so that means we never call the `sw_write()` method, is that correct?

spec/std/isa/csr/misa.yaml
```diff
@@ -182,18 +263,31 @@ fields:
...
-      # fall-through; write the intended value
-      return csr_value.Q;
+        # fall-through; write the intended value
+        return csr_value.Q;
+      } else {
+        return 0;
+      }
     legal?(csr_value): |
-      return !(csr_value.Q == 1 && csr_value.D == 0);
+      if (MISA_CSR_IMPLEMENTED) {
```
ThinkOpenly: Similar question to above, is this test needed in the `!MISA_CSR_IMPLEMENTED` case? (Do we ever need to test for legal values of a CSR if that CSR is read-only?)

spec/std/isa/csr/misa.yaml
```diff
@@ -214,34 +316,56 @@ fields:
...
+        return 0;
+      }
     definedBy: U
   V:
     location: 21
     description: |
       Indicates support for the `V` (vector) extension.
 
-      [when,"MUTABLE_MISA_V == true"]
+      [when,"MUTABLE_MISA_V == true && MISA_CSR_IMPLEMENTED == true"]
```
ThinkOpenly: I imagine we don't need this, since if any bits of the CSR are mutable, the CSR must be implemented. I'm not sure if we enforce that, though, which we should. To do that, we probably need some "EXTRA_VALIDATION" in `Sm.yaml` somewhere to check if any MUTABLE parameters are true when IMPLEMENTED is false. Is that something you want to include here, or shall we open a new issue?

spec/std/isa/ext/Sm.yaml
```diff
@@ -455,7 +455,8 @@ params:
         The `misa` CSR is read-only-0.
     schema:
       type: boolean
-  MTVEC_MODES:
+      default: true
+MTVEC_MODES:
```
ThinkOpenly: Is this related? I think this should be removed from this PR.

spec/std/isa/ext/Zhinx.yaml
ThinkOpenly, line 1: Why is this showing up in this PR?

spec/std/isa/isa/fp.idl
```diff
@@ -138,7 +138,7 @@ function check_f_ok {
     raise an exception.
   }
   body {
-    if (MUTABLE_MISA_F && CSR[misa].F == 0) {
+    if (MUTABLE_MISA_F && ((!MISA_CSR_IMPLEMENTED && !implemented?(ExtensionName::F)) || (CSR[misa].F == 0))) {
```
ThinkOpenly: This change is unnecessary. If the F bit is mutable, then the CSR is implemented.

spec/std/isa/isa/globals.isa
```diff
@@ -694,7 +694,7 @@ function raise_precise {
       $pc = {CSR[mtvec].BASE, 2'b00};
       CSR[mcause].INT = 1'b0;
       CSR[mcause].CODE = $bits(exception_code);
-      if (CSR[misa].H == 1) {
+      if ((!MISA_CSR_IMPLEMENTED) || (CSR[misa].H == 1)) {
```
ThinkOpenly: This needs the longer test:
```suggestion
      if ((!MISA_CSR_IMPLEMENTED && implemented?(ExtensionName::H)) || (CSR[misa].H == 1)) {
```

spec/std/isa/isa/globals.isa
```diff
@@ -714,7 +714,7 @@ function raise_precise {
         }
       }
       CSR[mstatus].MPP = $bits(from_mode);
-    } else if (CSR[misa].S == 1 && (handling_mode == PrivilegeMode::S)) {
+    } else if ((!MISA_CSR_IMPLEMENTED) || (CSR[misa].S == 1 )&& (handling_mode == PrivilegeMode::S)) {
```
ThinkOpenly: This needs the longer test, too.

spec/std/isa/isa/globals.isa
```diff
@@ -723,7 +723,7 @@ function raise_precise {
       CSR[scause].INT = 1'b0;
       CSR[scause].CODE = $bits(exception_code);
       CSR[mstatus].SPP = $bits(from_mode)[0];
-      if (CSR[misa].H == 1) {
+      if ((!MISA_CSR_IMPLEMENTED) || (CSR[misa].H == 1)) {
```
ThinkOpenly: This needs the longer test, too.

spec/std/isa/isa/globals.isa
```diff
@@ -752,7 +752,7 @@ function raise_precise {
           CSR[hstatus].GVA = 0;
         }
       }
-    } else if (CSR[misa].H == 1 && (handling_mode == PrivilegeMode::VS)) {
+    } else if ((!MISA_CSR_IMPLEMENTED) || (CSR[misa].H == 1 )&& (handling_mode == PrivilegeMode::VS)) {
```
ThinkOpenly: longer test

spec/std/isa/isa/globals.isa
```diff
@@ -874,9 +874,9 @@ function xlen {
       return 32;
     } else {
       if (mode() == PrivilegeMode::M) {
-        if (CSR[misa].MXL == $bits(XRegWidth::XLEN32)) {
+        if ((!MISA_CSR_IMPLEMENTED) || (CSR[misa].MXL == $bits(XRegWidth::XLEN32))) {
           return 32;
-        } else if (CSR[misa].MXL == $bits(XRegWidth::XLEN64)) {
+        } else if ((!MISA_CSR_IMPLEMENTED) || (CSR[misa].MXL == $bits(XRegWidth::XLEN64))) {
           return 64;
         }
```
ThinkOpenly: I'm not quite sure here, but pretty sure this needs to change. This always returns 32 if `!MISA_CSR_IMPLEMENTED`. @dhower-qc ?

spec/std/isa/isa/globals.isa
```diff
@@ -1161,15 +1161,15 @@ function effective_ldst_mode {
     # when the mode is M, loads and stores can be executed as if they were done from any other mode
     # with the use of mstatus.MPRV
     if (mode() == PrivilegeMode::M) {
-      if (CSR[misa].U == 1 && CSR[mstatus].MPRV == 1) {
+      if ((!MISA_CSR_IMPLEMENTED) || (CSR[misa].U == 1 )&& CSR[mstatus].MPRV == 1) {
```
ThinkOpenly: this needs a longer test, and be careful with how the logic is grouped. Using `A || B && C` could be confusing (and seems to be incorrect here).

spec/std/isa/isa/globals.isa
```diff
@@ -1161,15 +1161,15 @@ function effective_ldst_mode {
...
         if (CSR[mstatus].MPP == 0b00) {
-          if (CSR[misa].H == 1 && mpv() == 0b1) {
+          if ((!MISA_CSR_IMPLEMENTED) || (CSR[misa].H == 1 )&& mpv() == 0b1) {
```
ThinkOpenly: same

spec/std/isa/isa/globals.isa
```diff
@@ -1161,15 +1161,15 @@ function effective_ldst_mode {
...
             return PrivilegeMode::VU;
           } else {
             return PrivilegeMode::U;
           }
-        } else if (CSR[misa].S == 1 && CSR[mstatus].MPP == 0b01) {
-          if (CSR[misa].H == 1 && mpv() == 0b1) {
+        } else if ((!MISA_CSR_IMPLEMENTED) || (CSR[misa].S == 1 )&& CSR[mstatus].MPP == 0b01) {
+          if ((!MISA_CSR_IMPLEMENTED) || (CSR[misa].H == 1 )&& mpv() == 0b1) {
```
ThinkOpenly: same... there are more cases below where you are just checking for `MISA_CSR_IMPLEMENTED`, but you also need to check `implemented?...`, OK?

ThinkOpenly (changes requested):
I submitted a bunch of "single comments", above, rather than a single review. Hopefully it all makes sense.

spec/std/isa/csr/misa.yaml
jevinjojo: in the `Sm.yaml` file an extra_validation check under MISA_CSR_IMPLEMENTED and validation verifies that if any MISA bit (A-Z) is also need to marked as mutable. do we need to add error message if the condition violated

spec/std/isa/ext/Zhinx.yaml
jevinjojo: my bad! I accidentally included when I staged changes.

spec/std/isa/isa/globals.isa
jevinjojo: okay

spec/std/isa/csr/misa.yaml
jevinjojo: When MISA_CSR_IMPLEMENTED is false, the CSR is read-only-0, so no writes can actually occur. any write value is legal?

spec/std/isa/csr/misa.yaml
ThinkOpenly: > in the `Sm.yaml` file an extra_validation check under MISA_CSR_IMPLEMENTED and validation verifies that if any MISA bit (A-Z) is also need to marked as mutable.

I'm leaning toward adding the extra_validation in the extensions where the mutability parameter is defined. For example, in `spec/std/isa/ext/A.yaml`, under MUTABLE_MISA_A add something like:
```
  extra_validation: assert MISA_CSR_IMPLEMENTED == true
```
(untested). In that way, "everybody" only needs to know that one "implemented" parameter, rather than `Sm.yaml` having to know every other "mutability" parameter.

> do we need to add error message if the condition violated

No, that will come automatically when the architecture is resolved.

spec/std/isa/csr/misa.yaml
ThinkOpenly: I want to assert that in the read-only-0 case, we don't need to test for legal values because _no_ values will ever be accepted/written. I don't think you need any new code here.

spec/std/isa/csr/misa.yaml
ThinkOpenly: I think we need a little flow chart as to how "writable" is checked. I suspect something like:
1. "writable" attribute for the CSR
1. "type"/"type()" attribute/method for the field
1. "legal?()" method
1. "sw_write()" method logic

If that's reasonable accurate, then we won't get to the "sw_write()" method in the read-only-0 case, because the write is prevented by the "type()" method. So, no special code is needed here.

dhower-qc: > @dhower-qc, is this test needed? In the `!MISA_CSR_IMPLEMENTED` case, all fields are marked read-only, so that means we never call the `sw_write()` method, is that correct?

correct

dhower-qc: > I think we need a little flow chart as to how "writable" is checked. I suspect something like:
> 
> 1. "writable" attribute for the CSR
> 2. "type"/"type()" attribute/method for the field
> 3. "legal?()" method
> 4. "sw_write()" method logic
> 
> If that's reasonable accurate, then we won't get to the "sw_write()" method in the read-only-0 case, because the write is prevented by the "type()" method. So, no special code is needed here.

3. isn't actually used anywhere. we should remove it. Its purpose is handled through sw_write/sw_read, which together try to define what happens if the write is illegal.

Otherwise, the order is correct (1-2-4)

spec/std/isa/csr/misa.yaml
dhower-qc: As I said above, let's just remove the legal? field

spec/std/isa/isa/globals.isa
dhower-qc: yea, this can just check that the MXLEN parameter is 32/64. even when misa is implemented, MXL is read-only (this was a backward-looking, incompatible change made to the spec)

dhower-qc:
Keep up the good work!

spec/std/isa/ext/A.yaml
```diff
@@ -128,3 +128,4 @@ params:
       the extension can be disabled in the `misa.A` bit.
     schema:
       type: boolean
+      extra_validation: assert MISA_CSR_IMPLEMENTED == true
```
ThinkOpenly: Unindent one level to be an attribute of the field (not the "schema"):
```suggestion
    extra_validation: assert MISA_CSR_IMPLEMENTED == true
```

spec/std/isa/ext/B.yaml
```diff
@@ -43,3 +43,4 @@ params:
       Indicates whether or not the `B` extension can be disabled with the `misa.B` bit.
     schema:
       type: boolean
+      extra_validation: assert MISA_CSR_IMPLEMENTED == true
```
ThinkOpenly:
```suggestion
    extra_validation: assert MISA_CSR_IMPLEMENTED == true
```

spec/std/isa/ext/C.yaml
```diff
@@ -319,3 +319,4 @@ params:
       Indicates whether or not the `C` extension can be disabled with the `misa.C` bit.
     schema:
       type: boolean
+      extra_validation: assert MISA_CSR_IMPLEMENTED == true
```
ThinkOpenly:
```suggestion
    extra_validation: assert MISA_CSR_IMPLEMENTED == true
```
And more below. I won't mark them all.

spec/std/isa/ext/Sm.yaml
```diff
@@ -455,7 +455,7 @@ params:
         The `misa` CSR is read-only-0.
     schema:
       type: boolean
-  MTVEC_MODES:
+MTVEC_MODES:
```
ThinkOpenly: accidental change? please revert.

spec/std/isa/ext/Zhinx.yaml
ThinkOpenly, line 1: this is unexpected... any idea why this is showing up? I don't see any changes...?

spec/std/isa/isa/globals.isa
```diff
@@ -878,9 +878,9 @@ function xlen {
...
     } else {
       if (mode() == PrivilegeMode::M) {
-        if (CSR[misa].MXL == $bits(XRegWidth::XLEN32)) {
+        if (MXLEN == 32) {
           return 32;
-        } else if (CSR[misa].MXL == $bits(XRegWidth::XLEN64)) {
+        } else if (MXLEN == 64) {
           return 64;
         }
       } else if (implemented?(ExtensionName::S) && mode() == PrivilegeMode::S) {
```
ThinkOpenly: I think this section can be simplified:
```
    } else if (mode() == PrivilegeMode::M) {
      return MXLEN;
    } else if (implemented?(ExtensionName::S) && mode() == PrivilegeMode::S) {
...
```
with the rest of the sections also unindented one level, if that makes sense.

spec/std/isa/isa/interrupts.idl
```diff
@@ -338,12 +338,8 @@ function take_interrupt {
...
       CSR[mepc].PC = $pc;
       CSR[mstatus].MPP = $bits(mode())[1:0];
-      if (CSR[misa].H == 1'b1) {
-        if(MXLEN == 64) {
-          CSR[mstatus].MPV = $bits(mode())[2];
-        } else {
-          CSR[mstatush].MPV = $bits(mode())[2];
-        }
+      if ((!MISA_CSR_IMPLEMENTED && implemented?(ExtensionName::H)) || (CSR[misa].H == 1'b1)) {
+        CSR[mstatus].MPV = $bits(mode())[2];
```
ThinkOpenly: The original code can't be simplified here. Notice that the "then" block is using `mstatus` and the else block is using `mstatush`, so you'll need to revert that part of this change, but keep the "MISA_CSR_IMPLEMENTED" change.

ThinkOpenly (changes requested):
Some minor, but important changes.

spec/std/isa/ext/Zhinx.yaml
jevinjojo: I think it was due to line endings or whitespace normalization.

jevinjojo:
@ThinkOpenly @dhower-qc Any changes required.

spec/std/isa/csr/misa.yaml
```diff
@@ -8,10 +8,14 @@ kind: csr
...
 address: 0x301
+# writable when misa CSR is implemented; read-only-0 otherwise
 writable: true
 priv_mode: M
 length: MXLEN
-description: Reports the XLEN and "major" extensions supported by the ISA.
+description: |
+  Reports the XLEN and "major" extensions supported by the ISA.
+  [when,"MISA_CSR_IMPLEMENTED == false"] 
+  This CSR is read-only-0 (all bits hardwired to zero) and cannot be written.
```
dhower-qc, line 18: We've update how this is handled:


```suggestion
  - id: csr-misa-purpose
    text: |
      Reports the XLEN and "major" extensions supported by the ISA.
    when:
      param:
        name: MISA_CSR_IMPLEMENTED
        value: true
  - id: csr-misa-unimplemented
    text: |
      This CSR is read-only-0 (all bits hardwired to zero) and cannot be written.
    when:
      param:
        name: MISA_CSR_IMPLEMENTED
        value: false
```

Note that the schema will reject this until #891 ships

spec/std/isa/ext/A.yaml
```diff
@@ -128,3 +128,4 @@ params:
       the extension can be disabled in the `misa.A` bit.
     schema:
       type: boolean
+    extra_validation: assert MISA_CSR_IMPLEMENTED == true
```
dhower-qc, line 131: MISA_CSR_IMPLEMENTED only needs to be true if MUTABLE_MISA_A is true:


```suggestion
    extra_validation: assert MISA_CSR_IMPLEMENTED == true if MUTABLE_MISA_A
```

spec/std/isa/ext/Zhinx.yaml
dhower-qc, line 1: still here?

spec/std/isa/isa/globals.isa
```diff
@@ -876,39 +876,34 @@ function xlen {
...
+      if (CSR[hstatus].VSXL == $bits(XRegWidth::XLEN32)) {
+        return 32;
+      } else if (CSR[hstatus].VSXL == $bits(XRegWidth::XLEN64)) {
+        return 64;
+      }
+    } else if (implemented?(ExtensionName::H) && mode() == PrivilegeMode::VU) {
+      if (CSR[vsstatus].UXL == $bits(XRegWidth::XLEN32)) {
+        return 32;
+      } else if (CSR[vsstatus].UXL == $bits(XRegWidth::XLEN64)) {
+        return 64;
```
dhower-qc, line 903: Why is this diff here?

spec/std/isa/isa/globals.isa
```diff
@@ -876,39 +876,34 @@ function xlen {
...
       }
     }
+
```
dhower-qc, line 906: remove

tools/ruby-gems/udb/lib/udb/obj/csr.rb
```diff
@@ -341,7 +341,7 @@ def length_cond32
   def length_cond64
     case @data["length"]
     when "MXLEN"
-      "CSR[misa].MXL == 1"
+      "(!MISA_CSR_IMPLEMENTED) || (CSR[misa].MXL == 1)"
```
dhower-qc, line 344:
```suggestion
      "MXLEN == 64"
```

What you wrote isn't quite right, since it doesn't tell you what MXLEN is when misa isn't implemented. MXLEN can never change, so we don't need to consult misa at all here

dhower-qc (changes requested):
Great work. Still a bit more to do :) Check for references in spec/**/inst/*.yaml for misa; there should be a lot.

spec/std/isa/ext/Zhinx.yaml
jevinjojo: I checked with git diff to see if there line ending/permission changes, also tried resetting it.  but didn't work.  I'm not sure can you tell me what might be causing this?

jevinjojo:
@dhower-qc @ThinkOpenly Will do the changes in a few days.

ThinkOpenly:
Should we expect additional changes here?

spec/std/isa/csr/misa.yaml
ThinkOpenly: This is certainly obsolete, as we abandoned the "tagged text" used here.

Issue #361, opened 2024-12-18
Acquire and release instructions have a bug 


AFOliveira:
**Describe the bug**
The bug is the aq/rl bit not being correctly defined. This is due to my original parsing of instructions coming from riscv-opcodes _db_ and not from the generated .json it creates.

I'm fixing this and I'll PR soon.

AFOliveira:
Extending on this problem:

riscv-opcodes keeps acquire and release bits always as variable on atomic operations. This is compliant with the ISA standard description of this instructions:
![image](https://github.com/user-attachments/assets/d47b8735-379e-4953-ae36-40535988a55d)


However, LLVM is much more strict on these instructions, and separates them as, per example AMOADD.H - AQ/RL = 00 ---AMOADD.H.AQ - AQ/RL = 10 --- AMOADD.H.RL - AMOADD.H.RL = 01.

I think we could keep this as aq/rl variables for the purpose of the UDB being a DB and extend if we need, but since this caught my eye due to failing the #356 test, I want to ask for your opinion.

ThinkOpenly:
There is more to this. The AMO instructions include the "aq" and "rl" specification in the _mnemonics_. Taking just `amoadd.w` as an example, it actually has four associated mnemonics:
- `amoadd.w`
- `amoadd.w.aq`
- `amoadd.w.rl`
- `amoadd.w.aq.rl`

AFOliveira:
Then should we split the mnemonics from the variables? 

From what I can grasp, it is only one instruction that has several mnemonics. I think this is a hard-line on SW vs HW view of instructions. While for software they can be considered many, I believe on HW they are only one. Would it make sense if we change the assembly to acommodate several mnemonics?

lenary:
The problem is only partly the assembler, honestly.

The "orderings" are really another kind of operand that has been spliced onto the mnemonic - note how `fence` has ordering operands which aren't in the mnemonic, same goes for floating rounding modes which aren't in the mnemonic. In RISC-V, this is the only place where an assembler would need to slice up the first token to split a mnemonic (the instruction) from its operands.

Knowing when to do this slicing and when not to becomes hard, for an arbitrary mnemonic parsed from an assembly file.

I would highly suggest keeping them separate, I don't see the problem of them having identical pseudo-code, but with `aq`, `rl` parameters to that pseudocode set to constants.

This will make building a reference assembler from the riscv-unified-db much easier in the long run.

dhower-qc:
[...]

PR #921, opened 2025-07-23
feat: add amo instruction layout system for zaamo extension

### LINKED ISSUE #223: Add Zabha (Byte and Halfword Atomics)
**Describe the data**

Add isa spec and IDL

Sukuna0007Abhi:
**Add AMO Instruction Layout System for Zaamo Extension** 

fixes #223 

This PR implements a clean layout-based system for generating Atomic Memory Operation (AMO) instructions for the Zaamo extension, combining approaches suggested by @ddrcode and @ThinkOpenly.

✨ What's New
Layout Templates: Created 9 ERB templates (one per AMO operation) that automatically generate both word and doubleword variants
Organized Structure: Added layouts/ folder for templates and instructions/ folder for generated files
Automated Generation: Updated Rakefile to generate all 18 AMO instruction files from the layout templates
Better Descriptions: Each instruction now has proper long names and bullet-point descriptions
Schema Fixes: Fixed Zaamo.yaml extension definition with correct schema path

🎯 Result
Instead of manually maintaining 18 separate instruction files, we now have:

9 maintainable layout templates
Automatic generation of all AMO variants (.w and .d)
Consistent formatting and descriptions across all instructions
Easy to add new AMO operations in the future

🧹 Cleanup

Also updated .gitignore to exclude the 270+ generated CSR test files that were cluttering the repo.

Ready for review! 🎉 @AFOliveira @ThinkOpenly @dhower-qc


ThinkOpenly:
> **Add AMO Instruction Layout System for Zaamo Extension**
> 
> fixes #223

Thanks for your efforts!

> Layout Templates: Created 9 ERB templates (one per AMO operation) that automatically generate both word and doubleword variants

Sounds good, except that #223 is for Byte and Halfword atomics.

> Organized Structure: Added layouts/ folder for templates and instructions/ folder for generated files

I understand why you thought to do this, but this is not how we've been doing it so far. :slightly_smiling_face: The file extensions are sufficient to distinguish between layouts and instruction YAML. Let's leave everything where it is for now.

> Automated Generation: Updated Rakefile to generate all 18 AMO instruction files from the layout templates

OK

> 🎯 Result Instead of manually maintaining 18 separate instruction files, we now have:
> 
> 9 maintainable layout templates Automatic generation of all AMO variants (.w and .d) 

This is nice. This is also a step towards #361. Take a peek at that and see if that's something you wish to also include in this PR. Basically, all of these AMO instructions have not only the "width" variants, but also the "acquire/release" variants. For example, `amoadd` ends up with:
- `amoadd.b`, `amoadd.b.aq`, `amoadd.b.rl`, `amoadd.b.aq.rl`
- `amoadd.h`, `amoadd.h.aq`, `amoadd.h.rl`, `amoadd.h.aq.rl`
- `amoadd.w`, `amoadd.w.aq`, `amoadd.w.rl`, `amoadd.w.aq.rl`
- `amoadd.d`, `amoadd.d.aq`, `amoadd.d.rl`, `amoadd.d.aq.rl`
:exclamation: 

> Also updated .gitignore to exclude the 270+ generated CSR test files that were cluttering the repo.

These are files that we want to keep. The general process for "layout" files is to commit both the layout file and the generated files.

Sukuna0007Abhi:
> > **Add AMO Instruction Layout System for Zaamo Extension**
> > fixes #223
> 
> Thanks for your efforts!
> Thanks 
> > Layout Templates: Created 9 ERB templates (one per AMO operation) that automatically generate both word and doubleword variants
> 
> Sounds good, except that #223 is for Byte and Halfword atomics.
> 
> > Organized Structure: Added layouts/ folder for templates and instructions/ folder for generated files
> 
> I understand why you thought to do this, but this is not how we've been doing it so far. 🙂 The file extensions are sufficient to distinguish between layouts and instruction YAML. Let's leave everything where it is for now.
> Yes sir I thought the both layout flies and the instruction YAML will be mixed up that's but ok sir I will change that by today and should I leave the layout folder as it is and only deleting instruction folder.

> > Automated Generation: Updated Rakefile to generate all 18 AMO instruction files from the layout templates
> 
> OK
> 
> > 🎯 Result Instead of manually maintaining 18 separate instruction files, we now have:
> > 9 maintainable layout templates Automatic generation of all AMO variants (.w and .d)
> 
> This is nice. This is also a step towards #361. Take a peek at that and see if that's something you wish to also include in this PR. Basically, all of these AMO instructions have not only the "width" variants, but also the "acquire/release" variants. For example, `amoadd` ends up with:
> 
> * `amoadd.b`, `amoadd.b.aq`, `amoadd.b.rl`, `amoadd.b.aq.rl`
> * `amoadd.h`, `amoadd.h.aq`, `amoadd.h.rl`, `amoadd.h.aq.rl`
> * `amoadd.w`, `amoadd.w.aq`, `amoadd.w.rl`, `amoadd.w.aq.rl`
> * `amoadd.d`, `amoadd.d.aq`, `amoadd.d.rl`, `amoadd.d.aq.rl`
>   ❗
> 
> > Also updated .gitignore to exclude the 270+ generated CSR test files that were cluttering the repo.

> These are files that we want to keep. The general process for "layout" files is to commit both the layout file and the generated files.
> Ok then I will change the gitignore then both files will be there

Sukuna0007Abhi:
> > **Add AMO Instruction Layout System for Zaamo Extension**
> > fixes #223
> 
> Thanks for your efforts!
> Thanks 
> > Layout Templates: Created 9 ERB templates (one per AMO operation) that automatically generate both word and doubleword variants
> 
> Sounds good, except that #223 is for Byte and Halfword atomics.
> 
> > Organized Structure: Added layouts/ folder for templates and instructions/ folder for generated files
> 
> I understand why you thought to do this, but this is not how we've been doing it so far. 🙂 The file extensions are sufficient to distinguish between layouts and instruction YAML. Let's leave everything where it is for now.
> Yes sir I thought the both layout flies and the instruction YAML will be mixed up that's but ok sir I will change that by today and should I leave the layout folder as it is and only deleting instruction folder.

> > Automated Generation: Updated Rakefile to generate all 18 AMO instruction files from the layout templates
> 
> OK
> 
> > 🎯 Result Instead of manually maintaining 18 separate instruction files, we now have:
> > 9 maintainable layout templates Automatic generation of all AMO variants (.w and .d)
> 
> This is nice. This is also a step towards #361. Take a peek at that and see if that's something you wish to also include in this PR. Basically, all of these AMO instructions have not only the "width" variants, but also the "acquire/release" variants. For example, `amoadd` ends up with:
> 
> * `amoadd.b`, `amoadd.b.aq`, `amoadd.b.rl`, `amoadd.b.aq.rl`
> * `amoadd.h`, `amoadd.h.aq`, `amoadd.h.rl`, `amoadd.h.aq.rl`
> * `amoadd.w`, `amoadd.w.aq`, `amoadd.w.rl`, `amoadd.w.aq.rl`
> * `amoadd.d`, `amoadd.d.aq`, `amoadd.d.rl`, `amoadd.d.aq.rl`
>   ❗
> 
> > Also updated .gitignore to exclude the 270+ generated CSR test files that were cluttering the repo.

> These are files that we want to keep. The general process for "layout" files is to commit both the layout file and the generated files.

> Ok then I will change the gitignore then both files will be there

Sukuna0007Abhi:
Hi @ThinkOpenly,

I've implemented the AMO layout system with your byte/halfword approach and issue #361's acquire/release variants. The 9 layout files now generate all 144 instruction combinations (base, .aq, .rl, .aq.rl) automatically, replacing manual YAML maintenance. 

And also did the gitignore part and rewrite the folder structure 

Your conditional descriptions for "least-significant X of register" vs "value of register" are working perfectly across all operations.

Ready for Re-review! 🚀

Sukuna0007Abhi:
And @ThinkOpenly @dhower-qc @AFOliveira from the issue #361 I implemented the solution by these steps: 
Issue #361 Implementation:

Added acquire/release memory ordering variants to AMO instructions:

Generate 4 variants per operation: base, .aq, .rl, .aq.rl
Fixed bit encoding (hardcoded aq/rl bits instead of variables)
Dynamic instruction naming and descriptions based on memory ordering flags
Complete 144 instruction coverage (9 ops × 4 sizes × 4 aq/rl variants)

jordancarlin:
> I've implemented the AMO layout system with your byte/halfword approach (it will now work for 4 variants b, h w, d) and issue #361's acquire/release variants. The 9 layout files now generate all 144 instruction combinations (base, .aq, .rl, .aq.rl) automatically, replacing manual YAML maintenance.

Minor but important correction: the four suffixes should be nothing, `.aq`, `.rl`, and `.aqrl` (not `.aq.rl`).

Sukuna0007Abhi:
Sure,updating by today @ThinkOpenly 
& Thanks for suggesting @jordancarlin

Sukuna0007Abhi:
hi sir @ThinkOpenly and @jordancarlin I updated that and now all suffixes now follow the correct convention: .aqrl (not .aq.rl) for combined aq/rl. Thanks for catching this! Ready for re-review. @ThinkOpenly @dhower-qc

ThinkOpenly:
@Sukuna0007Abhi, could you check why this PR has 272 file changes? I think there is a lot of stuff here that shouldn't be here.

Also, there are some files missing: when we add layout generation, we commit the generated files here as well. So, while I certainly expect the `.layout` files in `spec/std/isa/inst/Zaamo`, I also want to see the generated `.yaml` files:

> Complete 144 instruction coverage (9 ops × 4 sizes × 4 aq/rl variants)

Sukuna0007Abhi:
Hi @ThinkOpenly sorry for delay I actually got stuck on rebase that's why there are several commits but I can't able to skip it now pls if only skip 29,july commits those are repeated commits and other members commits, that's why.But now the you can check there are not anymore 260+ files only there are required files I did.And now I will generate the files and commit here.Thank you

spec/std/isa/ext/Zaamo.yaml
```diff
@@ -1,12 +1,12 @@
...
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 
-# yaml-language-server: $schema=../../schemas/ext_schema.json
+# yaml-language-server: $schema=../../../schemas/ext_schema.json
 
 $schema: "ext_schema.json#"
 kind: extension
 name: Zaamo
-long_name: Load-acquire/Store-release atomic instructions
+long_name: Atomic Memory Operations
```
AFOliveira: Aren't this changes unrelated to the PR?

spec/std/isa/ext/Zaamo.yaml
```diff
@@ -134,3 +134,23 @@ description: |
   Specific compilation conventions may require both the _aq_ and _rl_
   bits to be set in either or both the LR and AMOSWAP instructions.
   ====
+  ====
+  We recommend the use of the AMO Swap idiom shown above for both lock
+  acquire and release to simplify the implementation of speculative lock
```
AFOliveira: I feel this should be a different PR, as the title suggests only layout introduction

spec/std/isa/inst/Zaamo/amoadd.d.yaml
```diff
@@ -1,24 +1,25 @@
...
 # Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 
 # yaml-language-server: $schema=../../../../schemas/inst_schema.json
 
 $schema: "inst_schema.json#"
 kind: instruction
-name: amoadd.d
-long_name: Atomic fetch-and-add doubleword
+name: amoadd.w
```
AFOliveira: This is wrong given the name of the file.

spec/std/isa/ext/Zaamo.yaml
Sukuna0007Abhi: Ok @AFOliveira I will reset that long_name but I will state the schema error that should be change

spec/std/isa/ext/Zaamo.yaml
Sukuna0007Abhi: So should I remove this? I'm a bit confused about it.

spec/std/isa/inst/Zaamo/amoadd.d.yaml
Sukuna0007Abhi: Thanks for pointing @AFOliveira I accidentally did wrong on that, will resolve by today.

Sukuna0007Abhi:
I updated those parts @AFOliveira thanks for suggesting. And yet ready for Re-review and possibility of merge. @ThinkOpenly @AFOliveira @dhower-qc

Rakefile
```diff
@@ -557,3 +599,19 @@ task "RVA20": "#{$root}/gen/profile/pdf/RVA20ProfileRelease.pdf"
 task "RVA22": "#{$root}/gen/profile/pdf/RVA22ProfileRelease.pdf"
 task "RVA23": "#{$root}/gen/profile/pdf/RVA23ProfileRelease.pdf"
 task "RVB23": "#{$root}/gen/profile/pdf/RVB23ProfileRelease.pdf"
+task "MC100-32-CTP-HTML": "#{$root}/gen/proc_ctp/pdf/MC100-32-CTP.html"
```
dhower-qc: Were these a mistake? I'm not sure why they get duplicated

spec/std/isa/inst/Zaamo/amoadd.d.yaml
```diff
@@ -6,19 +8,18 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: amoadd.d
-long_name: Atomic fetch-and-add doubleword
+long_name: Atomic fetch-and-add word
```
dhower-qc: I think this should stay as double word

dhower-qc (changes requested):
You're on the right path. Looks like a few changes aren't correct, though

Rakefile
```diff
@@ -5,6 +5,9 @@ require "sorbet-runtime"
 T.bind(self, T.all(Rake::DSL, Object))
 extend T::Sig
 
+require 'pathname'
+require 'erb'
```
ThinkOpenly, line 9: Why does _this_ PR need these changes? (I confess I don't know the answer.) If I read it correctly, there is existing code which depends on these. What changed?

Rakefile
```diff
@@ -16,6 +19,9 @@ require "etc"
 $root = Pathname.new(__dir__).realpath
 $lib = $root / "lib"
 
+# Add lib directory to load path
+$LOAD_PATH.unshift($lib) unless $LOAD_PATH.include?($lib)
```
ThinkOpenly: Same question: why does this PR need this change?

Rakefile
```diff
@@ -382,6 +414,16 @@ namespace :gen do
...
+
+    # Generate AMO instructions
+    %w[amoadd amoand amomax amomaxu amomin amominu amoor amoswap amoxor].each do |op|
+      ["b", "h", "w", "d"].each do |size|
+        # Generate all acquire/release variants
+        ["", ".aq", ".rl", ".aqrl"].each do |suffix|
+          Rake::Task["#{$resolver.std_path}/inst/Zaamo/#{op}.#{size}#{suffix}.yaml"].invoke
+        end
+      end
+    end
```
ThinkOpenly: I'm probably showing my ignorance, but this seems redundant with the new code block above, also with the same 3 nested loops. What does this do that the other block doesn't?

spec/std/isa/inst/Zaamo/amoaddN.layout
ThinkOpenly, line 1: I know this is tedious, but the `N` in the file name says "number" to me, but we're varying both the size and the aq/rl semantics. Could you rename all of the `N` files to something like `amo[op].W.AQRL.yaml`?

spec/std/isa/inst/Zaamo/amoaddN.layout
```diff
@@ -0,0 +1,167 @@
...
+  aq_bit = aq ? "1" : "0"
+  rl_bit = rl ? "1" : "0"
+-%>
+
+$schema: "inst_schema.json#"
+kind: instruction
+name: amoadd.<%= size %><%= aq_rl_suffix %>
+long_name: Atomic fetch-and-add <%= current_size[:name] %><%= aq ? " (acquire)" : "" %><%= rl ? " (release)" : "" %><%= aq && rl ? " (acquire-release)" : "" %>
+description: |
+  Atomically<%= aq ? " with acquire ordering" : "" %><%= rl ? " with release ordering" : "" %>:
```
ThinkOpenly: This doesn't read well when both are true:
> Atomically with acquire ordering with release ordering:

As with "long_name", you may want a double ternary where the "both true" case becomes "with acquire and release ordering".

Note that you can pull "ordering" out of the test and put it to the right, since it's in every case.

ThinkOpenly (changes requested):
This is fantastic, and very close to complete! Great work!

It looks like some unnecessary changes leaked in, and I have a few minor requests.

spec/std/isa/inst/Zaamo/amoaddN.layout
Sukuna0007Abhi: Hi sir @ThinkOpenly I'm updating changes but I'm a bit confused about should I rename those layout files like this amoadd.SIZE.AQRL.layout or amoadd[op].SIZE.AQRL.layout

ThinkOpenly: > should I rename those layout files like this amoadd.SIZE.AQRL.layout or amoadd[op].SIZE.AQRL.layout

Thanks for asking. I was trying to figure out the best way to express that, but maybe didn't find the best way. :-)

`amoadd.SIZE.AQRL.layout` is fine.

Rakefile
Sukuna0007Abhi: removed the part @dhower-qc

spec/std/isa/inst/Zaamo/amoadd.d.yaml
Sukuna0007Abhi: updated sir @dhower-qc

Rakefile
Sukuna0007Abhi: Hi sir @ThinkOpenly, These are defensive programming practices to ensure the AMO generation system works reliably.

The technical details:

While the existing codebase does use ERB for CSR templates, the AMO layout system is more complex - it generates 144 instruction variants (9 operations × 4 sizes × 4 acquire/release combinations) using more sophisticated ERB templating with conditional logic and variable binding.

Rather than relying on implicit requires that might exist elsewhere in the codebase, I added explicit requires to ensure:

require 'erb' - The AMO templates can access ERB functionality directly without depending on other parts of the code loading it first
[require 'pathname']File path operations work consistently across different platforms
LOAD_PATH modification - The UDB resolver and related modules are available for the generation tasks
Why be explicit? If someone refactors other parts of the Rakefile in the future, implicit dependencies could break. By being explicit about what the AMO system needs, we make the code more maintainable and less fragile.
and sir - these requires don't hurt anything and make the AMO generation more robust.

Rakefile
Sukuna0007Abhi: You're right @ThinkOpenly I updated that.

spec/std/isa/inst/Zaamo/amoaddN.layout
Sukuna0007Abhi: Updated @ThinkOpenly

spec/std/isa/inst/Zaamo/amoaddN.layout
Sukuna0007Abhi: updated

Sukuna0007Abhi:
Hi sir @ThinkOpenly @dhower-qc all cl checks will pass now except that regress smoke that check will be passed when we use my this PR #954

spec/std/isa/inst/Zaamo/amoadd.b.aq.yaml
```diff
@@ -0,0 +1,135 @@
...
+  s: always
+  u: always
+  vs: always
+  vu: always
+operation(): |
+  if (implemented?(ExtensionName::A) && (CSR[misa].A == 1'b0)) {
+    raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
+  }
+
+  XReg virtual_address = X[xs1];
```
dhower-qc: This also needs to handle the acquire semantics by calling `memory_model_acquire()` before the amo.

spec/std/isa/inst/Zaamo/amoadd.b.aqrl.yaml
```diff
@@ -0,0 +1,135 @@
...
+  u: always
+  vs: always
+  vu: always
+operation(): |
+  if (implemented?(ExtensionName::A) && (CSR[misa].A == 1'b0)) {
+    raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
+  }
+
+  XReg virtual_address = X[xs1];
+  X[xd] = amo<8>(virtual_address, X[xs2][7:0], AmoOperation::Add, 1'b1, 1'b1, $encoding);
```
dhower-qc: need acquire before the amo and release after

spec/std/isa/inst/Zaamo/amoadd.b.rl.yaml
```diff
@@ -0,0 +1,135 @@
...
+  u: always
+  vs: always
+  vu: always
+operation(): |
+  if (implemented?(ExtensionName::A) && (CSR[misa].A == 1'b0)) {
+    raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
+  }
+
+  XReg virtual_address = X[xs1];
+  X[xd] = amo<8>(virtual_address, X[xs2][7:0], AmoOperation::Add, 1'b0, 1'b1, $encoding);
```
dhower-qc: need release after

spec/std/isa/inst/Zaamo/amoadd.b.aq.yaml
```diff
@@ -0,0 +1,135 @@
...
+      location: 19-15
+    - name: xd
+      location: 11-7
+access:
+  s: always
+  u: always
+  vs: always
+  vu: always
+operation(): |
+  if (implemented?(ExtensionName::A) && (CSR[misa].A == 1'b0)) {
```
dhower-qc: I know you copied this from other files (the right thing to do) but....

It's not quite correct. We need to fix the other files, too. The misa CSR is optional, so it really needs to be:

```
if (implemented?(ExtensionName::A) && MISA_CSR_IMPLEMENTED && (CSR[misa].A == 1'b0))
```

Rakefile
dhower-qc: Another take: these are pulled in by other requires. Adding them doesn't hurt and makes things a bit more explicit, so I'm good keeping this change

Rakefile
dhower-qc: This actually isn't needed at all. Everything from `lib` has been moved anyway, so get rid of this.

Rakefile
```diff
@@ -356,6 +362,32 @@ file "#{$resolver.std_path}/csr/Zicntr/mcountinhibit.yaml" => [
   File.write(t.name, insert_warning(erb.result(binding), t.prerequisites.first))
 end
 
+# AMO instruction generation from layouts
+%w[amoadd amoand amomax amomaxu amomin amominu amoor amoswap amoxor].each do |op|
```
dhower-qc, line 363: Nice.

dhower-qc (changes requested):

Rakefile
Sukuna0007Abhi: Thanks sir @dhower-qc

Rakefile
Sukuna0007Abhi: Removing it

spec/std/isa/inst/Zaamo/amoadd.b.aq.yaml
Sukuna0007Abhi: Updated it sir but the issue is no source for misa issue giving another cl check failure regress gen appendix because I can't update the golden.adoc file , so when my #954 PR maybe merged then this misa part issue will be solved.Then i think all cl check will also pass.

spec/std/isa/inst/Zaamo/amoadd.b.rl.yaml
Sukuna0007Abhi: updated it

spec/std/isa/inst/Zaamo/amoadd.b.aqrl.yaml
Sukuna0007Abhi: updated it sir

spec/std/isa/inst/Zaamo/amoadd.b.aq.yaml
Sukuna0007Abhi: updated it sir

Sukuna0007Abhi:
Pls check @ThinkOpenly @dhower-qc

Sukuna0007Abhi:
Hello sir @ThinkOpenly I need some help, I am actually confused a bit for Zaamo it's d and w & for Zabha it's b and h right and I did aqrl,aq & rl variant for both, so should I arrange the yaml files on their exact folders because it all are currently in zaamo, I will arrange regenerate them the b/h yamls in Zabha and d/w yamls in Zaamo. 

And one more thing the layout files it's actually working for b,h,w,d so should I differentiate them same layout in Zaamo but only d/w and same layout in Zabha b/h.

ThinkOpenly:
> Hello sir @ThinkOpenly I need some help, I am actually confused a bit for Zaamo it's d and w & for Zabha it's b and h right

Very good point.

> and I did aqrl,aq & rl variant for both, so should I arrange the yaml files on their exact folders because it all are currently in zaamo, I will arrange regenerate them the b/h yamls in Zabha and d/w yamls in Zaamo.

That makes sense, yes.

> And one more thing the layout files it's actually working for b,h,w,d so should I differentiate them same layout in Zaamo but only d/w and same layout in Zabha but only b/h.

Ideally, we don't replicate code. Can you make it so there is still a single layout file for all, but you generate the respective files into the Zaamo and Zabha directories?

One thing I just noticed is that the "definedBy" attributes need to be different between the two sets: `definedBy: Zaamo` and `definedBy: Zabha` as appropriate.

This is really helpful! Thanks!

ThinkOpenly:
> in Zabha there is also amocas(b/h) files and i think those are connected with Zacas so should i create a aqrl layout for this ? & One more question the single layout files for all where to put means in which directory? Should I create a directory?

It's slightly complicated, and the UDB YAML files for `amocas.{b,h}` may be slightly inaccurate:
- Most of the `amo` instructions for {w,d,q} are defined by Zaamo.
- `amocas.{w,d,q}` are defined by Zacas. Zacas depends on Zaamo.
- Most of the `amo` instructions for {b,h} are defined by Zabha. Zabha depends on Zaamo.
- `amocas.{b,h}` are defined by Zabha if Zacas is supported.

In that last line is well-represented currently in UDB, specifically `amocas.{b,h}` are only:
```
definedBy: Zabha
```
when they should be defined by "allOf" Zabha and Zacas. Do you want to fix that in this PR, or shall I open a new issue?

So, while it's not critical what directory these files appear in, it probably makes the most sense for:
- Zaamo-only instructions in Zaamo.
- Zacas-only instructions in Zacas.
- Zabha-only instructions in Zabha.
- and `amocas.{b,h}` in Zabha.

spec/std/isa/inst/Zaamo/amoadd.d.aqrl.yaml
```diff
@@ -0,0 +1,140 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../../schemas/inst_schema.json
+
+$schema: "inst_schema.json#"
+kind: instruction
+name: amoadd.d.aqrl
+long_name: Atomic fetch-and-add doubleword (acquire) (release) (acquire-release)
```
ThinkOpenly: The "(acquire) (release) (acquire-release)" is a bit redundant. :-)
Could you put just a single parenthetical here, likely "(aquire-release)" in this case?

spec/std/isa/inst/Zaamo/amoadd.d.aqrl.yaml
```diff
@@ -0,0 +1,140 @@
...
+description: |
+  Atomically with acquire ordering with release ordering:
```
ThinkOpenly: This isn't what you intended, either, is it?

spec/std/isa/inst/Zaamo/amoadd.SIZE.AQRL.layout
```diff
@@ -0,0 +1,175 @@
...
+
+$schema: "inst_schema.json#"
+kind: instruction
+name: amoadd.<%= size %><%= aq_rl_suffix %>
+long_name: Atomic fetch-and-add <%= current_size[:name] %><%= aq && rl ? " (acquire-release)" : (aq ? " (acquire)" : (rl ? " (release)" : "")) %>
+description: |
+  Atomically<%= aq && rl ? " with acquire and release ordering" : (aq ? " with acquire ordering" : (rl ? " with release ordering" : "")) %>:
+
+    * Load the <%= current_size[:name] %> at address _xs1_
+    * Write the <%= %w[b h].include?(size) ? "sign-extended value" : (size == "w" ? "sign-extended value" : "loaded value") %> into _xd_
```
ThinkOpenly: Why not combine "b", "h", and "w" here if they all emit the same string.
Does the behavior of "w" depend on XLEN, though, where it's only sign-extended if XLEN > 32?

spec/std/isa/inst/Zaamo/amoadd.SIZE.AQRL.layout
```diff
@@ -0,0 +1,175 @@
...
+      location: 11-7
+access:
+  s: always
+  u: always
+  vs: always
+  vu: always
+operation(): |
+  if (implemented?(ExtensionName::A) && MISA_CSR_IMPLEMENTED && (CSR[misa].A == 1'b0)) {
+    raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
+  }
```
ThinkOpenly: Isn't the condition the opposite of what you want?

Although, it looks like you're just carrying forward the condition from the original code... I think it's wrong, though:
```
  if (!implemented?(ExtensionName::A) || (MISA_CSR_IMPLEMENTED && (CSR[misa].A == 0'b0))) {
  }
```
This applies to all of the other layouts as well.

Sukuna0007Abhi: ok sir i got it now, thanks for pointing

spec/std/isa/inst/Zaamo/amoadd.SIZE.AQRL.layout
Sukuna0007Abhi: Yes sir , thanks for pointing, After checking the RISC-V spec and existing implementations, word operations are actually always described as "sign-extended" regardless of XLEN.

I look at the existing lr.w instruction - its description says "places the sign-extended value in xd" and the operation code shows `sext(load_value[31:0], 32)` even on RV32. The Sail model in our templates also confirms this with` WORD => sign_extend(loaded[31:0])` for all XLEN values.

The  part  is that on RV32 it's conceptually a sign-extension (32→32 bits, identity) while on RV64 it's an actual sign-extension (32→64 bits), but the part consistently describes both as "sign-extended results."

So yes, combining b, h, and w is better since they all produce sign-extended results. Thanks for the suggestion.I will update that by today.

spec/std/isa/inst/Zaamo/amoadd.d.aqrl.yaml
Sukuna0007Abhi: ok sir updating it 👍

spec/std/isa/inst/Zaamo/amoadd.d.aqrl.yaml
Sukuna0007Abhi: I put the description as the same for it but ok sir i will change it a little bit

Sukuna0007Abhi:
I updated all pls check sir @ThinkOpenly @dhower-qc @AFOliveira

ThinkOpenly:
The no-aq-no-rl cases for "d" and "w" are now missing, like `amoadd.d`. Maybe I'm jumping in while you are still working?

Rakefile
```diff
@@ -382,6 +464,55 @@ namespace :gen do
...
+    end
+
+    # TODO: Uncomment when Zacas amocas instruction files are added in future PR
+    # %w[w d q].each do |size|
+    #   [
+    #     { suffix: ".aq", aq: true, rl: false },
+    #     { suffix: ".rl", aq: false, rl: true },
+    #     { suffix: ".aqrl", aq: true, rl: true }
+    #   ].each do |variant|
+    #     Rake::Task["#{$resolver.std_path}/inst/Zacas/amocas.#{size}#{variant[:suffix]}.yaml"].invoke
```
ThinkOpenly: This all duplicates the code above, if I understand correctly?

Sukuna0007Abhi: Sir I actually did that because some times it don't show genrated  variant files some error happened and not generating those yaml files that's why I created this manual invoke to atleast generate those properly as an addition , if the automation not work

Sukuna0007Abhi: But i will try to avoid this manual part though, but yet not automation thing not generating

ThinkOpenly: Hmm. In general the solution to inconsistent results is to fix the inconsistency, not to try again in a different way. If you have questions or need advice, ask for that, and provide details. Absent that, this is not really ready for another review.

Sukuna0007Abhi: I understood and sorry for it sir, but after some more research and work I fixed it now automation will generate all just by updating the dependency, no more manual or duplication , thanks for pointing me the right way.

Sukuna0007Abhi:
pls check sir @ThinkOpenly sir @dhower-qc sir @AFOliveira

Rakefile
Sukuna0007Abhi: Dependency Definition (Lines 162-192) & File Generation Rules (Lines 384-468) both will work by together perfectly

Rakefile
```diff
@@ -356,9 +389,88 @@ file "#{$resolver.std_path}/csr/Zicntr/mcountinhibit.yaml" => [
...
+    aq_rl_variants.each do |variant|
+      file "#{$resolver.std_path}/inst/#{extension_dir}/#{op}.#{size}#{variant[:suffix]}.yaml" => [
+        "#{$resolver.std_path}/inst/Zaamo/#{op}.SIZE.AQRL.layout",
+        __FILE__
+      ] do |t|
+        aq = variant[:aq]
+        rl = variant[:rl]
+        erb = ERB.new(File.read($resolver.std_path / "inst/Zaamo/#{op}.SIZE.AQRL.layout"), trim_mode: "-")
+        erb.filename = "#{$resolver.std_path}/inst/Zaamo/#{op}.SIZE.AQRL.layout"
+        File.write(t.name, insert_warning(erb.result(binding), t.prerequisites.first))
```
ThinkOpenly, line 385: I see you are using `insert_warning` here, but no warnings appear in the generated files. Do you know why?

Rakefile
```diff
@@ -155,8 +158,38 @@ end
 
 
 namespace :test do
```
ThinkOpenly, line 160: Maybe this isn't correct. Why is this located in the "test" namespace?

Rakefile
```diff
@@ -155,8 +158,38 @@ end
...
+      { suffix: ".rl", aq: false, rl: true },
+      { suffix: ".aqrl", aq: true, rl: true }
+    ].each do |variant|
+      amo_instruction_files << "#{$resolver.std_path}/inst/Zabha/amocas.#{size}#{variant[:suffix]}.yaml"
+    end
+  end
+
   desc "Check that instruction encodings in the DB are consistent and do not conflict"
-  task :inst_encodings do
+  task :inst_encodings => amo_instruction_files do
```
ThinkOpenly: Is this necessary? Does this task not find the generated instructions in the configured/resolved architecture without this dependency?

Rakefile
Sukuna0007Abhi: Oops ;) sorry sir I actually thought it will look bad that's why I removed that thing from all files by manually if you want I will regenerate and it will show again.

Rakefile
Sukuna0007Abhi: Yes sir at first it actually not finding those files but now I found a better one I am commiting it,

Rakefile
Sukuna0007Abhi: Ok sir , I'm changing the approach

Rakefile
ThinkOpenly: Yeah. As a general rule, avoid changing generated files. :-)

spec/std/isa/inst/Zaamo/amomaxu.w.yaml
```diff
@@ -35,77 +33,74 @@ access:
...
+  X[xd] = amo<32>(virtual_address, X[xs2][31:0], AmoOperation::Maxu, 1'b0, 1'b0, $encoding);
 
 # SPDX-SnippetBegin
 # SPDX-FileCopyrightText: 2017-2025 Contributors to the RISCV Sail Model <https://github.com/riscv/sail-riscv/blob/master/LICENCE>
 # SPDX-License-Identifier: BSD-2-Clause
 sail(): |
   {
     if extension("A") then {
-      /* Get the address, X(rs1) (no offset).
+      /* Get the address, X(xs1) (no offset).
```
ThinkOpenly: There shouldn't be any changes to any code in any of the `sail()` methods.

Rakefile
```diff
@@ -356,9 +359,88 @@ file "#{$resolver.std_path}/csr/Zicntr/mcountinhibit.yaml" => [
...
+#       __FILE__
+#     ] do |t|
+#       aq = variant[:aq]
+#       rl = variant[:rl]
+#       erb = ERB.new(File.read($resolver.std_path / "inst/Zabha/amocas.SIZE.AQRL.layout"), trim_mode: "-")
+#       erb.filename = "#{$resolver.std_path}/inst/Zabha/amocas.SIZE.AQRL.layout"
+#       File.write(t.name, insert_warning(erb.result(binding), t.prerequisites.first))
+#     end
+#   end
+# end
```
ThinkOpenly: If this code is no longer needed/used, remove it.

spec/std/isa/inst/Zaamo/amomax.d.aqrl.yaml
```diff
@@ -0,0 +1,142 @@
...
+
+# WARNING: This file is auto-generated from spec/std/isa/inst/Zaamo/amomax.SIZE.AQRL.layout
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../../schemas/inst_schema.json
+
+$schema: "inst_schema.json#"
+kind: instruction
+name: amomax.d.aqrl
+long_name: Atomic MAX doubleword (acquire) (release) (acquire-release)
```
ThinkOpenly: This probably isn't what you wanted. (It's not what I expected. ;-)

I expect something more like:
```suggestion
long_name: Atomic MAX doubleword (acquire-release)
```

spec/std/isa/inst/Zaamo/amomax.d.aqrl.yaml
```diff
@@ -0,0 +1,142 @@
...
+description: |
+  Atomically with acquire ordering with release ordering:
```
ThinkOpenly: This doesn't read well, either. Perhaps:
```suggestion
  Atomically with acquire ordering and release ordering:
```
or better:
```suggestion
  Atomically with acquire and release ordering:
```

ThinkOpenly (changes requested):

Rakefile
Sukuna0007Abhi: Sir it's actually for Zacas(w,q,d) amocas, for future work, I commented it. If you want I will remove it.

ThinkOpenly: If it's not for this PR, then please save a copy and remove from the PR content. Thanks!

spec/std/isa/inst/Zaamo/amomax.d.aqrl.yaml
Sukuna0007Abhi: It's got corrupted maybe sir I will fix it for continuous re-generation

Rakefile
Sukuna0007Abhi: Ok sir I'm removing it, will save in local

ThinkOpenly:
I think this only needs one more update to the "golden instruction appendix" to pass the CI tests. Are you willing to do so, @Sukuna0007Abhi ?

PR #922, opened 2025-07-24
fix(csr): add Q-bit in sw_read()

adingank-qualcomm:
Added bit 16 to sw_read for Q-bit according to Table 10 in Section 3.1.1. of [The RISC-V Instruction Set Manual Volume II] [1].

[1] <https://drive.google.com/file/d/17GeetSnT5wW3xNuAHI95-SI1gPGd5sJ_/view?usp=drive_link> 'Privileged Architecture'


adingank-qualcomm:
@BKhristenko FYI since you had contributed the original Q-bit definition.

Issue #704, opened 2025-04-30
Add "recommended" values for extension parameters


ThinkOpenly:
**Is your feature request related to a problem? Please describe.**
Currently, extension parameters are defined with types and possible values ("Sdtrig" extension, Machine Context (`mcontext`), "hcontext" field; from PR #524):
```
params:
  DBG_HCONTEXT_WIDTH:
    schema:
      type: integer
      minimum: 0
      maximum: 14
    description: Specifies the size of HCONTEXT
```
Some parameters have "recommended" values in the spec, for example (Machine Context, `mcontext`, "hcontext" field):
> An implementation may tie any number of upper bits in this field to 0. If the H extension is not implemented, it’s recommended to implement 6 bits on RV32 and 13 bits on RV64 (as visible through the mcontext register). If the H extension is implemented, it’s recommended to implement 7 bits on RV32 and 14 bits on RV64.

**Describe the solution you'd like**
Some means of expressing a recommended value for each type (integer, boolean, etc.), including with the conditional logic required by the example above, seems desirable to allow configurations to not have to determine the recommended values and explicitly set all of the configuration parameters to those values.

ThinkOpenly:
> [@ThinkOpenly](https://github.com/ThinkOpenly) I explored this issue and wanted to confirm if the following approach aligns with your expectations:
> 
> * Add a `recommended` field to the `param_data` definition in `spec/schemas/ext_schema.json`, supporting:
>   
>   * Static values (`integer`, `string`, `boolean`, etc.)
>   * Conditional objects (e.g., based on architecture width or extension presence)

We may want to see if the "default" attribute can be made to work here, rather than a custom "recommended" attribute. This is some subtlety between the linguistic definitions of "default" and "recommended", but since it can be easily overwritten in a config file, it may not matter.

> Example implementation:
> 
> * Update a real parameter in `spec/std/isa/ext/Sdtrig.yaml` (e.g., `MSTATEEN_CONTEXT_TYPE` or `HSTATEEN_CONTEXT_TYPE`)
> * Include both:
>   
>   * A simple static `recommended` value
>   * A conditional `recommended` example (with inline comments for clarity)

Interesting choice. What do you propose as the default value for these parameters?

xenonnn4w:
Thanks! 
[...]

PR #926, opened 2025-07-27
feat( extensions ) : add default values for Sdtrig extension parameters. 

xenonnn4w:
# Issue
* #704 
# Approach
* Added `default: "rw"` to the `MSTATEEN_CONTEXT_TYPE` and `HSTATEEN_CONTEXT_TYPE` parameters in `spec/std/isa/ext/Sdtrig.yaml`
* Chose `"rw"` as a flexible, least-restrictive default aligned with the spec's general behavior
* Uses the standard JSON Schema `default` field to allow tools to auto-fill if no explicit value is provided
* This serves as a minimal working example toward resolving #704 and avoids introducing a custom `recommended` keyword


spec/std/isa/ext/Sdtrig.yaml
```diff
@@ -40,6 +40,7 @@ params:
     schema:
       type: string
       enum: [rw, read-only-0, read-only-1]
+      default: "rw"
```
ThinkOpenly: I see the CI failure just now. Could you try removing the quotation marks here? Since it's an enum, maybe providing a string instead of the enum value is causing the issue.
```suggestion
      default: rw
```
Here and below.

xenonnn4w: Thanks, Fixed!

ThinkOpenly:
That last change would've been better as a new commit instead of a amend/force-push, so we could see the before (which failed) and the after (which still fails).

Thanks for trying. @dhower-qc ... CI is failing when using "default". Any ideas offhand?

dhower-qc:
> That last change would've been better as a new commit instead of a amend/force-push, so we could see the before (which failed) and the after (which still fails).
> 
> 
> 
> Thanks for trying. @dhower-qc ... CI is failing when using "default". Any ideas offhand?

The cpp generator wasn't expecting a string as a default value, so it didn't quote it.

@henrikg-qc, could you take a stab at a quick fix?

PR #932, opened 2025-07-30
Fix mathematical notation rendering in HTML documentation

### LINKED ISSUE #78: Address bug
In https://riscv-software-src.github.io/riscv-unified-db/manual/html/isa/20240411/chapters/rv64.html there is a bug in this note
image

Sukuna0007Abhi:
Fix math notation rendering issue in RV64 documentation

Fixes #78

This fixes the mathematical expressions showing up as raw LaTeX in the docs instead of properly rendered math.

The issue was reported by @AFOliveira on https://riscv-software-src.github.io/riscv-unified-db/manual/html/isa/20240411/chapters/rv64.html (though the link seems to be down now) where stuff like \({-}2^{31}{-}2^{11}\) was just showing as ugly text instead of actual math notation.

I went with the Antora/JS approach that was suggested by @dhower-qc - added stem: latexmath to enable math processing and threw in MathJax 3 to handle the rendering. Applied it to both the manual and HTML backends.

Should fix those broken math expressions so they actually look like math instead of random LaTeX code.                                                 Ready for Review @ThinkOpenly @AFOliveira @dhower-qc

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -489,6 +489,10 @@ task :portfolios do
   Rake::Task["#{$root}/gen/proc_ctp/pdf/MockProcessor-CTP.pdf"].invoke
   portfolio_start_msg("MockProfileRelease")
   Rake::Task["#{$root}/gen/profile/pdf/MockProfileRelease.pdf"].invoke
+  portfolio_start_msg("RVI20-32-CTP")
+  Rake::Task["#{$root}/gen/proc_ctp/pdf/RVI20-32-CTP.pdf"].invoke
+  portfolio_start_msg("RVI20-64-CTP")
+  Rake::Task["#{$root}/gen/proc_ctp/pdf/RVI20-64-CTP.pdf"].invoke
   portfolio_start_msg("MC100-32-CTP")
   Rake::Task["#{$root}/gen/proc_ctp/pdf/MC100-32-CTP.pdf"].invoke
   portfolio_start_msg("MC100-32-CRD")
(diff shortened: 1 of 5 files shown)


AFOliveira (commented):
@Sukuna0007Abhi Have you tried compiling the PDF and looking at the errors previously mentioned. Because as far as I can understand latex-math only works with the [stem] keyword

AFOliveira:
The one hiccup I can see is that Asciidoctor still should escape raw ` \(...\) `and `\[…\] `before MathJax ever sees them, so authors would still have to wrap every formula in a stem:[…] or passthrough, which doesn't quite solve the issue

What to do instead would be loading an extension that automatically picks up $…$, \(...\), and \[…\] as math:

```
asciidoc:
  extensions:
    - asciidoctor-kroki
    - '@asciidoctor/tabs'
    - asciidoctor-mathjax   # auto‑handles raw TeX delimiters
  attributes:
    stem: latexmath

```

With that in place, you can simply write: `[\({-}2^{31}{-}2^{11}\), \(2^{31}{-}2^{11}{-}1\)]` and it’ll just work without any extra wrapping.

Sukuna0007Abhi:
@AFOliveira Thank you for the detailed explanation and the exact configuration! 

To be completely honest, I haven't actually tested PDF generation yet - I focused on the HTML rendering based on the original issue description. You're absolutely right to question this.

Let me test the current approach with PDF generation first to see if it actually works, and then implement your suggested `asciidoctor-mathjax` extension approach.

A couple of questions:
1. Is there a specific rake task I should use to test PDF generation with mathematical expressions?
2. Should I look for the original mathematical expressions (like `\({-}2^{31}{-}2^{11}\)`) in the existing documentation to test with?

I'll test both approaches and report back with the results. Thanks for the guidance and for keeping me honest about proper testing!

AFOliveira:
In this case, it's always intended to test before PR so we actually ensure we are fixing what we say we are fixing. 

> 1. Is there a specific rake task I should use to test PDF generation with mathematical expressions?

Yes, you should look in the place it failed. Actually, its easier to look into the HTML manual, you need to use the command in the readme.md `./do gen:html_manual MANUAL_NAME=isa VERSIONS=all`. 

> 2. Should I look for the original mathematical expressions (like `\({-}2^{31}{-}2^{11}\)`) in the existing documentation to test with?

Yes! You should look into as many as you can find but truly the starting point is the one that failed. It's on RVI64 section.

Issue #929, opened 2025-07-28
Huge number of wavedrom failures when generating CTP


james-ball-qualcomm:
I get a massive number of these error messages when generating a CTP. I know we get a lot of warnings/errors normally when generating artifacts but this one is particularly egregious so I'm hoping you can fix it.

```
asciidoctor: ERROR: images/wavedrom/instruction-formats.edn: line 6: Failed to generate image: Could not find the 'wavedrom-cli' executable in PATH; add it to the PATH or specify its location using the 'wavedrom' document attribute
/workspaces/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/asciidoctor-diagram-2.3.2/lib/asciidoctor-diagram/diagram_source.rb:144:in `find_command'
/workspaces/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/asciidoctor-diagram-2.3.2/lib/asciidoctor-diagram/wavedrom/converter.rb:35:in `convert'
/workspaces/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/asciidoctor-diagram-2.3.2/lib/asciidoctor-diagram/diagram_processor.rb:196:in `create_image_block'
/workspaces/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/asciidoctor-diagram-2.3.2/lib/asciidoctor-diagram/diagram_processor.rb:106:in `process'
/workspaces/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/asciidoctor-2.0.23/lib/asciidoctor/parser.rb:1065:in `[]'
/workspaces/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/asciidoctor-2.0.23/lib/asciidoctor/parser.rb:1065:in `build_block'
/workspaces/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/asciidoctor-2.0.23/lib/asciidoctor/parser.rb:916:in `next_block'
/workspaces/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/asciidoctor-2.0.23/lib/asciidoctor/parser.rb:403:in `next_section'
/workspaces/riscv-unified-db/.home/.gems/ruby/3.2.0/gems/asciidoctor-2.0.23/lib/asciidoctor/parser.rb:388:in `next_section'
[...]

PR #937, opened 2025-07-30
fix : Solved Huge number of wavedrom failures when generating CTP #929

pankajbaid567:
### **The Problem**
The error occurred because:   Path resolution issue: Even when installed, asciidoctor-diagram couldn't find [wavedrom-cli](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) in the PATH

### **The Solution**
I implemented a comprehensive fix with multiple layers:   
Fixed path resolution in rake tasks: Added the -a wavedrom=#{$root}/node_modules/.bin/wavedrom-cli attribute to all asciidoctor commands that use the diagram extension:

[tasks.rake](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (both PDF and HTML generation)
[tasks.rake](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (PDF generation)
[tasks.rake](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (already had this fix)
Added robustness check: Created an [ensure_npm_dependencies](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) function that automatically installs npm dependencies if [wavedrom-cli](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is missing, and integrated it into the build process.

### DIFF
--- a/backends/portfolio/tasks.rake
+++ b/backends/portfolio/tasks.rake
@@ -13,6 +13,14 @@ require "idlc/passes/gen_adoc"
 
 require "udb/config"
 
+# Ensure npm dependencies are installed, particularly wavedrom-cli for diagram generation
+def ensure_npm_dependencies
+  unless File.exist?("#{$root}/node_modules/.bin/wavedrom-cli")
+    puts "Installing npm dependencies (wavedrom-cli not found)..."
+    system("cd #{$root} && npm install") || raise("Failed to install npm dependencies")
+  end
+end
+
 sig { returns(Udb::ConfiguredArchitecture) }
 def pf_create_arch
   $resolver.cfg_arch_for("_")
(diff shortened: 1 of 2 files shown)


backends/portfolio/tasks.rake
```diff
@@ -13,6 +13,14 @@ require "idlc/passes/gen_adoc"
 
 require "udb/config"
 
+# Ensure npm dependencies are installed, particularly wavedrom-cli for diagram generation
+def ensure_npm_dependencies
```
dhower-qc, line 17: In theory, we don't need this. It should already be installed via the setup script (bin/setup) that runs on every do command

dhower-qc (changes requested):
Any insight on why this works but the wavedrom attribute in the adoc template does not?

ThinkOpenly:
Is there an expectation as to if/when the review comments will be addressed?

PR #936, opened 2025-07-30
Fix unnecessary bullet points for single extensions in HTML docs

Sukuna0007Abhi:
## Summary
Fixes issue #927 where single extension definitions were showing unnecessary bullet points in the HTML documentation's "Defining Extension" section.

## Problem
- Single extensions like `definedBy: Sm` were displaying as `• Sm, version >= 1.11.0` with unwanted bullet points
- This made the documentation look cluttered for simple cases where only one extension is involved
- Multiple extensions correctly needed bullets, but single ones should appear clean

## Solution
- Modified `to_asciidoc` method in `req_expression.rb` to detect simple single extensions at root level
- Added `is_simple_single_extension?()` helper to identify when bullets aren't needed
- Added `to_asciidoc_no_bullet()` method for clean formatting of single extensions
- Preserved existing bullet behavior for complex multi-extension conditions (allOf, anyOf, etc.)

## Testing
- Single extensions now display cleanly: `Sm, version >= 1.11.0`
- Multiple extensions maintain proper structure with bullets
- No regression in complex extension requirement formatting

## Impact
- Improves readability of HTML documentation for CSRs with single defining extensions
- Maintains backward compatibility for all existing complex extension definitions

PR #938, opened 2025-08-01
docs: added conditional rendering of the bullet point in defining extension

### LINKED ISSUE #927: HTML docs: Attributes: Defining Extension has a bullet where none is needed
https://riscv-software-src.github.io/riscv-unified-db/manual/html/isa/isa_20240411/csrs/mtval.html

Image

Note how the content after "Defining Extension" has a bullet, but it's a single item and no bullet is needed (or desired).

Be careful not to break the cases where there are multiple defining extensions (e.g. `hstateen0`), although those may also have an unnecessary bullet:

Image

(Issue first mentioned in #63.)

shivansh-bhatnagar18:
Earlier the defining extension values were rendered with an additional bullet point which was not at all necessary so this PR makes the `to_asciidoc` function make the value render accordingly if it only has a single value (then no bullet point is needed) and if it has multiple values then the bullet points are getting rendered.
image
fixes #927


tools/ruby-gems/udb/lib/udb/req_expression.rb
```diff
@@ -155,19 +155,39 @@ def flat_versions
   def to_asciidoc(cond = @hsh, indent = 0, join: "\n")
     case cond
     when String
-      "#{'*' * indent}* #{cond}, version >= #{T.must(@arch.extension(cond)).min_version}"
+      if indent.zero?
```
dhower-qc: indent does not tell you if there are multiple conditions; it just tells you how much to indent the line in the rendered file.

As such, this patch does not do what the PR comment claims.

shivansh-bhatnagar18: Thanks for pointing out the confusion at my end, could you please review the code now @dhower-qc.

dhower-qc:
Can you provide a picture of what this looks like now in a PDF?

shivansh-bhatnagar18:
I am attaching below how the changes would look like to various CSRs' attributes
For cases like `hstateen0`
image
And for cases like 'mcycle`
image

shivansh-bhatnagar18:
@dhower-qc  Also we need to fix the tests which are failing for these bullet points in a different issue for these changes 
image

shivansh-bhatnagar18:
I suppose the issue is resolved earlier by the #936 so I am closing this PR. Sorry for the delay

PR #941, opened 2025-08-03
fix: printing object string in profile extension appendix

### LINKED ISSUE #853: bug: Profile extension appendix has text "#<DatabaseObject::Kind::Profile>"
Looks like a Ruby change was made and it is now not displaying the expected string.

Image

shivansh-bhatnagar18:
### **The Problem**
Using `to_s` on a `T::Enum` like `kind` returns the object reference (e.g., `#<DatabaseObject::Kind::Profile>`) instead of the actual string value.

### **The Solution**
Replaced `to_s` with `serialize`, which correctly returns the string value (e.g., `"profile"`).

### **Why**
`serialize` is the intended way to extract string values from `T::Enum`. 

Fixes #853

### DIFF
--- a/tools/ruby-gems/udb/lib/udb/obj/database_obj.rb
+++ b/tools/ruby-gems/udb/lib/udb/obj/database_obj.rb
@@ -49,7 +49,7 @@ class DatabaseObject
   attr_reader :long_name
 
   sig { returns(String) }
-  def kind = @kind.to_s
+  def kind = @kind.serialize
 
   # @return [Architecture] If only a specification (no config) is known
   # @return [ConfiguredArchitecture] If a specification and config is known


AFOliveira:
Quick question: am I right that `to_s` itself isn’t deprecated in Ruby and the real change is Sorbet dropping implicit `T::Enum` → String conversion, so we need `#serialize`? If so, could the PR description note that, and did you scan other `@kind.to_s` sites (e.g., in `generator.rb` and `reporter.rb`) to be sure they won’t hit the same runtime error?

shivansh-bhatnagar18:
Yeah @AFOliveira, you are right that `to_s` isn't deprecated in Ruby. 
Sorbet's `T::Enum` doesn't override `to_s` to return the serialized value rather it retains Ruby's default object stringification unless we explicitly call `.serialize`.

AFOliveira:
> I have made the required changes in the PR Description. I have verified the occurances of `.kind` in other classes like `portfolio.rb` etc but couldn't find them in `generator.rb` and `reporter.rb`.

Can you please just check the same bug happening here doesn't happen anywhere else?

AFOliveira (approved):
LGTM! But they ruby expert is definitely @dhower-qc :)

james-ball-qualcomm:
I'll try kind.serialize instead of the implicit kind.to_s. Thanks!

PR #942, opened 2025-08-03
Fix-resolved the profile part

### LINKED ISSUE #853: bug: Profile extension appendix has text "#<DatabaseObject::Kind::Profile>"

Sukuna0007Abhi:
Fix Profile appendix showing ugly enum string
Fixes #853

Profile extension docs were showing #<Udb::DatabaseObject::Kind::Profile> instead of "profile" in the table header. Not a good look! 😅

The fix:
`| <%= portfolio_kind.to_s.split('::').last.downcase %> |`
Root cause: The portfolio_kind variable was receiving a string representation of the enum object rather than the clean enum value.

Testing:

✅ Built portfolios successfully (./bin/rake portfolios)
✅ Verified table now shows clean "profile" text
✅ No breaking changes to other components
✅ All existing functionality preserved
Key advantages:

Minimal impact - Only touches the problematic template line
Safe approach - No changes to core DatabaseObject classes
Smart parsing - Extracts clean name from enum string representation
Reliable - Handles the actual data format being passed to template
Simple fix that gets the job done without breaking anything! 🎯

Ready for Review... @ThinkOpenly @dhower-qc @AFOliveira @james-ball-qualcomm

### DIFF
--- a/backends/portfolio/templates/ext_appendix.adoc.erb
+++ b/backends/portfolio/templates/ext_appendix.adoc.erb
@@ -12,7 +12,7 @@
 <% if portfolios.size > 1 -%>
 <%= ext.name %> Extension Presence
 |===
-| <%= portfolio_kind.to_s.split('::').last.downcase %> | v<%= ext.versions.map { |ext_ver| ext_ver.canonical_version.to_s }.join(" | v") %>
+| <%= portfolio_kind.to_s.split('::').last.gsub(/[<>]/, '').capitalize %> | v<%= ext.versions.map { |ext_ver| ext_ver.canonical_version.to_s }.join(" | v") %>
 
 <% portfolios.each do |portfolio| -%>
 | <%= portfolio.name %> | <%= portfolio.version_greatest_presence(ext.name, ext.versions).join(" | ") -%>
--- a/backends/portfolio/templates/ext_appendix.adoc.erb
+++ b/backends/portfolio/templates/ext_appendix.adoc.erb
@@ -12,7 +12,7 @@
 <% if portfolios.size > 1 -%>
 <%= ext.name %> Extension Presence
 |===
-| <%= portfolio_kind %> | v<%= ext.versions.map { |ext_ver| ext_ver.canonical_version.to_s }.join(" | v") %>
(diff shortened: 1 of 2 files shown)


AFOliveira:
Can you provide an example of how it was vs how it is now

AFOliveira:
@Sukuna0007Abhi sorry, but seems like @shivansh-bhatnagar18 fixed this issue in a previous PR

Sukuna0007Abhi:
Sir @AFOliveira ok I got it, but kindly could you check one time I think my approach is more robust and suitable than other PR solution. I will give you the example part.

Sukuna0007Abhi:
@shivansh-bhatnagar18 I disagree. My approach is actually more better sound.

String formatting belongs in templates, not data models. Your .serialize approach mixes presentation logic with core objects, while mine maintains proper separation of concerns.

My solution is defensive and future-proof - it works with any object type, not just T::Enum. What if the data pipeline changes or .serialize breaks? My .gsub(/[<>]/, '') also handles malformed strings while yours assumes perfect input.

I found and fixed this bug in multiple templates (both [ext_appendix.adoc.erb] AND profile.adoc.erb), while yours only addressed one instance. This shows a more comprehensive debugging approach. And Your "modularity" argument misunderstands software architecture. Template formatting logic should stay in templates where it belongs - this is basic separation of concerns. When you put .serialize calls in data models, you're mixing business logic with presentation formatting.

If we need the same formatting in multiple places, the correct approach is to create a dedicated helper method or utility function, not modify core data objects. My approach keeps the formatting concern isolated exactly where it's used, making the codebase more maintainable.

Sukuna0007Abhi:
Now which one you like your wish sir I will close then this Pr @AFOliveira

dhower-qc (commented):
Thanks for the contribution. However, #941 gets to the crux of the problem better -- portfolio_kind is supposed to be a pretty string, but it's not. That happened when we changed kind to a T::Enum for type checking. the `serialize` function call from #941 gets the pretty string back.

dhower-qc:
fixed by #941

PR #951, opened 2025-08-05
fix: ISS working, adds ability to run riscv-tests for defined configuration 

henrikg-qc:
Includes functionality to build, debug and run the cpp_hart backend with the riscv-tests or user provided RISC-V elf file.
Added MC100 configuration for riscv-tests with PMP and S support needed for csrs and instructions used in the riscv-tests suite.

This PR includes code changes identified after greater verification of the recent refactor and file layout change.


backends/cpp_hart_gen/cpp/include/udb/bitfield.hpp
```diff
@@ -164,11 +135,18 @@ namespace udb {
 
   template <unsigned Size>
   class Bitfield {
-   public:
+    public:
```
ThinkOpenly: revert this change to indentation, also to match "protected", below.

backends/cpp_hart_gen/cpp/include/udb/bitfield.hpp
```diff
@@ -164,11 +135,18 @@ namespace udb {
...
+
+    template <unsigned _Size>
+    using BitsType = PossiblyUnknownBits<_Size>;
+
+
```
ThinkOpenly: remove extra blank line

backends/cpp_hart_gen/cpp/include/udb/bits.hpp
```diff
@@ -270,6 +274,31 @@ namespace udb {
...
+    if (r < a || r < b) {
+      // overflow occurred, saturate
+      return std::numeric_limits<unsigned>::max();
+    }
+
+    if (r > std::numeric_limits<unsigned>::max()) {
+      return std::numeric_limits<unsigned>::max();
+    }
+
+    return r;
```
ThinkOpenly: Personal preference, but I think it reads better as a ternary here:
```suggestion
    return (r > std::numeric_limits<unsigned>::max()) ? std::numeric_limits<unsigned>::max() : r;
```

backends/cpp_hart_gen/cpp/include/udb/bits.hpp
```diff
@@ -270,6 +274,31 @@ namespace udb {
...
+  }
+
+  // saturating mul
+  template <unsigned A, unsigned B>
+  struct mulsat {
+    static constexpr unsigned value =
+        (((A * B) < A) || ((A * B) < B)) ? std::numeric_limits<unsigned>::max() : A * B;
```
ThinkOpenly, line 293: This is logically very similar to `addsat_unsigned`, above, but squashed into one line. Should we do some squashing there, or expanding here for consistency?

backends/cpp_hart_gen/cpp/include/udb/bits.hpp
```diff
@@ -281,6 +310,21 @@ namespace udb {
     { std::remove_cvref_t<decltype(a.value())>::IsABits == true };
     { std::remove_cvref_t<decltype(a.value())>::PossiblyUnknown == false };
     { std::same_as<std::remove_cvref_t<decltype(a.unknown_mask())>, typename T::MaskType> };
+
+    // cast the bits to a known (not undefined type). Might raise if there are undefined bits
```
ThinkOpenly:
```suggestion
    // cast the bits to a known (not undefined) type. Might raise if there are undefined bits
```
Also, what does "Might raise if there are undefined bits" mean?

backends/cpp_hart_gen/cpp/include/udb/bits.hpp
```diff
@@ -281,6 +310,21 @@ namespace udb {
...
+    { std::remove_cvref_t<decltype(a.template extract<0, 0>())>::IsABits == true };
+    { a.template extract<0, 0>().width() == 1 };
+
+    { std::remove_cvref_t<decltype(a.template at<0>())>::IsABits == true };
+    { a.template at<0>().width() == 1 };
+
+    // { std::remove_cvref_t<decltype(a.template replicate<1>())>::IsABits == true };
+    // { std::remove_cvref_t<decltype(a.template replicate<2>())>::IsABits == true };
+    // { a.replicate(1).width() == a.width() };
+    // { a.replicate(2).width() == a.width() * 2 };
```
ThinkOpenly: Does this need to be retained?

backends/cpp_hart_gen/cpp/include/udb/bits.hpp
```diff
@@ -898,63 +987,80 @@ namespace udb {
...
       }
     }
 
-    constexpr _Bits operator<<(const _Bits &shamt) const { return _Bits{m_val << shamt.get()}; }
-
     template <template <unsigned, bool> class ShiftBitsType, unsigned ShiftN>
       requires(BitsType<ShiftBitsType<ShiftN, false>>)
     constexpr _Bits operator<<(const ShiftBitsType<ShiftN, false> &shamt) const {
-      return _Bits{m_val << shamt.get()};
+      if (shamt.get() >= N) {
```
ThinkOpenly, line 995: What is `N` here?

ThinkOpenly: nevermind. template parameter.

backends/cpp_hart_gen/cpp/include/udb/version.hpp
```diff
@@ -192,6 +192,7 @@ namespace udb {
 
     constexpr bool satisfied_by(const Version& version) const {
       switch (m_op.kind()) {
+        case OpKind::COMPAT:
```
ThinkOpenly, line 195: Does this need more logic beyond acting the same as `GTE`?

backends/cpp_hart_gen/lib/gen_cpp.rb
```diff
@@ -240,14 +240,33 @@ def gen_cpp_argument_list(symtab)
...
+            "_#{arg.name}"
+          else
+            arg.name
+          end
+        list << "#{written ? '' : 'const'} #{arg_type} #{written ? '' : '&'} #{arg_name}"
+      end
+      # list = @argument_nodes.map do |arg|
+      #   written = (builtin? || generated?) || body.written?(symtab, arg.name)
+      #   "#{written ? '' : 'const'} #{arg.gen_cpp(symtab, 0, ref: !written)}"
+      # end.join(", ")
```
ThinkOpenly: Do this need to be retained?

backends/cpp_hart_gen/lib/gen_cpp.rb
```diff
@@ -414,9 +473,22 @@ def gen_cpp(symtab, indent = 0, indent_spaces: 2)
...
+          "#{' ' * indent}_RuntimeBits<#{symtab.cfg_arch.possible_xlens.max}, #{t.signed?}>{#{v}_b, __UDB_XLEN}"
+        else
+          "#{' ' * indent}_PossiblyUnknownRuntimeBits<#{symtab.cfg_arch.possible_xlens.max}, #{t.signed?}>{\"#{v}\"_xb, __UDB_XLEN}"
+        end
       else
-        "#{' ' * indent}_Bits<#{w}, #{t.signed?}>{#{v}_b}"
+        if t.known?
+          "#{' ' * indent}_Bits<#{t.width}, #{t.signed?}>(#{v}_b)"
+        else
+          # this is broken. it doesn't account for the known width
```
ThinkOpenly: I'm not sure what this means. This is in the `else` clause from `if t.known?`. What needs to happen here?

backends/cpp_hart_gen/templates/hart_impl.hxx.erb
```diff
@@ -24,31 +24,31 @@ namespace udb {
...
+  //   <%- cfg_arch.not_prohibited_csrs.each do |csr| -%>
+  //   <%- unless csr.address.nil? -%>
+  //   m_csr_addr_map[<%= csr.address %>_b] = &m_csrs.<%= csr.name %>;
+  //   <%- end -%>
+  //   <%- unless csr.indirect_address.nil? -%>
+  //   m_csr_indirect_addr_map[std::make_pair(<%= csr.indirect_address %>_b, <%= csr.indirect_slot %>_b)] = &m_csrs.<%= csr.name %>;
+  //   <%_ end -%>
+  //   m_csr_name_map["<%= csr.name %>"] = &m_csrs.<%= csr.name %>;
+  //   <%- end -%>
+  // }
```
ThinkOpenly: Do we need to retain this?

spec/std/isa/inst/I/fence.yaml
```diff
@@ -6,14 +6,14 @@
 $schema: "inst_schema.json#"
 kind: instruction
 name: fence
-long_name: Memory oxdering fence
+long_name: Memory ordering fence
```
ThinkOpenly, line 9: It would be nice to push these sorts of changes via a separate PR, but not a mortal sin.

backends/cpp_hart_gen/cpp/include/udb/bits.hpp
henrikg-qc: This is Dereks change but for readability I suggest:
    auto r = a + b;
    if (r < a || r < b || r > std::numeric_limits<unsigned>::max()) {
      // overflow occurred, saturate
      r = std::numeric_limits<unsigned>::max();
    }   

    return r;

backends/cpp_hart_gen/cpp/include/udb/bits.hpp
henrikg-qc: I suggest to expand as suggestion above

spec/std/isa/inst/V/vsetvli.yaml
```diff
@@ -14,7 +14,7 @@ assembly: xd, xs1, vtypei
 encoding:
   match: 0----------------111-----1010111
   variables:
-    - name: xs1
+    - name: zimm
```
ThinkOpenly, line 17: Ideally, via a separate PR. But, nice catch.

backends/cpp_hart_gen/cpp/include/udb/version.hpp
henrikg-qc: For this operation no. COMPAT == GTE

backends/cpp_hart_gen/templates/hart_impl.hxx.erb
henrikg-qc: This comes from Derek's refactor, I prefer to cleanup and not to retain

backends/cpp_hart_gen/lib/gen_cpp.rb
henrikg-qc: This comes from Derek's refactor, I prefer to cleanup and not to retain

backends/cpp_hart_gen/cpp/include/udb/bits.hpp
henrikg-qc: This comes from Derek's refactor, I prefer to cleanup and not to retain

tools/ruby-gems/idlc/lib/idlc/ast.rb
```diff
@@ -5412,6 +5490,10 @@ def value(symtab)
             end
           else
             if unsigned_value.bit_length > width
+              puts input_file
+              puts text_value
+              puts unsigned_value.bit_length
+              puts width
```
ThinkOpenly: left in debug?

backends/cpp_hart_gen/cpp/include/udb/bits.hpp
henrikg-qc: hmmmm....constexpr struct member assignment vs function. I'm usually a favor of expanding for readability, here I'm torn.

tools/ruby-gems/udb/python/yaml_resolver.py
```diff
@@ -160,6 +160,7 @@ def read_yaml(file_path: str | Path):
     """
     with open(file_path) as file:
         data = yaml.load(file)
+        file.close()
```
ThinkOpenly: I don't think this is necessary within a `with` block. I believe the scope of the `open` will provoke an automatic close at the end of the block.

tools/ruby-gems/udb/python/yaml_resolver.py
```diff
@@ -569,7 +573,7 @@ def write_resolved_file_and_validate(
         except ValidationError as e:
             print(f"JSON Schema Validation Error for {rel_path}:")
             print(best_match(schema.iter_errors(resolved_obj)).message)
-            exit(1)
+            # exit(1)
```
ThinkOpenly: What's the effect of this?

backends/cpp_hart_gen/cpp/include/udb/bitfield.hpp
henrikg-qc: Done

backends/cpp_hart_gen/cpp/include/udb/bitfield.hpp
henrikg-qc: Done

backends/cpp_hart_gen/cpp/include/udb/bits.hpp
henrikg-qc: Added commit with suggested change, please re-review

tools/ruby-gems/idlc/lib/idlc/ast.rb
henrikg-qc: Think so from Derek, I can remove it

tools/ruby-gems/udb/python/yaml_resolver.py
henrikg-qc: This looks to be a merge error when I rebased. This also explain why the build continued without required "writable:" in the cfg yaml files. 
I will change this back

backends/cpp_hart_gen/cpp/include/udb/bits.hpp
henrikg-qc: removed

backends/cpp_hart_gen/lib/gen_cpp.rb
henrikg-qc: removed

backends/cpp_hart_gen/templates/hart_impl.hxx.erb
henrikg-qc: removed

tools/ruby-gems/idlc/lib/idlc/ast.rb
henrikg-qc: removed

tools/ruby-gems/udb/python/yaml_resolver.py
henrikg-qc: At one point there was an issue where files were held open causing sequential builds to fail. I think was a a target of a blanket change. It has been removed and now relying on the auto closure of the file

backends/cpp_hart_gen/cpp/include/udb/bits.hpp
henrikg-qc: Trying to cast to a known type with bits undefined would throw an exception as you lose the ability to track the undefined bits.
Will update the comment

backends/cpp_hart_gen/lib/gen_cpp.rb
henrikg-qc: Derek will submit a patch for this

backends/cpp_hart_gen/cpp/include/udb/version.hpp
dhower-qc: More context: this class is just dealing with the version numbers ('~> 2.0') and isn't tied with an extension. Without the extension, there is nothing to distinguish COMPAT and GTE.

backends/cpp_hart_gen/lib/gen_cpp.rb
dhower-qc: context: the current code will spit out a user-defined literal for the unknown value (e.g., "0x111XXXX"_xb), which will result in a bits type that is only big enough to fit the value. That's a problem when we are compiling an IDL literal like `32'h111XXXX`. The fix will need to directly construct a PossiblyUnknown type with the correct data.

dhower-qc: Fixed in commit d94126133

backends/cpp_hart_gen/cpp/include/udb/bits.hpp
henrikg-qc: Talked to Derek and we agreed on leaving this one as is and change addsat according to my suggestion (current in this PR)

dhower-qc (changes requested):
This still needs to add a regression, correct?

ayosher (approved):
Xqci CSRs changes seems OK to me

henrikg-qc:
> I'm good. @ThinkOpenly, do you want to take a deeper look?

Planning on merging after lunch, ~1pm ET. @ThinkOpenly let me know if you want to see any changes.

PR #954, opened 2025-08-06
Fix CSR loading infrastructure: Add missing \ metadata

### LINKED ISSUE #953: CSR mismatch count and Cl checks not passed for every PR
CSR counts error : 
```
TestCli#test_list_csrs [test/test_cli.rb:74]:
Expected: 371
  Actual: 382
```

**Regress Smoke Github Cl did not passing Also for many PR**

Sukuna0007Abhi:
Fix missing $source metadata in YAML resolution pipeline

 Resolves "No $source" errors for all 371 CSR files
 Fixes infrastructure failures in regress-smoke tests
 Ensures proper source file tracking for database objects
 Maintains architectural separation between Python resolution and Ruby consumption

first mentioned on #936 & #933 
Fixes #953 
@ThinkOpenly @dhower-qc @AFOliveira
@adingank-qualcomm


AFOliveira (approved):
Thanks for looking into this. I tested it and I didn't get the error, but it's hard to say what fixed it. IMO we should merge this as it is a small change and looks like it really solves the error. 

I believe @adingank-qualcomm was also trying to fix this bug so his review would be helpful here.

Sukuna0007Abhi:
Thanks @AFOliveira, it actually solving this error `RuntimeError: No $source for misa
RuntimeError: No $source for mtvec` so that the misa/mtvec can be load for the yaml files, it's the issue occurred and CSR count mismatch and also regress smoke test not passing.

adingank-qualcomm:
> Thanks @AFOliveira, it actually solving this error `RuntimeError: No $source for misa RuntimeError: No $source for mtvec` so that the misa/mtvec can be load from the yaml files, it's the issue occurred and by doing this it solves this issues, e.g- CSR count mismatch and also regress smoke test not passing.

Thanks @Sukuna0007Abhi for fixing it at ($ ;-)source!

tools/ruby-gems/udb/lib/udb/architecture.rb
```diff
@@ -120,6 +120,7 @@ def self.generate_obj_methods(fn_name, arch_dir, obj_class)
         f.flock(File::LOCK_EX)
         obj_yaml = YAML.load(f.read, filename: obj_path, permitted_classes: [Date])
         f.flock(File::LOCK_UN)
+        obj_yaml["$source"] = Pathname.new(obj_path).realpath.to_s
```
dhower-qc: Unfortunately, this isn't how we want to fix the issue.

The `$source` key is supposed to point to the original, unresolved file that was used to generate a resolved file. The patch here is going to have $source pointing to the resolved file itself.

`$source` gets set during resolution in `tools/ruby-gems/udb/python/yaml_resolver.py#L553`

The question is why that intermittently does not appear.

Sukuna0007Abhi: @dhower-qc - Got it! You were pointing out that $source gets set in resolve_file(), but the issue was it was missing because write_resolved_file_and_validate() was calling resolve() again without preserving the $source. So I  fixed it by adding the missing assignment in the second phase at line 563, ensuring it correctly points to the original unresolved files in args.arch_dir (not the resolved files). Infrastructure working now - all 371 CSRs loading properly!

Sukuna0007Abhi:
Ready for Review @dhower-qc @ThinkOpenly

tools/ruby-gems/udb/python/yaml_resolver.py
```diff
@@ -560,6 +560,7 @@ def write_resolved_file_and_validate(
 ):
     resolved_path = os.path.join(resolved_dir, rel_path)
     resolved_obj = resolve(rel_path, args.arch_dir, do_checks)
+    resolved_obj["$source"] = os.path.join(args.arch_dir, rel_path)
```
dhower-qc, line 566: Bingo. Good job.

Sukuna0007Abhi: Thanks sir @dhower-qc

Sukuna0007Abhi:
Kindly could you merge this, @dhower-qc so that my other PR can be merged #933 , it stucked because of this issue 😞

Sukuna0007Abhi:
It's still not passing maybe some fixes will stable this test

Sukuna0007Abhi:
> > In short for resolver.rb : When Ruby calls Python resolver, this debug line would print the command being executed & broke CSR counting because the CSR count test reads ALL stdout lines, so it was counting this debug output as an extra CSR I think CLI commands should only output their actual results, not internal debug information. Like this :
> > ```
> >  Without commenting:
> > ./bin/udb list csrs
> > python3 /path/to/yaml_resolver.py resolve --quiet /path/to/arch /path/to/resolved  # ← This debug line
> > misa
> > mtvec
> > mcycle
> > Result: 374 lines counted instead of 371 CSRs
> > 
> > After commenting:
> > ./bin/udb list csrs
> > misa
> > mtvec  
> > mcycle
> > Result: 371 lines = 371 CSRs 
> > ```
> 
> If you just need to comment that line, why would you add `--quiet` arguments?

Oops @AFOliveira  thanks for pointing, I did wrong I thought the progress bar also mixed with stdout count but got it now it goes to stderr , so I fixed it  & I removed the --quiet flag entirely since progress bars and info messages already go to stderr, not stdout. The CSR count test only reads stdout, so there's no mixed issue. it will count the current and perfect exact csrs

Sukuna0007Abhi:
Pls again start the workflow it will again also pass now @AFOliveira

Sukuna0007Abhi:
Finally , Thanks @AFOliveira

Sukuna0007Abhi:
Ready for Review @dhower-qc @ThinkOpenly @AFOliveira

Sukuna0007Abhi:
Ok I think now I got it, just commenting out the Ruby puts wasn't enough. Python resolver was still printing progress bars to stdout, and there was also a "Using devcontainer environment" message from the setup script that was mixing in. I've now routed all of that to stderr so only the actual CSR names go to stdout.

So in Python Resolver : I routed all tqdm progress bars and info messages to stderr using file=sys.stderr and updated 
'Setup script' to fix the "Using devcontainer environment" message to go to stderr with >&2.

it now counts 373 CSRs matching the 373 YAML files exactly, with no stdout. Maybe it's correct, pls check @dhower-qc

Sukuna0007Abhi:
I think #967 is already solving the csr part so I think I should reset all these to that part where @dhower-qc first approved me

bin/setup
```diff
@@ -63,7 +63,7 @@ if [ ! -f ${ROOT}/.container-type ]; then
   echo ${CONTAINER_TYPE} > ${ROOT}/.container-type
 fi
 
-echo "Using ${CONTAINER_TYPE} environment"
+echo "Using ${CONTAINER_TYPE} environment" >&2
```
dhower-qc: revert this

dhower-qc (changes requested):

bin/setup
Sukuna0007Abhi: got it sir reverted

Sukuna0007Abhi:
pls check @dhower-qc

tools/ruby-gems/udb/lib/udb/resolver.rb
```diff
@@ -169,7 +169,7 @@ def any_newer?(target, deps)
     # run command in the shell. raise if exit is not zero
     sig { params(cmd: T::Array[String]).void }
     def run(cmd)
-      puts cmd.join(" ")
+      # puts cmd.join(" ")  # Uncomment for debugging
```
ThinkOpenly: This change seems unrelated to this PR.

Sukuna0007Abhi: Ok sir I will remove at first I thought it will also help to resolve the csr count mismatch but I think there is no need of it anymore.

Sukuna0007Abhi: updated

Sukuna0007Abhi:
Hi sir @dhower-qc kindly could you check and start the workflow

Sukuna0007Abhi:
I updated all @ThinkOpenly @dhower-qc @AFOliveira ready for check and the worflow

Sukuna0007Abhi:
Thanks @dhower-qc ,  pls check @ThinkOpenly happy if merged

Sukuna0007Abhi:
Thank you so much
 sir @dhower-qc sir @ThinkOpenly sir @AFOliveira 🙏

PR #912, opened 2025-07-20
Add Zhinx extension: ISA description and half-precision instruction YAMLs

blazethunderstorm:
This PR adds support for the unratified `Zhinx` extension in the RISC-V Unified Database.

✅ Includes:
- ISA description YAML for Zhinx: `spec/std/isa/zhinx.yaml`
- Instruction YAMLs for half-precision floating-point operations using integer registers:
  - `fadd.h`, `fsub.h`, `fmul.h`, `fdiv.h`, `fsqrt.h`
  - `fsgnj.h`, `fsgnjn.h`, `fsgnjx.h`
  - `fmin.h`, `fmax.h`
  - `feq.h`, `flt.h`, `fle.h`

ℹ️ These instructions mirror those in the `Zfh` extension but operate on integer (`x`) registers instead of floating-point (`f`) registers, per the Zhinx specification.

This contribution follows the format and structure used in other extensions like Zfinx and Zfh
---

Let me know if any changes are needed!

Resolves #410

PR #955, opened 2025-08-06
Feature/zhinx implementation

### LINKED ISSUE #410: Missing Zhinx Implementation
Zhinx extension has been added as needed by #340. However, only basic information has been added and needs completion.

HeerakKashyap:
### Summary
This PR implements the missing Zhinx extension by adding 37 instruction YAML files that define half-precision floating-point operations using integer registers.

### Changes
- Created `spec/std/isa/inst/Zhinx/` directory
- Added 37 instruction YAML files covering all half-precision operations
- Excluded transfer instructions (flh, fsh, fmv.h.x, fmv.x.h) as per Zhinx specification
- All instructions use `definedBy: Zhinx` field

### Technical Details
- Zhinx provides half-precision floating-point operations using x-registers instead of f-registers
- This allows systems without dedicated floating-point registers to perform half-precision math
- Implementation follows the same encoding and assembly format as Zfh instructions
- Excludes memory transfer operations which don't apply to x-register usage

### Related Issue
Fixes #410 - Missing Zhinx Implementation

### Testing
- All files have proper YAML syntax
- `definedBy: Zhinx` field correctly set in all 37 files
- Transfer instructions properly excluded
- Ready for integration with RISC-V Unified Database toolchain

### DIFF
--- a/spec/std/isa/inst/Zfh/fclass.h.yaml
+++ b/spec/std/isa/inst/Zfh/fclass.h.yaml
@@ -9,7 +9,7 @@ name: fclass.h
 long_name: No synopsis available
 description: |
   No description available.
-definedBy: Zfh
+definedBy: [Zfh, Zhinx]
 assembly: xd, fs1
 encoding:
   match: 111001000000-----001-----1010011
@@ -25,3 +25,10 @@ access:
   vu: always
 data_independent_timing: false
 operation(): |
+  if (implemented?(ExtensionName::Zhinx)) {
+    # Zhinx: use integer registers
+    X[rd] = fclass_h(X[rs1]);
+  } else {
+    # Zfh: use floating-point registers
+    X[rd] = fclass_h(F[rs1]);
+  }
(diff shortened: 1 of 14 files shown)


AFOliveira:
For the intructions - https://github.com/riscv-software-src/riscv-unified-db/pull/912#pullrequestreview-3039716657

For the IDL - I don't understand why would you set globals and not use them in the instructions.

HeerakKashyap:
> Also, please note that this is one of the harder extensions in IDL, so proceed with care. Probably, trying to do all at once will be too much work with too many errors, which will also be hard to review. I would advise you to start small

yeah sure, I was looking after this. For now kindly see if my suggested changes work, if not i'll split into smaller sections..

HeerakKashyap:
> This is mixing Zhinx and some additions to H; they will need to be split into multiple PRs.
> 
> On Zhinx, we can't create _new_ files since the instructions already exist in Zfh.
> 
> Instead, we have to modify the existing files to do a different operation when Zhinx is implemented.

Okay sir , I'll look for this

Issue #563, opened 2025-03-27
Add missing S CSRs


dhower-qc:
Need to add:

- SEDELEG
- SIDELEG
- SIE
- SCOUNTINHIBIT

sudo-apt-abdullah:
@ThinkOpenly, I'll try to give it a try as well. Can you please assign it to me ?

PR #910, opened 2025-07-18
Add Smaia/Ssaia (Advanced Interrupt Architecture) support

7908837174:
## Summary

This PR implements support for the RISC-V Advanced Interrupt Architecture (AIA) extensions Smaia and Ssaia as requested in issue #216.

### New CSR Definitions

**Smaia (Machine-level AIA):**
- `mvien` (0x308) - Machine Virtual Interrupt Enable
- `mvip` (0x309) - Machine Virtual Interrupt Pending
- `mtopei` (0x35C) - Machine Top External Interrupt

**Ssaia (Supervisor-level AIA):**
- `sieh` (0x114) - Supervisor Interrupt Enable High
- `siph` (0x154) - Supervisor Interrupt Pending High
- `stopei` (0x15C) - Supervisor Top External Interrupt
- `stopi` (0x15B) - Supervisor Top Interrupt

### IDL Hart-side Logic

- Extended `interrupts.idl` with IMSIC (Incoming Message-Signaled Interrupt Controller) support
- Added interrupt file management for machine, supervisor, and virtual supervisor levels
- Implemented TOPEI register update logic
- Added functions for setting/clearing IMSIC interrupts
- Integrated AIA virtual interrupts into the main interrupt handling flow

### Key Features

- **IMSIC Support**: Full interrupt file management with 2048 entries per privilege level
- **Priority-based Interrupt Handling**: Automatic TOPEI register updates with highest priority pending interrupts
- **Virtual Interrupt Support**: Integration with existing interrupt enable/pending mechanisms
- **CSR Access Functions**: Proper sw_read/sw_write implementations for TOPEI registers with claim-and-clear semantics

## Testing

The implementation follows the existing UDB patterns and schemas. All new CSR files conform to the `csr_schema.json` specification.

## Related Issues

Fixes #216
Contributes to #508 (Add YAML files for all missing extensions)

## Notes

- The implementation provides a foundation for AIA support that can be extended as needed
- TOPEI registers implement the standard claim-and-clear behavior when written
- The IDL extensions are backward compatible with existing interrupt handling
- All CSRs include comprehensive documentation following UDB conventions

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) with guidance from the PR author

PR #958, opened 2025-08-07
docs(S csr): sie and scountinhibit - issue #563

sudo-apt-abdullah:


spec/std/isa/csr/scountinhibit.yaml
```diff
@@ -0,0 +1,671 @@
...
+      } else {
+        return csr_value.HPM31
+      }
+sw_read(): |
+  if ((mode() == PrivilegeMode::M && CSR[menvcfg].CDE == 0) {
+    raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+  } else if ((mode() == PrivilegeMode::VS && CSR[menvcfg].CDE == 1'b1) ||
+             (mode() == PrivilegeMode::VU && CSR[menvcfg].CDE == 1'b1)) {
+    raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
+  }
```
dhower-qc: sw_read needs to return a value

spec/std/isa/csr/scountinhibit.yaml
dhower-qc, line 1: This is a great start.

scountinhibit is a restricted alias of mcountinhibit. As such, it has no state of its own. So, most of what you've written in IDL is good except that the writes need to propagate to mcountinhibit (e.g., `CSR[mcountinhibit].CY = csr_value.CY`) and the read needs to return the masked view of mcountinhibit.

spec/std/isa/csr/sie.yaml
dhower-qc, line 1: since this is also an alias register, you need to implement sw_read/sw_writes to forward to `mie`

spec/std/isa/csr/scountinhibit.yaml
sudo-apt-abdullah: > This is a great start.
Thank you for your review and feedback. 

> scountinhibit is a restricted alias of mcountinhibit. As such, it has no state of its own. So, most of what you've written in IDL is good except that the writes need to propagate to mcountinhibit (e.g., `CSR[mcountinhibit].CY = csr_value.CY`) and the read needs to return the masked view of mcountinhibit.

I updated the file to propagate writes to mcountinhibit as you suggested.

spec/std/isa/csr/scountinhibit.yaml
sudo-apt-abdullah: I added code to return a value. Let me know if there's still something missing.

spec/std/isa/csr/sie.yaml
sudo-apt-abdullah: Yes, I’ve implemented sw_read and sw_write to forward to mie. Please have a look—I'd appreciate it if you could check in case I missed anything.

spec/std/isa/csr/scountinhibit.yaml
```diff
@@ -0,0 +1,704 @@
...
+length: SXLEN
+definedBy: Smcdeleg
+fields:
+  CY:
+    location: 0
+    description: |
+      Cycle inhibit. When 1, the cycle counter does not increment. Only meaningful if
+      mcycle is delegated to S-mode; otherwise reads as 0.
+    type(): |
+      return (CSR[menvcfg].CDE == 1'b1 && CSR[mcounteren].CY == 1'b1) ? CsrFieldType::RW : CsrFieldType::RO;
```
dhower-qc: type() is supposed to capture the configured, rather than dynamic/runtime, type of the field. Since CSR[menvcfg].CDE and CSR[mcounteren].CY can both change at runtime, they shouldn't affect the type.

The COUNTINHIBIT_EN parameter _will_ affect the type here, though.

spec/std/isa/csr/scountinhibit.yaml
```diff
@@ -0,0 +1,704 @@
...
+    sw_write(csr_value): |
+      if (mode() == PrivilegeMode::M && CSR[menvcfg].CDE == 0) {
+        raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+      } else if ((mode() == PrivilegeMode::VS && CSR[menvcfg].CDE == 1'b1) ||
+                 (mode() == PrivilegeMode::VU && CSR[menvcfg].CDE == 1'b1)) {
+        raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
+      } else {
+        CSR[mcountinhibit].CY = csr_value.CY;
+        return csr_value.CY
+      }
```
dhower-qc: Just a style preference: I think it's more clear to raise the CDE check:

```
if (CSR[menvcfg].CDE == 0) {
  if (mode() == M) {
  } else if (mode() == VS ..) {
  }
}
```

spec/std/isa/csr/scountinhibit.yaml
```diff
@@ -0,0 +1,704 @@
...
+    location: 0
+    description: |
+      Cycle inhibit. When 1, the cycle counter does not increment. Only meaningful if
+      mcycle is delegated to S-mode; otherwise reads as 0.
+    type(): |
+      return (CSR[menvcfg].CDE == 1'b1 && CSR[mcounteren].CY == 1'b1) ? CsrFieldType::RW : CsrFieldType::RO;
+    reset_value(): |
+      return COUNTINHIBIT_EN[0] ? UNDEFINED_LEGAL : 0;
+    sw_write(csr_value): |
+      if (mode() == PrivilegeMode::M && CSR[menvcfg].CDE == 0) {
```
dhower-qc: Unfortunately, menvcfg is not guaranteed to be implemented. According to the menvcfg data:

```yaml
definedBy:
  allOf:
    - name: Sm
      version: ">=1.12"
    - name: U
```

Thus, to truly represent the spec, this needs to check the above condition.

You can do that with:

```
if (implemented_csr?(CsrName::menvcfg) && (CSR[menvcfg].CDE == 1'b1))
```

spec/std/isa/csr/scountinhibit.yaml
```diff
@@ -0,0 +1,704 @@
...
+        return csr_value.IR
+      }
+  HPM3:
+    location: 3
+    description: |
+      [when="COUNTINHIBIT_EN[3] == true"]
+      When bit is set `hpmcounter3.COUNT` does not increment; when 0, it increments normally.
+
+      [when="COUNTINHIBIT_EN[3] == false"]
+      Since hpmcounter3 is not implemented, this field is read-only zero.
```
dhower-qc: These `[when ..]` tags are deprecated. Instead, use:

```yaml
description:
  - id: csr-field-hpmcounter3.COUNT-set-effect
    normative: true
    when:
      param:
        name: COUNTINHIBIT_EN
        index: 3
        value: true
    text: |
      When bit is set `hpmcounter3.COUNT` does not increment; when 0, `hpmcounter3.COUNT` increments normally.
```

and similar for the false case

(this depends on #891)

spec/std/isa/csr/sie.yaml
```diff
@@ -0,0 +1,96 @@
...
+    definedBy: Sscofpmf
+    description: |
+      If the Sscofpmf extension is implemented, bit sie.LCOFIE is the interrupt-enable
+      bit for local-counter-overflow interrupts. If the Sscofpmf extension is not
+      implemented, sie.LCOFIE is read-only zeros.
+    sw_write(csr_value): |
+      CSR[mie].LCOFIE = csr_value.LCOFIE;
+      return csr_value.LCOFIE;
+sw_read(): |
+  return $bits(CSR[mie]);
```
dhower-qc: This needs to be masked, since only a subset of bits from mie appear in sie

spec/std/isa/csr/scountinhibit.yaml
```diff
@@ -0,0 +1,704 @@
...
+        return csr_value.HPM31
+      }
+sw_read(): |
+  if (mode() == PrivilegeMode::M && CSR[menvcfg].CDE == 0) {
+    raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+  } else if ((mode() == PrivilegeMode::VS && CSR[menvcfg].CDE == 1'b1) ||
+             (mode() == PrivilegeMode::VU && CSR[menvcfg].CDE == 1'b1)) {
+    raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
+  } else {
+    return $bits(CSR[mcountinhibit]);
```
dhower-qc: This needs to be masked since counters not delegated to s mode must read as 0.

spec/std/isa/csr/sie.yaml
ThinkOpenly, line 1: Do we need to define a field(s) for the "other" interrupt enable bits, that are for "platform use"? PR #910 depends on that, since it's defining CSR `sieh`, the high-order bits of `sie`.

spec/std/isa/csr/scountinhibit.yaml
```diff
@@ -0,0 +1,704 @@
...
+      if (mode() == PrivilegeMode::M && CSR[menvcfg].CDE == 0) {
+        raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+      } else if ((mode() == PrivilegeMode::VS && CSR[menvcfg].CDE == 1'b1) ||
+                 (mode() == PrivilegeMode::VU && CSR[menvcfg].CDE == 1'b1)) {
+        raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
+      } else {
+        CSR[mcountinhibit].HPM30 = csr_value.HPM30;
+        return csr_value.30
+      }
+  HPM31:
```
ThinkOpenly, line 1102: Given all the repetition, it might be nice to use the "layout" functionality used by other CSRs like `spec/std/isa/csr/I/mcounteren.layout`, which is very similar in that it also has "HPM<n>" fields.

spec/std/isa/csr/sie.yaml
sudo-apt-abdullah: Tbh I'm not sure, but since spec doesn’t define any standard bits in sieh and their semantics are platform-specific, I believe there's no need to define explicit fields. Instead, we could just model it as a single value field, with a note that these bits are reserved for platform use:
```
fields:
  VALUE:
    location_rv32: 31-0
    description: |
       The base spec doesn’t define any standard bits in sieh, those upper interrupt enables are reserved for platform use.
```
Let me know if I’m overlooking something!

spec/std/isa/csr/scountinhibit.yaml
sudo-apt-abdullah: Thanks for clarification, I've made the required changes.

spec/std/isa/csr/scountinhibit.yaml
sudo-apt-abdullah: Yes, it will be better.

spec/std/isa/csr/scountinhibit.yaml
sudo-apt-abdullah: I've added this condition as well.

spec/std/isa/csr/scountinhibit.yaml
sudo-apt-abdullah: At the moment, the schema doesn’t support this, but I went ahead and used it anyway. Please let me know if any improvements are needed.

spec/std/isa/csr/sie.yaml
sudo-apt-abdullah: Alright! Now, masked it to read only a subset of bits from mie.

spec/std/isa/csr/scountinhibit.yaml
sudo-apt-abdullah: Have a look and let me know if still something missing, thanks!

spec/std/isa/csr/scountinhibit.yaml
sudo-apt-abdullah: Thanks for suggestion, it's done.

spec/std/isa/csr/sie.yaml
ThinkOpenly: One complication is that `sieh` needs to serve as an alias for the high-order bits of `sie`, although it looks like there is a precedent:
```
spec/std/isa/csr/Zihpm/hpmcounter26h.yaml:    alias: mhpmcounter26h.COUNT[63:32]
```
So, you can move forward with that approach.

sudo-apt-abdullah: I believe, putting `alias: sie[63:32]` in sieh.yaml file will be a right thing to do? Let me know if I'm overlooking or something needs to be added in sie.yaml.

ThinkOpenly: Agreed.

spec/std/isa/csr/scountinhibit.layout
```diff
@@ -0,0 +1,119 @@
...
+            value: true
+        text: |
+          When bit is set `hpmcounter<%= hpm_num %>.COUNT` does not increment; when 0, `hpmcounter<%= hpm_num %>.COUNT` increments normally.
+      - id: unimplemented-hpmcounter<%= hpm_num %>
+        normative: false
+        when:
+          param:
+            name: COUNTINHIBIT_EN
+            index: <%= hpm_num %>
+            value: flase
```
ThinkOpenly:
```suggestion
            value: false
```

spec/std/isa/csr/scountinhibit.layout
```diff
@@ -0,0 +1,119 @@
...
+      if (mode() == PrivilegeMode::M) {
+      raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+      }
+    } else if (implemented_csr?(CsrName::menvcfg) && CSR[menvcfg].CDE == 1'b1) {
+      if ((mode() == PrivilegeMode::VS) || (mode() == PrivilegeMode::VU)) {
+      raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
+      }
+    } else {
+    return ($bits(CSR[mcountinhibit]) & ($bits(CSR[scounteren]) & ~32'd2)) ;
+    }
```
ThinkOpenly:
```suggestion
    }
    return ($bits(CSR[mcountinhibit]) & ($bits(CSR[scounteren]) & ~32'd2)) ;
```
although, what purpose is the `~32'd2`?

sudo-apt-abdullah: > although, what purpose is the `~32'd2`?

Just to read second bit as zero.

spec/std/isa/csr/scountinhibit.layout
sudo-apt-abdullah: Oops, fixed the typo :)

sudo-apt-abdullah:
> Note that to _realize_ the layout results (to generate `scountinhibit.yaml`), you'll need to add a new stanza in `/Rakefile`. There are examples there already, but ask if you need some help.

Thanks for guidance! I've added that as well, please have a look and let me know if something is wrong, thank!

spec/std/isa/csr/scountinhibit.yaml
```diff
@@ -0,0 +1,1178 @@
...
+name: scountinhibit
+long_name: Supervisor Counter-Inhibit Register
+address: 0x120
+writable: true
+priv_mode: S
+description: |
+  For counters delegated to S-mode, the associated `mcountinhibit` bits can be accessed
+  via `scountinhibit`. For counters not delegated to S-mode, the associated bits in
+  `scountinhibit` are read-only zero.
+length: SXLEN
```
ThinkOpenly: Since this is a "masked alias of `mcountinhibit`", we need some adjustments here, including:
```suggestion
length: 32
```

Further, all of the fields need to be defined as `alias` of the corresponding field of `mcountinhibit`.

spec/std/isa/csr/scountinhibit.layout
```diff
@@ -0,0 +1,117 @@
...
+        raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+        }
+      } else if (implemented_csr?(CsrName::menvcfg) && CSR[menvcfg].CDE == 1'b1) {
+        if ((mode() == PrivilegeMode::VS) || (mode() == PrivilegeMode::VU)) {
+        raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
+        }
+      } else {
+        CSR[mcountinhibit].CY = csr_value.CY;
+        return csr_value.CY;
+      }
```
ThinkOpenly: Let's make it obvious that this always returns _something_:
```suggestion
      }
      CSR[mcountinhibit].CY = csr_value.CY;
      return csr_value.CY;
```

spec/std/isa/csr/scountinhibit.layout
```diff
@@ -0,0 +1,117 @@
...
+        raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+        }
+      } else if (implemented_csr?(CsrName::menvcfg) && CSR[menvcfg].CDE == 1'b1) {
+        if ((mode() == PrivilegeMode::VS) || (mode() == PrivilegeMode::VU)) {
+        raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
+        }
+      } else {
+        CSR[mcountinhibit].IR = csr_value.IR;
+        return csr_value.IR;
+      }
```
ThinkOpenly:
```suggestion
      }
      CSR[mcountinhibit].IR = csr_value.IR;
      return csr_value.IR;
```

spec/std/isa/csr/scountinhibit.layout
```diff
@@ -0,0 +1,117 @@
...
+      Instructions-retired inhibit. When 1, the instret counter does not increment. Only
+      meaningful if minstret is delegated to S-mode; else reads 0.
+    type(): |
+      return COUNTINHIBIT_EN[2] ? CsrFieldType::RW : CsrFieldType::RO;
+    reset_value(): |
+      return COUNTINHIBIT_EN[2] ? UNDEFINED_LEGAL : 0;
+    sw_write(csr_value): |
+      if (implemented_csr?(CsrName::menvcfg) && CSR[menvcfg].CDE == 1'b0) {
+        if (mode() == PrivilegeMode::M) {
+        raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
```
ThinkOpenly:
```suggestion
          raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
```

spec/std/isa/csr/scountinhibit.layout
```diff
@@ -0,0 +1,117 @@
...
+        }
+      } else if (implemented_csr?(CsrName::menvcfg) && CSR[menvcfg].CDE == 1'b1) {
+        if ((mode() == PrivilegeMode::VS) || (mode() == PrivilegeMode::VU)) {
+        raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
```
ThinkOpenly:
```suggestion
          raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
```

spec/std/isa/csr/scountinhibit.layout
```diff
@@ -0,0 +1,117 @@
...
+        text: |
+          Since hpmcounter<%= hpm_num %> is not implemented, this field is read-only zero.
+    type(): |
+      return COUNTINHIBIT_EN[<%= hpm_num %>] ? CsrFieldType::RW : CsrFieldType::RO;
+    reset_value(): |
+      return COUNTINHIBIT_EN[<%= hpm_num %>] ? UNDEFINED_LEGAL : 0;
+    sw_write(csr_value): |
+      if (implemented_csr?(CsrName::menvcfg) && CSR[menvcfg].CDE == 1'b0) {
+        if (mode() == PrivilegeMode::M) {
+        raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
```
ThinkOpenly:
```suggestion
          raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
```

spec/std/isa/csr/scountinhibit.layout
```diff
@@ -0,0 +1,117 @@
...
+        }
+      } else if (implemented_csr?(CsrName::menvcfg) && CSR[menvcfg].CDE == 1'b1) {
+        if ((mode() == PrivilegeMode::VS) || (mode() == PrivilegeMode::VU)) {
+        raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
```
ThinkOpenly:
```suggestion
          raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
```

spec/std/isa/csr/scountinhibit.layout
```diff
@@ -0,0 +1,117 @@
...
+        }
+      } else {
+        CSR[mcountinhibit].HPM<%= hpm_num %> = csr_value.HPM<%= hpm_num %>;
+        return csr_value.HPM<%= hpm_num %>;
+      }
```
ThinkOpenly:
```suggestion
      } else {
      CSR[mcountinhibit].HPM<%= hpm_num %> = csr_value.HPM<%= hpm_num %>;
      return csr_value.HPM<%= hpm_num %>;
```

spec/std/isa/csr/scountinhibit.layout
```diff
@@ -0,0 +1,117 @@
...
+  <%- end -%>
+
+sw_read(): |
+    if (implemented_csr?(CsrName::menvcfg) && CSR[menvcfg].CDE == 1'b0) {
+      if (mode() == PrivilegeMode::M) {
+      raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
```
ThinkOpenly:
```suggestion
        raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
```

spec/std/isa/csr/scountinhibit.layout
```diff
@@ -0,0 +1,117 @@
...
+      }
+    } else if (implemented_csr?(CsrName::menvcfg) && CSR[menvcfg].CDE == 1'b1) {
+      if ((mode() == PrivilegeMode::VS) || (mode() == PrivilegeMode::VU)) {
+      raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
```
ThinkOpenly:
```suggestion
        raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
```

spec/std/isa/csr/scountinhibit.layout
```diff
@@ -0,0 +1,117 @@
...
+      }
+    } return ($bits(CSR[mcountinhibit]) & ($bits(CSR[scounteren]) & ~32'd2)) ;
```
ThinkOpenly:
```suggestion
    }
    return ($bits(CSR[mcountinhibit]) & ($bits(CSR[scounteren]) & ~32'd2));
```

spec/std/isa/csr/sie.yaml
```diff
@@ -0,0 +1,96 @@
...
+    definedBy: Sscofpmf
+    description: |
+      If the Sscofpmf extension is implemented, bit sie.LCOFIE is the interrupt-enable
+      bit for local-counter-overflow interrupts. If the Sscofpmf extension is not
+      implemented, sie.LCOFIE is read-only zeros.
+    sw_write(csr_value): |
+      CSR[mie].LCOFIE = csr_value.LCOFIE;
+      return csr_value.LCOFIE;
+sw_read(): |
+  return ($bits(CSR[mie]) & 64'h2222);
```
ThinkOpenly: Personal preference for my software peeps:
```suggestion
  return $bits(CSR[mie]) & 0x2222;
```

spec/std/isa/csr/scountinhibit.layout
```diff
@@ -0,0 +1,117 @@
...
+      Cycle inhibit. When 1, the cycle counter does not increment. Only meaningful if
+      mcycle is delegated to S-mode; otherwise reads as 0.
+    type(): |
+      return COUNTINHIBIT_EN[0] ? CsrFieldType::RW : CsrFieldType::RO;
+    reset_value(): |
+      return COUNTINHIBIT_EN[0] ? UNDEFINED_LEGAL : 0;
+    sw_write(csr_value): |
+      if (implemented_csr?(CsrName::menvcfg) && CSR[menvcfg].CDE == 1'b0) {
+        if (mode() == PrivilegeMode::M) {
+        raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
```
ThinkOpenly:
```suggestion
          raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
```

spec/std/isa/csr/scountinhibit.layout
```diff
@@ -0,0 +1,117 @@
...
+        }
+      } else if (implemented_csr?(CsrName::menvcfg) && CSR[menvcfg].CDE == 1'b1) {
+        if ((mode() == PrivilegeMode::VS) || (mode() == PrivilegeMode::VU)) {
+        raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
```
ThinkOpenly:
```suggestion
          raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
```

spec/std/isa/csr/scountinhibit.yaml
ThinkOpenly: You may need to rebase your branch on #891 for the CI to pass when we get to that point.

spec/std/isa/csr/scountinhibit.yaml
ThinkOpenly: If done like the others, there should be a "WARNING" at the top of the generated YAML file (inserted by `insert_warning`), but I don't see that. Did you add the generated YAML file here?

spec/std/isa/csr/scountinhibit.yaml
sudo-apt-abdullah: Done!

sudo-apt-abdullah:
I forgot to include the generated file initially, but I’ve pushed it now. I also attempted to rebase on PR #891, but things are getting a bit messy and I’m not sure if I did it correctly. Should I undo the rebase and go back to my last commit? I’m fine waiting for PR #891 to merge first.

spec/std/isa/csr/sie.yaml
ThinkOpenly: I don't see `sieh.yaml` in the PR. Were you planning on adding that?

sudo-apt-abdullah: It's already in #910. Let me know if I'm supposed to add it in this PR, thanks!

sudo-apt-abdullah:
> Agree it looks messy. It's not quite what I expected. When looking at the commit history, I expected to see the commits from #891 and then your 7 or so commits on top, but they're scattered. It's up to you if you want to wait. I'm not aware of anyone actively waiting for these CSR definitions. #891 may take a while to land, though.

I'm fine with waiting for #891 to land :)

spec/std/isa/csr/sie.yaml
ThinkOpenly: Oh, there is no `sieh` definition yet, waiting on #910 to be merged. I see. Do you want to wait for #891 _and_ #910, then? :-)

sudo-apt-abdullah: Perfectly fine for me :)

PR #960, opened 2025-08-07
feat: Add missing platform-specific parameters to Zihpm extension

### LINKED ISSUE #69: Zihpm.yaml missing several parameters and has vague description for HPM_COUNTER_EN
Couple of issues I noticed with Zihpm.

The Unpriv ISA manual chapter on Zihpm says "The implemented number and width of these additional counters, and the set of events they count, is platform-specific. Accessing an unimplemented or ill-configured counter may cause an illegal-instruction exception or may return a constant value."

Seems like we need a bunch more of parameters in the Zihpm.yaml to capture all these possibilities. Should we add them?

The other issue is that the Zihpm.yaml has this description for HPM_COUNTER_EN:
        "List of HPM counters that are enabled."
What does "enabled" mean? Counting some event (i.e., run-time controlled) or provides a non-zero value when the associated CSR is read (since all 29 perf counters are always implemented but can be read-only-0 if not actually functional).

Sukuna0007Abhi:
## Summary
Adds missing platform-specific configuration parameters to Zihpm extension as requested in issue #69.

## Changes
- **NUM_HPM_COUNTERS**: Specifies number of additional HPM counters (0-29)
- **UNIMPLEMENTED_HPM_BEHAVIOR**: Defines behavior for unimplemented counter access
- **UNIMPLEMENTED_HPM_CONSTANT_VALUE**: Constant value returned when applicable

## Addresses
- Platform-specific counter implementation requirements from ISA manual
- ThinkOpenly's final parameter specifications in issue comments

Closes #69 

Ready for Review @ThinkOpenly @dhower-qc @AFOliveira


spec/std/isa/ext/Zihpm.yaml
```diff
@@ -7,11 +7,76 @@ $schema: "ext_schema.json#"
...
+  an illegal-instruction exception or may return a constant value.
 type: unprivileged
 versions:
   - version: "2.0.0"
     state: ratified
     ratification_date: null
     requires:
       name: Smhpm
+params:
+  NUM_HPM_COUNTERS:
```
dhower-qc: We are actually tracking this through HPM_COUNTER_EN (Smhpm.yaml), which lets you enable/disable counters individually rather than assuming all the enabled ones are sequential. As such, we don't need this parameter.

spec/std/isa/ext/Zihpm.yaml
```diff
@@ -7,11 +7,76 @@ $schema: "ext_schema.json#"
 kind: extension
 name: Zihpm
 long_name: Programmable hardware performance counters
-description: Programmable hardware performance counters
+description:
```
dhower-qc, line 10: Zihpm actually just defines the user-mode CSRs hpmcounter[3-31] that alias mhpmcounter[3-31].

The real counters are defined by Smhpm (an extension for purposes of UDB, but not a ratified extension [yet]).

spec/std/isa/ext/Zihpm.yaml
```diff
@@ -7,11 +7,76 @@ $schema: "ext_schema.json#"
...
+      the upper 32 bits are accessible via the hpmcounter<N>h CSRs. The actual hardware 
+      implementation may utilize fewer bits internally, but must present a 64-bit interface.
+      
+      When NUM_HPM_COUNTERS is less than 29, the unimplemented counters (hpmcounter<3+NUM_HPM_COUNTERS>
+      through hpmcounter31) will behave according to UNIMPLEMENTED_HPM_BEHAVIOR.
+    schema:
+      type: integer
+      minimum: 0
+      maximum: 29
+  UNIMPLEMENTED_HPM_BEHAVIOR:
```
dhower-qc: The spec says:

> All counters should be implemented, but a legal implementation is to make both the counter and its corresponding event selector be read-only 0.

So we don't need a parameter for 'unimplemented' counters.

dhower-qc (changes requested):

spec/std/isa/ext/Zihpm.yaml
Sukuna0007Abhi: Ok sir got it

Sukuna0007Abhi:
😔sir @dhower-qc  there is only work I can do for it, is just update the description, other than is there anything I can do for it either? Could you pls suggest?

ThinkOpenly:
The CI tests for the current content of this PR are failing due to trailing spaces, however...

The text you've suggested is OK, but it differs from the content in the spec. If you wanted to include the content from the spec (chapter 7.2 "Zihpm" Extension for Hardware Performance Counters), that would probably be preferred. It's only about 7 paragraphs.

Issue #211, opened 2024-10-30
Implement H in IDL


dhower-qc:
Much of H has already been implemented in IDL, but more work and all testing is needed.

omsuneri:
@dhower-qc can you please assign me this issue and what is your highest priority for H to idl conversion

PR #961, opened 2025-08-08
Feature/hypervisor only

HeerakKashyap:
Implements the RISC-V H (Hypervisor) extension with complete IDL support and all hypervisor instructions.

 Changes
- Added `H.idl` with hypervisor operational semantics
- Added 15 hypervisor instruction YAMLs (hlv.*, hsv.*, hfence.*)
- Added H extension definition with test procedures
- Connected IDL functions to instruction operations
- Fixed compilation issues (VMID_WIDTH, VmaOrderType, pc references)

 Key Features
**Two-stage address translation**: VS-stage and G-stage translation support
- **VMID management**: Virtual Machine ID handling
- **Exception handling**: Proper hypervisor exception setup
- **TLB management**: Translation Lookaside Buffer invalidation
- **Privilege mode support**: VS, VU, HS mode operations

IDL Functions Implemented
- `hypervisor_enabled()` - Check H extension availability
- `get_virtualization_mode()` / `set_virtualization_mode()` - V bit management
- `translate_two_stage()` - Two-stage address translation
- `h_translate_gstage()` - G-stage translation wrapper
- `hfence_vvma()` / `hfence_gvma()` - TLB invalidation
- `hlv_*()` / `hsv_*()` / `hlvx_*()` - Memory operations
- `get_current_vmid()` / `set_current_vmid()` - VMID management

Related
- Addresses issue #211 "Implement H in IDL"
- Zhinx extension handled in separate PR (modifying existing Zfh files) - https://github.com/riscv-software-src/riscv-unified-db/pull/955


spec/std/isa/isa/H.idl
```diff
@@ -0,0 +1,421 @@
...
+}
+
+# Get current virtualization mode (V bit)
+function get_virtualization_mode {
+  returns Boolean
+  description {
+    Get the current virtualization mode (V bit)
+  }
+  body {
+    return V;
```
dhower-qc, line 25: Where is V defined?

spec/std/isa/isa/H.idl
```diff
@@ -0,0 +1,421 @@
...
+# This file contains hypervisor-specific functions and utilities
+
+# Check if hypervisor extension is enabled
+function hypervisor_enabled {
+  returns Boolean
+  description {
+    Check if the hypervisor extension is enabled via misa.H bit
+  }
+  body {
+    return implemented?(ExtensionName::H);
```
dhower-qc, line 14: also needs to check for dynamic disabling:

```
return
  implemented?(ExtensionName::H) &&
  (!MISA_CSR_IMPLEMENTED || (CSR[misa].H == 1'b1));
```

spec/std/isa/isa/H.idl
```diff
@@ -0,0 +1,421 @@
...
+  description {
+    Check if currently in HS-mode
+  }
+  body {
+    return (mode() == PrivilegeMode::HS);
+  }
+}
+
+# Get guest physical address from virtual address
+function translate_guest_physical {
```
dhower-qc, line 74: Does this do something not already handled through translation in globals.isa?

spec/std/isa/isa/H.idl
```diff
@@ -0,0 +1,421 @@
...
+      TranslationResult gs_result = translate_gstage(vs_result.paddr, vaddr, op, mode, 0);
+      return gs_result;
+    } else {
+      return vs_result;
+    }
+  }
+}
+
+# Two-stage address translation for hypervisor
+function translate_two_stage {
```
dhower-qc, line 96: same question

dhower-qc (changes requested):
I'm excited you are taking this on. It will be a long process, so let's do a few things:

1. Mark this PR as a draft.
2. We need to set up a phone call. I'll email you on the side.

spec/std/isa/isa/H.idl
HeerakKashyap: V is defined as a global state variable in globals.isa. It's part of the core RISC-V state that tracks the virtualization mode (V bit). The V bit indicates whether the hart is currently executing in a guest (V=1) or in host mode (V=0)

spec/std/isa/isa/H.idl
HeerakKashyap: Thank you for catching that Sir! You're right. The current implementation only checks if the H extension is implemented, but doesn't account for dynamic disabling via the misa.H bit.

spec/std/isa/isa/H.idl
HeerakKashyap: These functions provide hypervisor-specific wrappers around the existing translation infrastructure in globals.isa, but they serve a different purpose than the base translation functions.

translate_guest_physical(): This function specifically handles two-stage address translation for hypervisor operations. While globals.isa has the base translate() function for single-stage translation, this function:
- First performs VS-stage translation (virtual → guest physical)
 - Then performs G-stage translation (guest physical → supervisor physical) only when V=1
 - Returns the final supervisor physical address for hypervisor memory operations.

translate_two_stage(): This is a hypervisor-specific entry point that:
 - Checks virtualization mode (V bit)
 - Handles the complete two-stage translation flow
 - Provides proper error handling for hypervisor-specific exceptions
-  Is called by hypervisor load/store instructions (hlv.*, hsv.*)

The existing translate() in globals.isa only handles single-stage translation - Hypervisor operations need conditional two-stage translation based on V bit.
 Both functions call the existing translate() and translate_gstage() functions from globals.isa
  They add the hypervisor-specific control logic on top of the base translation, This follows the same pattern as other extension-specific translation functions

PR #965, opened 2025-08-08
Add normative rules for I extension

james-ball-qualcomm:
This PR contains a number of commits over the last few weeks while working with Jamie (@jrgrimm11) to add normative rule tags in the ISA manuals for the I extension. I edited many YAML files for all the base RV32I and RV64I instructions to add normative rules to them. Also added in coverage points from the CVW project (https://github.com/openhwgroup/cvw-arch-verif) to demonstrate to them how UDB can be used to generate a test plan (CTP) instead of having a hand-written test plan that duplicates information in UDB.

In the process of adding normative rules for the I extension, since this is the first use of the normative rule tooling UDB I made many changes/enhancements to the tooling to get good results. Going forward I'll keep adding normative rules to UDB YAML files to work towards having everything in the RVI20 Profiles be covered. There should be fewer changes to the tooling and mostly just additions to existing YAML files.

### DIFF
--- a/backends/proc_crd/tasks.rake
+++ b/backends/proc_crd/tasks.rake
@@ -31,20 +31,7 @@ Dir.glob("#{$resolver.std_path}/proc_cert_model/*.yaml") do |f|
     "#{$root}/backends/proc_cert/templates/rev_history.adoc.erb",
     "#{PROC_CRD_DOC_DIR}/templates/proc_crd.adoc.erb"
   ] do |t|
-    begin
-      proc_cert_create_adoc("#{PROC_CRD_DOC_DIR}/templates/proc_crd.adoc.erb", t.name, model_name, nil)
-    rescue => e
-      # Send to stdout since UDB sends tons of cr*p to stderr that floods one with useless information.
-      # Note that the $logger sends to stdout so anything send to $logger actually gets displayed as useful
-      # information if one just redirects stderr to /dev/null (e.g., in bash, run "./do <task-name> 2>/dev/null).
-      puts "Caught error: #{e.message}"
-
-      # Only print out 1st two lines of stack backtrace to stdout.
-      puts e.backtrace.take(2)
-
-      # Send full stacktrace to stderr with "warn".
-      warn e.backtrace
-    end
+    proc_cert_create_adoc("#{PROC_CRD_DOC_DIR}/templates/proc_crd.adoc.erb", t.name, model_name, nil)
   end
 
   file "#{PROC_CRD_GEN_DIR}/pdf/#{model_name}-CRD.pdf" => [
--- a/backends/proc_ctp/tasks.rake
+++ b/backends/proc_ctp/tasks.rake
@@ -32,44 +32,31 @@ Dir.glob("#{$resolver.std_path}/proc_cert_model/*.yaml") do |f|
     "#{$root}/backends/proc_cert/templates/priv_modes.adoc.erb",
     "#{PROC_CTP_DOC_DIR}/templates/proc_ctp.adoc.erb"
   ] do |t|
-    begin
-      pf_get_latest_csc_isa_manual(t.name)
+    pf_get_latest_csc_isa_manual(t.name)
 
-      # Figure out some pathnames/filenames.
-      norm_tags_json_suffix = "-norm-tags.json"
-      isa_manual_dir = "#{PROC_CTP_GEN_DIR}/adoc/ext/riscv-isa-manual"
-      unpriv_adoc = "#{isa_manual_dir}/src/riscv-unprivileged.adoc"
-      priv_adoc = "#{isa_manual_dir}/src/riscv-privileged.adoc"
-      unpriv_tags_json = "#{PROC_CTP_GEN_DIR}/adoc/riscv-unprivileged#{norm_tags_json_suffix}"
-      priv_tags_json = "#{PROC_CTP_GEN_DIR}/adoc/riscv-privileged#{norm_tags_json_suffix}"
-      norm_rules_json = "#{PROC_CTP_GEN_DIR}/adoc/norm-rules.json"
+    # Figure out some pathnames/filenames.
+    norm_tags_json_suffix = "-norm-tags.json"
+    isa_manual_dir = "#{PROC_CTP_GEN_DIR}/adoc/ext/riscv-isa-manual"
+    unpriv_adoc = "#{isa_manual_dir}/src/riscv-unprivileged.adoc"
+    priv_adoc = "#{isa_manual_dir}/src/riscv-privileged.adoc"
+    unpriv_tags_json = "#{PROC_CTP_GEN_DIR}/adoc/riscv-unprivileged#{norm_tags_json_suffix}"
+    priv_tags_json = "#{PROC_CTP_GEN_DIR}/adoc/riscv-privileged#{norm_tags_json_suffix}"
+    norm_rules_json = "#{PROC_CTP_GEN_DIR}/adoc/norm-rules.json"
 
-      # Extract normative rule tags from ISA manuals into JSON files.
-      pf_adoc2norm_tags(unpriv_adoc, unpriv_tags_json, isa_manual_dir, norm_tags_json_suffix)
-      pf_adoc2norm_tags(priv_adoc, priv_tags_json, isa_manual_dir, norm_tags_json_suffix)
+    # Extract normative rule tags from ISA manuals into JSON files.
+    pf_adoc2norm_tags(unpriv_adoc, unpriv_tags_json, isa_manual_dir, norm_tags_json_suffix)
+    pf_adoc2norm_tags(priv_adoc, priv_tags_json, isa_manual_dir, norm_tags_json_suffix)
 
-      # Create normative rules using ISA manual repository.
-      pf_build_norm_rules(isa_manual_dir, unpriv_tags_json, priv_tags_json, norm_rules_json)
+    # Create normative rules using ISA manual repository.
+    pf_build_norm_rules(isa_manual_dir, unpriv_tags_json, priv_tags_json, norm_rules_json)
 
-      # Read in normative rule JSON file to Ruby object.
-      data = JSON.parse(File.read(norm_rules_json))
+    # Read in normative rule JSON file to Ruby object.
+    data = JSON.parse(File.read(norm_rules_json))
 
-      # Load normative rules into a Ruby class to provide access to rules when generating CTP.
-      normative_rules = Udb::NormativeRules.new(data)
+    # Load normative rules into a Ruby class to provide access to rules when generating CTP.
+    normative_rules = Udb::NormativeRules.new(data)
 
-      proc_cert_create_adoc("#{PROC_CTP_DOC_DIR}/templates/proc_ctp.adoc.erb", t.name, model_name, normative_rules)
-    rescue => e
-      # Send to stdout since UDB sends tons of cr*p to stderr that floods one with useless information.
-      # Note that the $logger sends to stdout so anything send to $logger actually gets displayed as useful
-      # information if one just redirects stderr to /dev/null (e.g., in bash, run "./do <task-name> 2>/dev/null).
-      puts "Caught error: #{e.message}"
-
-      # Only print out 1st two lines of stack backtrace to stdout.
-      puts e.backtrace.take(2)
-
-      # Send full stacktrace to stderr with "warn".
-      warn e.backtrace
-    end
+    proc_cert_create_adoc("#{PROC_CTP_DOC_DIR}/templates/proc_ctp.adoc.erb", t.name, model_name, normative_rules)
   end
 
   file "#{PROC_CTP_GEN_DIR}/pdf/#{model_name}-CTP.pdf" => [
--- a/backends/profile/tasks.rake
+++ b/backends/profile/tasks.rake
@@ -39,58 +39,45 @@ Dir.glob("#{$resolver.std_path}/profile_release/*.yaml") do |f|
     "#{$root}/backends/portfolio/templates/beginning.adoc.erb",
     "#{PROFILE_DOC_DIR}/templates/profile.adoc.erb"
   ].concat(profile_pathnames) do |t|
-    begin
-      # Create architecture object without any knowledge of the profile release.
-      # Just used to get the PortfolioGroup object.
-      arch = pf_create_arch
+    # Create architecture object without any knowledge of the profile release.
+    # Just used to get the PortfolioGroup object.
+    arch = pf_create_arch
 
-      # Create ProfileRelease for specific profile release as specified in its arch YAML file.
-      # The Architecture object also creates all other portfolio-related object instances from their arch YAML files.
-      # None of these objects are provided with a AbstractConfig or Design object when created.
-      $logger.info "Creating ProfileRelease with only an Architecture object for #{release_name}"
-      profile_release_with_arch = arch.profile_release(release_name)
+    # Create ProfileRelease for specific profile release as specified in its arch YAML file.
+    # The Architecture object also creates all other portfolio-related object instances from their arch YAML files.
+    # None of these objects are provided with a AbstractConfig or Design object when created.
+    $logger.info "Creating ProfileRelease with only an Architecture object for #{release_name}"
+    profile_release_with_arch = arch.profile_release(release_name)
 
-      # Now create a ConfiguredArchitecture object for the PortfolioDesign.
-      cfg_arch = pf_create_cfg_arch(profile_release_with_arch.portfolio_grp)
+    # Now create a ConfiguredArchitecture object for the PortfolioDesign.
+    cfg_arch = pf_create_cfg_arch(profile_release_with_arch.portfolio_grp)
 
-      $logger.info "Creating ProfileRelease with a ConfiguredArchitecture object for #{release_name}"
-      profile_release_with_cfg_arch = cfg_arch.profile_release(release_name)
+    $logger.info "Creating ProfileRelease with a ConfiguredArchitecture object for #{release_name}"
+    profile_release_with_cfg_arch = cfg_arch.profile_release(release_name)
 
-      # Create the one PortfolioDesign object required for the ERB evaluation.
-      # Provide it with all the profiles in this ProfileRelease.
-      $logger.info "Creating PortfolioDesign object using profile release #{release_name}"
-      portfolio_design = Udb::PortfolioDesign.new(
-        release_name,
-        cfg_arch,
-        Udb::PortfolioDesign.profile_release_type,
-        profile_release_with_cfg_arch.profiles,
-        profile_release_with_cfg_arch.profile_family,
-        nil # no normative rules
-      )
+    # Create the one PortfolioDesign object required for the ERB evaluation.
+    # Provide it with all the profiles in this ProfileRelease.
+    $logger.info "Creating PortfolioDesign object using profile release #{release_name}"
+    portfolio_design = Udb::PortfolioDesign.new(
+      release_name,
+      cfg_arch,
+      Udb::PortfolioDesign.profile_release_type,
+      profile_release_with_cfg_arch.profiles,
+      profile_release_with_cfg_arch.profile_family,
+      nil # no normative rules
+    )
 
-      # Create empty binding and then specify explicitly which variables the ERB template can access.
-      # Seems to use this method name in stack backtraces (hence its name).
-      def evaluate_erb
-        binding
-      end
-      erb_binding = evaluate_erb
-      portfolio_design.init_erb_binding(erb_binding)
-      erb_binding.local_variable_set(:profile_release, profile_release_with_cfg_arch)
-      erb_binding.local_variable_set(:profile_family, profile_release_with_cfg_arch.profile_family)
-
-      pf_create_adoc("#{PROFILE_DOC_DIR}/templates/profile.adoc.erb", erb_binding, t.name, portfolio_design)
-    rescue => e
-      # Send to stdout since UDB sends tons of cr*p to stderr that floods one with useless information.
-      # Note that the $logger sends to stdout so anything send to $logger actually gets displayed as useful
-      # information if one just redirects stderr to /dev/null (e.g., in bash, run "./do <task-name> 2>/dev/null).
-      puts "Caught error: #{e.message}"
-
-      # Only print out 1st two lines of stack backtrace to stdout.
-      puts e.backtrace.take(2)
-
-      # Send full stacktrace to stderr with "warn".
-      warn e.backtrace
+    # Create empty binding and then specify explicitly which variables the ERB template can access.
+    # Seems to use this method name in stack backtraces (hence its name).
+    def evaluate_erb
+      binding
     end
+    erb_binding = evaluate_erb
+    portfolio_design.init_erb_binding(erb_binding)
+    erb_binding.local_variable_set(:profile_release, profile_release_with_cfg_arch)
+    erb_binding.local_variable_set(:profile_family, profile_release_with_cfg_arch.profile_family)
+
+    pf_create_adoc("#{PROFILE_DOC_DIR}/templates/profile.adoc.erb", erb_binding, t.name, portfolio_design)
   end
 
   file "#{PROFILE_GEN_DIR}/pdf/#{release_name}ProfileRelease.pdf" => [


james-ball-qualcomm:
Based on discussions with @dhower-qc, having the normative rules in the extension/instruction/CSR YAML files isn't acceptable. Instead Derek wants me to create top-level database objects for normative rules. So, I'll do that in this branch/PR and then use it for the normative rules for the I extension.

james-ball-qualcomm:
To avoid having thousands of tiny YAML files (i.e., one file for each normative rule), I'd like each YAML file to contain one or more related normative rules. So, here's an example:

$schema: "normative_rules_schema.json#"
kind: normative_rules
name: I
type: extension
rules:
  R-type_ops:
    description: R-type instruction source and destination operands
    doc_links: ["norm:basegrp:all:R-type_operands"]
    coverage_points: [cp_rs1, cp_rs2, cp_rd, cp_rs1_edges]
  x0eq0:
    description: All X0eq0
    doc_links: ["norm:basegrp:all:x0eq0"]
    coverage_points: [cp_rs1, cp_rs2, cp_rd]
  imm_always_sex:
    description: place holder
    doc_links: ["norm:basegrp:all:imm_always_sex"]
    coverage_points: [cp_imm_edges]
  pcreg:
    description: All PCREG
    doc_links: ["norm:basegrp:all:pcreg"]
    coverage_points: [cp_imm_edges]
  rv32i_rv64i_other_xregs:
    description: Registers x1 to x31
    doc_links: ["norm:bases:rv32i_rv64i:other-xregs"]

  #######################
  # RV32I (MXLEN == 32) #
  #######################

  rv32_xregwidth:
    description: RV32I XLEN
    doc_links: ["norm:base:rv32i:xregwidth"]
    # TODO: add "when" statement saying MXLEN == 32

```

**Comment by james-ball-qualcomm** [comment]
And here is the JSON schema file for this
[normative_rules_schema.json](https://github.com/user-attachments/files/21873599/normative_rules_schema.json)

**Comment by dhower-qc** [comment]
The invariant I'd like to maintain is that each rule has its own file. The example you give would be a fine way to drive generation of those files. Make sense?

**Comment by dhower-qc** [comment]
> extra complexity and effort I'd like to avoid creating, explaining, and maintaining

Isn't this a one-time cost? We haven't had issues with layout files we use.

**Comment by james-ball-qualcomm** [comment]
https://stackoverflow.com/questions/31839578/how-to-define-a-json-schema-that-requires-at-least-one-of-many-properties

File: spec/std/isa/ext/I.yaml [hunk 1]
```diff
@@ -16,3 +16,62 @@ versions:
...
+  - name: rv64_w_sex
+    # TODO: add "when" statement saying MXLEN == 64
+
+  #CTI and LS
+  - name: taken_cti_misaligned_exc
+    # TODO: add "when" statement referencing the taken CTI instruction group
+  - name: ldst_addr_misaligned_or_access_fault_exc
+    # TODO: add "when" statement referencing the load/store instruction group
+  - name: ldst_atomicity_for_aligned
+    # TODO: add "when" statement referencing the load/store instruction group
```
dhower-qc, line 45: I'm not following what you intend the "when" statement to do here. Is there ever an instance where extension I would be defined but the load/store instruction group wouldn't?

backends/portfolio/tasks.rake
```diff
@@ -152,7 +154,117 @@ def pf_adoc2html(adoc_file, target_pname)
...
-  sh cmd
+  sh(cmd)
 
   $logger.info "Generated HTML in #{target_pname}"
 end
+
+# @param adoc_file [String] Full pathname of source adoc file
+# @param target_pname [String] Full name of tags file being generated
+# @param isa_manual_dirname [String] Full pathname of ISA manual root directory
+def pf_adoc2norm_tags(adoc_file, target_pname, isa_manual_dirname)
```
ThinkOpenly: It would be helpful to have at least a one-liner description of what this function does.

backends/portfolio/tasks.rake
```diff
@@ -152,7 +154,117 @@ def pf_adoc2html(adoc_file, target_pname)
...
+  target_dirname = File.dirname(target_pname)
+
+  # Ensure target directory is present.
+  FileUtils.mkdir_p(target_dirname)
+
+  # The tags backend will put the tags file in the same directory as the input adoc file.
+  backend_tags_pname = adoc_file.sub(/\.adoc$/, "-norm-tags.json")
```
ThinkOpenly: since the string "-norm-tags.json" is used multiple times, it might help to use a variable, in case there is a need to change it later, so it can be changed once.

backends/portfolio/tasks.rake
```diff
@@ -152,7 +154,117 @@ def pf_adoc2html(adoc_file, target_pname)
...
+    adoc_file
+  ].join(" ")
+
+  $logger.info "bundle exec #{cmd}"
+
+  # Write out command used to extract tags to allow running this manually during development.
+  run_pname = "#{target_dirname}/adoc2norm_tags.sh"
+  sh "rm -f #{run_pname}"
+  sh "echo '#!/bin/bash' >#{run_pname}"
+  sh "echo >>#{run_pname}"
```
ThinkOpenly, line 200: nit: this line is not needed.

backends/portfolio/tasks.rake
```diff
@@ -152,7 +154,117 @@ def pf_adoc2html(adoc_file, target_pname)
...
+  FileUtils.mv(backend_tags_pname, target_pname)
+
+  $logger.info "Moved normative rule tags to #{target_pname}"
+end
+
+# @param isa_manual_dirname [String] Full pathname of ISA manual root directory
+# @param unpriv_tags_json [String] Full pathname of unpriv ISA manual JSON tags file
+# @param priv_tags_json [String] Full pathname of priv ISA manual JSON tags file
+# @param target_pname [String] Full pathname where normative rules should end up
+def pf_build_norm_rules(isa_manual_dir, unpriv_tags_json, priv_tags_json, target_pname)
```
ThinkOpenly, line 221: a one-liner description of the purpose of the function would be helpful.

backends/portfolio/tasks.rake
```diff
@@ -152,7 +154,117 @@ def pf_adoc2html(adoc_file, target_pname)
...
+        instances: [add]
+    tags:
+      - "norm:add_op"
+  - name: Xmock_nr2
+    description: |
+      Normative rule without any tags.
+      Should have lots of room to display this description in the CTP tables.
+TEXT
+
+  File.write("#{defs_dir}/mock.yaml", mock_nr_def_yaml)
```
ThinkOpenly, line 259: what's this?

backends/portfolio/templates/normative_rules.adoc.erb
```diff
@@ -1,22 +1,51 @@
-<% unless db_obj.cert_normative_rules.empty? -%>
+<% unless db_obj.normative_rule_reqs.empty? -%>
 <% nice_name = db_obj.is_a?(Udb::CsrField) ? "#{db_obj.parent.name}.#{db_obj.name}" : db_obj.name -%>
-<% if defined?(use_description_list) && use_description_list -%>
+<%   if defined?(use_description_list) && use_description_list -%>
```
ThinkOpenly, line 3: is this indented for a reason?

spec/std/isa/csr/mie.yaml
ThinkOpenly, line 1: Nice to have this additional text, but why is this in this PR? Are there normative rules hidden within?

ThinkOpenly:
I have not been following the implementation of normative rules. Is there a few sentence description of how the references here refer to a rule, presumably somewhere else? For example, the normative rule name "X0eq0" appears exactly once in this PR, so the rule itself must be elsewhere.

spec/std/isa/ext/I.yaml
james-ball-qualcomm: After this PR is merged, I'll be taking another pass at the I.yaml norm rules and most will replaced with references in each instruction so the when statement comments will go away.

backends/portfolio/tasks.rake
james-ball-qualcomm: Added

backends/portfolio/tasks.rake
james-ball-qualcomm: A way to improve test coverage by adding in a mock normative rule in UDB since it helps test the CTP PDF generated by UDB. Not needed in ISA manual repo.

backends/portfolio/templates/normative_rules.adoc.erb
james-ball-qualcomm: Yes and line 2 should be indented 2 spaces too so I've fixed that. They are both inside an "unless" statement in line 1.

spec/std/isa/csr/mie.yaml
james-ball-qualcomm: Fixing a problem I happened to notice. It's in this PR because I'm trying to balance my productivity against the effort to make a new issue, branch, and PR for every little thing I see wrong. Hopefully you understand this need.

backends/proc_crd/tasks.rake
```diff
@@ -31,7 +31,20 @@ Dir.glob("#{$resolver.std_path}/proc_cert_model/*.yaml") do |f|
...
+      # Note that the $logger sends to stdout so anything send to $logger actually gets displayed as useful
+      # information if one just redirects stderr to /dev/null (e.g., in bash, run "./do <task-name> 2>/dev/null).
+      puts "Caught error: #{e.message}"
+
+      # Only print out 1st two lines of stack backtrace to stdout.
+      puts e.backtrace.take(2)
+
+      # Send full stacktrace to stderr with "warn".
+      warn e.backtrace
+    end
```
dhower-qc: please remove this. you can use ruby's "--backtrace-limit" option to control how much you'd like to see

backends/proc_ctp/tasks.rake
```diff
@@ -25,15 +26,50 @@ Dir.glob("#{$resolver.std_path}/proc_cert_model/*.yaml") do |f|
...
+      # Note that the $logger sends to stdout so anything send to $logger actually gets displayed as useful
+      # information if one just redirects stderr to /dev/null (e.g., in bash, run "./do <task-name> 2>/dev/null).
+      puts "Caught error: #{e.message}"
+
+      # Only print out 1st two lines of stack backtrace to stdout.
+      puts e.backtrace.take(2)
+
+      # Send full stacktrace to stderr with "warn".
+      warn e.backtrace
+    end
```
dhower-qc: remove

backends/profile/tasks.rake
```diff
@@ -39,44 +39,58 @@ Dir.glob("#{$resolver.std_path}/profile_release/*.yaml") do |f|
...
+      # Send to stdout since UDB sends tons of cr*p to stderr that floods one with useless information.
+      # Note that the $logger sends to stdout so anything send to $logger actually gets displayed as useful
+      # information if one just redirects stderr to /dev/null (e.g., in bash, run "./do <task-name> 2>/dev/null).
+      puts "Caught error: #{e.message}"
+
+      # Only print out 1st two lines of stack backtrace to stdout.
+      puts e.backtrace.take(2)
+
+      # Send full stacktrace to stderr with "warn".
+      warn e.backtrace
```
dhower-qc: remove

backends/profile/tasks.rake
```diff
@@ -39,44 +39,58 @@ Dir.glob("#{$resolver.std_path}/profile_release/*.yaml") do |f|
...
+        nil # no normative rules
+      )
+
+      # Create empty binding and then specify explicitly which variables the ERB template can access.
+      # Seems to use this method name in stack backtraces (hence its name).
+      def evaluate_erb
+        binding
+      end
+      erb_binding = evaluate_erb
+      portfolio_design.init_erb_binding(erb_binding)
```
dhower-qc: Is this just to get a different name in the backtrace? If so, ERB has a #filename method that.

tools/ruby-gems/udb/lib/udb/obj/can_have_normative_rule_reqs.rb
dhower-qc, line 1: please move up a directory. `obj` is supposed to be the classes that directly represent a DB object.

tools/ruby-gems/udb/lib/udb/obj/database_obj.rb
```diff
@@ -49,7 +47,7 @@ class Kind < T::Enum
   attr_reader :long_name
 
   sig { returns(String) }
-  def kind = @kind.serialize
+  def kind = @kind.serialize  # Enum string (e.g., "profile") vs. full object reference (e.g., #<DatabaseObject::Kind::Profile>)
```
dhower-qc, line 50: Do we need this comment? Isn't it implied by the return type?

tools/ruby-gems/udb/lib/udb/obj/instruction.rb
```diff
@@ -522,7 +523,7 @@ def reachable_exceptions_str(effective_xlen=nil)
         else
           effective_xlen = cfg_arch.mxlen
           pruned_ast = pruned_operation_ast(effective_xlen)
-          puts " #{name}..."
+          # puts " #{name}..."
```
dhower-qc: just remove this line.

tools/ruby-gems/udb/lib/udb/resolver.rb
```diff
@@ -169,7 +169,7 @@ def any_newer?(target, deps)
     # run command in the shell. raise if exit is not zero
     sig { params(cmd: T::Array[String]).void }
     def run(cmd)
-      puts cmd.join(" ")
+      # puts cmd.join(" ")
```
dhower-qc: just remove

tools/ruby-gems/udb_helpers/lib/udb_helpers/backend_helpers.rb
```diff
@@ -160,19 +159,29 @@ def link_to_udb_doc_idl_func(func_name)
...
-  # @param org [String] Organization of normative rules and test procedures (sep=separate chapters, combo=combined chapters, appendix=appendix)
-  # @param id [String] ID of the normative rule
-  def link_to_udb_doc_cov_pt(org, id)
-    raise ArgumentError, "Unknown org value of '#{org}' for ID '#{id}'" unless org == "sep" || org == "combo" || org == "appendix"
-    "%%UDB_DOC_COV_PT_LINK%#{org};#{id.sanitize};#{id}%%"
+  # @return [String] A hyperlink to a UDB normative rule
+  # @param fmt [String] Is link to normative rule defined in its own separate chapter (fmt = "sep")
+  #                     or in a chapter that combines (fmt = "combo") normative rules, coverage points, and test procedures
+  # @param name [String] Unique name of the normative rule
+  def link_to_udb_doc_norm_rule(fmt, name)
```
dhower-qc, line 184: please add a Sorbet signature (all places in file). E.g.:

```ruby
sig { params(fmt: String, name: String).returns(String) }
def link_to_udb_doc_norm_rule(fmt, name)
```

After doing so, you can remove the type in the comment ("[String]") since the documentation generator uses Sorbet for that

backends/proc_crd/tasks.rake
james-ball-qualcomm: --backtrace-limit won't address the whole problem. There is way too much non-error information spewed out of UDB and AsciiDoctor to stderr when generating files making it very hard to find the nugget of information related to a failure. So, I'm keeping this.

backends/proc_ctp/tasks.rake
james-ball-qualcomm: Keeping

backends/profile/tasks.rake
james-ball-qualcomm: Keeping

backends/profile/tasks.rake
james-ball-qualcomm: No, this is to be explicit about what variables an ERB can use. Your previous solution wasn't explicit making it harder to understand and debug.

tools/ruby-gems/udb/lib/udb/obj/database_obj.rb
james-ball-qualcomm: Yes, we need this comment. Just saying String isn't descriptive enough.

tools/ruby-gems/udb/lib/udb/obj/instruction.rb
james-ball-qualcomm: It is your code so I just politely commented it out. I can remove it though. It was spewing tons out output to stdout.

tools/ruby-gems/udb/lib/udb/resolver.rb
james-ball-qualcomm: Removed

tools/ruby-gems/udb/lib/udb/obj/can_have_normative_rule_reqs.rb
james-ball-qualcomm: This code is common code is "included" by DB objects that are in this directory. So, it is part of the DB object code and so I'd prefer to keep it in the obj directory (seems like a better organization).

image

tools/ruby-gems/udb_helpers/lib/udb_helpers/backend_helpers.rb
james-ball-qualcomm: Hey, I did this by asking the GitHub AI "Add Sorbet signatures for all functions in this file" and it did a really good job.

spec/std/isa/csr/mie.yaml
ThinkOpenly: Honestly, not for something of this magnitude. It's a large change tucked inside of a 73-commit PR. It's confusing for this reviewer, at least, and reviewers' time is important, too. When scanning through 2,395 lines of changes in unfamiliar code and something apparently orthogonal appears, a reviewer should be asking "why is this here? should this be here? is this correct?" That in turn requires research and time.

backends/proc_crd/tasks.rake
dhower-qc: What is printed that causes the problems? I'd rather focus on that then cover it up like this.

Also, there are practical problems with the code. Since it doesn't restrict the type being rescued, you wind up with odd things like being unable to terminate the program since ctrl+c won't make it back to the interpreter.

backends/profile/tasks.rake
dhower-qc: How does renaming `binding` to `evaluate_erb` and then again to `erb_binding` help?

tools/ruby-gems/udb/lib/udb/obj/can_have_normative_rule_reqs.rb
dhower-qc: `Helpers::WavedromUtil`, right below, has the same property, but it is not defined in `obj`. I've been trying to move to a 1:1 match between files/classes in `obj` and database objects. I'd appreciate the help achieving that goal.

tools/ruby-gems/udb_helpers/lib/udb_helpers/backend_helpers.rb
dhower-qc: nifty

dhower-qc:
> > ```
> > * Please add sorbet signatures to all of the new Ruby files.
> > ```
> 
> 
> Will do. When do you plan to add Sorbet to all the existing Ruby code? Seems like that should be done for all code, not just new code. This is an example of where resources that you influence can be used to help clean up the code instead of adding new functionality.

The beauty of Sorbet is that it is [incremental](https://sorbet.org/docs/gradual). I think we've been good at making sure any new/touched Ruby code gets signatures.

backends/proc_crd/tasks.rake
james-ball-qualcomm: I'm totally open to a better solution as long as it actually solves the problems I'm having.
[MockCRD-stdout.log](https://github.com/user-attachments/files/22361802/MockCRD-stdout.log)
[MockCRD-stderr.log](https://github.com/user-attachments/files/22361804/MockCRD-stderr.log)

I've attached the stdout and stderr created when running "./do MockCRD 1>MockCRD-stdout.log 2>MockCRD-stderr.log". Normally I run "./do MockCRD 2>&1 | tee Mock.log" to capture stdout and stderr to a log file so they are mixed together in one file but I've attached stdout/stderr separately so you can see what is produced. I usually combine stdout/stderr together in one log file so they both have information needed to debug failures.

Now that I've commented out a few "puts" statements in UDB in this PR, it generates a nice summary to stdout. However, the problem is stderr. I can't ignore since that's where raise statements and stack traces are written. Look at all the lines UDB is adding with these "resolving arch" and "validating arch" messages. Are these really needed? However, Asciidoctor is the main culprit since the UDB stuff is early in program execution (not typically close to error messages at the end of the stderr). Asciidoctor even prints messages that say "error" but then apparently this isn't fatal.

backends/proc_ctp/tasks.rake
james-ball-qualcomm: Will replace with better solution if one available.

backends/profile/tasks.rake
james-ball-qualcomm: Will replace with better solution if one available.

backends/profile/tasks.rake
james-ball-qualcomm: I can't remember the details but I've been doing this for months in my code. Maybe you are just noticing it for the first time? It works well and I'd have to spend considerable time finding out why I wrote this but it was for good reasons (not arbitrary). So, can you please live with this?

tools/ruby-gems/udb/lib/udb/obj/can_have_normative_rule_reqs.rb
james-ball-qualcomm: Okay, will move (I tried).

backends/proc_ctp/tasks.rake
james-ball-qualcomm: I'll just remove this and in the other 2 files too. Need to work on more productive items before vacation ...

backends/profile/tasks.rake
james-ball-qualcomm: Removed

backends/proc_crd/tasks.rake
james-ball-qualcomm: Removed

dhower-qc: Since the vast majority of the messages come from asciidoctor-pdf, it seems to me what you want is to tell it to be quiet via its "--quiet" option.

For the progress messages, this diff will remove them when you redirect to a file (from my copy, lines numbers may be a bit off for you):

```
diff --git a/tools/ruby-gems/udb/python/yaml_resolver.py b/tools/ruby-gems/udb/python/yaml_resolver.py
index 120ac3525..f3a0925c9 100644
--- a/tools/ruby-gems/udb/python/yaml_resolver.py
+++ b/tools/ruby-gems/udb/python/yaml_resolver.py
@@ -633,7 +639,8 @@ if __name__ == "__main__":
             arch_paths,
             ascii=True,
             desc="Merging arch",
-            file=sys.stderr,
+            file=sys.stdout,
+            disable=None # disable for non-TTY
         ):
             merged_arch_path = (
                 os.path.join(args.merged_dir, arch_path)
@@ -663,7 +670,8 @@ if __name__ == "__main__":
                 arch_paths,
                 ascii=True,
                 desc="Resolving arch",
-                file=sys.stderr,
+                file=sys.stdout,
+                disable=None # disable for non-TTY
             )
         )
         abs_resolved_dir = (
@@ -684,7 +692,8 @@ if __name__ == "__main__":
                 arch_paths,
                 ascii=True,
                 desc="Validating arch",
-                file=sys.stderr,
+                file=sys.stdout,
+                disable=None # disable for non-TTY
             )
         )
         for arch_path in iter:
```

backends/profile/tasks.rake
james-ball-qualcomm: Trying to wrap this up before vacation. So the "def evaluate_erb" is obviously defining a function that returns a blank binding instance (because binding is called inside the evaluate_erb function) and then the function is called and the return value is put into a local variable called erb_binding and then passed to init_erb_binding. I use the local variable so I can put each of these statements on their own line (useful for putting in breakpoints in Ruby debugger). So, I think this is on pretty sound footing. Can you please approve it?

dhower-qc: I'd rather see it removed. Especially since the comment in the code doesn't match your description here, nor my intuition, and is likely to cause confusion.

```suggestion
      portfolio_design.init_erb_binding(binding)
```

PR #969, opened 2025-08-09
fix(setup): fix container detection and image reference issues

rpsene:
Updated bin/setup to use POSIX-compatible variable checks.

Corrected Docker image reference formatting to prevent invalid reference format errors.

### DIFF
--- a/bin/python
+++ b/bin/python
@@ -1,8 +1,21 @@
 #!/usr/bin/env bash
+set -euo pipefail
 
-ROOT=$(dirname $(realpath ${BASH_SOURCE[0]}))
-if [ ! -v NO_SOURCE ]; then
-  source $ROOT/setup
+ROOT="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd -P)"
+REPO_ROOT="$(cd -- "$ROOT/.." && pwd -P)"
+
+# Source setup unless NO_SOURCE is set
+if [ -z "${NO_SOURCE+x}" ]; then
+  # shellcheck disable=SC1091
+  source "$REPO_ROOT/bin/setup"
 fi
 
-$PYTHON "$@"
+if [ -n "${PYTHON+x}" ] && [ -n "$PYTHON" ]; then
+  # $PYTHON is a *command string* (e.g., `docker run ... python3`).
+  # Append our args safely and exec it.
+  printf -v _args ' %q' "$@"
+  eval "exec ${PYTHON}${_args}"
+else
+  # Fallback to local venv interpreter
+  exec "$REPO_ROOT/.home/.venv/bin/python3" "$@"
+fi
--- a/bin/setup
+++ b/bin/setup
@@ -1,10 +1,12 @@
 #!/usr/bin/env bash
 
-ROOT=$(dirname $(dirname $(realpath ${BASH_SOURCE[0]})))
+# Resolve ROOT like before, but robustly
+ROOT="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")/.." && pwd -P)"
 OLDWD=$PWD
-cd $ROOT
+cd "$ROOT"
 
-CONTAINER_TAG=`cat ${ROOT}/bin/.container-tag`
+# Trim newline(s) from the tag to avoid "invalid reference format"
+CONTAINER_TAG="$(tr -d '\n' < "${ROOT}/bin/.container-tag")"
 
 # Supported container types:
 #
@@ -25,7 +27,6 @@ CONTAINER_TAG=`cat ${ROOT}/bin/.container-tag`
 # ALL THREE CONTAINERS ARE (in theory) IDENTICAL. Which one you use is a preference of your setup and installed
 # software.
 
-
 # get the container type
 #
 # use, in priority order:
@@ -34,41 +35,41 @@ CONTAINER_TAG=`cat ${ROOT}/bin/.container-tag`
 #  3. Docker, if DOCKER is set in the environment
 #  4. Singularity, if SINGULARITY is set in the environment
 #  5. The choice made by the user, which will be cached in .container-type
-if [ -v DEVCONTAINER_ENV ]; then
-CONTAINER_TYPE=devcontainer
-elif [ -f ${ROOT}/.container-type ]; then
-CONTAINER_TYPE=`cat ${ROOT}/.container-type`
-elif [ -v DOCKER ]; then
-CONTAINER_TYPE=docker
-elif [ -v SINGULARITY ]; then
-CONTAINER_TYPE=singularity
+if [ -n "${DEVCONTAINER_ENV+x}" ]; then
+  CONTAINER_TYPE=devcontainer
+elif [ -f "${ROOT}/.container-type" ]; then
+  CONTAINER_TYPE="$(tr -d '\n' < "${ROOT}/.container-type")"
+elif [ -n "${DOCKER+x}" ]; then
+  CONTAINER_TYPE=docker
+elif [ -n "${SINGULARITY+x}" ]; then
+  CONTAINER_TYPE=singularity
 else
-  echo -e "UDB tools run in a container. Both Docker and Singularity/Apptainer are supported.\\n\\n1. Docker\\n2. Singularity\\n"
+  echo -e "UDB tools run in a container. Both Docker and Singularity/Apptainer are supported.\n\n1. Docker\n2. Singularity\n"
   while true; do
     echo "Which would you like to use? (1/2) "
     read ans
-    case $ans in
-        [1]* ) CONTAINER_TYPE=docker; break;;
-        [2]* ) CONTAINER_TYPE=singularity; break;;
-        * ) echo -e "\\nPlease answer 1 or 2.";;
+    case "$ans" in
+      [1]* ) CONTAINER_TYPE=docker; break;;
+      [2]* ) CONTAINER_TYPE=singularity; break;;
+      * ) echo -e "\nPlease answer 1 or 2.";;
     esac
   done
 fi
 
-if [ "${CONTAINER_TYPE}" != "docker" -a "${CONTAINER_TYPE}" != "singularity" -a "${CONTAINER_TYPE}" != "devcontainer" ]; then
+if [ "$CONTAINER_TYPE" != "docker" -a "$CONTAINER_TYPE" != "singularity" -a "$CONTAINER_TYPE" != "devcontainer" ]; then
   echo "BAD CONTAINER TYPE: ${CONTAINER_TYPE}"
 fi
 
-if [ ! -f ${ROOT}/.container-type ]; then
-  echo ${CONTAINER_TYPE} > ${ROOT}/.container-type
+if [ ! -f "${ROOT}/.container-type" ]; then
+  echo "${CONTAINER_TYPE}" > "${ROOT}/.container-type"
 fi
 
 echo "Using ${CONTAINER_TYPE} environment"
 
-if [ -v GITHUB_ACTIONS ]; then
+if [ -n "${GITHUB_ACTIONS+x}" ]; then
     echo "Running in a GitHub Action"
-    CONTAINER_PATH=${ROOT}/.singularity/image.sif
-    HOME_PATH=${GITHUB_WORKSPACE}
+    CONTAINER_PATH="${ROOT}/.singularity/image.sif"
+    HOME_PATH="${GITHUB_WORKSPACE}"
     HOME_OPT="--home ${ROOT}/.home"
     SINGULARITY_CACHE=--disable-cache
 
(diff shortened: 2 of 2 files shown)


bin/python
```diff
@@ -1,8 +1,21 @@
 #!/usr/bin/env bash
+set -euo pipefail
```
ThinkOpenly, line 2: Why is this needed?

bin/python
```diff
@@ -1,8 +1,21 @@
...
 
-ROOT=$(dirname $(realpath ${BASH_SOURCE[0]}))
-if [ ! -v NO_SOURCE ]; then
-  source $ROOT/setup
+ROOT="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd -P)"
+REPO_ROOT="$(cd -- "$ROOT/.." && pwd -P)"
```
ThinkOpenly, line 5: What's the motivation for the change?
Can these be combined? ROOT is not used anywhere else.

bin/python
```diff
@@ -1,8 +1,21 @@
...
+
+# Source setup unless NO_SOURCE is set
+if [ -z "${NO_SOURCE+x}" ]; then
```
ThinkOpenly, line 8: My shell-foo is weak. Why "+x"?

bin/python
```diff
@@ -1,8 +1,21 @@
...
+  # shellcheck disable=SC1091
```
ThinkOpenly, line 9: I'm not a fan of magic numbers. Would it be better (more durable) to say
```suggestion
  # shellcheck source=/dev/null
```
(reference: https://www.shellcheck.net/wiki/SC1091)

bin/python
```diff
@@ -1,8 +1,21 @@
...
+  source "$REPO_ROOT/bin/setup"
 fi
 
-$PYTHON "$@"
+if [ -n "${PYTHON+x}" ] && [ -n "$PYTHON" ]; then
```
ThinkOpenly, line 13: Similar to above, how do these two tests differ, and why are we using this double test here, and not above?

bin/python
```diff
@@ -1,8 +1,21 @@
...
+  # $PYTHON is a *command string* (e.g., `docker run ... python3`).
+  # Append our args safely and exec it.
+  printf -v _args ' %q' "$@"
+  eval "exec ${PYTHON}${_args}"
+else
+  # Fallback to local venv interpreter
+  exec "$REPO_ROOT/.home/.venv/bin/python3" "$@"
```
ThinkOpenly, line 20: I'm not sure why these two invocations are so different?

bin/setup
```diff
@@ -1,10 +1,12 @@
 #!/usr/bin/env bash
 
-ROOT=$(dirname $(dirname $(realpath ${BASH_SOURCE[0]})))
+# Resolve ROOT like before, but robustly
```
ThinkOpenly, line 3: This comment only seems to have relevant context if the deleted line is preserved. "like before" with the changes has no "before".

bin/setup
```diff
@@ -1,10 +1,12 @@
...
+ROOT="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")/.." && pwd -P)"
 OLDWD=$PWD
-cd $ROOT
+cd "$ROOT"
 
-CONTAINER_TAG=`cat ${ROOT}/bin/.container-tag`
+# Trim newline(s) from the tag to avoid "invalid reference format"
+CONTAINER_TAG="$(tr -d '\n' < "${ROOT}/bin/.container-tag")"
```
ThinkOpenly, line 9: Where does that error occur?
Is this an "always" error, or only with certain tags? If only certain tags contain newlines, will those tags work at all if the newlines are removed?

bin/setup
```diff
@@ -25,7 +27,6 @@ CONTAINER_TAG=`cat ${ROOT}/bin/.container-tag`
 # ALL THREE CONTAINERS ARE (in theory) IDENTICAL. Which one you use is a preference of your setup and installed
 # software.
 
-
```
ThinkOpenly, line 28: avoid unnecessary whitespace changes in commits, please

bin/setup
```diff
@@ -34,147 +35,148 @@ CONTAINER_TAG=`cat ${ROOT}/bin/.container-tag`
...
-elif [ -f ${ROOT}/.container-type ]; then
-CONTAINER_TYPE=`cat ${ROOT}/.container-type`
-elif [ -v DOCKER ]; then
-CONTAINER_TYPE=docker
-elif [ -v SINGULARITY ]; then
-CONTAINER_TYPE=singularity
+if [ -n "${DEVCONTAINER_ENV+x}" ]; then
+  CONTAINER_TYPE=devcontainer
+elif [ -f "${ROOT}/.container-type" ]; then
+  CONTAINER_TYPE="$(tr -d '\n' < "${ROOT}/.container-type")"
```
ThinkOpenly, line 41: Does this need the same comment as above, or does the case above not need a comment?
Similar questions here... is this always an issue, or what causes the issue to manifest?

bin/setup
```diff
@@ -34,147 +35,148 @@ CONTAINER_TAG=`cat ${ROOT}/bin/.container-tag`
...
   while true; do
     echo "Which would you like to use? (1/2) "
     read ans
-    case $ans in
-        [1]* ) CONTAINER_TYPE=docker; break;;
-        [2]* ) CONTAINER_TYPE=singularity; break;;
-        * ) echo -e "\\nPlease answer 1 or 2.";;
+    case "$ans" in
+      [1]* ) CONTAINER_TYPE=docker; break;;
+      [2]* ) CONTAINER_TYPE=singularity; break;;
```
ThinkOpenly, line 53: seems a little strange to allow an answer of "12", but that's what the unchanged code did. hrm

bin/setup
```diff
@@ -34,147 +35,148 @@ CONTAINER_TAG=`cat ${ROOT}/bin/.container-tag`
...
-CONTAINER_TYPE=devcontainer
-elif [ -f ${ROOT}/.container-type ]; then
-CONTAINER_TYPE=`cat ${ROOT}/.container-type`
-elif [ -v DOCKER ]; then
-CONTAINER_TYPE=docker
-elif [ -v SINGULARITY ]; then
-CONTAINER_TYPE=singularity
+if [ -n "${DEVCONTAINER_ENV+x}" ]; then
+  CONTAINER_TYPE=devcontainer
+elif [ -f "${ROOT}/.container-type" ]; then
```
ThinkOpenly, line 40: Why do we need `{}` around `ROOT`? I guess you're just carrying that from the unchanged code?

bin/setup
```diff
@@ -34,147 +35,148 @@ CONTAINER_TAG=`cat ${ROOT}/bin/.container-tag`
...
 
-if [ "${CONTAINER_TYPE}" != "docker" -a "${CONTAINER_TYPE}" != "singularity" -a "${CONTAINER_TYPE}" != "devcontainer" ]; then
+if [ "$CONTAINER_TYPE" != "docker" -a "$CONTAINER_TYPE" != "singularity" -a "$CONTAINER_TYPE" != "devcontainer" ]; then
   echo "BAD CONTAINER TYPE: ${CONTAINER_TYPE}"
 fi
 
-if [ ! -f ${ROOT}/.container-type ]; then
-  echo ${CONTAINER_TYPE} > ${ROOT}/.container-type
+if [ ! -f "${ROOT}/.container-type" ]; then
+  echo "${CONTAINER_TYPE}" > "${ROOT}/.container-type"
```
ThinkOpenly, line 64: You removed the braces above. I don't think we need them here, either. We should probably pick a consistent style.

ThinkOpenly (commented):
Thanks for the contribution, @rpsene ! I have a number of questions mostly showing my ignorance.

bin/python
rpsene: @ThinkOpenly:

-e, exit immediately if a command fails
-u, treat unset variables as an error and exit, for instance, without it, something like ${CONTAINER_TYPE} could be empty and cause the wrong container setup. With it, the script stops immediately.
-o, pipefail, makes a pipeline fail if any command fails, not just the last one

bin/python
rpsene: @ThinkOpenly, yes, it can be combined in somthing like REPO_ROOT="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")/.." && pwd -P)"

bin/python
rpsene: @ThinkOpenly, +x in ${NO_SOURCE+x} is a bash parameter expansion trick to safely way test whether a variable is set at all, it avoids errors from set -u when a variable might not be defined.

[ -z "${NO_SOURCE+x}" ] means:
If NO_SOURCE is unset, it becomes empty (true for -z), otherwise substitute 'x' (non-empty, false for -z).

bin/python
rpsene: @ThinkOpenly SC1091 is the ShellCheck rule for  'Not following: file not found'.

Using # shellcheck source=/dev/null tells ShellCheck explicitly 'yes, this file exists, don’t warn', we can keep as is or add a better documentation for sc1091.

bin/python
rpsene: @ThinkOpenly fair question, if you only care whether it’s set and non-empty, the first test isn’t strictly needed, because the second will fail anyway if it’s unset (with set -u on).

bin/setup
rpsene: @ThinkOpenly sure!

bin/setup
rpsene: @ThinkOpenly we can add a better comment like `# Using ${VAR+x} to check if it's set, avoids 'unbound variable' errors with 'set -u'`

bin/setup
rpsene: @ThinkOpenly we can adjust it with something like:

`case $ans in
    1) CONTAINER_TYPE=docker; break;;
    2) CONTAINER_TYPE=singularity; break;;
    *) echo -e "\nPlease answer 1 or 2.";;
esac`

bin/setup
rpsene: @ThinkOpenly this is on picking one style (with or without braces) and use it consistently.  The ${} around ROOT here are just a style choice and aren’t strictly needed. ${ROOT} and $ROOT behave the same unless we are appending characters right after the variable name.

bin/setup
rpsene: @ThinkOpenly  this is on picking one style (with or without braces) and use it consistently like the previous case.

rpsene:
@ThinkOpenly btw, my debug started when I tried to setup UDB in fresh environment and got issues like:

`./do --tasks
/Users/rpsene/Documents/Source/riscv-unified-db/bin/setup: line 37: [: -v: unary operator expected
/Users/rpsene/Documents/Source/riscv-unified-db/bin/setup: line 41: [: -v: unary operator expected
/Users/rpsene/Documents/Source/riscv-unified-db/bin/setup: line 43: [: -v: unary operator expected
UDB tools run in a container. Both Docker and Singularity/Apptainer are supported.

1. Docker
2. Singularity

Which would you like to use? (1/2) `

and 

`./do --tasks
Using docker environment
docker: invalid reference format

Run 'docker run --help' for more information`

and

`./do --tasks
Using docker environment
TAG="0.9"
IMAGE="riscvintl/udb:0.9"
docker: invalid reference format

Run 'docker run --help' for more information`

and 

`git commit -s -m "Fix container detection and image reference issues in setup script"
/Users/rpsene/Documents/Source/riscv-unified-db/bin/python: line 4: [: -v: unary operator expected
/Users/rpsene/Documents/Source/riscv-unified-db/bin/python: line 8: -mpre_commit: command not found`

bin/python
ThinkOpenly: > -e, exit immediately if a command fails
> -u, treat unset variables as an error and exit, for instance, without it, something like ${CONTAINER_TYPE} could be empty and cause the wrong container setup. With it, the script stops immediately.
> -o, pipefail, makes a pipeline fail if any command fails, not just the last one

I see. We aren't using pipes here. Is this just boilerplate / best practices?

bin/python
ThinkOpenly: Let's do that, if you don't mind.

bin/python
ThinkOpenly: > Using # shellcheck source=/dev/null tells ShellCheck explicitly 'yes, this file exists, don’t warn'

Isn't that the desired behavior? My understanding is that ShellCheck is an offline check, and won't be used in a "production" path with an actual script invocation. (Do we even use ShellCheck as part of CI here? Maybe we should?)

bin/python
ThinkOpenly: OK. Let's drop the first check, then.

bin/setup
ThinkOpenly: In this case, I was referring to the `tr` command use.

bin/setup
ThinkOpenly: My vote would be either:
- no innocuous changes, or
- make all the innocuous changes needed for consistency

Having some, but not all, makes it challenging to review. "Why is this changed, but that isn't?"

ThinkOpenly:
> anything else to add on this one?

I was hoping for another pass of changes. If @rpsene is unable, I can adopt this one, or just do a follow-up (or just let it go). :-)

rpsene:
@ThinkOpenly  and @dhower-qc I will work on the updates tomorrow. Sorry about the delay.

PR #970, opened 2025-08-11
feat: add "udb create instruction/extension" commands to CLI

dhower-qc:


tools/ruby-gems/udb/lib/udb/cli/create.rb
```diff
@@ -0,0 +1,689 @@
...
+    end
+
+    sig { params(prompt: TTY::Prompt).returns(String) }
+    def self.get_copyright(prompt)
+      prompt.say \
+        "First, let's get the legal necessities out of the way."
+
+      copyright = prompt.ask \
+      "What copyright do you want to assign to newly created files?",
+      default: "Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.",
```
ThinkOpenly, line 43: Hmm. I'm trying to think of a better approach. There is not an obvious one. "RISC-V International"? I worry about perception, and the reality of casually assigning ownership. And does the assignee need to accept the assignment? I also wonder if we need more explanation, like "What entity/entities is/are creating this content and will own the copyright?" But I also worry about "I am not a lawyer. This is not legal advice." Complicated.

tools/ruby-gems/udb/lib/udb/cli/create.rb
```diff
@@ -0,0 +1,689 @@
...
+      required: true,
+      modify: [:trim]
+
+      license_info = <<~INFO
+        Great. The copyright will read:
+           #{copyright}
+
+        The license defaults to BSD-3-Clear.
```
ThinkOpenly, line 51: I wonder if this should be a link the license text.

Also, should we allow other licenses anyway, or at least describe how to utilize another license: what would need to change, etc.?

And, did you mean "BSD-3-Clause"? https://opensource.org/license/bsd-3-clause

tools/ruby-gems/udb/lib/udb/cli/create.rb
```diff
@@ -0,0 +1,689 @@
...
+      value = prompt.ask \
+        "What is the value of the opcode, as a binary number (e.g., 0111011)?",
+        validate: (proc do |v|
+          v =~ /^[01]+$/ && \
+            v.to_i(2).bit_length <= size
+        end)
+
+      display = prompt.ask \
+        "How should the opcode be displayed in documentation?",
+        default: name
```
ThinkOpenly, line 139: Shall we describe how this differs from the "name of the opcode" entered above?

tools/ruby-gems/udb/lib/udb/cli/create.rb
```diff
@@ -0,0 +1,689 @@
...
+      erb.filename = template_path.to_s
+
+      CreatedFile.new(path: dest_path, contents: erb.result(binding), next_steps: [])
+    end
+
+
+    sig { params(prompt: TTY::Prompt, copyright: String, outdir: Pathname).returns(CreatedFile) }
+    def self.create_inst_type(prompt, copyright, outdir)
+      name = prompt.ask \
+        "What is the instruction type name? It should be a single letter (e.g., 'R')",
```
ThinkOpenly, line 154: What about "R4", "CR", "CSS", "CIW", ...?

tools/ruby-gems/udb/lib/udb/cli/create.rb
```diff
@@ -0,0 +1,689 @@
...
+      erb = ERB.new(template_path.read, trim_mode: "-")
+      erb.filename = template_path.to_s
+
+      CreatedFile.new(path: dest_path, contents: erb.result(binding), next_steps: [])
+    end
+
+    sig { params(prompt: TTY::Prompt, copyright: String, outdir: Pathname, inst_type: T.nilable(String)).returns(CreatedFile) }
+    def self.create_inst_var(prompt, copyright, outdir, inst_type)
+      name = prompt.ask \
+        "What is the instruction var name (e.g., 'xs1')?",
```
ThinkOpenly, line 258: We may need to explain exactly what this is... an operand, or a field name?

Do we need to explain some conventions here? Maybe a list of examples?

tools/ruby-gems/udb/lib/udb/cli/create.rb
```diff
@@ -0,0 +1,689 @@
...
+      inst_types.each do |itype|
+        itype[:contents]["variables"].each_key do |itype_var_name|
+          opcode_fields["#{itype[:name]}:#{itype_var_name}"] =
+            { type: itype, var: itype_var_name }
+        end
+      end
+
+      opcode_field = prompt.select \
+        "What opcode field does this variable correspond with?",
+        opcode_fields.keys
```
ThinkOpenly, line 278: so this is applying a variable name to a defined bit field in the opcode subtype?

tools/ruby-gems/udb/lib/udb/cli/create.rb
```diff
@@ -0,0 +1,689 @@
...
+
+      name = T.let(nil, T.nilable(String))
+      loop do
+        name = prompt.ask \
+          "What is the subtype name? It should begin with the type name and then a dash (e.g., #{type_data["name"]}-new): ",
+          required: true,
+          modify: [:trim],
+          validate: /^#{type_data["name"]}-.*$/
+        if File.exist?(outdir / "inst_subtype" / "#{type_data["name"]}" / "#{name}.yaml")
+          prompt.say "Instruction subtype '#{name}' already exists. Please select another name."
```
ThinkOpenly, line 333: "Do you want to use the existing subtype, or pick another name?" ?

tools/ruby-gems/udb/lib/udb/cli/create.rb
```diff
@@ -0,0 +1,689 @@
...
+        var_type = prompt.select \
+          "What kind of variable is the '#{var_name}' slot for subtype #{name}?",
+          inst_vars,
+          filter: true
+
+        if var_type == "Create new"
+          prompt.say "Ok. Let's get some information on the new instruction variable type"
+          files << CreationActions.create_inst_var(prompt, copyright, outdir, type_data["name"])
+          var_type = T.must(files.last).path.basename(".yaml")
+          prompt.say "That's all I need for the instruction variable type. Back to the subtype variable.\n\n"
```
ThinkOpenly, line 356: There are no more prompts here. Do we need this?

tools/ruby-gems/udb/lib/udb/cli/create.rb
```diff
@@ -0,0 +1,689 @@
...
+          subtype_name = subtype_contents["name"]
+          subtypes["#{subtype_name.ljust(10)} (#{subtype_contents["long_name"]})"] = subtype_name
+        end
+
+        type_subtype_info = <<~INFO
+          Next, you will identify the instruction _subtype_, which points to an instruction _type_.
+
+          An instruction _type_ represents the general structure of an instruction encoding:
+            - where fixed _opcode_ fields occur (e.g., OP-32 in bits 6:0)
+            - where variable _operand_ fields occur (e.g., rd in bits 11:7).
```
ThinkOpenly, line 559:
```suggestion
            - where variable fields based on _operands_ occur (e.g., rd in bits 11:7).
```

tools/ruby-gems/udb/lib/udb/cli/create.rb
```diff
@@ -0,0 +1,689 @@
...
+          Everything needed for variables is provided by the subtype.
+
+          For opcodes, you will provide a _display name_ for documentation (such as encoding diagrams)
+          and a specific _value_ for the instruction.
+
+          We distinguish between two classes of opcodes:
+            - Those that are shared by multiple instructions (e.g., OP, OP-32 in encoding[6:0])
+            - Those that are specific to an instruction (e.g., ADDI in encoding[14:12])
+
+          Common opcodes become their own object in the database and referenced by instructions.
```
ThinkOpenly:
```suggestion
          Shared opcodes become their own object in the database and referenced by instructions.
```
...to match the preceding language.

tools/ruby-gems/udb/lib/udb/cli/create.rb
```diff
@@ -0,0 +1,689 @@
...
+          The specific opcodes are defined directly in the instruction, and do not elevate to a
```
ThinkOpenly:
```suggestion
          The instruction-specific opcodes are defined directly in the instruction, and do not elevate to a
```

ThinkOpenly (changes requested):
There's a lot of extra stuff here... is that due to basing on an unmerged branch?

tools/ruby-gems/udb/lib/udb/cli/create.rb
dhower-qc: I think I can use REUSE to collect the copyrights that exist in the project, and present them alphabetically along with a "Something else" option.  I like the suggestion of "What entity/entities is/are creating this content and will own the copyright?".

tools/ruby-gems/udb/lib/udb/cli/create.rb
dhower-qc: I can add a link. And it should actually be [BSD-3-Clause-Clear](https://spdx.org/licenses/BSD-3-Clause-Clear.html); it's a small tweak on BSD-3-Clause that Qualcomm wanted me to use.

tools/ruby-gems/udb/lib/udb/cli/create.rb
dhower-qc: Good idea.

tools/ruby-gems/udb/lib/udb/cli/create.rb
dhower-qc: Indeed ;)

tools/ruby-gems/udb/lib/udb/cli/create.rb
dhower-qc: I'm thinking we should switch to the following (expanding on the discussion in email):

- Field - contiguous set of bits in an encoding (e.g., rs1, imm12_hi, imm12_lo)
- Opcode - A field that has a fixed value in an instruction (e.g., funct3)
- Operand - A potentially non-contiguous set of bits in the encoding (e.g., xs1, imm12), along with an ordering (e.g., 31|25-20|...), transformations (e.g., left_shift), types (e.g., X source reg), and relocations (e.g., REL_PC_HI)

Instruction formats are non-overlapping collections of one or more opcodes and zero or more operands.

Instruction types are non-overlapping collections of fields.

Instruction formats point to an instruction type, but don't go on to use anything from it (it's a suggestion, not a rule). We can talk about this later, but my finding is that we will wind up with a lot of types if we make it a rule.

Obviously, this change (or whatever we settle on) needs to happen before this PR.

tools/ruby-gems/udb/lib/udb/cli/create.rb
dhower-qc: deferring this until we settle on the format/type discussion

PR #984, opened 2025-08-13
fix: update IDL highlights to reflect language changes

dhower-qc:
Fixes problem where backticks showed up in red in PDFs:

image

now looks like:

image


dhower-qc:
> Seem Copilot did some fix that I am not sure it is correct. Can you double-check that?!

That's neat. It caught a legit bug. Fix coming.

ayosher (approved):

tools/ruby-gems/idl_highlighter/lib/idl_highlighter.rb
```diff
@@ -46,9 +46,9 @@ def self.keywords_type
         rule %r/0x[0-9a-f]+[lu]*/i, Num::Hex
         rule %r/0[0-7]+[lu]*/i, Num::Oct
         rule %r{\d+}, Num::Integer
-        rule %r{(?:true|false|\$encoding|\$pc|\$signed|\$bits)}, Name::Builtin
+        rule %r{(?:true|false|\$encoding|\$pc|\$signed|\$bits|\$width|\$enum_size|\$enum_element_size|\$enum_to_a|\$enum|\$array_size)}, Name::Builtin
         rule %r{[.,;:\[\]\(\)\}\{]}, Punctuation
-        rule %r([~!%^&*+=\|?:<>/-]), Operator
+        rule %r([~!%^&*+=\|?:<>\/\-`]), Operator
```
ThinkOpenly: I see you chose to add backslash escapes rather than reordering:
```suggestion
        rule %r([~!%^&*+=\|?:<>/`-]), Operator
```
I'm curious if you thought the backslash approach was more durable, perhaps?

dhower-qc: Good question. I was testing it with [rubular.com](https://rubular.com/), which assumes you use the `/regex/` syntax that would need forward slash escaped. That doesn't apply here, though.

PR #999, opened 2025-08-19
fix: use VS field when updating mstatus.VS

jordancarlin:
The mstatus.VS field was incorrectly written with the value of the FS field from a CSR write. Update it to use the VS field as expected.

### DIFF
--- a/spec/std/isa/csr/mstatus.yaml
+++ b/spec/std/isa/csr/mstatus.yaml
@@ -476,13 +476,13 @@ fields:
       }
     sw_write(csr_value): |
       if (CSR[misa].V == 1'b1){
-        return ary_includes?<$array_size(MSTATUS_VS_LEGAL_VALUES), 2>(MSTATUS_VS_LEGAL_VALUES, csr_value.FS) ? csr_value.FS : UNDEFINED_LEGAL_DETERMINISTIC;
+        return ary_includes?<$array_size(MSTATUS_VS_LEGAL_VALUES), 2>(MSTATUS_VS_LEGAL_VALUES, csr_value.VS) ? csr_value.VS : UNDEFINED_LEGAL_DETERMINISTIC;
       } else if ((CSR[misa].S == 1'b0) && (CSR[misa].V == 1'b0)) {
         # must be read-only-0
         return 0;
       } else {
         # there will be no hardware update in this case because we know the V extension isn't implemented
-        return ary_includes?<$array_size(MSTATUS_VS_LEGAL_VALUES), 2>(MSTATUS_VS_LEGAL_VALUES, csr_value.FS) ? csr_value.FS : UNDEFINED_LEGAL_DETERMINISTIC;
+        return ary_includes?<$array_size(MSTATUS_VS_LEGAL_VALUES), 2>(MSTATUS_VS_LEGAL_VALUES, csr_value.VS) ? csr_value.VS : UNDEFINED_LEGAL_DETERMINISTIC;
       }
   SPP:
     location: 8


spec/std/isa/csr/mstatus.yaml
```diff
@@ -476,13 +476,13 @@ fields:
...
-        return ary_includes?<$array_size(MSTATUS_VS_LEGAL_VALUES), 2>(MSTATUS_VS_LEGAL_VALUES, csr_value.FS) ? csr_value.FS : UNDEFINED_LEGAL_DETERMINISTIC;
+        return ary_includes?<$array_size(MSTATUS_VS_LEGAL_VALUES), 2>(MSTATUS_VS_LEGAL_VALUES, csr_value.VS) ? csr_value.VS : UNDEFINED_LEGAL_DETERMINISTIC;
       } else if ((CSR[misa].S == 1'b0) && (CSR[misa].V == 1'b0)) {
         # must be read-only-0
         return 0;
       } else {
         # there will be no hardware update in this case because we know the V extension isn't implemented
-        return ary_includes?<$array_size(MSTATUS_VS_LEGAL_VALUES), 2>(MSTATUS_VS_LEGAL_VALUES, csr_value.FS) ? csr_value.FS : UNDEFINED_LEGAL_DETERMINISTIC;
+        return ary_includes?<$array_size(MSTATUS_VS_LEGAL_VALUES), 2>(MSTATUS_VS_LEGAL_VALUES, csr_value.VS) ? csr_value.VS : UNDEFINED_LEGAL_DETERMINISTIC;
       }
```
ThinkOpenly, line 486: Unrelated to the issue/PR, but isn't this code equivalent to:
```
      if (CSR[misa].V == 1'b1 || CSR[misa].S == 1'b1) {
        return ary_includes?<$array_size(MSTATUS_VS_LEGAL_VALUES), 2>(MSTATUS_VS_LEGAL_VALUES, csr_value.VS) ? csr_value.VS : UNDEFINED_LEGAL_DETERMINISTIC;
      } else {
        # must be read-only-0
        return 0;
      }
```
(Whether the current code is correct or not is another question.)

ThinkOpenly (approved):
unrelated question inline.

I am curious how you found this issue? Nice catch, regardless!

spec/std/isa/csr/mstatus.yaml
jordancarlin: Agreed that it seems equivalent in its current form.

The current `else` case should technically only be writing to the `mstatus.VS` field if `MSTATUS_VS_WRITABLE` is true. I'm not sure if that is already handled by the type being read-only in that case (see below), but if not, then the `sw_write` section needs to be updated.

```
type(): |
      if (CSR[misa].V == 1'b1){
        return CsrFieldType::RWH;
      } else if ((CSR[misa].S == 1'b0) && (CSR[misa].V == 1'b0)) {
        # must be read-only-0
        return CsrFieldType::RO;
      } else {
        # there will be no hardware update in this case because we know the V extension isn't implemented
        return MSTATUS_VS_WRITABLE ? CsrFieldType::RW : CsrFieldType::RO;
      }
```

ThinkOpenly: > The current `else` case should technically only be writing to the `mstatus.VS` field if `MSTATUS_VS_WRITABLE` is true. I'm not sure if that is already handled by the type being read-only in that case

Yes, a type of read-only will prevent sw_write from being invoked.

jordancarlin: Great. I'll open another PR with that cleanup once this is merged. Would rather not mix them since this one is a behavior change and the other is just a refactor.

PR #944, opened 2025-08-04
docs(ssp): shadow stack pointer - issue #560

sudo-apt-abdullah:

PR #1009, opened 2025-08-21
feat: Added operation fields for Zicfiss instructions.

### LINKED ISSUE #995: Implemention of Shadow stack instructions
Sub-issue of #560

sudo-apt-abdullah:
Closes  #995 which depends on #944.

### DIFF
--- a/spec/std/isa/inst/Zicfiss/c.sspopchk.x5.yaml
+++ b/spec/std/isa/inst/Zicfiss/c.sspopchk.x5.yaml
@@ -38,8 +38,8 @@ operation(): |
   if ((CSR[menvcfg].SSE == 1'b1) || (CSR[henvcfg].SSE == 1'b1) || (CSR[senvcfg].SSE == 1'b1)) {
     Bits<XLEN> ssp = csr_sw_read(csr_handle);
     XReg temp = read_memory<XLEN>(ssp, $encoding);
-    if (temp != X[1]) {
+    if (temp != X[5]) {
       raise(ExceptionCode::SoftwareCheck, mode(), $encoding);
-    } else {
+    }
     csr_sw_write(csr_handle, (ssp + (XLEN/8)));
   }
--- a/spec/std/isa/inst/Zicfiss/sspopchk.x1.yaml
+++ b/spec/std/isa/inst/Zicfiss/sspopchk.x1.yaml
@@ -35,11 +35,11 @@ operation(): |
     raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
   }
 
-  if ((CSR[menvcfg].SSE == 1'b1) || (CSR[henvcfg].SSE == 1'b1) || (CSR[senvcfg].SSE == 1'b1)) {
+  if ((CSR[menvcfg].SSE == 1) || (CSR[henvcfg].SSE == 1) || (CSR[senvcfg].SSE == 1)) {
     Bits<XLEN> ssp = csr_sw_read(csr_handle);
     XReg temp = read_memory<XLEN>(ssp, $encoding);
     if (temp != X[1]) {
       raise(ExceptionCode::SoftwareCheck, mode(), $encoding);
-    } else {
+    }
     csr_sw_write(csr_handle, (ssp + (XLEN/8)));
   }
--- a/spec/std/isa/inst/Zicfiss/sspopchk.x5.yaml
+++ b/spec/std/isa/inst/Zicfiss/sspopchk.x5.yaml
@@ -35,11 +35,11 @@ operation(): |
     raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
   }
 
-  if ((CSR[menvcfg].SSE == 1'b1) || (CSR[henvcfg].SSE == 1'b1) || (CSR[senvcfg].SSE == 1'b1)) {
+  if ((CSR[menvcfg].SSE == 1) || (CSR[henvcfg].SSE == 1) || (CSR[senvcfg].SSE == 1)) {
     Bits<XLEN> ssp = csr_sw_read(csr_handle);
     XReg temp = read_memory<XLEN>(ssp, $encoding);
     if (temp != X[5]) {
       raise(ExceptionCode::SoftwareCheck, mode(), $encoding);
-    } else {
+    }
     csr_sw_write(csr_handle, (ssp + (XLEN/8)));
   }


spec/std/isa/inst/Zicfiss/ssamoswap.d.yaml
```diff
@@ -6,9 +6,11 @@
 $schema: inst_schema.json#
 kind: instruction
 name: ssamoswap.d
-long_name: No synopsis available
+long_name: Atomic Swap from a Shadow Stack Location
 description: |
-  No description available.
+  The `ssamoswap.w` atomically loads a 64-bit data value from address of a shadow
+  stack location in rs1, puts the loaded value into register rd, and stores the
+  64-bit value held in rs2 to the original address in rs1.
```
ThinkOpenly:
```suggestion
  stack location in xs1, puts the loaded value into register xd, and stores the
  64-bit value held in xs2 to the original address in xs1.
```

spec/std/isa/inst/Zicfiss/ssamoswap.d.yaml
```diff
@@ -31,3 +33,21 @@ access:
   vu: always
 data_independent_timing: false
 operation(): |
+  Csr csr_handle = direct_csr_lookup(12'h011);
+
+  if (csr_handle.valid == false) {
+    unimplemented_csr($encoding);
```
ThinkOpenly: What's the purpose of this code?

spec/std/isa/inst/Zicfiss/ssamoswap.d.yaml
```diff
@@ -31,3 +33,21 @@ access:
...
+  } else if (!implemented?(ExtensionName::S)) {
+    raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
```
ThinkOpenly: I don't see this code matching the spec:
> use of Zicfiss in U-mode requires S-mode to be implemented.

So, you need to check the mode here.

spec/std/isa/inst/Zicfiss/ssamoswap.d.yaml
```diff
@@ -31,3 +33,21 @@ access:
...
+             (mode() == PrivilegeMode::U && CSR[senvcfg].SSE == 0)) {
+    raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+  } else if ((mode() == PrivilegeMode::VS && CSR[henvcfg].SSE == 0) ||
+             (mode() == PrivilegeMode::VU && (CSR[henvcfg].SSE == 0 || CSR[senvcfg].SSE == 0))) {
+    raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
+  }
+
+  if ((CSR[menvcfg].SSE == 1'b1) || (CSR[henvcfg].SSE == 1'b1) || (CSR[senvcfg].SSE == 1'b1)) {
+    X[xd] = read_memory<64>(X[xs1], $encoding);
+    write_memory<64>(X[xs1], X[xs2], $encoding);
```
ThinkOpenly: I think this needs to use an atomic swap operation. See `spec/std/isa/inst/Zaamo/amoswap.d.yaml` for an example.

spec/std/isa/inst/Zicfiss/ssamoswap.d.yaml
```diff
@@ -6,9 +6,11 @@
 $schema: inst_schema.json#
 kind: instruction
 name: ssamoswap.d
-long_name: No synopsis available
+long_name: Atomic Swap from a Shadow Stack Location
```
ThinkOpenly:
```suggestion
long_name: Atomic Swap Doubleword from a Shadow Stack Location
```

spec/std/isa/inst/Zicfiss/ssamoswap.w.yaml
```diff
@@ -6,9 +6,11 @@
 $schema: inst_schema.json#
 kind: instruction
 name: ssamoswap.w
-long_name: No synopsis available
+long_name: Atomic Swap from a Shadow Stack Location
```
ThinkOpenly:
```suggestion
long_name: Atomic Swap Word from a Shadow Stack Location
```

spec/std/isa/inst/Zicfiss/ssamoswap.w.yaml
```diff
@@ -6,9 +6,11 @@
...
 description: |
-  No description available.
+  The `ssamoswap.w` atomically loads a 32-bit data value from address of a shadow
+  stack location in rs1, puts the loaded value into register rd, and stores the
+  32-bit value held in rs2 to the original address in rs1.
```
ThinkOpenly:
```suggestion
  stack location in xs1, puts the loaded value into register xd, and stores the
  32-bit value held in xs2 to the original address in xs1.
```

spec/std/isa/inst/Zicfiss/ssrdp.yaml
```diff
@@ -8,7 +8,7 @@ kind: instruction
 name: ssrdp
 long_name: Read ssp into a Register
 description: |
-  No description available.
+  The SSRDP instruction is provided to move the contents of ssp to a destination register.
```
ThinkOpenly:
```suggestion
  Move the contents of ssp to a destination register.
```

spec/std/isa/inst/Zicfiss/ssrdp.yaml
```diff
@@ -24,3 +24,20 @@ access:
   vu: always
 data_independent_timing: false
 operation(): |
+  Csr csr_handle = direct_csr_lookup(12'h011);
```
ThinkOpenly: We could probably use a better way of looking up arbitrary CSRs, by name perhaps... ignoring that, could we maybe add a bit of context so it reads better?
```suggestion
  Bits<12> ssp_addr = 0x011;
  Csr csr_handle = direct_csr_lookup(ssp_addr);
```

spec/std/isa/inst/Zicfiss/sspush.x1.yaml
```diff
@@ -6,9 +6,11 @@
 $schema: inst_schema.json#
 kind: instruction
 name: sspush.x1
-long_name: No synopsis available
+long_name: Push to the Shadow Stack
```
ThinkOpenly:
```suggestion
long_name: Push to the Shadow Stack from register x1
```

spec/std/isa/inst/Zicfiss/sspush.x5.yaml
```diff
@@ -6,9 +6,11 @@
 $schema: inst_schema.json#
 kind: instruction
 name: sspush.x5
-long_name: No synopsis available
+long_name: Push to the Shadow Stack
```
ThinkOpenly:
```suggestion
long_name: Push to the Shadow Stack from register x5
```

spec/std/isa/inst/Zicfiss/sspush.x1.yaml
```diff
@@ -6,9 +6,11 @@
...
 description: |
-  No description available.
+  A shadow stack push operation is defined as decrement of the ssp by XLEN/8
+  followed by a store of the value in the link register to memory at the new
```
ThinkOpenly:
```suggestion
  followed by a store of the value in the link register x1 to memory at the new
```

spec/std/isa/inst/Zicfiss/sspush.x5.yaml
```diff
@@ -6,9 +6,11 @@
...
 description: |
-  No description available.
+  A shadow stack push operation is defined as decrement of the ssp by XLEN/8
+  followed by a store of the value in the link register to memory at the new
```
ThinkOpenly:
```suggestion
  followed by a store of the value in the link register x5 to memory at the new
```

spec/std/isa/inst/Zicfiss/ssamoswap.d.yaml
sudo-apt-abdullah: Spec says:
```
if S-mode not implemented
        raise illegal-instruction exception
```
```
if (!implemented?(ExtensionName::S)) {
```
Isn't checking the same thing. Let me know if I'm missing something!

spec/std/isa/inst/Zicfiss/ssamoswap.d.yaml
sudo-apt-abdullah: A quick check that ssp is indeed implemented. It might be redundant, and can be removed. Let me know if it can be removed.

spec/std/isa/inst/Zicfiss/ssamoswap.d.yaml
ThinkOpenly: Can you point to where in the spec you found that? I see (33.2.7 Atomic Swap from a Shadow Stack Location):
```
if privilege_mode != M && menvcfg.SSE == 0
    raise illegal-instruction exception
else if S-mode not implemented
    raise illegal-instruction exception
[...]
```

spec/std/isa/inst/Zicfiss/ssamoswap.d.yaml
ThinkOpenly: Since this instruction doesn't access `ssp`, you don't need to check it here.

spec/std/isa/inst/Zicfiss/ssamoswap.d.yaml
sudo-apt-abdullah: > I think this needs to use an atomic swap operation. See `spec/std/isa/inst/Zaamo/amoswap.d.yaml` for an example.

Thanks, I've now used amo operation for both ssamoswap.w/d and removed the condition checking for csr implementation.

spec/std/isa/inst/Zicfiss/ssamoswap.d.yaml
sudo-apt-abdullah: > Can you point to where in the spec you found that? I see (33.2.7 Atomic Swap from a Shadow Stack Location):

Yes, we're on the same page.

spec/std/isa/inst/Zicfiss/ssamoswap.d.yaml
sudo-apt-abdullah: Removed the condition.

spec/std/isa/inst/Zicfiss/ssamoswap.d.yaml
ThinkOpenly: ah, I see now. both instances results in illegal-instruction, so the order doesn't matter.

spec/std/isa/inst/Zicfiss/ssamoswap.d.yaml
ThinkOpenly, line 1: could you add:
```
base: 64
```
to this file?

sudo-apt-abdullah: Sure, I've added this in ssamoswap.d.yaml file. One more question, am I supposed to add new files for C.SSPOPCHK and C.SSPUSH under the Zicfiss extension?

ThinkOpenly: These are present as `sspopchk.x1.yaml` and similar, although they is not really correct. I have PR #872 that needs attention which fixes these, but I haven't finished getting it to pass CI yet. (And haven't had a chance to do so recently.) You can proceed as you already have.

sudo-apt-abdullah: In that PR, I can see that you merged sspopchk.x1.yaml & sspopchk.x5.yaml into single file sspopchk.yaml and sspush.x1.yaml & sspush.x5.yaml into file sspush.yaml, but can't see 16 bit version of sspush and sspopchk. Do we need to create separate files for that or I'm missing something? 

> You can proceed as you already have.
You mean I should pick the changes from there, right?

ThinkOpenly: Ah, I conflated the two. Yes, we do need C.SSPOPCHK and C.SSPUSH. Note that they're a little weird because their encoding overrides some C.MOP encodings.

sudo-apt-abdullah: Ok, I'll give it some time to understand and then will try to write those files as well.

sudo-apt-abdullah:
@ThinkOpenly, please have a look at your convinience, thanks!

ThinkOpenly:
There are some CI failures. Are you able to look at those?

sudo-apt-abdullah:
> There are some CI failures. Are you able to look at those?

It couldn't find `CSR[Xenvcfg].SSE` which I've added in PR #995. So, it depends on #995 that depends on #891 :)

PR #1011, opened 2025-08-21
feat(ci): mark all new PRs as draft

dhower-qc:


ThinkOpenly:
Is there any way to get a notification when the CI tests have completed? I can see this muddying our workflow some.

dhower-qc:
Fair question. I know we can run an action when checks pass, so in theory we can send a notification. I will see if I can cook up something

dhower-qc:
@ThinkOpenly, did this get discussed in the UDB SIG meeting, and if so, any conclusions?

PR #1023, opened 2025-08-25
fix inconsistent destination register naming in compressed shift instructions

hansfbaier:
In all the other instructions the destination register is named xd, not rd, except the shift instructions.
Furthermore in the assembly attribute of the shift instructions the destination register is designated as xd, not rd,
thus causing code generators to generate incorrect code.

### DIFF
--- a/backends/instructions_appendix/all_instructions.golden.adoc
+++ b/backends/instructions_appendix/all_instructions.golden.adoc
@@ -5206,18 +5206,18 @@ This instruction has different encodings in RV32 and RV64
 RV32::
 [wavedrom, ,svg,subs='attributes',width="100%"]
 ....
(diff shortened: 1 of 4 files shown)


hansfbaier:
This PR fixes this issue: https://github.com/riscv-software-src/riscv-unified-db/issues/1029#event-19329062552

ThinkOpenly (changes requested):
Looks good, but could you revert all of the Sail changes? The Sail code here was a one-time import from the RISC-V Sail repository, and we eventually want some sort of yet-to-be-determined better means of keeping them in sync. For now, we're leaving it unchanged. (Not ideal, I know.)

hansfbaier:
@ThinkOpenly OK, makes sense, done

dhower-qc (approved):

ThinkOpenly:
One of the CI tests is failing. When updating instruction content, there is an additional step required to ensure that if the content of the instruction appendix change, it is intentional, so we maintain a "golden" reference version. This needs to be regenerated and included in this PR. We have a "chore" to do this:
```
$ ./do chore:update_golden_appendix
```
Then, `git add... commit... push`.

PR #1028, opened 2025-08-25
fix: always recreate bundle config by default #980

### LINKED ISSUE #980: Guarding creating of bundle config is overly cautious
https://github.com/riscv-software-src/riscv-unified-db/blob/fe86e703431c9f84ef702d29ed4ee7c04b919a68/bin/setup#L131

When a config file already exists, the config step is skipped.
If ever you want to change something about the future, users who already have a config file will not get that update.
This does not seem to be very time consuming in the grand scheme of things, so it might be better to create a fresh config every time.

Sukuna0007Abhi:
- Change default behavior to always recreate .bundle/config
- Users will now always get configuration updates
- Add --preserve-config option for backward compatibility
- Add --help option for usage information
- Fix argument processing conflicts when script is sourced by other tools
- Prioritizes reliability over performance concerns 
- Addresses overly cautious guard that prevented config updates
- Uses sir @rpsene's argument parsing approach with improved defaults

Fixes #980

Ready for Review sir @ThinkOpenly sir @dhower-qc sir @christian-herber-nxp

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -9,27 +9,30 @@ CONTAINER_TAG=`cat ${ROOT}/bin/.container-tag`
 # Parse command line arguments
 PRESERVE_CONFIG=0
 
-for arg in "$@"; do
-  case $arg in
-    --preserve-config)
-      PRESERVE_CONFIG=1
-      shift
-      ;;
-    --help|-h)
-      echo "Usage: $0 [OPTIONS]"
-      echo ""
-      echo "Options:"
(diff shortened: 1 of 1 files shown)


Sukuna0007Abhi:
Thanks @christian-herber-nxp , pls also review sir @ThinkOpenly @dhower-qc

Sukuna0007Abhi:
Thanks sir @ThinkOpenly @dhower-qc @christian-herber-nxp

PR #1040, opened 2025-08-26
fix(bug): fixed type of excepetion cbo.inval should throw when mode!=M and menvfg.CBIE==0

### LINKED ISSUE #1031: What type of exception should the cbo.inval instruction report when menvcfg.CBIE==00 in vu/vs mode
https://github.com/riscv/riscv-isa-manual/issues/2227

sudo-apt-abdullah:
Closes #1031


dhower-qc (approved):
Thanks!

ThinkOpenly:
One of the CI tests is failing. When updating instruction content, there is an additional step required to ensure that if the content of the instruction appendix change, it is intentional, so we maintain a "golden" reference version. This needs to be regenerated and included in this PR. We have a "chore" to do this:
```
$ ./do chore:update_golden_appendix
```
Then, `git add... commit... push`.

PR #1043, opened 2025-08-26
fix: henvcfg upper-half fields: add base: 64

### LINKED ISSUE #1041: `henvcfg` upper-half fields need `base: 64`
`spec/std/isa/csr/H/henvcfg.yaml`

**Expected value**
All fields in the "upper-half" of `henvcfg` need to be restricted to XLEN 64 (by adding `base: 64`). When XLEN == 32, these fields are not present in `henvcfg` and must be accessed via `henvcfgh`.

(It's probably worth checking other CSRs which have "upper-half" semantics as well.)

Sukuna0007Abhi:
Add base: 64 to henvcfg fields with location >= 32:
- STCE (location: 63)
- PBMTE (location: 62)
- ADUE (location: 61)

This ensures these fields are only present when XLEN == 64. On 32-bit systems, they must be accessed via henvcfgh.

Fixes #1041

Note: Other CSRs (menvcfg, mstatus, etc.) need similar fixes and will be addressed in follow-up PRs.

Ready for check sir @ThinkOpenly @dhower-qc

### DIFF
--- a/spec/std/isa/csr/H/henvcfg.yaml
+++ b/spec/std/isa/csr/H/henvcfg.yaml
@@ -103,6 +103,7 @@ definedBy:
 fields:
   STCE:
     location: 63
+    base: 64
     description: |
       *STimecmp Enable*
 
@@ -126,6 +127,7 @@ fields:
       return (implemented?(ExtensionName::Sstc)) ? UNDEFINED_LEGAL : 0;
   PBMTE:
     location: 62
+    base: 64
     description: |
       *Page Based Memory Type Enable*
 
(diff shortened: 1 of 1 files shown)


Sukuna0007Abhi:
Hm, I checked it but sir I think this cl checks failing are related to upstream, I think it's happening because it's hitting an existing bug in the IDL compiler when it tries to process floating-point instructions.

`The value of 'expZ - 1' is truncated from -1 to 255 because the result is only 8 bits.`

also , Ruby NoMethodError
Location: idlc/lib/idlc/ast.rb:3292 in AST pruning pass
Issue maybe : Code attempts to call .const? method on a nil object

`qualifiers << :const if lhs_type.const? && (short_circuit || rhs_type.const?)
                                ^^^^^^^  # lhs_type is nil`

Sukuna0007Abhi:
Ok I think a I can solve this failure it's need some changes in fp.idl and ast.rb, becuase here the floating-point exponent calculation doesn't handle the edge case where the exponent becomes zero and needs to be decremented.

If you want, sir I can create a issue and separate PR to fix this, although it's unrelated to this PR @ThinkOpenly

PR #1046, opened 2025-08-28
feat: added sie and scountinhibit yaml files

### LINKED ISSUE #563: Add missing S CSRs

sudo-apt-abdullah:
Closes #563 
Depends on #891 
#910 has sieh.yaml that needs to serve as alias for sie which is in this PR!


spec/std/isa/csr/scountinhibit.layout
```diff
@@ -0,0 +1,118 @@
...
+    description:
+      - id: csr-field-hpmcounter<%= hpm_num %>.COUNT-effect
+        normative: true
+        when:
+          param:
+            name: COUNTINHIBIT_EN
+            index: <%= hpm_num %>
+            value: true
+        text: |
+          When bit is set `hpmcounter<%= hpm_num %>.COUNT` does not increment; when 0, `hpmcounter<%= hpm_num %>.COUNT` increments normally.
```
ThinkOpenly:
```suggestion
          When set, `hpmcounter<%= hpm_num %>.COUNT` does not increment; when 0, `hpmcounter<%= hpm_num %>.COUNT` increments normally.
```

spec/std/isa/csr/scountinhibit.layout
```diff
@@ -0,0 +1,118 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
dhower-qc: Give yourself credit!

spec/std/isa/csr/scountinhibit.yaml
```diff
@@ -0,0 +1,1181 @@
...
+    alias: mcountinhibit.CY
+    description: |
+      Cycle inhibit. When 1, the cycle counter does not increment. Only meaningful if
+      mcycle is delegated to S-mode; otherwise reads as 0.
+    type(): |
+      return COUNTINHIBIT_EN[0] ? CsrFieldType::RW : CsrFieldType::RO;
+    reset_value(): |
+      return COUNTINHIBIT_EN[0] ? UNDEFINED_LEGAL : 0;
+    sw_write(csr_value): |
+      if (implemented_csr?(CsrName::menvcfg) && CSR[menvcfg].CDE == 1'b0) {
```
dhower-qc: Checking if `menvcfg` is implemented is correct, but it strays from how we have been handling these types of cases. Usually, we check for the extension that defines what is needed (in this case, S ~> 1.11). In IDL, this would be `implemented_version?(ExtensionName::S, "~> 1.11")`

spec/std/isa/csr/scountinhibit.yaml
```diff
@@ -0,0 +1,1181 @@
...
+        if (mode() == PrivilegeMode::M) {
```
dhower-qc: The spec says:

> When menvcfg.CDE=0, attempts to access scountinhibit raise an illegal instruction exception.

So an Illegal Instruction should raise even in M mode.

spec/std/isa/csr/scountinhibit.yaml
```diff
@@ -0,0 +1,1181 @@
...
+          raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+        }
+      } else if (implemented_csr?(CsrName::menvcfg) && CSR[menvcfg].CDE == 1'b1) {
```
dhower-qc: This check is too restrictive. It won't raise a Virtual Instruction if menvcfg is not implemented, but it should.

To fix, I think you just need to make this an "else".

spec/std/isa/csr/scountinhibit.yaml
```diff
@@ -0,0 +1,1181 @@
...
+    alias: mcountinhibit.IR
+    description: |
+      Instructions-retired inhibit. When 1, the instret counter does not increment. Only
+      meaningful if minstret is delegated to S-mode; else reads 0.
+    type(): |
+      return COUNTINHIBIT_EN[2] ? CsrFieldType::RW : CsrFieldType::RO;
+    reset_value(): |
+      return COUNTINHIBIT_EN[2] ? UNDEFINED_LEGAL : 0;
+    sw_write(csr_value): |
+      if (implemented_csr?(CsrName::menvcfg) && CSR[menvcfg].CDE == 1'b0) {
```
dhower-qc: Same comments as above

spec/std/isa/csr/scountinhibit.yaml
```diff
@@ -0,0 +1,1181 @@
...
+  if (implemented_csr?(CsrName::menvcfg) && CSR[menvcfg].CDE == 1'b0) {
+    if (mode() == PrivilegeMode::M) {
+      raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+    }
+  } else if (implemented_csr?(CsrName::menvcfg) && CSR[menvcfg].CDE == 1'b1) {
+    if ((mode() == PrivilegeMode::VS) || (mode() == PrivilegeMode::VU)) {
+      raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
+    }
+  }
+  return ($bits(CSR[mcountinhibit]) & ($bits(CSR[scounteren]) & ~32'd2)) ;
```
dhower-qc: I think this should be:

```suggestion
  return $bits(CSR[mcountinhibit]) & $bits(CSR[mcounteren]);
```

Which will return everything in mcountinihibit that has been delegated to S.

spec/std/isa/csr/sie.yaml
```diff
@@ -0,0 +1,96 @@
...
+  STIE:
+    location: 5
+    type: RW-H
+    reset_value: UNDEFINED_LEGAL
+    alias:
+      - mie.STIE
+    definedBy: S
+    description: |
+      sie.STIE is the interrupt-enable bit for supervisor-level timer interrupts.
+    sw_write(csr_value): |
```
dhower-qc, line 68: This (and all other bits in sie) needs to account for `mideleg` to determine if the bit has any effect

dhower-qc (changes requested):

spec/std/isa/csr/scountinhibit.yaml
sudo-apt-abdullah: I've updated the condition `(implemented_version?(ExtensionName::Sm, ">=1.12"))` to first check if menvcfg is there, then raise exceptions based on menvcfg.CDE.

spec/std/isa/csr/scountinhibit.yaml
sudo-apt-abdullah: The condition is already there for M mode. I think you're talking about S mode which was missing, right? Let me know if I'm overlooking something!

spec/std/isa/csr/scountinhibit.yaml
sudo-apt-abdullah: I've updated the conditions. Please have a look and suggest any further improvement if needed.

spec/std/isa/csr/scountinhibit.yaml
sudo-apt-abdullah: Makes sense. Done!

spec/std/isa/csr/sie.yaml
sudo-apt-abdullah: I placed a check of respective bits from `mideleg`.

PR #1049, opened 2025-08-30
feat: add zvfofp8min extension

7908837174:


spec/std/isa/ext/Zvfofp8min.yaml
```diff
@@ -0,0 +1,27 @@
...
+kind: extension
+name: Zvfofp8min
+long_name: Vector OFP8 Converts
+description: |
+  This extension provides the minimal set of instructions needed to enable vector support of the
+  OFP8 format.
+  It enables OFP8 as an interchange format as it provides conversion between OFP8 values and FP32 values.
+  The extension offers conversion between OFP8 and BF16 (both directions), and from FP32 to OFP8
+  (only quad narrowing conversion). The narrowing conversions support both saturate and non-saturate
+  modes. Both OFP8 formats, E4M3 and E5M2, are supported.
```
ThinkOpenly, line 16: Where did this text originate? It would be best to use text directly from the spec.

spec/std/isa/ext/Zvfofp8min.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+  This extension depends upon either the V extension or the Zve32f embedded vector extension.
+type: unprivileged
+versions:
+  - version: "0.2.1"
+    state: ratified
```
ThinkOpenly, line 22:
```suggestion
    state: development
```

spec/std/isa/ext/Zvfofp8min.yaml
```diff
@@ -0,0 +1,27 @@
...
+    ratification_date: null
+    requires:
+      anyOf:
+        - V
+        - Zve32f
```
ThinkOpenly, line 27: Per my understanding:
```suggestion
    requires: Zve32f
```
The "Zve*" extensions are designed for minimal vector support in embedded-class implementations, not full "V"/RVV support.

ThinkOpenly (changes requested):

ThinkOpenly:
What are the status and expectations here?

PR #1051, opened 2025-08-30
fix: add backward compatibility for new instruction schema format

AFOliveira:
UDB currently allows for 2 schemas for instructions with the on-going sub-type development. Some instructions now use a 'format' field with 'opcodes' instead of the traditional 'encoding' field with 'match' and 'variables'. This caused generators to fail when processing these instructions as they couldn't extract the bit pattern matching information.

Changes:
- Added build_match_from_format() function to convert format.opcodes to match strings compatible with existing generator logic
- Enhanced encoding detection in load_instruction() to handle both old schema (encoding.match) and new schema (format.opcodes)
- Maintains full backward compatibility with existing instructions
- No functional changes to generated output format

The fix ensures generators can process the complete UDB instruction set regardless of which schema format individual instructions use.


AFOliveira:
Fixes #893

backends/generators/generator.py
```diff
@@ -17,6 +17,55 @@ def check_requirement(req, exts):
...
+    Build a match string from the format field in the new schema.
+    The format field contains opcodes with specific bit fields.
+    """
+    if not format_field or "opcodes" not in format_field:
+        return None
+
+    opcodes = format_field["opcodes"]
+
+    # Initialize a 32-bit match string with all variable bits
+    match_bits = ["-"] * 32
```
ThinkOpenly: This also needs support for instructions that are not 32 bits, correct?

AFOliveira: Correct. I'll add it.

AFOliveira: Fixed.

backends/generators/generator.py
```diff
@@ -17,6 +17,74 @@ def check_requirement(req, exts):
...
+                else:
+                    high = int(location)
+                valid_locations.append(high)
+            except (ValueError, IndexError):
+                continue  # Skip invalid location formats
+
+    if not valid_locations:
+        return None
+
+    max_bit = max(valid_locations)
```
ThinkOpenly, line 71: This will be based entirely on the position of the rightmost _opcode_ in the encoding. I imagine this generally works, and might be good enough for now, but I don't think it would be too difficult to look at the locations of all of the opcodes _and_ variables. That should give you the size of the instruction exactly such that the "width" assignment below is determinative:
```
width = max_bit + 1
```

Indeed, all S-type instructions (32-bit) will be found to be 16-bit opcodes currently, if I read things correctly, because their highest opcode bit is bit 14.

AFOliveira: I'm half asleep, you're completely right. Sorry about it.

AFOliveira: With the help of @AliAlaa88, this is now fixed.

backends/generators/generator.py
```diff
@@ -17,6 +17,94 @@ def check_requirement(req, exts):
...
+
+    # Determine instruction width by finding maximum bit position
+    valid_locations = []
+
+    # Check opcodes
+    for field_data in opcodes.values():
+        if (
+            isinstance(field_data, dict)
+            and "location" in field_data
+            and isinstance(field_data["location"], str)
```
ThinkOpenly, line 39: I think the schema (for "possibly_split_field_location" allows strings ("31-24", "31") or integers (31). I presume we need to handle the latter case here, too?

backends/generators/generator.py
```diff
@@ -17,6 +17,94 @@ def check_requirement(req, exts):
...
+                valid_locations.append(high)
+            except (ValueError, IndexError):
+                continue  # Skip invalid location formats
+
+    # Check variables
+    for var_data in variables.values():
+        if (
+            isinstance(var_data, dict)
+            and "location" in var_data
+            and isinstance(var_data["location"], str)
```
ThinkOpenly, line 56: and here?

backends/generators/generator.py
```diff
@@ -17,6 +17,94 @@ def check_requirement(req, exts):
...
+            except (ValueError, IndexError):
+                continue  # Skip invalid location formats
+
+    if not valid_locations:
+        return None
+
+    max_bit = max(valid_locations)
+
+    # Set instruction width based on maximum bit position
+    width = max(valid_locations) + 1
```
ThinkOpenly, line 74: I'm guessing you meant to use `max_bit` here? (Otherwise, it is never used.) So, either remove the assignment above, or:
```suggestion
    width = max_bit + 1
```

PR #1054, opened 2025-09-01
feat: add SystemVerilog backend generator

AFOliveira:
Implements a new backend generator for SystemVerilog output, matching
the exact format used by riscv-opcodes/inst.sverilog. This provides
direct compatibility with hardware designs using the riscv-opcodes
SystemVerilog package format.

Features:
- Generates SystemVerilog package with instruction and CSR definitions
- Outputs 32-bit instruction patterns with proper bit encoding
- Handles compressed (16-bit) instructions correctly
- Supports all standard RISC-V extensions
- Integrated with the ./do build system as gen:sverilog task

The generator produces output identical to riscv-opcodes format:
- Instructions as 'localparam [31:0] NAME = 32'bpattern'
- CSRs as 'localparam logic [11:0] CSR_NAME = 12'haddr'
- Proper alignment and formatting for readability

Tested against riscv-opcodes/inst.sverilog to ensure format compatibility.


jordancarlin:
> CC @jordancarlin If you can test this or spot any errors/improvements, it would be awesome. If you are already working on something similar I can drop this, I just wanted to give a little help and improve UDB's outputs

Thanks @AFOliveira. I started on it a while ago, but haven't touched it since I ran into #893, so let's go with this version since it is more up to date.

This is looking great so far! I have a few comments with differences from the riscv-opcodes version and then some additional comments for things we should improve compared to the riscv-opcodes version since we're reimplementing it anyway:
- Differences from riscv-opcodes to address
  - There are a number of instructions that have `_RV32` versions even though the encoding is the same for both rv32 and rv64. It would be good to avoid this if possible. I noticed it for `AMOCAS_D_RV32`, `C_LD_RV32`, `C_LDSP_RV32`, `C_SD_RV32`, `S_SDSP_RV32`, `LD_RV32`, and `SD_RV32`.
  - All of the *h CSRs have `_RV32` at the end of the name. This convention makes sense for instruction encodings when they are different for rv32/rv64, but the *h CSRs only have one version so it is unnecessary. The riscv-opcodes version also does not do this, so migration will be harder.
  - Beyond missing extensions (which just need to get added to UDB), I also noticed that riscv-opcodes has `C_ILLEGAL` as an instruction and it is missing from the UDB generated version.
- Improvements that we should make over the riscv-opcodes version
  - Instructions are of type `localparam` while CSRs are `localparam logic`. It would be best to standardize on one, and `localparam logic` is probably the right choice here.
  - `*.sverilog` is an unusual file extension to use. `*.svh` would be a more traditional choice for a SystemVerilog header file.
  - Similar to the C header, it might be worth including the exception causes in the generated SystemVerilog header as localparams.

I'll leave some implementation specific comments inline.

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,179 @@
...
+from generator import load_instructions, load_csrs
+
+
+def format_instruction_name(name):
+    """Format instruction name for SystemVerilog (uppercase with underscores)."""
+    # Handle compressed instructions
+    if name.startswith("c."):
+        name = "C_" + name[2:]
+    # Replace dots with underscores and convert to uppercase
+    return name.replace(".", "_").upper()
```
jordancarlin, line 21: Seems like there is no reason to special-case compressed instructions here? The `c.` --> `C_` conversion should be handled by the standard `name.replace(".", "_").upper()`.

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,179 @@
...
+
+def format_csr_name(name):
+    """Format CSR name for SystemVerilog (uppercase with underscores)."""
+    return "CSR_" + name.replace(".", "_").upper()
+
+
+def match_to_sverilog_bits(match_str, is_compressed=False):
+    """Convert a match string to SystemVerilog bit pattern."""
+    if not match_str:
+        return "32'b" + "?" * 32
```
jordancarlin, line 32: Shouldn't this throw an error/warning instead of silently generating a match string of all wildcards that will match any instruction? That would be very problematic in use because it might shadow an actual match later.

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,179 @@
...
+    
+    # For compressed instructions (16-bit), we need to handle them differently
+    # The riscv-opcodes format puts the 16-bit pattern in the lower 16 bits
+    # with the upper 16 bits as wildcards
+    if is_compressed or len(match_str) == 16:
+        # Pad with wildcards on the left for 16-bit instructions
+        match_str = "?" * 16 + match_str
+    elif len(match_str) < 32:
+        # For other cases, pad on the right
+        match_str = match_str + "-" * (32 - len(match_str))
```
jordancarlin, line 42: What cases do you envision this matching? If none, we're probably better off throwing an error here.

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,179 @@
...
+    return "CSR_" + name.replace(".", "_").upper()
+
+
+def match_to_sverilog_bits(match_str, is_compressed=False):
+    """Convert a match string to SystemVerilog bit pattern."""
+    if not match_str:
+        return "32'b" + "?" * 32
+    
+    # For compressed instructions (16-bit), we need to handle them differently
+    # The riscv-opcodes format puts the 16-bit pattern in the lower 16 bits
```
jordancarlin, line 35: Probably better for comments to stand alone instead of justifying choices based on riscv-opcodes.

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,179 @@
...
+    
+    # Convert to SystemVerilog format (0, 1, or ?)
+    result = []
+    for bit in match_str:
+        if bit == "0":
+            result.append("0")
+        elif bit == "1":
+            result.append("1")
+        else:  # '-' or any other character
+            result.append("?")
```
jordancarlin, line 52: Could this whole block be simplified to just `match_str.replace("-", "?")`?

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,179 @@
...
+    
+    return "32'b" + "".join(result)
+
+
+def generate_sverilog(instructions, csrs, output_file):
+    """Generate SystemVerilog package file."""
+    with open(output_file, "w") as f:
+        # Write header
+        f.write("\n/* Automatically generated by parse_opcodes */\n")
```
jordancarlin, line 61: Maybe this is more clear?
```suggestion
        f.write("\n/* Automatically generated by UDB */\n")
```

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,179 @@
...
+            sv_name = format_instruction_name(name)
+            # Pad the name for alignment
+            padded_name = sv_name.ljust(max_len)
+            
+            # Get the match pattern
+            if isinstance(encoding, dict) and "match" in encoding:
+                match = encoding["match"]
+            else:
+                # If no match field, use all wildcards
+                match = "-" * 32
```
jordancarlin, line 81: As above, please don't generate all wildcards if we don't find a match. A warning/error seems like a better choice here.

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,179 @@
...
+            
+            # Check if this is a compressed instruction
+            is_compressed = name.startswith("c.")
```
jordancarlin, line 84: Is the `is_compressed` logic necessary since you also check if the instruction encoding is 16 bits in the `match_to_sverilog_bits` function? Seems like it should be fine to just use that to check instead of having both the string-based and encoding-based approaches.

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,179 @@
...
+        help="Directory containing instruction YAML files",
+    )
+    parser.add_argument(
+        "--csr-dir",
+        default="../../../gen/resolved_spec/_/csr/",
+        help="Directory containing CSR YAML files",
+    )
+    parser.add_argument(
+        "--output", 
+        default="inst.sverilog",
```
jordancarlin, line 119: What about something like this instead?
```suggestion
        default="riscv_decode_package.svh", 
```

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,179 @@
...
+    parser.add_argument(
+        "--output", 
+        default="inst.sverilog", 
+        help="Output SystemVerilog file name"
+    )
+    parser.add_argument(
+        "--extensions",
+        default="A,D,F,I,M,Q,Zba,Zbb,Zbs,S,System,V,Zicsr,Smpmp,Sm,H,U,Zicntr,Zihpm,Smhpm",
+        help="Comma-separated list of enabled extensions. Default includes standard extensions.",
+    )
```
jordancarlin, line 126: This should not default to a subset of (seemingly arbitrary) extensions. Not sure what the comments means when it says it defaults to the _standard_ extensions. Similar to the generated C header, the default should be to include all extensions.

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,179 @@
...
+    parser.add_argument(
+        "--arch",
+        default="RV64",
+        choices=["RV32", "RV64", "BOTH"],
+        help="Target architecture (RV32, RV64, or BOTH). Default is RV64.",
+    )
```
jordancarlin, line 132: The generated C header does not have this option. I think it makes sense to keep the same CLI interface for both of them. The default output should also include everything, not just RV64.

AFOliveira:
> * Once this is ready, I think it'd be a good candidate for golden output checking like we do with the instruction appendix

Sure, no problem.

> * Not a must, but I do have a long-term vision of transitioning everything out of backends/ and into tools/ to finish up the file reorg. If it isn't too much work, we could move it there now and save some work down the road.

To keep the commit history organized, can we first merge these two (this PR and #1051) and then I may move the full generator folder from one place to another?

jordancarlin:
> > * Once this is ready, I think it'd be a good candidate for golden output checking like we do with the instruction appendix
> 
> Sure, no problem.

We should also add the generated C header as a golden output to avoid issues like #893 from popping up again.

AFOliveira:
Superseeded by #1090

PR #1063, opened 2025-09-04
fix: incorrect pseudoinstruction name for bge (fixes #1062)

### LINKED ISSUE #1062: incorrect pseudoinstruction name in bge instruction
**Describe the bug**
The semantics of the bge xs1, xs2, imm instruction is described as follows:

>   Branch to PC + imm if
>   the signed value in register xs1 is greater than or equal to the signed value in register xs2.

But in the pseudoinstructions we have:
```
pseudoinstructions:
  - when: xs1 == 0
    to: blez xs2,imm
  - when: xs2 == 0
    to: blez xs1,imm
```
So, when xs2 equals zero then the branch is taken if xs1 is greater than or equal than zero,
so per definition the name should be bgez like this:
```
pseudoinstructions:
  - when: xs1 == 0
    to: blez xs2,imm
  - when: xs2 == 0
    to: bgez xs1,imm
```

**To Reproduce**
n/a

n/a

**Screenshots**
not necessary

**Additional context**
not necessary

hansfbaier:
the second pseudoinstruction name is incorrect (see #1062)

### DIFF
--- a/spec/std/isa/inst/I/bge.yaml
+++ b/spec/std/isa/inst/I/bge.yaml
@@ -33,7 +33,7 @@ pseudoinstructions:
   - when: xs1 == 0
     to: blez xs2,imm
   - when: xs2 == 0
-    to: blez xs1,imm
+    to: bgez xs1,imm
 operation(): |
   XReg lhs = X[xs1];
   XReg rhs = X[xs2];


ThinkOpenly (approved):
Thank you!

hansfbaier:
My pleasure!

PR #1068, opened 2025-09-08
fix: rakefile and Readme file

### LINKED ISSUE #1039: Not able to generate implementation-specific spec
**Describe the bug**
`./do gen:arch[rv32]`
This command does not create any files in gen folder.

**To Reproduce**
1. Run './do gen:arch[rv32]'

None

**Screenshots**

Image

**Additional context**
Run command is taken from https://github.com/riscv-software-src/riscv-unified-db/blob/main/README.adoc

ShashankVM:
Fixes #1039

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -84,9 +84,9 @@ namespace :gen do
   desc "Resolve the configuration CFG in arch/, and write it to gen/resolved_arch/<CFG>. Default CFG is the standard, \"_\"."
   task "resolved_arch" do
     cfg = ENV["CFG"]
-      if cfg.nil?
-        cfg = "_"
-      end
+    if cfg.nil?
+      cfg = "_"
+    end
     $resolver.cfg_arch_for(cfg)
   end
 end


dhower-qc:
Hmm. I see what you are saying, and it makes sense.

That said, it's not how things are currently working. The resolution step:

1. Applies an overlay, if used
2. Performs the `$inherits` YAML operator

`1.` is why you sometimes need to specify a config, since that's where an overlay gets specified.

The actual replacement of MXLEN with the MXLEN parameter happens when the resolved config is read/interpreted by the Ruby view.

Getting it to do more at resolution is a very big change. We should avoid that.

Perhaps we just want to rename some things to make it clearer what is happening. Like changing the command so its:

```bash
./do gen:resolved_arch USE_OVERLAY_FROM_CFG=example_rv64_with_overlay
```

Or something.

Rakefile
```diff
@@ -81,10 +81,15 @@ namespace :gen do
     end
   end
 
-  desc "Resolve the standard in arch/, and write it to gen/resolved_arch/_"
-  task "resolved_arch" do
-    $resolver.cfg_arch_for("_")
-  end
+desc "Resolve the configuration CFG in arch/, and write it to gen/resolved_arch/<CFG>. Default CFG is the standard, \"_\"."
+   task "resolved_arch" do
+     $resolver.cfg_arch_for("_")
```
ThinkOpenly: remove this line.
```suggestion
```

Rakefile
```diff
@@ -81,10 +81,14 @@ namespace :gen do
...
-    $resolver.cfg_arch_for("_")
-  end
+desc "Resolve the configuration CFG in arch/, and write it to gen/resolved_arch/<CFG>. Default CFG is the standard, \"_\"."
+   task "resolved_arch" do
+     cfg = ENV["CFG"]
+     if cfg.nil?
+       cfg = "_"
+     end
+     $resolver.cfg_arch_for(cfg)
+   end
```
ThinkOpenly: this is a minor nit... could you change the indentation to 2 spaces for these 7 lines to match the previous code and the project conventions? Thanks!

ShashankVM: Thanks for the feedback, done!

Rakefile
```diff
@@ -81,9 +81,13 @@ namespace :gen do
...
   end
 
-  desc "Resolve the standard in arch/, and write it to gen/resolved_arch/_"
+  desc "Resolve the configuration CFG in arch/, and write it to gen/resolved_arch/<CFG>. Default CFG is the standard, \"_\"."
   task "resolved_arch" do
-    $resolver.cfg_arch_for("_")
+    cfg = ENV["CFG"]
+      if cfg.nil?
+        cfg = "_"
+      end
```
ThinkOpenly: almost there. These three lines need to be de-indented by 2 spaces to be even with the preceding line.
```suggestion
    if cfg.nil?
      cfg = "_"
    end
```

PR #1078, opened 2025-09-10
fix: added MockProcessor-CTP to deploy pages

james-ball-qualcomm:


ThinkOpenly:
We discussed this on the UDB SIG PR call today.

Is this needed? Is it as proof-of-concept? If so, can't MCP100, which is already deployed, serve that purpose?

PR #1090, opened 2025-09-17
feat: add SystemVerilog backend generator

AliAlaa88:
This pull request adds support for generating a SystemVerilog package containing RISC-V instruction and CSR definitions, and enhances the instruction encoding logic to support a new schema format. The changes include introducing a new generator script, updating the encoding extraction logic to handle new formats, and integrating the SystemVerilog generation into the build process.

**SystemVerilog Generation:**
- Added a new script `sverilog_generator.py` that generates a SystemVerilog package (`riscv_decode_package.svh`) with instruction and CSR encodings, including formatting and argument parsing for flexible generation.
- Integrated a new Rake task (`gen:sverilog`) to invoke the SystemVerilog generator, with options for configuration and output directory.
- Created the `gen/sverilog` output directory as part of the build process.

**Instruction Encoding Enhancements:**
- Added `build_match_from_format` function to extract encoding match strings from the new `format` field in instruction definitions, supporting opcodes with specific bit fields.
- Updated instruction loading logic to use the new `build_match_from_format` when the legacy `encoding` field is missing, improving compatibility with new schema definitions.
- Improved handling of RV32/RV64 instruction encodings to ensure both variants are processed and named correctly.

**CSR Handling:**
- Simplified CSR naming logic by always using the uppercase name without `.RV32` suffix, regardless of architecture.

### DIFF
--- a/backends/generators/sverilog/sverilog_generator.py
+++ b/backends/generators/sverilog/sverilog_generator.py
@@ -42,7 +42,7 @@ def parse_args():
         help="Include all instructions and CSRs regardless of extensions",
     )
     parser.add_argument(
-        "--debug", "-v", action="store_true", help="Enable debug logging"
+        "--debug", "-d", action="store_true", help="Enable debug logging"
     )
     parser.add_argument(
         "--extensions",


backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,178 @@
...
+def format_csr_name(name):
+    """Format CSR name for SystemVerilog (uppercase with underscores)."""
+    return "CSR_" + name.replace(".", "_").upper()
+
+
+def match_to_sverilog_bits(match_str):
+    """Convert a match string to SystemVerilog bit pattern."""
+    if not match_str:
+        logging.error(f"Empty match string encountered.")
+        # return "32'b" + "?" * 32
```
jordancarlin: Remove all of the commented out code all throughout this file.

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,178 @@
...
+
+    # For compressed instructions (16-bit), we need to handle them differently.
+    # The 16-bit pattern is in the lower 16 bits,
+    # with the upper 16 bits as wildcards
+    if len(match_str) == 16:
+        # Pad with wildcards on the left for 16-bit instructions
+        match_str = "?" * 16 + match_str
+    elif len(match_str) < 32:
```
jordancarlin:
```suggestion
    elif len(match_str) != 32:
```

We’ll need to do additional work down the road to support instructions >32 bits, so for now let’s just fail for anything other than 16 or 32.

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,178 @@
...
+        # For other cases, pad on the right
```
jordancarlin: Outdated comment

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,178 @@
...
+            # if not name.startswith("v"):
+            #     logging.info(f"Processing instruction: {name}")
+            #     logging.info(f"Formatted instruction: {sv_name}")
+            #     logging.info(f"Padded instruction: {padded_name}")
+
+            # Get the match pattern
+            if isinstance(encoding, dict) and "match" in encoding:
+                match = encoding["match"]
+            else:
+                # If no match field, use all wildcards
```
jordancarlin: Outdated comment

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,178 @@
...
+                logging.warning(f"No match field for instruction {name}.")
```
jordancarlin: This should probably be an error, not a warning.

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,178 @@
...
+        # Write instruction parameters
+        for name in sorted(instructions.keys()):
+            encoding = instructions[name]
+            sv_name = format_instruction_name(name)
+            # Pad the name for alignment
+            padded_name = sv_name.ljust(max_len)
+            # if not name.startswith("v"):
+            #     logging.info(f"Processing instruction: {name}")
+            #     logging.info(f"Formatted instruction: {sv_name}")
+            #     logging.info(f"Padded instruction: {padded_name}")
```
jordancarlin: Extra debug info that should be deleted?

backends/generators/generator.py
```diff
@@ -211,15 +276,17 @@ def load_instructions(
...
                             }  # RV64 gets the default name
 
-                        # Process RV32 encoding with a _rv32 suffix
-                        rv32_match = rv32_encoding.get("match")
-                        if rv32_match:
-                            instr_dict[f"{name}_rv32"] = {"match": rv32_match}
+                        if rv32_match and rv32_match != rv64_match:
+                            # Process RV32 encoding with a _rv32 suffix
+                            if rv32_match:
+                                instr_dict[f"{name}_rv32"] = {"match": rv32_match}
```
jordancarlin: Redundant if conditions
```suggestion
                        if rv32_match and rv32_match != rv64_match:
                            # Process RV32 encoding with a _rv32 suffix
                            instr_dict[f"{name}_rv32"] = {"match": rv32_match}
```

backends/generators/generator.py
```diff
@@ -377,10 +444,11 @@ def load_csrs(csr_root, enabled_extensions, include_all=False, target_arch="RV64
...
 
                 # For BOTH architecture, add suffix to RV32-specific CSRs
-                if target_arch == "BOTH" and base == 32:
-                    csrs[addr_int] = f"{name.upper()}.RV32"
-                else:
-                    csrs[addr_int] = name.upper()
+                # if target_arch == "BOTH" and base == 32:
+                #     csrs[addr_int] = f"{name.upper()}.RV32"
+                # else:
+                #     csrs[addr_int] = name.upper()
```
jordancarlin: Remove commented out code.

backends/generators/tasks.rake
```diff
@@ -87,4 +88,28 @@ namespace :gen do
...
+    FileUtils.mkdir_p output_dir
+
+    # Get the arch paths based on the config
+    resolver = Udb::Resolver.new
+    cfg_arch = resolver.cfg_arch_for(config_name)
+    inst_dir = cfg_arch.path / "inst"
+    csr_dir = cfg_arch.path / "csr"
+
+    # Run the SystemVerilog generator script using the same Python environment
+    sh "#{$root}/.home/.venv/bin/python3 #{$root}/backends/generators/sverilog/sverilog_generator.py --inst-dir=#{inst_dir} --csr-dir=#{csr_dir} --output=#{output_dir}riscv_decode_package.svh"
```
jordancarlin: The C header uses the `--include-all` flag in the rake task. This one should too for consistency.

```suggestion
    sh "#{$root}/.home/.venv/bin/python3 #{$root}/backends/generators/sverilog/sverilog_generator.py --inst-dir=#{inst_dir} --csr-dir=#{csr_dir} --output=#{output_dir}riscv_decode_package.svh --include-all"
```

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,178 @@
...
+        help="Target architecture (RV32, RV64, or BOTH). Default is RV64.",
+    )
+    parser.add_argument(
+        "--verbose", "-v", action="store_true", help="Enable verbose logging"
+    )
+    parser.add_argument(
+        "--include-all",
+        action="store_true",
+        help="Include all instructions and CSRs regardless of extensions",
+    )
```
jordancarlin: The C and SystemVerilog headers should have the same CLI and same defaults. They produce similar artifacts, so having different interfaces introduces unnecessary confusion and complexity.

For example, the C generator has a `--debug` flag while this uses a `--verbose` flag. Both flags seem to do the same thing. Another example is the C header includes an empty list as the default argument for `--extensions` but this includes an arbitrary list of extensions by default.

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,178 @@
...
+
+    return "32'b" + "".join(result)
+
+
+def generate_sverilog(instructions, csrs, output_file):
+    """Generate SystemVerilog package file."""
+    with open(output_file, "w") as f:
+        # Write header
+        f.write("\n/* Automatically generated by UDB */\n")
+        f.write("package riscv_instr;\n")
```
jordancarlin: Let's use the generated file name (without the extension) as the package name.

jordancarlin (changes requested):
This is looking great and has addressed most of my feedback from #1054. I've left some new comments inline (and carried over any from #1054 that haven't been addressed yet.

The other things mentioned in #1054 that haven't been handled here yet and don't fit into an inline comment are the following:

- [x] Similar to the C header, it might be worth including the exception causes in the generated SystemVerilog header as localparams.
- [ ] Add a task to the regression GitHub Action to generate the SystemVerilog header (similar to what is done for the C header).
- [ ] Add a golden output comparison to ensure the output is not unintentionally changed (similar to what is done for the instruction appendix).

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,204 @@
...
+            sv_name = format_csr_name(csr_name)
+            # Pad the name for alignment
+            padded_name = sv_name.ljust(max_len)
+
+            # Format CSR address as 12-bit hex
+            f.write(f"  localparam logic [11:0] {padded_name} = 12'h{addr:03x};\n")
+
+        # Write exception code parameters
+        for num, name in causes:
+            sv_name = "CAUSE_" + name.upper()
```
jordancarlin: Why doesn't this use the `format_cause_name` function you defined above?

backends/generators/tasks.rake
```diff
@@ -87,4 +88,57 @@ namespace :gen do
...
+          'var' => ecode.var,
+          'ext' => ext.name
+        }
+      end
+    end
+
+    # Write resolved exception codes to a temporary JSON file
+    resolved_codes_file = Tempfile.new(['resolved_exception_codes', '.json'])
+    resolved_codes_file.write(JSON.pretty_generate(resolved_exception_codes))
+    resolved_codes_file.flush
```
jordancarlin: We should avoid duplicating all of this from the C task. I'm not familiar enough with rake files, but presumably we can define another task that both the C and SystemVerilog generators depend on that runs this. Check with @dhower-qc if you need help with this.

backends/generators/generator.py
```diff
@@ -403,6 +466,124 @@ def load_csrs(csr_root, enabled_extensions, include_all=False, target_arch="RV64
...
+
+            return unique_codes
+
+        except Exception as e:
+            logging.error(
+                f"Error loading resolved codes file {resolved_codes_file}: {e}"
+            )
+            # Fall back to processing YAML files
+
+    for dirpath, _, filenames in os.walk(ext_dir):
```
jordancarlin, line 567: We should probably do this in another PR, but we should drop the fallback to directly processing YAML files. It produces invalid code (which is why the other version was added in the first place), so it would be better to just throw an error or skip the exception cause section. As is it makes the entire generated artifact unusable if it reaches this code.

jordancarlin:
> Thanks for the review Mr. Jordan, We need to discuss some points about the exception causes.
> 
> Should it be an int or a 6 bit logic signal? You stated that both will work, but you need to decide which is more sense.

I'd probably be in favor of using `logic` just for consistency with the other items in the package, but I could be convinced either way if anyone else has an opinion on this.

> What should be the number format? I made it Firstly in a 0x format but you clarified that it doesn't work in sverilog, so i changed it to 6'h format, is there any other problem?

The new format should be fine.

backends/generators/sverilog/sverilog_generator.py
ThinkOpenly: Looks like `--extensions` was changed to match, but there is still `--debug` in one and `--verbose` in the other. Shall we unify them?

backends/generators/generator.py
ThinkOpenly: Shall we open an issue for this concern?

ThinkOpenly: Ali opened #1145.

backends/generators/sverilog/sverilog_generator.py
jordancarlin: What about moving a version of the `parse_args` function into the shared `generators.py` file? I imagine we will want similar arguments for most (if not all) of the generators. Then individual generators can add any extra arguments that they need on top of that.

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,204 @@
...
+    result = match_str.replace("-", "?")
+
+    return "32'b" + "".join(result)
+
+
+def generate_sverilog(instructions, csrs, causes, output_file):
+    """Generate SystemVerilog package file."""
+    with open(output_file, "w") as f:
+        # Write header
+        f.write("\n/* Automatically generated by UDB */\n")
```
jordancarlin:
```suggestion
        f.write("/* Automatically generated by UDB */\n")
```
No need for a leading newline at the beginning of the file.

backends/generators/sverilog/sverilog_generator.py
ThinkOpenly: > What about moving a version of the `parse_args` function into the shared `generators.py` file?

Will this be included in this PR?

jordancarlin: I’d be in favor of merging this PR as is and then we can do this as a future optimization.

ThinkOpenly: We can refactor in a later PR, but we should unify the incompatible options now. Since this is the newer code, to my understanding, let's change the `--verbose` option here to `--debug` to match the C header generation.

ThinkOpenly: I also ask that the `parse_args` function be moved to near the top of the file. (Personal preference, but I find it handy/expected to find the command line options early/first in the file.)

backends/generators/sverilog/sverilog_generator.py
```diff
@@ -0,0 +1,204 @@
...
+        default="riscv_decode_package.svh",
+        help="Output SystemVerilog file name",
+    )
+    parser.add_argument(
+        "--include-all",
+        action="store_true",
+        help="Include all instructions and CSRs regardless of extensions",
+    )
+    parser.add_argument(
+        "--debug", "-v", action="store_true", help="Enable debug logging"
```
ThinkOpenly:
```suggestion
        "--debug", "-d", action="store_true", help="Enable debug logging"
```

ThinkOpenly (approved):

PR #1092, opened 2025-09-19
fix: add backward compatibility for new instruction schema format

AliAlaa88:
This pull request enhances support for a new instruction schema in the code generator by adding logic to interpret and convert the newer `format` field into the legacy `encoding` format when loading instructions. This ensures backward compatibility and smoother migration to the new schema without breaking existing functionality.

**Support for new instruction schema:**

* Added a new function `build_match_from_format` in `generator.py` to construct a match string from the `format` field, supporting 16-, 32-, and 48-bit instructions. This function interprets opcode and variable bit locations and values to generate the appropriate match string.
* Updated the instruction loading logic in `load_instructions` to detect when only a `format` field (and not an `encoding` field) is present. If found, it uses `build_match_from_format` to synthesize an `encoding` entry, allowing the rest of the pipeline to function unchanged. If conversion fails, the instruction is filtered out with an error log.

### DIFF
--- a/backends/generators/generator.py
+++ b/backends/generators/generator.py
@@ -20,87 +20,98 @@ def check_requirement(req, exts):
 def build_match_from_format(format_field):
     """
     Build a match string from the format field in the new schema.
-    Supports 16-bit, 32-bit, and 48-bit instructions.
     """
     if not format_field or "opcodes" not in format_field:
         return None
 
-    opcodes = format_field["opcodes"]
-    variables = format_field["variables"]
-
     # Determine instruction width by finding maximum bit position
     valid_locations = []
 
+    opcodes = format_field["opcodes"]
     # Check opcodes
     for field_data in opcodes.values():
-        if (
-            isinstance(field_data, dict)
-            and "location" in field_data
-            and isinstance(field_data["location"], str)
-        ):
-            try:
-                location = field_data["location"]
-                if "-" in location:
-                    high = int(location.split("-")[0])
-                else:
-                    high = int(location)
-                valid_locations.append(high)
-            except (ValueError, IndexError):
-                continue  # Skip invalid location formats
+        if isinstance(field_data, dict) and "location" in field_data:
+            if isinstance(field_data["location"], str):
+                try:
+                    location = field_data["location"]
+                    split_location = location.split("|")
+                    high = max(
+                        (
+                            int(location.split("-")[0])
+                            if "-" in location
+                            else int(location)
+                        )
+                        for location in split_location
+                    )
+                    valid_locations.append(high)
+                except (ValueError, IndexError):
+                    raise ValueError(
+                        f"Invalid location format: {field_data['location']}"
+                    )
+            elif isinstance(field_data["location"], int):
+                try:
+                    valid_locations.append(field_data["location"])
+                except (ValueError, IndexError):
+                    raise ValueError(
+                        f"Invalid location format: {field_data['location']}"
+                    )
+            else:
+                raise ValueError(f"Unknown location format: {field_data['location']}")
 
-    # Check variables
-    for var_data in variables.values():
-        if (
-            isinstance(var_data, dict)
-            and "location" in var_data
-            and isinstance(var_data["location"], str)
-        ):
-            try:
-                location = var_data["location"]
-                if "-" in location:
-                    high = int(location.split("-")[0])
+    if "variables" in format_field:
+        variables = format_field["variables"]
+        # Check variables
+        for var_data in variables.values():
+            if isinstance(var_data, dict) and "location" in var_data:
+                if isinstance(var_data["location"], str):
+                    try:
+                        location = var_data["location"]
+                        if "-" in location:
+                            high = int(location.split("-")[0])
+                        else:
+                            high = int(location)
+                        valid_locations.append(high)
+                    except (ValueError, IndexError):
+                        raise ValueError(
+                            f"Invalid location format: {var_data['location']}"
+                        )
+                elif isinstance(var_data["location"], int):
+                    try:
+                        valid_locations.append(var_data["location"])
+                    except (ValueError, IndexError):
+                        raise ValueError(
+                            f"Invalid location format: {var_data['location']}"
+                        )
                 else:
-                    high = int(location)
-                valid_locations.append(high)
-            except (ValueError, IndexError):
-                continue  # Skip invalid location formats
+                    raise ValueError(f"Invalid location format: {var_data['location']}")
 
     if not valid_locations:
-        return None
+        raise ValueError("No valid bit locations found in format field")
 
     max_bit = max(valid_locations)
 
     # Set instruction width based on maximum bit position
-    width = max(valid_locations) + 1
+    width = max_bit + 1
     match_bits = ["-"] * width
 
     # Populate match string with opcode bits
     for field_data in opcodes.values():
-        if (
-            isinstance(field_data, dict)
-            and "location" in field_data
-            and "value" in field_data
-            and isinstance(field_data["location"], str)
-            and isinstance(field_data["value"], int)
-        ):
-
+        if isinstance(field_data, dict):
             try:
                 location = field_data["location"]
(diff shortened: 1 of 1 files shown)


backends/generators/generator.py
```diff
@@ -17,6 +17,102 @@ def check_requirement(req, exts):
...
+def build_match_from_format(format_field):
+    """
+    Build a match string from the format field in the new schema.
+    Supports 16-bit, 32-bit, and 48-bit instructions.
+    """
+    if not format_field or "opcodes" not in format_field:
+        return None
+
+    opcodes = format_field["opcodes"]
+    variables = format_field["variables"]
```
ThinkOpenly: We probably need to be a bit more careful here. While it is impossible to have an instruction without "opcode" fields, it is possible to have instruction without "variables" fields.

backends/generators/generator.py
```diff
@@ -17,6 +17,102 @@ def check_requirement(req, exts):
...
+            if isinstance(field_data["location"], str):
+                try:
+                    location = field_data["location"]
+                    if "-" in location:
+                        high = int(location.split("-")[0])
+                    else:
+                        high = int(location)
+                    valid_locations.append(high)
+                except (ValueError, IndexError):
+                    continue  # Skip invalid location formats
```
ThinkOpenly: I suggest letting this fail. I don't think we want to ignore errors here. "location", if present, should be valid.

Note, also, that we have "split location" fields:
```
spec/std/isa/inst/Zcf/c.fsw.yaml:      location: 5|12-10|6
```

backends/generators/generator.py
```diff
@@ -17,6 +17,102 @@ def check_requirement(req, exts):
...
+            elif isinstance(field_data["location"], int):
+                try:
+                    valid_locations.append(field_data["location"])
+                except (ValueError, IndexError):
+                    continue  # Skip invalid location formats
+            else:
+                continue  # Skip invalid location formats
```
ThinkOpenly: These would appear to be error conditions we shouldn't ignore.

backends/generators/generator.py
```diff
@@ -17,6 +17,102 @@ def check_requirement(req, exts):
...
+
+    # Check variables
+    for var_data in variables.values():
+        if isinstance(var_data, dict) and "location" in var_data:
+            if isinstance(var_data["location"], str):
```
ThinkOpenly: The several comments for "opcodes" above also apply here.

backends/generators/generator.py
```diff
@@ -17,6 +17,102 @@ def check_requirement(req, exts):
...
+            elif isinstance(var_data["location"], int):
+                try:
+                    valid_locations.append(var_data["location"])
+                except (ValueError, IndexError):
+                    continue  # Skip invalid location formats
+            else:
+                continue  # Skip invalid location formats
+
+    if not valid_locations:
+        return None
```
ThinkOpenly: Also seems like an error condition we shouldn't ignore.

backends/generators/generator.py
```diff
@@ -17,6 +17,102 @@ def check_requirement(req, exts):
...
+
+    # Populate match string with opcode bits
+    for field_data in opcodes.values():
+        if (
+            isinstance(field_data, dict)
+            and "location" in field_data
+            and "value" in field_data
+            and isinstance(field_data["location"], str)
+            and isinstance(field_data["value"], int)
+        ):
```
ThinkOpenly: I think we want to fail if any of these aren't true, so I think we can remove the tests.

However, might "location" be an int (for a single bit opcode)? You'll need to handle that below, like you did well above.

backends/generators/generator.py
```diff
@@ -17,6 +17,102 @@ def check_requirement(req, exts):
...
+
+            try:
+                location = field_data["location"]
```
ThinkOpenly, line 101: might be an int, so be careful.

We don't support "split fields" for opcodes, though, so the code below which presumes formats comply with "X" or "X-Y" is fine.

backends/generators/generator.py
```diff
@@ -17,6 +17,102 @@ def check_requirement(req, exts):
     return False
 
 
+def build_match_from_format(format_field):
+    """
+    Build a match string from the format field in the new schema.
+    Supports 16-bit, 32-bit, and 48-bit instructions.
```
ThinkOpenly: We don't check or enforce this in the code. You could, of course, or remove this line.

backends/generators/generator.py
```diff
@@ -17,6 +17,102 @@ def check_requirement(req, exts):
...
+                    high = low = int(location)
+
+                if high < low or high >= width:
+                    continue  # Skip invalid bit ranges
+
+                binary_value = format(field_data["value"], f"0{high - low + 1}b")
+                for i, bit in enumerate(binary_value):
+                    pos = high - i
+                    if 0 <= pos < width:
+                        match_bits[width - 1 - pos] = bit
```
ThinkOpenly: I haven't thought this through fully, but can you just do a full substring replace?
```
match_bits[low:high-1] = binary_value
```
?

backends/generators/generator.py
```diff
@@ -17,6 +17,102 @@ def check_requirement(req, exts):
...
+            except (ValueError, IndexError):
+                continue  # Skip invalid field data
```
ThinkOpenly: I think this also is a case where we do not want to ignore errors.

backends/generators/generator.py
```diff
@@ -177,11 +273,27 @@ def load_instructions(
...
+                            f"Could not build encoding from format field in instruction {name} in {path}"
+                        )
+                        encoding_filtered += 1
+                        continue
+                else:
+                    logging.error(
+                        f"Missing 'encoding' field in instruction {name} in {path}"
+                    )
+                    encoding_filtered += 1
+                    continue
```
ThinkOpenly: Let's swap the `if` and `else` cases to reduce the indentation:
```suggestion
                if not format_field:
                    logging.error(
                        f"Missing 'encoding' field in instruction {name} in {path}"
                    )
                    encoding_filtered += 1
                    continue

                # Try to build a match string from the format field
                match_string = build_match_from_format(format_field)
                if not match_string:
                    logging.error(
                        f"Could not build encoding from format field in instruction {name} in {path}"
                    )
                    encoding_filtered += 1
                    continue

                # Create a synthetic encoding compatible with existing logic
                encoding = {"match": match_string, "variables": []}
                logging.debug(f"Built encoding from format field for {name}")
```

ThinkOpenly (changes requested):

backends/generators/generator.py
AliAlaa88: Should I log the error using logging.error then continue or fail the whole process

backends/generators/generator.py
AliAlaa88: what tests to remove?

backends/generators/generator.py
ThinkOpenly: My preference is an ugly, messy failure that will fail any CI tests. I could be talked out of that if we find that UDB has far too many problems such that this approach will fail without a lot of effort. So, I suggest we start with a hard failure and see how things look.

backends/generators/generator.py
ThinkOpenly: I'm thinking we should remove them all. Change `if (tests) code else error-handling` to `code`.

AliAlaa88: The dict instance check for field_data can't be removed because $child_of attribute of opcodes is not a dict and have no location or value
So I will just keep this condition to skip this case and let other conditions cause a failure.

backends/generators/generator.py
ThinkOpenly: I lied. We don't support "split fields" for _variables_. We do support split fields for opcodes, so similar changes to those you made in the first `for` loop above, for opcodes, are also needed here.

ThinkOpenly: Sorry. Ignore my last comment. It was easy for me to create a split field opcode for testing, but my first comment was correct: we do not support that.

PR #1134, opened 2025-09-25
feat(idl): validated support for virtual memory and S-Mode

henrikg-qc:
- Added support for Sv32 in mc100-32-riscv-tests configuration.
- Several bug fixes including comparators and bit extraction.
- Added instruction tracing to hit basic block handling.
- Added regression tests for M, C, S extensions.

### DIFF
--- a/backends/generators/generator.py
+++ b/backends/generators/generator.py
@@ -17,6 +17,105 @@ def check_requirement(req, exts):
     return False
 
 
+def build_match_from_format(format_field):
+    """
+    Build a match string from the format field in the new schema.
+    """
+    if not format_field or "opcodes" not in format_field:
+        return None
+
+    # Determine instruction width by finding maximum bit position
+    valid_locations = []
+
+    opcodes = format_field["opcodes"]
+    # Check opcodes
+    for field_data in opcodes.values():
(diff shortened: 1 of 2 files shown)


henrikg-qc:
Note that wfi test is excluded from the riscv-tests for the regression as SIE CSR is not yet supported.
Awaiting #1046

dhower-qc (approved):
👍

cfgs/mc100-32-riscv-tests.yaml
jordancarlin, line 6: Should we consider renaming this config since it is moving further and further away from mc100? Maybe something like `rv32-riscv-tests`?

henrikg-qc: I brought this up a couple of weeks ago with Derek as I also wanted to separate this out from MC100. I think we decided to take the tested MC100-riscv-tests config and make it MC100, however I agree with you this is now widening scope of what it is supporting.
What is MC100 supposed to contain? Is there a definition for MC100/200/x00 is supposed to support?

I will merge this as is for now and create an issue for myself to rename and/or merge content with MCx100 configurations

PR #1135, opened 2025-09-27
feat: add support for Svrsw60t59b extension

jordancarlin:


spec/std/isa/isa/globals.isa
```diff
@@ -1834,10 +1834,16 @@ function gstage_page_walk {
...
+      # Sv32 has no reserved bits, and Sv39/48/57 all have reserved bits at 58:54
+      if ((VA_SIZE != 32) && (pte[58:54] != 0)) {
         raise_guest_page_fault(op, gpaddr, vaddr, tinst, effective_mode);
       }
+      if (!implemented?(ExtensionName::Svrsw60t59b)) {
+        if ((PTESIZE >= 64) && pte[60:59] != 0) {
+          # 60:59 are reserved if Svrsw60t59b is not supported
+          raise_guest_page_fault(op, gpaddr, vaddr, tinst, effective_mode);
+        }
+      }
```
ThinkOpenly, line 1855: since the body is the same, shall we combine these?
```suggestion
      if (
          # Sv32 has no reserved bits, and Sv39/48/57 all have reserved bits at 58:54
          ((VA_SIZE != 32) && (pte[58:54] != 0))
          ||
          # 60:59 are reserved if Svrsw60t59b is not supported
          ((!implemented?(ExtensionName::Svrsw60t59b)) &&
           ((PTESIZE >= 64) && pte[60:59] != 0))
      ) {
          raise_guest_page_fault(op, gpaddr, vaddr, tinst, effective_mode);
      }
```

spec/std/isa/isa/globals.isa
```diff
@@ -2147,6 +2153,13 @@ function stage1_page_walk {
         }
       }
 
+      if (!implemented?(ExtensionName::Svrsw60t59b)) {
+        if ((PTESIZE >= 64) && pte[60:59] != 0) {
+          # 60:59 are reserved if Svrsw60t59b is not supported
+          raise (page_fault_code, effective_mode, vaddr);
+        }
+      }
```
ThinkOpenly, line 2170: Similarly, shall we combine this with the earlier reserved bit test, or do the intervening checks need to be in a particular order?

ThinkOpenly (commented):
Questions, possibly borne of ignorance.

spec/std/isa/isa/globals.isa
jordancarlin: I was following the existing pattern used for all of the other extensions that impose additional checks on the page tables. Almost all of them have the same body. Happy to merge if you'd prefer.

ThinkOpenly: The awkward part, of course, is that Sv39/Sv48/Sv57 are defined as having reserved bits 60:54, and Svrsw60t59b unreserves 60:59. It's not terribly important, but I'd prefer to combine so that the initial set of reserved bits is handled as a unit.

jordancarlin: Updated

spec/std/isa/isa/globals.isa
```diff
@@ -1834,8 +1834,13 @@ function gstage_page_walk {
...
 
       # check if any reserved bits are set
-      # Sv32 has no reserved bits, and Sv39/48/57 all have reserved bits at 60:54
-      if ((VA_SIZE != 32) && (pte[60:54] != 0)) {
+      if (
+        # Sv32 has no reserved bits, and Sv39/48/57 all have reserved bits at 58:54
+        ((VA_SIZE != 32) && (pte[58:54] != 0))
+        ||
+        # 60:59 are reserved if Svrsw60t59b is not supported
+        ((!implemented?(ExtensionName::Svrsw60t59b)) && ((PTESIZE >= 64) && pte[60:59] != 0))
```
ThinkOpenly: Since these conditions are all ANDed, we don't need the extra parentheses, here, and below, but this is a nit. I'll approve now, without any further changes. We can merge with or without.
```suggestion
        (!implemented?(ExtensionName::Svrsw60t59b) && (PTESIZE >= 64) && pte[60:59] != 0)
```

ThinkOpenly (approved):

spec/std/isa/isa/globals.isa
```diff
@@ -2118,9 +2123,14 @@ function stage1_page_walk {
...
-      if ((VA_SIZE != 32) && (pte[60:54] != 0)) {
-        raise(page_fault_code, effective_mode, vaddr);
+      if (
+        # Sv32 has no reserved bits, and Sv39/48/57 all have reserved bits at 58:54
+        ((VA_SIZE != 32) && (pte[58:54] != 0))
+        ||
+        # 60:59 are reserved if Svrsw60t59b is not supported
+        ((!implemented?(ExtensionName::Svrsw60t59b)) && ((PTESIZE >= 64) && pte[60:59] != 0))
+      ) {
+        raise_guest_page_fault(op, gpaddr, vaddr, tinst, effective_mode);
```
ThinkOpenly: maybe a cut-and-pasto?

```suggestion
        raise(page_fault_code, effective_mode, vaddr);
```

ThinkOpenly (changes requested):

spec/std/isa/isa/globals.isa
jordancarlin: I'm surprised it's not short circuiting the conditional here, but unfortunately this approach doesn't work:
```
In the code:

  1840: |
  1841:         # 60:59 are reserved if Svrsw60t59b is not supported
  1842:         (!implemented?(ExtensionName::Svrsw60t59b) && (PTESIZE >= 64) && pte[60:59] != 0)
  1843:       ) {
  1844:         raise_guest_page_fault(op, gpaddr, vaddr, tinst, effective_mode);
  

A type error occurred
  Range too large for bits (msb = 60, range size = 32)
```

jordancarlin: I've reverted to the previous implementation for now.

ThinkOpenly: @dhower-qc , shall we open an issue for this? (I'm thinking 'yes'.)

ThinkOpenly (approved):

spec/std/isa/isa/globals.isa
ThinkOpenly: #1140 opened

ThinkOpenly:
I think this just needs a final rebase?

PR #1138, opened 2025-09-29
fix: remove all tagging from descriptions

sudo-apt-abdullah:


spec/custom/non_isa/preface_demo.yaml
```diff
@@ -28,85 +25,65 @@ license:
...
-          The manual assumes a basic familiarity with computer architecture and programming concepts, but provides background information and references for further study.
-      - id: sec-preface-audience-conditional
-        normative: false
-        text: |
-          This section is only included for educational distributions.
-        when(): "distribution_type == 'educational'"
+      The manual assumes a basic familiarity with computer architecture and programming concepts, but provides background information and references for further study.
+
+      This section is only included for educational distributions.
+    when(): "distribution_type == 'educational'"
```
ThinkOpenly: In the previous content, this `when()` method applied only to the preceding line.

sudo-apt-abdullah: > In the previous content, this `when()` method applied only to the preceding line.

I'm sorry, but I couldn't get it. Can you please elaborate?

ThinkOpenly: The content without the changes here is:
```
      - id: sec-preface-audience-conditional
        normative: false
        text: |
          This section is only included for educational distributions.
        when(): "distribution_type == 'educational'"
```
So, the `when()` conditional only applies to that single line.

With these changes, I think that the "description" is still defined as an array of "tagged_text", but that definition was renamed to "conditional_text", so I'm a bit confused how the CI tests passed. I must be missing something.

Anyway, in these changes, the `when()` clause is associated with _all_ of the previous text, which is not desired.

sudo-apt-abdullah: There were some tagged description with `when():` condition, especially in `hstatus.yaml`. So, I updated the definition from tagged_text to conditional_text after removing tags, so that it can make more sense now.

ThinkOpenly: But with these latest changes (`when()` removed), that phrase is no longer suppressed, as it was previously. Can we get back to the original behavior, where that clause, and only that clause, is conditionally displayed?

PR #1139, opened 2025-09-29
feat(backends): add GNU Assembler Test Generator for RISC-V

AnimeshAgarwal28:
This PR introduces a test generator that creates GNU Assembler test files from the RISC-V unified database. The generator produces test files in the format expected by the binutils gas test suite.

Currently the generated test files although very close to upstream test files can be incorrect due to issues like:

1. Extension Dependencies: Instructions like ld and sd require the zilsd extension for RV32 but are incorrectly included in the base i extension tests.

2. Assembly Field Validation: Some instructions have malformed assembly patterns in the database, leading to invalid test cases (e.g., sspopchk.x1).

For usage instructions and more information about the generator check gas_test_generator_readme.md

---

*This is a draft PR intended to gather community feedback/suggestions on approach and priorities before investing in accuracy improvements.*

### DIFF
--- a/backends/generators/binutils/gas_test_generator.py
+++ b/backends/generators/binutils/gas_test_generator.py
@@ -26,6 +26,22 @@ from typing import Dict, List, Tuple, Set
 sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 from generator import (parse_extension_requirements, load_csrs)
 
+# Named constants for fallback 12-bit signed immediate range.
+# The 12-bit signed immediate range is from -2048 to 2047 (inclusive).
+DEFAULT_12BIT_SIGNED_IMM_MIN = -2048
+DEFAULT_12BIT_SIGNED_IMM_MAX = 2047
+
+# Named constants for fallback 12-bit unsigned immediate range.
+# The 12-bit unsigned immediate range is from 0 to 4095 (inclusive).
+DEFAULT_12BIT_UNSIGNED_IMM_MIN = 0
+DEFAULT_12BIT_UNSIGNED_IMM_MAX = 4095
+
+# Maximum number of CSR example names to keep for replacements
+MAX_CSR_EXAMPLES = 10
+
+# Maximum length for sanitized extension filenames
+MAX_EXTENSION_NAME_LENGTH = 20
+
 
 def calculate_location_width(location) -> int:
     """Calculate the total bit width from a location string or integer."""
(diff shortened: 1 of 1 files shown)


AFOliveira (changes requested):
Hello, @AnimeshAgarwal28.

Thanks for the PR, can you change it so that instead of merging into main, you merge it into my branch that has binutils generator? It's AFOliveira/addBinutils.

Also, if possible, can you post the artifacts this approach generates? Even if just an exapmle for the I extension.

AFOliveira:
Are you only generating valid tests? I.e. not generating test that can fail?

backends/generators/generator.py
```diff
@@ -3,11 +3,48 @@
...
 import pprint
+import re
+import glob
+from typing import Dict, List, Tuple, Optional
 
 pp = pprint.PrettyPrinter(indent=2)
 logging.basicConfig(level=logging.INFO, format="%(levelname)s:: %(message)s")
 
 
+def calculate_location_width(location) -> int:
```
AFOliveira: Is there any reason for this to be here? It is not used anywhere else in the other generators, only in the gas code.

backends/generators/binutils/gas_test_generator.py
```diff
@@ -0,0 +1,1143 @@
...
+                
+                instruction_data[name] = data
+                
+                constraints = extract_instruction_constraints(name, data)
+                if constraints:
+                    instruction_constraints[name] = constraints
+                
+                defined_by = data.get('definedBy')
+                if defined_by:
+                    ext_name = self._extract_extension_name(defined_by)
```
AFOliveira: This function is private to another class, why did you call it here? It's not working, but I suppose the code is still producing functional ouput. Does this mean this is not required and could be removed? Or extract extesion name should be made public?

backends/generators/generator.py
AnimeshAgarwal28: I thought of adding this function here so that future generators could use it, but since currently only gas-test-generator uses it, it can be removed and added later if required.

backends/generators/binutils/gas_test_generator.py
AnimeshAgarwal28: _extract_extension_name is not required here, I'll fix this.

AFOliveira:
@AnimeshAgarwal28 Is this the version that you already tested on binutils and that compiled?

AnimeshAgarwal28:
Yes, a lot of the generated tests work as expected, there are still issues with some of the extensions. For eg. c-d.s does not pass because of a hyphen in the march string, this needs to be addressed.

Issue #1015, opened 2025-08-22
fix jsonschema lint failures: Setting `type` alongside `const` is considered an anti-pattern, as the constant already implies its respective type (const_with_type)


kbroch-rivosinc:
Why?: If ruby code uses the "type" to decide the type, then this turns this optional keyword in json schema into a mandatory one.

```
kbroch@kbroch-mac.local:~/rvi/non-repo-repos/riscv-unified-db/spec/schemas on  main
❯ /opt/homebrew/bin/jsonschema lint  --ignore json-schema-draft-07.json --resolve *.json . | rg -B 1  const_with_type
/Users/kbroch/rvi/non-repo-repos/riscv-unified-db/spec/schemas/proc_cert_class_schema.json:
  Setting `type` alongside `const` is considered an anti-pattern, as the constant already implies its respective type (const_with_type)
--
/Users/kbroch/rvi/non-repo-repos/riscv-unified-db/spec/schemas/proc_cert_class_schema.json:
  Setting `type` alongside `const` is considered an anti-pattern, as the constant already implies its respective type (const_with_type)
--
/Users/kbroch/rvi/non-repo-repos/riscv-unified-db/spec/schemas/proc_cert_class_schema.json:
  Setting `type` alongside `const` is considered an anti-pattern, as the constant already implies its respective type (const_with_type)
--
/Users/kbroch/rvi/non-repo-repos/riscv-unified-db/spec/schemas/proc_cert_class_schema.json:
  Setting `type` alongside `const` is considered an anti-pattern, as the constant already implies its respective type (const_with_type)
--
/Users/kbroch/rvi/non-repo-repos/riscv-unified-db/spec/schemas/proc_cert_model_schema.json:
  Setting `type` alongside `const` is considered an anti-pattern, as the constant already implies its respective type (const_with_type)
--
/Users/kbroch/rvi/non-repo-repos/riscv-unified-db/spec/schemas/proc_cert_model_schema.json:
  Setting `type` alongside `const` is considered an anti-pattern, as the constant already implies its respective type (const_with_type)
--
/Users/kbroch/rvi/non-repo-repos/riscv-unified-db/spec/schemas/proc_cert_model_schema.json:
  Setting `type` alongside `const` is considered an anti-pattern, as the constant already implies its respective type (const_with_type)
--
/Users/kbroch/rvi/non-repo-repos/riscv-unified-db/spec/schemas/proc_cert_model_schema.json:
[...]

PR #1150, opened 2025-10-02
feat: add GPR Information to UDB

### LINKED ISSUE #1085: Add GPR Information to UDB
Yes. The UDB currently lacks information about General Purpose Registers in YAML format. The information currently is hardcoded in downstream projects. [RISC-V ABIs Specification's]( https://riscv-non-isa.github.io/riscv-elf-psabi-doc/) first few sections define these neatly.  

Ideally ABI specification's should be a part of the UDB but it is a big change, we can start with RISC-V Calling Conventions and add information about all the GPRs. 

The only alternative is the current situation, where each software tool maintains its own internal, hardcoded mapping of the RISC-V GPRs. This approach lacks a single source of truth and is less flexible for automation and verification.

AnimeshAgarwal28:
This PR implements GPR definition in the Unified Database, addressing issue #1085.

## Problem

The UDB currently lacks information about General Purpose Registers in YAML format.

## Solution

This PR adds structured register file support to UDB, starting with RISC-V general purpose registers as a foundation for future register file additions.

### New Files
- **`spec/schemas/register_schema.json`**: JSON schema defining the structure for YAML register file.
- **`spec/std/isa/register/gpr.yaml`**: Complete RISC-V general purpose register file with all 32 registers, proper ABI mnemonics, calling convention roles, and conditional support for RV32E (16 registers)
- **`tools/ruby-gems/udb/lib/udb/obj/register_file.rb`**: RegisterFile class extending TopLevelDatabaseObject for programmatic access to register file data

### Modified Files
- **`spec/schemas/schema_defs.json`**: Added register-specific schema definitions
- **`spec/std/isa/README.adoc`**: Updated architecture documentation to include register files alongside extensions/instructions/CSRs with usage examples
- **`tools/ruby-gems/udb/lib/udb/obj/database_obj.rb`**: Added RegisterFile kind to DatabaseObject::Kind enum for type system integration
- **`tools/ruby-gems/udb/lib/udb/architecture.rb`**: Added register file loading support to architecture system

## Register Details

The GPR implementation includes:
- All 32 general purpose registers (x0-x31) with standard names
- Proper ABI mnemonics (zero, ra, sp, gp, tp, t0-t6, s0-s11, a0-a7)
- Calling convention classifications (caller/callee saved, arguments, return values)
- Special register roles (zero register, stack pointer, frame pointer, etc.)
- Conditional presence for RV32E embedded profile

## Benefits

- **Single Source of Truth**: Eliminates need for hardcoded register mappings in downstream tools
- **Consistency**: Ensures all tools use identical register information
- **Automation Ready**: Structured format enables automatic code generation

## Future Work

This establishes the foundation for adding other register files mentioned in issue #1085:
- Floating Point Registers
- Vector Registers

Closes #1085

### DIFF
--- a/tools/ruby-gems/udb/lib/udb/obj/register_file.rb
+++ b/tools/ruby-gems/udb/lib/udb/obj/register_file.rb
@@ -5,7 +5,7 @@
 # frozen_string_literal: true
 
 require_relative "database_obj"
-require_relative "../req_expression"
+require_relative "../condition"
 
 module Udb
 
@@ -82,18 +82,18 @@ class RegisterFile < TopLevelDatabaseObject
     sig { returns(T.nilable(String)) }
     def sw_write = arch_write
 
-    sig { returns(T.nilable(ExtensionRequirementExpression)) }
+    sig { returns(T.nilable(AbstractCondition)) }
     def defined_by_condition
       return nil unless @data.key?("definedBy")
 
-      @defined_by_condition ||= ExtensionRequirementExpression.new(@data.fetch("definedBy"), @file.arch)
+      @defined_by_condition ||= Condition.new(@data.fetch("definedBy"), @file.arch)
     end
 
-    sig { returns(T.nilable(ExtensionRequirementExpression)) }
+    sig { returns(T.nilable(Condition)) }
     def when_condition
       return nil unless @data.key?("when")
 
-      @when_condition ||= ExtensionRequirementExpression.new(@data.fetch("when"), @file.arch)
+      @when_condition ||= Condition.new(@data.fetch("when"), @file.arch)
     end
   end
 


spec/schemas/register_file_schema.json
ThinkOpenly, line 1: Since CSRs are also "registers", maybe we use a slightly more specific file name, like "register_file_schema.json" or "regfile_schema.json"?

spec/schemas/register_schema.json
```diff
@@ -0,0 +1,157 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+
+  "title": "Register File Schema",
+  "description": "Schema for describing a register file",
+
+  "$defs": {
+    "count_variant": {
```
ThinkOpenly: I'm not sure the best way to do this. I'd prefer to put the burden of defining the dimension of the register file on whatever needs to change it from a default of 32, like the "E" extension in this case. I'd prefer to avoid the definition of "variants" here, allow the register files to specify a default value, and have a configuration parameter override this. You could even have the extension include a "params_contraints" to enforce the setting, but then the burden is on every configuration that includes support for "E" to define the dimension of the GPR regfile. I don't think this is ideal, either. Might need some discussion.

spec/schemas/register_file_schema.json
```diff
@@ -0,0 +1,157 @@
...
+              "thread_pointer",
+              "frame_pointer",
+              "return_value",
+              "argument",
+              "caller_saved",
+              "callee_saved",
+              "temporary"
+            ]
+          },
+          "uniqueItems": true
```
ThinkOpenly, line 66: This doesn't hurt. I don't think it will reject multiple register definitions with the same index if they have different names (or same name and different ABI mnemonics), right? We may want to validate that externally, OR, do we even need to have an index attribute, given that this is an (ordered) array?

spec/schemas/schema_defs.json
```diff
@@ -43,6 +43,29 @@
...
+      "oneOf": [
+        {
+          "type": "integer",
+          "minimum": 1
+        },
+        {
+          "type": "string",
+          "minLength": 1
+        }
+      ]
```
ThinkOpenly, line 58: Do we need both? Is "integer" sufficient?

spec/std/isa/register_file/X.yaml
```diff
@@ -0,0 +1,217 @@
+# SPDX-License-Identifier: BSD-3-Clause-Clear
```
ThinkOpenly, line 2: Drop your own copyright statement here! :-)

spec/std/isa/register_file/X.yaml
```diff
@@ -0,0 +1,217 @@
...
+      when: { name: E }
+      description: |
+        The embedded base architecture (E) reduces the integer register file to 16 entries.
+registers:
+  - name: x0
+    index: 0
+    abi_mnemonic: zero
+    roles: [zero]
+    description: |
+      Register x0 is hardwired with all bits equal to 0.
```
ThinkOpenly, line 23: We need to add `sw_read()` and `sw_write()` methods to enforce this. Examples of the schema definitions for similar methods for CSRs can be found in `spec/schemas/csr_schema.json`.

spec/std/isa/register/gpr.yaml
```diff
@@ -0,0 +1,217 @@
...
+    roles: [temporary, caller_saved]
+    description: Temporary register.
+  - name: x7
+    index: 7
+    abi_mnemonic: t2
+    roles: [temporary, caller_saved]
+    description: Temporary register.
+  - name: x8
+    index: 8
+    abi_mnemonic: s0
```
ThinkOpenly: This can also be "fp". So, I suggest making this attribute an array, "abi_mnemonics". I wonder if each mnemonic needs its own "description".

spec/schemas/register_schema.json
```diff
@@ -0,0 +1,157 @@
...
+              "zero",
+              "return_address",
+              "stack_pointer",
+              "global_pointer",
+              "thread_pointer",
+              "frame_pointer",
+              "return_value",
+              "argument",
+              "caller_saved",
+              "callee_saved",
```
ThinkOpenly: I wonder if these should be attributes for every register_entry, with a default of "no"?

spec/std/isa/README.adoc
```diff
@@ -29,8 +29,8 @@ Three standard configs are present:
  * `rv64`: A configuration where only `MXLEN` is known to be 64, i.e., the RV64 ISA.
 
 The architecture is specified in a series of https://en.wikipedia.org/wiki/YAML[YAML]
-files for _Extensions_, _Instructions_, and _Control and Status Registers (CSRs)_.
-Each extension/instruction/CSR has its own file.
+files for _Extensions_, _Instructions_, _Registers_, and _Control and Status Registers (CSRs)_.
+Each extension/instruction/register/CSR has its own file.
```
ThinkOpenly: "RegFiles", "regfile"?
```suggestion
files for _Extensions_, _Instructions_, _RegFiles_, and _Control and Status Registers (CSRs)_.
Each extension/instruction/regfile/CSR has its own file.
```

spec/std/isa/README.adoc
```diff
@@ -142,6 +142,39 @@ H: # <1>
 
 === Instructions
 
+=== Registers
```
ThinkOpenly:
```suggestion
=== Register Files
```

spec/schemas/register_schema.json
```diff
@@ -0,0 +1,157 @@
...
+        },
+        "definedBy": {
+          "$ref": "schema_defs.json#/$defs/requires_entry"
+        },
+        "when": {
+          "$ref": "schema_defs.json#/$defs/requires_entry"
+        },
+        "length": {
+          "$ref": "schema_defs.json#/$defs/bit_length_value"
+        },
```
ThinkOpenly: I'm not sure we want to allow this. Having a register file where each entry can have its own length might make the code which uses them harder to write correctly. Maybe we remove this unless and until it's needed?

spec/std/isa/README.adoc
```diff
@@ -142,6 +142,39 @@ H: # <1>
...
+    abi_mnemonic: zero # <3>
+    roles: [zero]
+  - name: x1
+    index: 1
+    abi_mnemonic: ra
+    roles: [return_address]
+----
+
+<1> Register files are defined independently of the resolved XLEN; individual register entries
+    may override the default bit length when necessary.
```
ThinkOpenly: Is that first phrase true?

I suggest we remove that second phrase and make all registers in each register file the same bit width.

tools/ruby-gems/udb/lib/udb/obj/database_obj.rb
```diff
@@ -26,6 +26,7 @@ class Kind < T::Enum
       Csr = new("csr")
       CsrField = new("csr_field")
       Extension = new("extension")
+      RegisterFile = new("register file")
```
ThinkOpenly:
```suggestion
      RegisterFile = new("register_file")
```
?

tools/ruby-gems/udb/lib/udb/obj/register_file.rb
```diff
@@ -0,0 +1,80 @@
+# SPDX-License-Identifier: BSD-3-Clause-Clear
```
ThinkOpenly, line 2: Add your personal (or organizational) copyright here, too! :-)

ThinkOpenly (changes requested):
This is, frankly, pretty awesome! Great work, @AnimeshAgarwal28 !

There are a few comments that need discussion, but this is an impressive start!

spec/schemas/register_file_schema.json
AnimeshAgarwal28: Sure.

spec/schemas/schema_defs.json
AnimeshAgarwal28: I put string there to accept  XLEN as a valid length.

spec/std/isa/register_file/X.yaml
AnimeshAgarwal28: Sure :)

tools/ruby-gems/udb/lib/udb/obj/register_file.rb
AnimeshAgarwal28: Sure, will do.

spec/std/isa/README.adoc
AnimeshAgarwal28: I think a better way to convey the first phrase is to say: Register files use symbolic width values (XLEN) that are resolved during configuration.

As for the second phrase, yes you are right I shall drop it.

tools/ruby-gems/udb/lib/udb/obj/database_obj.rb
AnimeshAgarwal28: This was an overlook on my side, I shall correct this.

spec/schemas/register_schema.json
AnimeshAgarwal28: I don't see any problem in removing this.

spec/schemas/register_schema.json
AnimeshAgarwal28: Are you suggesting we should have them as boolean attributes and make both of them true/false for each register entry?

spec/std/isa/register/gpr.yaml
AnimeshAgarwal28: Ahh thanks I missed that, I was following the risc-v abi document which does not specify x8 as fp but The RISC-V Instruction Set Manual(ratified) does. 

Since there is only one register with two ABI mnemonics we can either describe both the mnemonics in the description of the register entry itself, or do as you suggested.

spec/schemas/register_schema.json
ThinkOpenly: yes, with default values of "false".

spec/schemas/register_schema.json
dhower-qc: Yea, I think this is the trickiest part; namely, where to put the condition(s).

At least two "generic" options that would probably be future-proofed:

1. Condition the entire "meat" of the definition, so you would basically define everything twice (or more) when variants are needed.
2. Condition each register separately. I see you already have this ("when" in the register definition).

I could go either way.

spec/schemas/register_schema.json
```diff
@@ -0,0 +1,157 @@
...
+          "$ref": "schema_defs.json#/$defs/requires_entry"
+        },
+        "description": {
+          "$ref": "schema_defs.json#/$defs/spec_text"
+        }
+      }
+    },
+    "register_entry": {
+      "type": "object",
+      "required": ["name", "index"],
```
dhower-qc: Something to think about:

If we condition the entire "meat", the index key may not be needed since it could be inferred from the register array.

spec/schemas/register_schema.json
```diff
@@ -0,0 +1,157 @@
...
+      "additionalProperties": false,
+      "properties": {
+        "name": {
+          "$ref": "schema_defs.json#/$defs/register_name"
+        },
+        "abi_mnemonic": {
```
dhower-qc: @lenary, are you aware of any cases where a register might have different abi names under different conditions?

spec/schemas/register_schema.json
```diff
@@ -0,0 +1,157 @@
...
+          "description": "Register index"
+        },
+        "description": {
+          "$ref": "schema_defs.json#/$defs/spec_text"
+        },
+        "definedBy": {
+          "$ref": "schema_defs.json#/$defs/requires_entry"
+        },
+        "when": {
+          "$ref": "schema_defs.json#/$defs/requires_entry"
```
dhower-qc: I don't think we'll need both definedBy and when; with #891, they would both be generic conditions

spec/schemas/register_schema.json
dhower-qc: that would be a new thing for us; we don't generally use defaults in the existing schemas.

spec/schemas/register_file_schema.json
dhower-qc: That's right. Duplicate indices would need to be validated outside JSON schema, especially since it might be valid to have entries with the same index before conditions are applied.

spec/schemas/register_schema.json
```diff
@@ -0,0 +1,157 @@
...
+          "$ref": "schema_defs.json#/$defs/spec_text"
+        },
+        "definedBy": {
+          "$ref": "schema_defs.json#/$defs/requires_entry"
+        },
+        "register_class": {
+          "type": "string",
+          "enum": ["general_purpose", "floating_point", "vector"]
+        },
+        "length": {
```
dhower-qc: "length" could mean register width or the number of entries in the register file.

Given you have "count" below, I'm guessing it is the former. Maybe register_length?

spec/schemas/register_schema.json
```diff
@@ -0,0 +1,157 @@
...
+          "$ref": "schema_defs.json#/$defs/bit_length_value"
+        },
+        "count": {
```
dhower-qc: Might not be needed if we condition the entire "meat".

spec/schemas/register_schema.json
```diff
@@ -0,0 +1,157 @@
...
+          }
+        },
+        "$source": {
+          "type": "string",
+          "format": "uri-reference"
+        }
+      }
+    }
+  },
+  "$ref": "#/$defs/register_file"
```
dhower-qc: I know you probably copied this from something I did, but I've since learned that our JSON schema linter flags this as bad form.

Instead of "$ref"ing the register file, the register file should just be at the top level of the schema.

spec/schemas/schema_defs.json
dhower-qc: Good thinking. We should tighten it up, though: instead of string, make it an enum: ["MXLEN"]. (In UDB "XLEN" is considered dynamic -- might change at runtime -- while MXLEN is the immutable maximum XLEN of a configuration)

spec/std/isa/register/gpr.yaml
```diff
@@ -0,0 +1,217 @@
...
+
+# yaml-language-server: $schema=../../../schemas/register_schema.json
+
+$schema: register_schema.json#
+kind: register_file
+name: X
+long_name: General Purpose Registers
+description: |
+  The 'X' register file contains the general-purpose integer registers accessible to
+  scalar instructions. Register widths track the current XLEN. Register 'x0' is hardwired
```
dhower-qc: Not just scalar instructions. For example, vector instruction also use X registers.

"track the current XLEN" is sort of correct but ambiguous. There are actually rules about what happens to the un-viewable bits when XLEN < MXLEN.

spec/std/isa/register_file/X.yaml
```diff
@@ -0,0 +1,217 @@
...
+$schema: register_schema.json#
+kind: register_file
+name: X
+long_name: General Purpose Registers
+description: |
+  The 'X' register file contains the general-purpose integer registers accessible to
+  scalar instructions. Register widths track the current XLEN. Register 'x0' is hardwired
+  to zero, while the remaining registers hold architecturally visible integer state used
+  by the calling convention and execution environment.
+definedBy: I
```
dhower-qc, line 15: After #891, this will be:

```yaml
definedBy:
  extension:
    name: I
```

Or, if you want to add E too:

```yaml
definedBy:
  extension:
    oneOf:
      - name: I
      - name: E
```

spec/std/isa/register/gpr.yaml
```diff
@@ -0,0 +1,217 @@
...
+definedBy: I
+register_class: general_purpose
+length: XLEN
+count:
+  default: 32
+  variants:
+    - value: 16
+      when: { name: E }
+      description: |
+        The embedded base architecture (E) reduces the integer register file to 16 entries.
```
dhower-qc: After #891, this would look more like:

```yaml
count:
  oneOf:
    -  value: 32
       when:
         extension: { name: I }
    -  value: 16
       when:
         extension: { name: E }
```

spec/std/isa/register_file/X.yaml
dhower-qc: ...and we need to think about whether we want sw_read/write to be per-register or per-register file.

I could see either working fine, but we should make a decision either way.

spec/std/isa/register_file/X.yaml
```diff
@@ -0,0 +1,217 @@
...
+    index: 0
+    abi_mnemonic: zero
+    roles: [zero]
+    description: |
+      Register x0 is hardwired with all bits equal to 0.
+  - name: x1
+    index: 1
+    abi_mnemonic: ra
+    roles: [return_address, caller_saved]
+    description: Return address register used by CALL/JAL instructions.
```
dhower-qc, line 32: very nice.

spec/std/isa/register/gpr.yaml
dhower-qc: ...and there is the answer to my question earlier. ;)

spec/std/isa/register_file/X.yaml
```diff
@@ -0,0 +1,217 @@
...
+    abi_mnemonic: a5
+    roles: [argument, caller_saved]
+    description: Function argument register 5.
+  - name: x16
+    index: 16
+    abi_mnemonic: a6
+    roles: [argument, caller_saved]
+    description: Function argument register 6.
+    when:
+      not: { name: E }
```
dhower-qc, line 106: After #891, this would be:

```yaml
  when:
    not:
      extension:
        name: E
```

spec/std/isa/README.adoc
dhower-qc: "Register files are defined independently of the resolved XLEN"

That's murky. I would just be more explicit: registers can have either a fixed architecture width or take their width from a parameter (e.g., MXLEN, VLEN).

tools/ruby-gems/udb/lib/udb/obj/register_file.rb
```diff
@@ -0,0 +1,80 @@
...
+module Udb
+
+class RegisterFile < TopLevelDatabaseObject
+  extend T::Sig
+
+  class RegisterEntry
+    extend T::Sig
+
+    sig { returns(T::Hash[String, T.untyped]) }
+    attr_reader :raw
```
dhower-qc: For consistency, let's name this `#data`, which will correspond to the same thing you get from TopLevelDatabaseObject

tools/ruby-gems/udb/lib/udb/obj/register_file.rb
```diff
@@ -0,0 +1,80 @@
...
+    def abi_mnemonic = @raw["abi_mnemonic"]
+
+    sig { returns(T.any(Integer, String)) }
+    def index = @raw.fetch("index")
+
+    sig { returns(T.nilable(String)) }
+    def description = @raw["description"]
+
+    sig { returns(T::Array[String]) }
+    def roles = @raw.fetch("roles", [])
```
dhower-qc: It's more work, but also better checked if you return a Sorbet enum here.

spec/schemas/register_schema.json
```diff
@@ -0,0 +1,157 @@
...
+          "format": "uri-reference",
+          "const": "register_schema.json#",
+          "description": "Path to schema, relative to <UDB ROOT>/schemas"
+        },
+        "kind": {
+          "type": "string",
+          "const": "register_file"
+        },
+        "name": {
+          "$ref": "schema_defs.json#/$defs/register_name"
```
dhower-qc: register_file_name, which should be different than register_name.

So far, all register file names are a single uppercase letter.

dhower-qc (changes requested):
Excellent first draft!

In addition to the inline comments, a few high-level thoughts:

To really get this integrated, we'll want to reflect it in a few places where the X registers are built in concepts. A few that come to mind: the IDL symbol table and the generated ISS.

In the symbol table, you'll find this:

https://github.com/riscv-software-src/riscv-unified-db/blob/5d6c397acee0e2ff4e22fcf1958b562437117b4d/tools/ruby-gems/idlc/lib/idlc/symbol_table.rb#L221-L226

That's where "X" gets added in global scope as an array of Bits<MXLEN>. Instead of that, we'll want to construct (all of) the register files based on the YAML. Hopefully you can get some ideas of to do so looking around here where the symbol table is instantiated:

 https://github.com/riscv-software-src/riscv-unified-db/blob/5d6c397acee0e2ff4e22fcf1958b562437117b4d/tools/ruby-gems/udb/lib/udb/cfg_arch.rb#L462-L488

In the ISS, the X registers are defined on a "hart" object, as you can see here:

https://github.com/riscv-software-src/riscv-unified-db/blob/5d6c397acee0e2ff4e22fcf1958b562437117b4d/backends/cpp_hart_gen/templates/hart.hxx.erb#L378-L415

https://github.com/riscv-software-src/riscv-unified-db/blob/5d6c397acee0e2ff4e22fcf1958b562437117b4d/backends/cpp_hart_gen/templates/hart.hxx.erb#L529-L534

We'll want to think this one through a bit more with @henrikg-qc

spec/schemas/register_file_schema.json
AnimeshAgarwal28: Now when I think of it, index is not needed even if it is not an "ordered" array because we always have the register name as the "unique key".

spec/schemas/register_schema.json
AnimeshAgarwal28: This is true, name is already enough of a unique key.

spec/schemas/register_schema.json
AnimeshAgarwal28: You are right, it is the the register width. I'll change it.

spec/schemas/register_schema.json
ThinkOpenly: > So far, all register file names are a single uppercase letter.

Is that true? We only have two so far, right? `X` and `f`. From `fadd.s.yaml`, for example:
```
f[fd] = f32_add(f[fs1], f[fs2], mode);
```

I guess we're expecting the vector register file to be `V`, although I don't know if that exists yet.

spec/std/isa/register/gpr.yaml
AnimeshAgarwal28: Should I make this change now? Or, wait till #891 gets merged.

spec/std/isa/register_file/X.yaml
ThinkOpenly, line 1: Let's put this in the directory `spec/std/isa/register_file`.

spec/std/isa/register/gpr.yaml
```diff
@@ -0,0 +1,201 @@
...
+
+# yaml-language-server: $schema=../../../schemas/register_file_schema.json
+
+$schema: register_file_schema.json#
+kind: register_file
+name: X
+long_name: General Purpose Registers
+description: |
+  The 'X' register file contains the general-purpose integer registers. Each register
+  is architecturally XLEN bits wide, unless XLEN<MXLEN. Register 'x0' is hardwired to
```
ThinkOpenly: This says "A unless B", but doesn't say what "B" implies.
Should we say "Each register is architecturally MXLEN bits wide"?

spec/std/isa/register/gpr.yaml
```diff
@@ -0,0 +1,201 @@
...
+registers:
+  - name: x0
+    abi_mnemonics: [zero]
+    roles: [zero]
+    description: |
+      Register x0 is hardwired with all bits equal to 0.
+    sw_read(): |
+      return 0;
+    sw_write(value): |
+      # x0 ignores all writes
```
ThinkOpenly: `sw_write()` is expected to return the value to be written, so this needs a `return 0`.

Alternatively, we could add a `writable` attribute as is done for CSRs, so you wouldn't need a `sw_write()` method for read-only registers.

spec/std/isa/README.adoc
```diff
@@ -142,6 +142,40 @@ H: # <1>
...
+      # x0 ignores all writes
+  - name: x1
+    abi_mnemonics: [ra]
+    roles: [return_address]
+  - name: x8
+    abi_mnemonics: [s0, fp] # <5>
+    roles: [callee_saved, frame_pointer]
+----
+
+<1> Registers can have either a fixed architecture width or take their width from a parameter (e.g., MXLEN, VLEN).
```
ThinkOpenly: I propose expanding this a tad, just to help make it more obvious that every register in a register file is the same width:
> The registers in a register file have identical fixed widths, either an explicit width or from a parameter (e.g. MXLEN or VLEN).

... or something like that.

spec/schemas/register_schema.json
ThinkOpenly: We now have `definedBy` for the register files, and `when` for the registers. Either way, this will need to be revisited when #891 gets merged. I'm going to resolve this comment for now, since we no longer have `definedBy` and `when` for any one entity.

spec/schemas/register_schema.json
ThinkOpenly: I think this is resolved, but please confirm.

spec/std/isa/register/gpr.yaml
ThinkOpenly: That is a good question. I think both are close to being ready. Let's get this one ready to merge without the #891-related changes and then see where #891 is.

tools/ruby-gems/udb/lib/udb/obj/register_file.rb
ThinkOpenly: This might be resolved? Please confirm.

tools/ruby-gems/udb/lib/udb/obj/register_file.rb
ThinkOpenly: I think this is resolved. Please confirm.

spec/schemas/schema_defs.json
```diff
@@ -43,6 +43,35 @@
...
+    },
+    "register_name": {
+      "type": "string",
+      "pattern": "^[A-Za-z][A-Za-z0-9_.-]*$",
+      "description": "Register name"
+    },
+    "register_file_name": {
+      "type": "string",
+      "pattern": "^[A-Za-z][A-Za-z0-9_.-]*$",
```
ThinkOpenly: Let's restrict this to a single character. Since we already have `X` and `f`, either case:

```suggestion
      "pattern": "^[A-Za-z]$",
```

tools/ruby-gems/udb/lib/udb/obj/register_file.rb
AnimeshAgarwal28: Yes this is resolved, I'll mark this as resolved

tools/ruby-gems/udb/lib/udb/obj/register_file.rb
AnimeshAgarwal28: Yes, this is resloved. Marking this as resolved.

spec/std/isa/register/gpr.yaml
AnimeshAgarwal28: I prefer the alternative approach of adding a writable attribute, makes it explicit.

spec/std/isa/register/gpr.yaml
ThinkOpenly: Let's go with "Each register is MXLEN bits wide".

spec/std/isa/register_file/X.yaml
```diff
@@ -1,30 +1,29 @@
...
 $schema: register_file_schema.json#
 kind: register_file
 name: X
 long_name: General Purpose Registers
 description: |
   The 'X' register file contains the general-purpose integer registers. Each register
-  is architecturally XLEN bits wide, unless XLEN<MXLEN. Register 'x0' is hardwired to
-  zero, while the remaining registers hold architecturally visible integer state used
-  by the calling convention and execution environment.
+  is MXLEN bits wide, unless XLEN<MXLEN. Register 'x0' is hardwired to zero, while
```
ThinkOpenly:
```suggestion
  is MXLEN bits wide, but will appear to be XLEN bits wide if XLEN<MXLEN. Register 'x0' is hardwired to zero, while
```

spec/schemas/schema_defs.json
dhower-qc: Agree. And perhaps to take it a step further: make them all single letter uppercase? Right now, we have the unfortunate situation that X is uppercase but f and v are lowercase. That's out of necessity since we declare f and v as generic globals right now, and uppercase variable names are const by definition in IDL. X is special since it's builtin.

By having dedicated register file definitions, we can avoid the discrepancy.

spec/std/isa/register_file/X.yaml
```diff
@@ -0,0 +1,200 @@
...
+    roles: [global_pointer]
+    description: Global pointer.
+  - name: x4
+    abi_mnemonics: [tp]
+    roles: [thread_pointer]
+    description: Thread pointer or TLS base.
+  - name: x5
+    abi_mnemonics: [t0]
+    caller_saved: true
+    roles: [temporary]
```
dhower-qc: Isn't x5 sometimes a return address?

ThinkOpenly: The only reference I can find relates to the use of `x5` with shadow stack operations. Is that what you are referring to?

spec/schemas/schema_defs.json
ThinkOpenly: So, further constrain this pattern to single-character upper case, and we'll need to change the current `f` and `v` register files to `F` and `V` in the YAML and in the Ruby code?

spec/std/isa/register_file/X.yaml
lenary: No. X5 is also called the "alternate link register", and is used for calling some functions where they don't want to trash `ra` (millicode, and outlining).

lenary: I guess part of the reason this is an alternate link register is because it is a temporary register, and functions that use `x5` this way know it when they are constructed.

X5 does have specific support in Zicfiss for being used as a return address.

ThinkOpenly: > X5 is also called the "alternate link register"

Called by whom? I can't find it in the RISC-V APIs Specification (Version 1.0: Ratified), nor in the RISC-V Assembly Programmer's Manual (2025-02-05).

I'm just wondering if this is an ad-hoc classification, since it doesn't seem to be officially documented as anything other than a temporary register.

spec/schemas/schema_defs.json
ThinkOpenly: @AnimeshAgarwal28, could you make the change so the pattern only matches "single upper-case character"?

AnimeshAgarwal28: Sure

AnimeshAgarwal28: Done

spec/schemas/register_file_schema.json
```diff
@@ -0,0 +1,130 @@
...
+        },
+        "writable": {
+          "type": "boolean",
+          "default": true,
+          "description": "Whether the register is writable. If false, no sw_write() method is needed."
+        }
```
ThinkOpenly: I'm second-guessing this. `x0` is technically "writable" in that you can issue a store to it without raising an exception, but the store is ignored.

Let's go with `sw_read()`/`sw_write()` and remove this attribute.

spec/schemas/register_file_schema.json
```diff
@@ -0,0 +1,130 @@
...
+        },
+        "description": {
+          "$ref": "schema_defs.json#/$defs/spec_text"
+        },
+        "when": {
+          "$ref": "schema_defs.json#/$defs/requires_entry"
+        },
+        "sw_read()": {
+          "type": "string",
+          "description": "Function that returns the value of the register when read by software. Use this to define special behavior for registers (e.g., x0 always reads as zero)."
```
ThinkOpenly: In a recent ad-hoc review in a Unified DB SIG meeting, we preferred that these `sw_read()` and `sw_write()` methods be renamed to something akin to "architectural read/write", so maybe `arch_read()` and `arch_write()`. This distinguishes configuration-dependent behavior from architecturally defined behavior. Please update the description to match.

The implementations can be kept the same, which is basically all for the case of `x0` to always be write-any-read-0.

spec/schemas/register_file_schema.json
```diff
@@ -0,0 +1,130 @@
...
+        "name",
+        "long_name",
+        "description",
+        "register_length",
+        "registers"
+      ],
+      "additionalProperties": false,
+      "properties": {
+        "$schema": {
+          "type": "string",
```
ThinkOpenly: When "const" is present, "type" is unnecessary. (Per #1015)

spec/schemas/register_file_schema.json
```diff
@@ -0,0 +1,130 @@
...
+          "format": "uri-reference",
+          "const": "register_file_schema.json#",
+          "description": "Path to schema, relative to <UDB ROOT>/schemas"
+        },
+        "kind": {
+          "type": "string",
```
ThinkOpenly: "type" unnecessary here as well.

spec/std/isa/register_file/X.yaml
jordancarlin: It's in a non-normative note in section 2.1 of the ISA manual:
> There is no dedicated stack pointer or subroutine return address link register in the Base Integer ISA; the instruction encoding allows any x register to be used for these purposes. However, the standard software calling convention uses register x1 to hold the return address for a call, with register x5 available as an alternate link register. The standard calling convention uses register x2 as the stack pointer.
>
> Hardware might choose to accelerate function calls and returns that use x1 or x5. See the descriptions of the JAL and JALR instructions.

ThinkOpenly: Thanks for that reference, @jordancarlin.

Let's add "alternate link register" as a role here. Are you willing and able, @AnimeshAgarwal28?

AnimeshAgarwal28: Just returned from an amazing vacation :). I will do this for sure in some time.

tools/ruby-gems/udb/lib/udb/obj/register_file.rb
```diff
@@ -0,0 +1,118 @@
+# Copyright (c) Animesh Agarwal
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# typed: true
+# frozen_string_literal: true
+
+require_relative "database_obj"
+require_relative "../req_expression"
```
ThinkOpenly:
```suggestion
require_relative "../condition"
```

tools/ruby-gems/udb/lib/udb/obj/register_file.rb
```diff
@@ -0,0 +1,118 @@
...
+      @data["arch_write(value)"] || @data["sw_write(value)"]
+    end
+
+    sig { returns(T.nilable(String)) }
+    def sw_read = arch_read
+
+    sig { returns(T.nilable(String)) }
+    def sw_write = arch_write
+
+    sig { returns(T.nilable(ExtensionRequirementExpression)) }
```
ThinkOpenly:
```suggestion
    sig { returns(T.nilable(AbstractCondition)) }
```

tools/ruby-gems/udb/lib/udb/obj/register_file.rb
```diff
@@ -0,0 +1,118 @@
...
+      return nil unless @data.key?("definedBy")
+
+      @defined_by_condition ||= ExtensionRequirementExpression.new(@data.fetch("definedBy"), @file.arch)
+    end
+
+    sig { returns(T.nilable(ExtensionRequirementExpression)) }
+    def when_condition
+      return nil unless @data.key?("when")
+
+      @when_condition ||= ExtensionRequirementExpression.new(@data.fetch("when"), @file.arch)
```
ThinkOpenly:
```suggestion
      @when_condition ||= Condition.new(@data.fetch("when"), @file.arch)
```

tools/ruby-gems/udb/lib/udb/obj/register_file.rb
```diff
@@ -0,0 +1,118 @@
...
+    def sw_read = arch_read
+
+    sig { returns(T.nilable(String)) }
+    def sw_write = arch_write
+
+    sig { returns(T.nilable(ExtensionRequirementExpression)) }
+    def defined_by_condition
+      return nil unless @data.key?("definedBy")
+
+      @defined_by_condition ||= ExtensionRequirementExpression.new(@data.fetch("definedBy"), @file.arch)
```
ThinkOpenly:
```suggestion
      @defined_by_condition ||= Condition.new(@data.fetch("definedBy"), @file.arch)
```

tools/ruby-gems/udb/lib/udb/obj/register_file.rb
```diff
@@ -0,0 +1,118 @@
...
+    end
+
+    sig { returns(T.nilable(ExtensionRequirementExpression)) }
```
ThinkOpenly:
```suggestion
    sig { returns(T.nilable(Condition)) }
```

ThinkOpenly (changes requested):
@AnimeshAgarwal28, if you are willing and able to incorporate these suggested changes needed after the changes in #891, I think we can get this through CI and merged. If not, let me know and I'll create a new branch and cherry-pick from this branch and try to move it forward that way. Thanks!

spec/std/isa/register_file/X.yaml
ThinkOpenly: I don't think these changes are required.

The `register_file_schema` has:
```
        "definedBy": {
          "$ref": "schema_defs.json#/$defs/requires_entry"
        },
```
and `requires_entry` is:
```
    "requires_entry": {
      "oneOf": [
        { "$ref": "#/$defs/extension_requirement" },
```
and `extension_requirement`:
```
    "extension_requirement": {
      "description": "A requirement on an extension. Can either specify just an extension name, in which case version '>= 0' is implied, or both a name and a requirement",
      "type": "object",
      "properties": {
        "name": {
          "$ref": "#/$defs/extension_name"
        },
        "version": {
          "$ref": "#/$defs/version_requirements"
        }
      },
      "required": ["name"],
      "additionalProperties": false
    },
```
So, just a name works here (and CI is passing). I'm going to close this. If we do need schema changes, post here, or open an issue.

spec/std/isa/register/gpr.yaml
ThinkOpenly: This PR abandoned `count` in favor of listing the registers as an array, so the "count" can be easily determined and need not be specified. So, I'm resolving this comment thread.

spec/std/isa/register_file/X.yaml
ThinkOpenly: See https://github.com/riscv/riscv-unified-db/pull/1150/changes#r2402109626. Marking resolved.

Issue #513, opened 2025-03-19
Add S{m,s}ctr extension IDL


dhower-qc:
The recently ratified (Nov24) extension S{m,s}ctr Control Transfer Records needs IDL:

registers (m, s, vs)
1x new system instruction (SCTRCLR)

Zain2050:
@ThinkOpenly please assign this issue to me.

Zain2050:
I want to work on its IDL. When will PR #554 get merged. Should I create new yaml files for smctr/ssctr (to define config params) & sctrclr (to describe it in IDL) and leave out rest of the fields as they will be pushed by that PR?

ThinkOpenly:
> When will PR [#554](https://github.com/riscv-software-src/riscv-unified-db/pull/554) get merged.

Looking at it now, I think it just needs a rebase.

> Should I create new yaml files for smctr/ssctr (to define config params) & sctrclr (to describe it in IDL) and leave out rest of the fields as they will be pushed by that PR?

If you wanted to build on top of this PR, base your branch on the branch for this PR.

PR #1164, opened 2025-10-03
fix(IDL): implement Smctr/Ssctr in IDL

Zain2050:
Related to Issue #513. 
This PR is based upon PR #554. Can be merged after 554 is merged.

**New Changes**:
params: CTR_CYCLE_COUNTER and CCV (Smctr.yaml)
IDL: (Sctrclr.yaml)


arch/ext/Smctr.yaml
```diff
@@ -39,6 +39,25 @@ description: |
...
+  CTR_CYCLE_COUNTER:
+    description: |
+      The elapsed cycle counter includes a count of CPU cycles elapsed since the prior CTR record.
+      It is represented by the CC field, which has a 12-bit mantissa component
+      (Cycle Count Mantissa, or CCM) and a 4-bit exponent component (Cycle Count Exponent, or CCE).
+      It increments at the same rate as the mcycle counter.
+      The CtrCycleCounter is reset on writes to xctrctl , and on execution of SCTRCLR, to ensure that any
+      accumulated cycle counts do not persist across a context switch.
+    schema:
+      type: integer
```
ThinkOpenly: Is this to represent the optionality of the the CC field in the `ctrdata` CSR? If so, this seems like this should be a boolean?

Also, it's not clear that this is what the parameter is for, so maybe include a bit more about the purpose of the parameter:
> The ctrdata register may optionally include a count of CPU cycles elapsed since the prior CTR record.

arch/ext/Smctr.yaml
```diff
@@ -39,6 +39,25 @@ description: |
...
+  CCV:
```
ThinkOpenly: Similarly, what is this parameter used for? Is it the "should" in the following description, as to whether that "should" is respected or not?

arch/inst/Smdbltrp/sctrclr.yaml
```diff
@@ -0,0 +1,80 @@
+# yaml-language-server: $schema=../../../schemas/inst_schema.json
+
+$schema: inst_schema.json#
+kind: instruction
+name: sctrclr
+long_name: Supervisor Control Transfer Record (CTR) clear
+description: |
+  - id: inst-sctrclr-operation
+    normative: false
+    text: |
```
ThinkOpenly: We have abandoned this text tagging, so these lines can be deleted. Similarly elsewhere.

ThinkOpenly (changes requested):
In addition to the inline comments, the following needs associated parameters:
> An implementation that supports cycle counting must implement CCV and all CCM bits, but may implement 0..4 exponent bits in CCE. Unimplemented CCE bits are read-only 0. For implementations that
support transfer type filtering, it is recommended to implement at least 3 exponent bits. This allows capturing the full latency of most functions, when recording only calls and returns.

arch/ext/Smctr.yaml
Zain2050: This is a bit vague in the spec. From what I've understood. CtrCycleCounter and ctrdata.CC are different. CtrCycleCounter is an independant counter whose value is stored in ctrdata in case of a control transfer. From 
 section 11.5.3
> The CC field is encoded such that CCE holds 0 if the CtrCycleCounter value is less than 4096, otherwise it holds the index of the most significant one bit in the CtrCycleCounter value, minus 11.

ctrdata for each entry is just a copy of CtrCycleCounter at the time of control transfer. Also, 

> The SCTRCLR instruction performs the following operations:
> Zeroes all CTR Entry Registers, for all DEPTH values
> Zeroes the CTR cycle counter and CCV

Here, zeroing out entry registers meaning zeroing out ctrdata as well. But it explicitly states to zero out CTR cycle counter and CCV.
As it's a counter therefore, it's type would be integer, right? 
I'll improve the description of the param as it seems a bit unclear.

arch/inst/Smdbltrp/sctrclr.yaml
Zain2050: It's the changes from PR 554. I'll remove these.

arch/ext/Smctr.yaml
Zain2050: This represents the value that will be store in the CCV field of ctrdata when the next control transfer occurs. I have updated its description.

arch/ext/Smctr.yaml
ThinkOpenly: I'm not following, sorry. The purpose of parameters is for configurations to resolve optionality in the spec. (For example, when the spec says "should", "may", or recommends a behavior, but does not mandate it with "must".) What optionality is being resolved with this parameter?

arch/ext/Smctr.yaml
ThinkOpenly: I'm not following here, either. Are you trying to parameterize the "should" in
> This flag should additionally be cleared after any other implementation-specific scenarios where active cycles might not be counted in CTR_CYCLE_COUNTER.

?

And, it's defined as boolean, so what would be the value to be stored? Regardless, I'm not sure this really needs to be parameterized, but I want to understand your thinking first.

arch/ext/Smctr.yaml
Zain2050: I guess I might be using parameters the wrong way. 
Spec states that STRCLR zeros out the CTR cycle counter.

> The SCTRCLR instruction performs the following operations:
> Zeroes all CTR Entry Registers, for all DEPTH values
> Zeroes the CTR cycle counter and CCV

 Therefore, I used params to represent the hardware counter and CCV flag. What would be the correct way to represent the counter so that I could zero it out in IDL?

ThinkOpenly: OK, so "CCV" is a field in the `ctrdata` register. This register is not yet defined in UDB, but needs to be for this work to proceed. It would probably be defined much like other CSRs, but it might be a little weird in that I think it can only be accessed indirectly (`siselect` + `sireg3`). And there isn't a single set of CTR registers, it's a buffer of up to 256 records that can be accessed by being mapped to indirect register selections 0x200-0x2ff. And, how many records are actually supported is implementation-dependent (needs a configuration parameter) -- see `sctrdepth`.

So, it's a little complicated. #554 is a start on defining the extension, but isn't merged, and doesn't have the parameters defined.

ThinkOpenly: ...reading...
I read the spec as saying CtrCycleCounter _is_ the value represented by CCE (exponent) + CCM (mantissa).

Zain2050: This is a bit confusing in the spec. Sctrclr clears out ctrdata therefore clearing out CC and CCV. But also explicitly states that it will zero out ctrcyclecounter and CCV. This indicates that it's talking about hardware counter and its associated CCV flag, not the csr fields as they are automatically cleared if entry registers are cleared. 
What I understood is that it will reset the hardware counter and as the counter has been reset the next entry will have CCV zero. I am fully aware that CCV is a csr field. But perhaps the spec is talking about CCV in sense of a hardware bit, such that this bit will be what's stored in the next entry register (Therefore say to zero out CCV and the next entry will have CCV bit in ctrdata equal to 0, indicating the counter was reset and the CC is not correctly representing the cycle count).
From this line of the spec, it seems that the CC field in ctrdata is encoded depending on the value in a hardware counter.

> The CC field is encoded such that CCE holds 0 if the CtrCycleCounter value is less than 4096, otherwise it holds the index of the most significant one bit in the CtrCycleCounter value, minus 11.

Why do you think that the spec mentioned to clear out counter and CCV again if clearing out entry registers (including ctrdata) already cleared those fields?

> The SCTRCLR instruction performs the following operations:
> Zeroes all CTR Entry Registers, for all DEPTH values
> Zeroes the CTR cycle counter and CCV

ThinkOpenly: I think I understand better now. It does require some careful reading. This mysterious "CtrCycleCounter" is arguably a CSR akin to `mcycle`, but it's never explicitly defined that way. It certainly has a "value", and a (minimum?) width in bits (relative to bits in CCE definition).

We should probably go ahead and define it as a CSR, but we might also want to clarify what it is architecturally (mostly likely via tech-privileged@lists.riscv.org).

Zain2050: Great! 
Should I just mail at tech-privileged@lists.riscv.org? Or am I supposed to ask it in their weekly meeting (if there is any)?

ThinkOpenly: I wouldn't wait for a meeting. If you are able to ask on the mailing list, please do. (You may have to be a RISC-V member and join that group. Let me know if I can help.)

Zain2050: Unfortunately, I'm not a RISC-V member :( 
Can you ask it?

ThinkOpenly: https://lists.riscv.org/g/tech-privileged/message/2679

ThinkOpenly: The response was essentially that CtrCycleCounter is not directly visible machine state, only indirectly through `ctrdata` register(s). So, I think this needs to be implemented using a callout to a builtin function, akin to the use of `read_mcycle()` in `cycle.yaml`.

Zain2050: Okay. What about clearing out the entry registers? We don't have direct access to those registers. Should I use a builtin for that as well?

ThinkOpenly: At this point, I don't think it's terribly important whether it is represented as a set of (indirect) actual CSRs, or as a set of indirect CSRs that have `sw_write()` and `sw_read()` methods that access some builtin memory buffer. The former is pretty straight-forward. The latter would require defining a builtin accessor function.

Zain2050: So, my current approach makes sense, right?

ThinkOpenly: If you are talking about configuration parameters, then no. :-)

Flipping through the extension documentation, I see at least these as needed parameters:
- what DEPTH values are supported
- whether MISP is supported
- the optionality of each ctrdata field
- number of supported CCE bits

For registers, you'll need to use indirect access as noted previously.

For CtrCycleCounter, you should implement as a call to a builtin function.

For CCV, each of those bits is contained in each Control Transfer Record, shadowed by the `ctrdata` register, accessed indirectly through `sireg3` (when `siselect` = buffer offset/register index), so if you're already setting the register to zero, you have also set CCV to zero.

I hope that makes sense. :-)

arch/inst/Smdbltrp/sctrclr.yaml
```diff
@@ -0,0 +1,69 @@
...
+  else if (implemented?(ExtensionName::H) && mode() == PrivilegeMode::VU) {
+      raise (ExceptionCode::VirtualInstruction, mode(), $encoding);
+  }
+  else {
+    for (U32 i = 0; i < (16 << CSR[sctrdepth].DEPTH); i++) {
+      CSR[siselect] = (0x200 + i);
+      CSR[sireg1] = 0;
+      CSR[sireg2] = 0;
+      CSR[sireg3] = 0;
+    }
```
ThinkOpenly: Unfortunately, there is no `sireg1`... it's `sireg`. :-/

While this should accomplish the goal of clearing the CTR buffer, it has side-effect of changing the value of `siselect`. This should probably be saved and restored.

Now I'm wondering, though, if we also need to be concerned with changing the value of `siselect` at all here, because it would be an observable change during a trap, for example. The only way I can think to avoid that issue would be to bypass the indirect register usage and only manipulate the internal state directly.

So, let's define all of the CTR registers as CSRs. This is in line with how they are described in the spec under "CSR LIsting":

Image
Image

You can use a "layout" to generate the 256 definitions.

Then, you can manipulate them directly without needing to change the value of `siselect`.

This is a nice case in support of https://github.com/riscv-software-src/riscv-unified-db/issues/1169, but we don't have a means to treat these individual registers as an array. You could also use a layout to initialize/reset them all.

ThinkOpenly:
@Zain2050, #554 has (finally) been merged (via #1238). Are you interested in restarting the work here?

Zain2050:
Hi @ThinkOpenly, you suggested to define the CTR entry registers (0x200 - 0x2FF) as CSRs. However, these addresses overlap with other CSRs. For example indirect CSR ctr*0 has an address of 0x200 which is same as the address of vsstatus. Do we need to define a new kind?
```
kind: indirect-csr
```
Also, writing 0x200 in siselect gives us access to 3 CSRs ctrsource, ctrtarget & ctrdata. Will we be using the same address (i.e. 0x200) for these 3 CSRs? 

```
kind: csr
name: ctrsource0
long_name: Control Transfer Records Source 0
address: 0x200
```
```
kind: csr
name: ctrtarget0
long_name: Control Transfer Records Target 0
address: 0x200
```

ThinkOpenly:
So sorry it took so long to respond!

> you suggested to define the CTR entry registers (0x200 - 0x2FF) as CSRs. However, these addresses overlap with other CSRs. For example indirect CSR ctr*0 has an address of 0x200 which is same as the address of vsstatus. Do we need to define a new kind?

The direct addresses and indirect addresses are independent. Some CSRs have _both_. `hpmcounter3`, for example, has direct address 0xC03 and indirect address 0x43 (register `sireg`).

(I'm wondering whether the indirect addresses should be considered a single number space or divided into multiple spaces for `miselect`, `siselect`, and `vsiselect`. As defined, they are identical with some addresses left undefined in some selection contexts. I think proceeding with a single "indirect" space is fine.)

Given that other CSRs have _both_ direct and indirect addresses (e.g. `hpmcounter3`), I think we just need a new field in the CSR schema, maybe "address_indirect"?

Currently, the schema requires "address". It will need to redefine the current "address" and the new "address_indirect" as "at least one of" required. I found an example (unverified) at https://stackoverflow.com/questions/31839578/how-to-define-a-json-schema-that-requires-at-least-one-of-many-properties, although AI can probably figure it out, too.

> Also, writing 0x200 in siselect gives us access to 3 CSRs ctrsource, ctrtarget & ctrdata. Will we be using the same address (i.e. 0x200) for these 3 CSRs?

Apparently so, as the same _indirect_ address. Note that even though they have the same indirect address, they are accessed via different indirect registers: `sireg`, `sireg2`, and `sireg3` respectively.

Are you willing to continue with this PR?

PR #1165, opened 2025-10-04
feat(IDL): single precision floating point multiply add instruction

ShashankVM:


ShashankVM (commented):
Thanks for fixing this IDL compiler issue with the correct fix. I now understand the error.

tools/ruby-gems/idlc/lib/idlc/ast.rb
ShashankVM, line 1: Let's keep this as a separate PR https://github.com/riscv/riscv-unified-db/pull/1690, since this PR focuses on Floating point IDL

PR #1167, opened 2025-10-05
feat: add a custom_* field to extend instruction's schema

AFOliveira:
This adds a way of adding _custom_*_  fields into instructions. It's very handy to describe data UDB doesn't support.

### DIFF
--- a/spec/schemas/inst_schema.json
+++ b/spec/schemas/inst_schema.json
@@ -504,6 +504,9 @@
             }
           ]
         }
+      },
+      "patternProperties": {
+        "^custom_[A-Za-z0-9_]+$": {}
       }
     }
   },


spec/schemas/inst_schema.json
```diff
@@ -504,6 +504,9 @@
             }
           ]
         }
+      },
+      "patternProperties": {
```
ThinkOpenly, line 508: Wouldn't it need to be within the "properties" section? (I don't know, but I thought everything needed to be within "properties", and
```
 "additionalProperties": false,
```
prevented anything additional from being added.)

AFOliveira: Well, not in the properties, as this field doesn't have any properties. It just specifies that anything starting with _custom__ can be inserted in the schema. Which is the behaviour that I think was intended.

Issue #1145, opened 2025-09-29
Drop the fallback to directly processing YAML files in loading exception codes if there is no resolved codes file.


AliAlaa88:
**Describe the bug**
Fallback to directly processing YAML files in loading exception codes if there is no resolved codes file produces invalid code.

**To Reproduce**
1. On Branch **'Ali/Sverilog'**
2. Remove the `--resolved-codes` from the `sh` in the **sverilog task** in the **Tasks.rake** file
3. Run `do gen:sverilog`

**Text output/Error Messages**
This is an example of output error in the generated .svh
```
  localparam logic [5:0] CAUSE_ENVIRONMENT_CALL_FROM_<%__IF_EXT?(:H)__%>V<%__END__%>U_MODE = 6'h08;
  localparam logic [5:0] CAUSE_ENVIRONMENT_CALL_FROM_<%__IF_EXT?(:H)__%>H<%__END__%>S_MODE = 6'h09;
```

AlgoArtist06:
I tried to reproduce this and it no longer reproduces. The YAML fallback was already removed in commit 9b8701a5 (PR #1233), so `load_exception_codes` no longer touches raw YAML at all.

To check, I ran `./do gen:sverilog` normally (works, 22 exception codes), then ran both `sverilog_generator.py` and `generate_encoding.py` directly against `gen/resolved_spec/_` with `--resolved-codes` omitted. Both log `Resolved codes file not found: None` and then die with a `TypeError` before writing anything, so no artifact is produced. There are no `<%__IF_EXT?...%>` markers in any generated output.

Looks safe to close. One leftover I noticed while in there: the `except` branch in `load_exception_codes` logs the error and then falls through into now-unreachable code, so a corrupt resolved-codes JSON silently yields zero exception codes instead of failing. Different path, same silent-bad-output class of problem. I filed that separately.

ThinkOpenly:
OK, closing this as fixed by #1233

PR #1189, opened 2025-10-14
fix: remove processing raw yaml files in exception cause generation

AliAlaa88:
This PR related to PR #1090 which requests removal of processing raw yaml files in exception cause generation to prevent generation of wrong codes #1145


AliAlaa88:
> @AliAlaa88 I think your commit is not adding what you would like to add, right?

It is overriding the load_exception_codes function in the generator.py
I think if the sverilog PR is merged then this one after it it will be OK or what?

PR #1196, opened 2025-10-22
fix: set owner of Node.js dependencies

qwersem:
Few files in a node_modules/ folder belongs to user who doesn’t exist in a host environmnet when `npm` installs external packages inside container. This change explicitly sets the owner for these files.

### Steps to reproduce the issue

```shell
$ git clone https://github.com/riscv-software-src/riscv-unified-db.git
$ cd riscv-unified-db
$ export DOCKER=1
$ ./do test:smoke
Fetching gem metadata from https://rubygems.org/......
$ ls -l node_modules/min-document/dom-fragment.js
-rw-r--r-- 1 100999 100999 876 Oct 23 12:07 node_modules/min-document/dom-fragment.js
# "100999" IS UNKNOWN OWNER OUTSIDE CONTAINER
```

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -216,7 +216,12 @@ if [[ ! -z "$DEVELOPMENT" && $DEVELOPMENT -eq 1 ]]; then
 fi
 
 if [ ! -d ${ROOT}/node_modules ]; then
-    ${RUN} npm i
+    ${RUN} sh -c '
+      npm i
+      npm_res=$?
+      chown -R "$(whoami):$(whoami)" ./node_modules
+      exit $npm_res
+    '
 fi
 
 if [ "${CONTAINER_TYPE}" == "devcontainer" ]; then


AFOliveira (commented):
Hi @qwersem Thanks for your contribution!  

PR titles need to follow an appropriate structure, since they are used as part of the commits message in UDB, I changed the  title from "Set owner of Node.js dependencies" to "fix: set owner of Node.js dependencies" so it passes the CI. If you prefer any other, feel free to change it.

I'm very unaware of the problem you described, so I'll let others chime in to approve.

qwersem:
@AFOliveira thanks

ThinkOpenly:
This doesn't seem like a bug that should be fixed in this repository. Do all other node.js repos have this workaround? What provokes it, and are we doing something wrong that leads to these ownership states? Is it because we're pretending to be root within the container?

qwersem:
@ThinkOpenly , 

_Do all other node.js repos have this workaround?_ 
Yes, I see quite a lot of similar solutions: https://github.com/search?q=chown+-R+%24%28whoami%29+%22node_modules%2F%22&type=code

jordancarlin:
I was working on it for other reasons, but I think #1220 will resolve this.

PR #1205, opened 2025-10-27
feat: add support for podman as the container runtime in addition to docker

jordancarlin:


ThinkOpenly:
> When a fresh copy of UDB is cloned, the first time you run `./do ...` it prompts you for which container runtime you want to use. Previously the choices it gave were Docker and Singularity/Apptainer. Now Podman is one of those choices. There is no special setup needed for it other than selecting Podman.
> 
> To rerun this setup, you can delete the `.container-type` file.

And, I learned something today. :-)

PR #1210, opened 2025-10-27
fix(data): use hint for `pause` instruction

### LINKED ISSUE #1181: Add `pause` instruction from Zihintpause extension
The `Zihintpause.yaml` extension yaml exists, but the `pause` instruction is not defined in the database yet. It should be defined as a hint of `fence`.

jordancarlin:
Replace `pseudoinstructions` inline in `fence.yaml` with hint references.

Fixes #1181


ThinkOpenly (approved):

ThinkOpenly:
tsk, tsk. CI found trailing whitespace. :-)

henrikg-qc:
There is a global fix for this pending. The ISS built for the regression tests will run with compiler definition IGNOREDEFAULTS. This will not throw the UndefinedValueError exception.
Everything is implemented in the build flow but I have a rake issue I am working through. I will push this for review shortly.

ThinkOpenly:
> The menvcfg CSR is only defined for `Sm >= 1.12.0`, but the `rv32-riscv-tests` config specifies that it implements Sm 1.11.0.

That's a bit unexpected, to me at least. Do we know why that restriction exists?

jordancarlin:
> > The menvcfg CSR is only defined for `Sm >= 1.12.0`, but the `rv32-riscv-tests` config specifies that it implements Sm 1.11.0.
> 
> That's a bit unexpected, to me at least. Do we know why that restriction exists?

From the "Preface to Version 20211203" documenting the changes since priv 1.11:
> Defined menvcfg, henvcfg, and senvcfg CSRs (and RV32-only menvcfgh and henvcfgh CSRs), which control various characteristics of the execution environment.

I assume it just wasn't needed by any features before priv 1.12 was introduced.

PR #1218, opened 2025-10-29
feat: attempt to pull docker image before building it

jordancarlin:
The docker/podman image was always built from scratch before even though a precompiled version is automatically uploaded to Docker Hub. This attempts to pull the prebuilt image and only builds it from scratch if that fails.


dhower-qc (approved):

ThinkOpenly (approved):
Nice! Thanks!

PR #1221, opened 2025-10-30
feat: treat wheel group as sudo

jordancarlin:
Many distros use `wheel` as the group that grants sudo permissions. Check for it in addition to the `sudo` group in the `build_container` script.

### DIFF
--- a/bin/build_container
+++ b/bin/build_container
@@ -36,7 +36,7 @@ if [ $NEED_SUDO -eq 0 ]; then
   FAKEROOT=--fakeroot
   echo "Using fakeroot"
 else
-  if [[ ! -z "$GITHUB_RUN_ID" || `groups` == *"sudo"* || `groups` == *"wheel"* ]]; then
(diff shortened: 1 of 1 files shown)


bin/build_container
```diff
@@ -36,7 +36,7 @@ if [ $NEED_SUDO -eq 0 ]; then
   FAKEROOT=--fakeroot
   echo "Using fakeroot"
 else
-  if [[ ! -z "$GITHUB_RUN_ID" || `groups` == *"sudo"* ]]; then
+  if [[ ! -z "$GITHUB_RUN_ID" || `groups` == *"sudo"* || `groups` == *"wheel"* ]]; then
```
ThinkOpenly: I know you're replicating a case that was already there, but I wonder if we should use more precise matching here, something like:

```suggestion
  if [[ ! -z "$GITHUB_RUN_ID" || ' '$(groups)' ' =~ ' sudo ' || ' '$(groups)' ' =~ ' wheel ' ]]; then
```
Otherwise, we'll match if `groups` reports "notsudo" or "wheelies".

dhower-qc (approved):

PR #1225, opened 2025-10-30
fix: run docker/podman container once

qwersem:
Instead of running multiple containers, it is enough to use just one.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -9,6 +9,15 @@ CONTAINER_TAG=`cat ${ROOT}/bin/.container-tag`
 # Parse command line arguments
 PRESERVE_CONFIG=0
 
+CONTAINER_SETUP_CMD=()
+run() {
+  if [ "${CONTAINER_TYPE}" == "docker" -o "${CONTAINER_TYPE}" == "podman" ]; then
+    CONTAINER_SETUP_CMD+=( "$*" )
+  else
+    ${RUN} "$@"
+  fi
+}
+
 # Only parse arguments if this script is being run directly (not sourced)
 if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
   for arg in "$@"; do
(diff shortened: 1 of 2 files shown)


bin/setup
```diff
@@ -180,16 +187,16 @@ if [ $PRESERVE_CONFIG -eq 0 ] || [ ! -f $ROOT/.bundle/config ]; then
...
+    run bundle config set --local cache_path ${ROOT}/.home/.cache
+    run bundle config set --local with development
     cd $OLDDIR
 fi
 
 if [ ! -d $ROOT/.home/.gems ]; then
     OLDDIR=$PWD
     cd $ROOT
-    ${RUN} bundle install
+    run bundle install
```
dhower-qc, line 199: I think this (and other) setup commands already run just once since they are conditioned on the existence of file (`$ROOT/.home/.gems`) in this case. Are you seeing them run multiple times anyway?

bin/setup
```diff
@@ -198,13 +205,13 @@ if [ ! -f $ROOT/ext/rbi-central/README.md ]; then
 fi
 
 if [ ! -d $ROOT/.home/.venv ]; then
-    ${RUN} /usr/bin/python3 -m venv ${ROOT}/.home/.venv
+    run /usr/bin/python3 -m venv ${ROOT}/.home/.venv
 fi
 
-source ${ROOT}/.home/.venv/bin/activate
+run source ${ROOT}/.home/.venv/bin/activate
```
dhower-qc: Since this (I think) sets environment variables in the host, it might need to run every time anyway.

dhower-qc (commented):
I think several of these commands are already elided because they are gated by a file existence check (for example, `bundle install` only runs if `$ROOT/.home/.gems` is missing).

Is that also your understanding, or is there something I'm missing here?

Also,

bin/setup
qwersem: This change doesn't refuse existence checks. If we shouldn't do something based on checks then this command won't be accumulated to `CONTAINER_SETUP_CMD` and won't be executed. It does full build inside only one container instead of multiple needless creation and removing of containers.

bin/setup
qwersem: I guess if we build docs inside container so we need to activate Python venv there. Activation on the host doesn't share environment variables implicitly into container.
```shell
riscv-unified-db$ export ROOT=$PWD
riscv-unified-db$ source ${ROOT}/.home/.venv/bin/activate
(.venv) riscv-unified-db$ env | grep VIRTUAL_ENV
VIRTUAL_ENV=/home/q/workspace/sync_upstream/riscv-unified-db/.home/.venv
VIRTUAL_ENV_PROMPT=(.venv)
(.venv) riscv-unified-db$ docker run --rm -v ${ROOT}:${ROOT} -w ${ROOT} riscvintl/udb:0.9 bash -c "env"
HOSTNAME=2fe0d945961c
PWD=/home/q/workspace/sync_upstream/riscv-unified-db
container=podman
HOME=/root
TERM=xterm
SHLVL=0
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
DEBIAN_FRONTEND=noninteractive
_=/usr/bin/env
# WRONG: We don't have venv variables inside container.
```

dhower-qc: I see. I use singularity, where environment variables from the host appear in the container by default (and do not persist in the container). In Docker, will the environment variable persist?

qwersem: There is `--env` option which allows to set variable inside. But it's also possible to define through shell call to use it one time.

PR #1228, opened 2025-10-31
fix(data): extension Svadu does not conflict with Svade

### LINKED ISSUE #1217: Svadu should not be incompatible with Svade
`Svadu.yaml` currently lists `Svade` as a conflicting extension.

**Expected value**
`Svadu` and `Svade` should be able to co-exist:
- There is no mention in the spec of the extensions being incompatible with each other.
- The RVA23 profile requires `Svade` and has `Svadu` as an optional, additional extension. This isn't possible if they conflict with each other.
- The existing profiles & bases & extensions spreadsheet does not list them as conflicting extensions
- From the Linux kernel device tree documentation:
  > Both Svade and Svadu extensions control the hardware behavior when the PTE A/D bits need to be set. The default behavior for the four possible combinations of these extensions in the device tree are:
  > 1) Neither Svade nor Svadu present in DT => It is technically unknown whether the platform uses Svade or Svadu. Supervisor software should be prepared to handle either hardware updating of the PTE A/D bits or page faults when they need updated.
  > 2) Only Svade present in DT => Supervisor must assume Svade to be always enabled.
  > 3) Only Svadu present in DT => Supervisor must assume Svadu to be always enabled.
  > 4) Both Svade and Svadu present in DT => Supervisor must assume Svadu turned-off at boot time. To use Svadu, supervisor must explicitly enable it using the SBI FWFT extension.

ThinkOpenly:
Svadu, roughly, preempts Svade:
> If the Svadu extension is implemented, the ADUE bit controls whether hardware updating of PTE A/D bits
> is enabled for S-mode and G-stage address translations. When ADUE=1, hardware updating of PTE A/D
> bits is enabled during S-mode address translation, and the implementation behaves as though the Svade
> extension were not implemented for S-mode address translation. When the hypervisor extension is
> implemented, if ADUE=1, hardware updating of PTE A/D bits is enabled during G-stage address
> translation, and the implementation behaves as though the Svade extension were not implemented for G-
> stage address translation. When ADUE=0, the implementation behaves as though Svade were implemented
> for S-mode and G-stage address translation.

Fixes #1217

### DIFF
--- a/spec/std/isa/ext/Svade.yaml
+++ b/spec/std/isa/ext/Svade.yaml
@@ -43,11 +43,11 @@ doc_license:
   url: https://creativecommons.org/licenses/by/4.0/
 requirements:
   allOf:
-    # Svade was defined with Sm 1.13
-    # it needs at least Sm 1.11 for menvcfg defintion
     - extension:
         name: Sm
-        version: ">= 1.13"
-    - not:
-        extension:
-          name: Svadu
+        version: ">= 1.12"
+    - anyOf:
+      - extension:
+          name: Sv32
+      - extension:
+          name: Sv39


jordancarlin (approved):

ThinkOpenly:
In the UnifiedDB SIG PR review meeting, we decided to wait for #891 before diving deeper into the issue seen here.

spec/std/isa/ext/Svadu.yaml
```diff
@@ -128,9 +128,6 @@ requirements:
     - extension:
         name: Sm
```
jordancarlin, line 127: I think we discussed this during one of our meetings, but the Sm requirement needs to be changed to Sv32/Sv39.

ThinkOpenly: Would that be "anyOf" Sv32 or Sv39, and ignoring the others like Sv48, Sv57?

jordancarlin: Yes because Sv48 and Sv57 both require Sv39.

spec/std/isa/ext/Svadu.yaml
```diff
@@ -124,9 +124,15 @@ versions:
 requirements:
   # Svadu was defined with Sm 1.13
   # it needs at least Sm 1.11 for menvcfg defintion
-  extension:
-    name: Sm
-    version: ">= 1.13"
+  allOf:
+    - extension:
+        name: Sm
+        version: ">= 1.13"
```
jordancarlin: This should be 1.12 because that is when `menvcfg` was added and that is the only dependency on Machine mode.

spec/std/isa/ext/Svadu.yaml
```diff
@@ -124,9 +124,15 @@ versions:
 requirements:
   # Svadu was defined with Sm 1.13
   # it needs at least Sm 1.11 for menvcfg defintion
```
jordancarlin: Comment needs to be updated (or maybe just dropped).

jordancarlin (changes requested):
Svade still needs to be updated.

PR #1229, opened 2025-10-31
feat: add additional ecosystems to Dependabot

jordancarlin:
Adds the following ecosystems to Dependabot for automatic dependency updates:

- [ ] npm
- [ ] pip
- [ ] bundler
- [ ] 3rd party GitHub Actions
- [ ] Docker

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,3 +6,28 @@ updates:
     directory: /
     schedule:
       interval: weekly
+
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: weekly
+
+  - package-ecosystem: "pip"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -6,3 +6,28 @@ updates:
     directory: /
     schedule:
       interval: weekly
+
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: weekly
```
dhower-qc, line 13: Do you know if this is smart enough to look in the container files now that we are installing things there?

jordancarlin: It updates the versions in `package.json` and `package-lock.json`, which are then used to install the dependencies in the container.

PR #1230, opened 2025-11-03
fix(data): profiles: clean up after multiple inheritance

### LINKED ISSUE #1195: RVA23 Profile doesn't exactly match RVA23 Profile spec (e.g., Zkn)
Just noticed that the RVA23 Profile has some extensions that were optional in RVA22 but no longer options in RVA23 (like scalar crypto). Also Za128rs shouldn't be mandatory or optional any more.

```
The scalar crypto extensions Zkn and Zks that were options in RVA22 are not
options in RVA23. The goal is for both hardware and software vendors to
move to use vector crypto, as vectors are now mandatory and vector crypto is
substantially faster than scalar crypto.

### PR DESCRIPTION
**Pull request by ThinkOpenly**
Most profiles don't automatically incorporate optional extensions from "ancestors", so more extensions need to be removed after profile inheritance.

For example, in UDB, profile RVA23S64 inherits from both RVB23S64 and RVA23U64. However, in the profiles documentation it (roughly) inherits from only RVA23U64, and then only "all the mandatory unprivileged extensions". So, the optional extensions need to be handled more explicitly, and many need to be removed.

Fixes #1195

### DIFF
--- a/spec/std/isa/ext/Za64rs.yaml
+++ b/spec/std/isa/ext/Za64rs.yaml
@@ -17,11 +17,13 @@ description: |
   [NOTE]
   The minimum reservation set size is effectively determined by the size of atomic accesses in
   the A extension.
-conflicts: Za128rs
 versions:
   - version: "1.0.0"
     state: ratified
     ratification_date: null
+    implies:
+      - name: Za128rs
+        version: "1.0.0"
     param_constraints:
       LRSC_RESERVATION_STRATEGY:
         schema:
--- a/spec/std/isa/profile/RVA22S64.yaml
+++ b/spec/std/isa/profile/RVA22S64.yaml
@@ -21,7 +21,6 @@ extensions:
     - "profile/RVA20S64.yaml#/extensions"
     - "profile/RVA22U64.yaml#/extensions"
   $remove:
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by jordancarlin — APPROVED** [review]

File: spec/std/isa/ext/Za64rs.yaml [hunk 1]
```diff
@@ -21,6 +21,9 @@ versions:
   - version: "1.0.0"
     state: ratified
     ratification_date: null
+    implies:
+      - name: Za128rs
+        version: "1.0.0"
```
dhower-qc, line 26: Based on Greg's reply, seems like this should be removed.

ThinkOpenly: I added it based on this part of Greg's reply:
> > Za64rs: Reservation sets are contiguous, naturally aligned, and a maximum of 64 bytes.
> > Za128rs: Reservation set size of at most 128 bytes
> > 
> > Is it possible to claim support for both extensions?
> 
> Yes, since supporting the former means that one inherently also supports the latter.

dhower-qc: ah, yes, you are right

PR #1231, opened 2025-11-04
feat: add Python example to list extensions associated with profiles

ThinkOpenly:
```
$ ./tools/python/profile-extensions.py --profiles RVA23U64 gen/resolved_spec/_/
RVA23U64:
-  A = 2.1 mandatory
-  C = 2.0 mandatory
-  D = 2.2 mandatory
-  F = 2.2 mandatory
-  I ~> 2.1 mandatory
-  M = 2.0 mandatory
-  Supm ~>1.0 mandatory
-  U ~> 2.0 mandatory
-  V ~> 1.0 mandatory
-  Za64rs ~>1.0 mandatory
[...]
```

### DIFF
--- a/tools/python/udb.py
+++ b/tools/python/udb.py
@@ -7,21 +7,19 @@ from pathlib import Path
 import yaml
 
 
-database = []
-
-
 def find_and_load_yaml(path, kinds=None):
     """Load the YAML files in a directory hierarchy into an array of dictionaries.
 
     Optionally, restrict to specific "kinds" of YAML files.
     """
+    database = []
 
     p = Path(path)
-    for path in p.rglob("*.yaml"):
-        with open(path, encoding="utf-8") as f:
+    for file in p.rglob("*.yaml"):
+        with open(file, encoding="utf-8") as f:
             y = yaml.safe_load(f)
             if "kind" in y:
                 if len(kinds) == 0 or y["kind"] in kinds:
-                    y["file"] = path
+                    y["file"] = file
                     database.append(y)
     return database


dhower-qc (commented):
- Dumb Python question: What's the Python equivalent of a Ruby Gem? Should this wrapped in whatever that is (with other functions, too?)?
- I would suggest a new regression in regress.yml so this continues to work.

dhower-qc:
> Refactoring would likely make the example even simpler by some measurement, since the "load YAML" part would be refactored out. I'll work on that.

Excellent


> I suppose we might want to use the "golden reference" approach.

We might also get away with comparing the python "list extension" to the ruby "list extension"

ThinkOpenly:
> > I suppose we might want to use the "golden reference" approach.
> 
> We might also get away with comparing the python "list extension" to the ruby "list extension"

Does the ruby "list extension" tag or filter by containing profile?

tools/python/UDB.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python3
+# Copyright (c) Ventana Micro Systems
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+import yaml
+from pathlib import Path
+
+
+def store_yaml(path, kinds=[]):
```
jordancarlin: We should avoid using an empty list as a default argument. It can create some very strange results if a function is called multiple times. See https://docs.astral.sh/ruff/rules/mutable-argument-default/ for more details.

ThinkOpenly: Good catch. I learned something today.

Fixed.

dhower-qc:
> > > I suppose we might want to use the "golden reference" approach.
> > 
> > 
> > We might also get away with comparing the python "list extension" to the ruby "list extension"
> 
> Does the ruby "list extension" tag or filter by containing profile?

Yes, in that it takes a config. For example:

```bash
./bin/udb list extensions --config rva23_full_example
```

Will give you the extensions present in rva23_full_example

tools/python/udb.py
```diff
@@ -0,0 +1,31 @@
...
+
+    Optionally, restrict to specific "kinds" of YAML files.
+    """
+
+    p = Path(path)
+    if p.is_dir():
+        for dirent in p.iterdir():
+            find_and_load_yaml(dirent, kinds)
+    else:
+        if str(path).endswith(".yaml"):
```
jordancarlin:
```suggestion
        if path.suffix == ".yaml":
```

tools/python/udb.py
```diff
@@ -0,0 +1,31 @@
...
+            with open(path, encoding="utf-8") as f:
+                y = yaml.safe_load(f)
+                if "kind" in y:
```
jordancarlin: Could this all be simplified to the following? I don't think the recursion is needed.
```python
for path in p.rglob("*.yaml"):
    with open(path, encoding="utf-8") as f:
        ...
```

ThinkOpenly: nice. I think we're noting the difference between "competent" and "proficient". ;-)

tools/python/udb.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python3
+# Copyright (c) Ventana Micro Systems
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+"""Python utilities for using UDB"""
+
+from pathlib import Path
+import yaml
+
+
+database = []
```
jordancarlin: I think this can be local to the `find_and_load_yaml` function now.

PR #1233, opened 2025-11-06
fix: remove processing raw yaml files in exception cause generation

AliAlaa88:
This PR related to PR https://github.com/riscv-software-src/riscv-unified-db/pull/1090 which requests removal of processing raw yaml files in exception cause generation to prevent generation of wrong codes https://github.com/riscv-software-src/riscv-unified-db/issues/1145


ThinkOpenly:
I think this matches your expectation, right, @jordancarlin ?

ThinkOpenly (approved):

PR #1245, opened 2025-11-12
First draft vector parameters

jacassidy:
Trying to pickup on any mentions of "implementation specific" choices that could be made by a designer of the RISC-V Vector Extension. Descriptions are direct quotes from the spec for easy search. Some parameters may be unneeded as they wouldnt affect instruction-level output behavior but I figured including more rather than less could only be benificial. Names are tentative.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -106,7 +106,7 @@ if [ -v GITHUB_ACTIONS ]; then
     echo "Running in a GitHub Action"
     CONTAINER_PATH=${ROOT}/.singularity/image.sif
     HOME_PATH=${GITHUB_WORKSPACE}
-    HOME_OPT="--home ${ROOT}/.home"
+    GEM_HOME=${ROOT}/.home/.gems && GEM_PATH=$GEM_HOME
     SINGULARITY_CACHE=--disable-cache
 
     # needed to get singularity working on Ubuntu 24.04
@@ -137,7 +137,7 @@ elif [ "${CONTAINER_TYPE}" == "docker" -o "${CONTAINER_TYPE}" == "podman" ]; the
 elif [ "${CONTAINER_TYPE}" == "singularity" ]; then
     CONTAINER_PATH=${ROOT}/.singularity/image-$CONTAINER_TAG.sif
     HOME_PATH=${HOME}
-    HOME_OPT="--bind ${ROOT}/.home:${HOME_PATH}"
+    GEM_HOME=${ROOT}/.home/.gems && GEM_PATH=$GEM_HOME
     SINGULARITY_CACHE=
     if [ ! -f ${CONTAINER_PATH} ]; then
       echo "Fetching container..."
--- a/tools/scripts/pages.html.erb
+++ b/tools/scripts/pages.html.erb
@@ -156,6 +156,7 @@
           <h3>Repository development</h3>
           <ul>
             <li><a href="<%= ENV["PAGES_URL"] %>/reuse_bom.txt">Reuse Bill of Materials (LICENSE information)</a></li>
+            <li><a href="https://dhower-qc.github.io/udb-stats/riscv-software-src/riscv-unified-db/latest-report/report.html">GitHub usage statistics (updated daily)</a></li>
           </ul>
         </div>
       </div>


spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+    schema:
+      type: integer
+      enum: [8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536]
+  VLEN:
+    description: |
+      The number of bits in a single vector register, _VLEN_ {ge} ELEN, which must be a power of 2, and must be no greater than 2^16^.
+    schema:
+      type: integer
+      enum: [8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536]
+  SEWmin:
```
ThinkOpenly: We may want to observe an invariant that all params consist only of upper case and underscore. Looking at the schema, though, I think we already do.
```suggestion
  SEW_MIN:
```

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+      Implementations must provide fractional LMUL settings that allow the
+      narrowest supported type to occupy a fraction of a vector register
+      corresponding to the ratio of the narrowest supported type's width to
+      that of the largest supported type's width.  In general, the
+      requirement is to support LMUL {ge} SEW~MIN~/ELEN, where SEW~MIN~ is
+      the narrowest supported SEW value and ELEN is the widest supported SEW
+      value.
+    schema:
+      type: integer
+      enum: [8, 16, 32, 64]
```
ThinkOpenly, line 44: The spec has this:
> In the standard extensions, SEW<sub>MIN</sub>=8.

So, do we need to support a parameter for this, given that only 8 is "standard"?

Related, where did these 4 values come from?

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+  VECTOR_AGNOSTIC_BEHAVIOR:
+    description:
+      When a set is marked agnostic, the corresponding set of destination
+      elements in any vector destination operand can either retain the value
+      they previously held, or are overwritten with 1s.
+    schema:
+      type: string
+      enum: ["never_overwritten", "custom"]
```
ThinkOpenly: This appears to be parameterizing a specific possible implementation of "agnostic" for a behavior which can even be non-deterministic. Is there value in doing so? (Honest question.)

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+  SUPPORT_FRACTIONAL_LMUL_BEYOND_REQUIRED:
+    description:
+      For a given supported fractional LMUL setting, implementations must support
+      SEW settings between SEW~MIN~ and LMUL * ELEN, inclusive.
+    schema:
+      type: string
+      enum: ["no_unrequied_supported", "custom"]
```
ThinkOpenly: spell-o
```suggestion
      enum: ["no_unrequired_supported", "custom"]
```

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+      When a set is marked agnostic, the corresponding set of destination
+      elements in any vector destination operand can either retain the value
+      they previously held, or are overwritten with 1s.
+    schema:
+      type: string
+      enum: ["never_overwritten", "custom"]
+  SUPPORT_FRACTIONAL_LMUL_BEYOND_REQUIRED:
+    description:
+      For a given supported fractional LMUL setting, implementations must support
+      SEW settings between SEW~MIN~ and LMUL * ELEN, inclusive.
```
ThinkOpenly, line 51: This would seem to be a normative rule, rather than a parameter. (But don't trust my interpretation! 😉 )

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+  VILL_SET_ON_RESERVED_VTYPE:
+    description:
+      The use of `vtype` encodings with LMUL < SEW~MIN~/ELEN is
+      __reserved__, but implementations can set `vill` if they do not
+      support these configurations.
+    schema:
+      type: boolean
+  VILL_IMPLICIT_ENCODING:
+    description: |
+      A small implementation supporting ELEN=32 requires only seven bits of state in `vtype`: two bits for `ma` and `ta`, two bits for `vsew[1:0]` and three bits for `vlmul[2:0]`.  The illegal value represented by `vill` can be internally encoded using the illegal 64-bit combination in `vsew[1:0]` without requiring an additional storage
```
ThinkOpenly: nit: missing a period at the end.
```suggestion
      A small implementation supporting ELEN=32 requires only seven bits of state in `vtype`: two bits for `ma` and `ta`, two bits for `vsew[1:0]` and three bits for `vlmul[2:0]`.  The illegal value represented by `vill` can be internally encoded using the illegal 64-bit combination in `vsew[1:0]` without requiring an additional storage.
```

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+      bit to hold `vill`.:
+    schema:
+      type: boolean
+  MUTABLE_MISA_V:
+    description: Implementations may have a writable `misa.V` field.
+    schema:
+      type: boolean
```
ThinkOpenly: This parameter is already present.
```suggestion
```

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+  HW_MSTATUS_VS_DIRTY_UPDATE:
+    description: |
+      Implementations may also change `mstatus.VS` or `vsstatus.VS` from Initial or
+      Clean to Dirty at any time, even when there is no change in vector state.
+    schema:
+      type: string
+      enum: ["never", "precise", "imprecise", "custom"]
```
ThinkOpenly: What's the difference between "imprecise" and "custom"?

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+  MSTATUS_VS_EXISTS:
+    description: |
+      Analogous to the way in which the floating-point unit is handled, the `mstatus.VS` field may exist even if `misa.V` is clear.
+    schema:
+      type: boolean
+  RESERVED_VILL_SET:
+    description: |
+      When _rs1_=`x0` and _rd_=`x0`, the instructions operate as if the current vector length in `vl` is used as the AVL...
+      Use of the instructions with a new SEW/LMUL ratio that would result in a change of VLMAX is reserved. 
+      Use of the instructions is also reserved if `vill` was 1 beforehand. Implementations may set `vill` in either case.
```
ThinkOpenly: I wonder if these two cases should be separately parameterized? But, I guess you handle each case independently via the enum below. So, maybe fine.

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+    schema:
+      type: string
+      enum: ["never", "on_vlmax_change", "on_vill_set", "always", "custom"]
+  VECTOR_LS_INDEX_MAX_EEW:
+    description:
+      A profile may place an upper limit on the maximum supported index
+      EEW (e.g., only up to XLEN) smaller than ELEN.
+    schema:
+      type: integer
+      enum: [8, 16, 32, 64]
```
ThinkOpenly: I wonder if this additionally needs "XLEN" as an option, given the description suggests that.

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+  VECTOR_FF_PAST_TRAP:
```
ThinkOpenly: My reading is that fault-only-first loads don't trap. They "trim" instead. So, suggest something like this:
```suggestion
  VECTOR_FF_UPDATE_PAST_TRIM:
```

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+    description: 
+      Similarly, fault-only-first load instructions may update active destination elements past the element that causes trimming 
+      of the vector length (but not past the original vector length).  The values of these spurious updates do not have to correspond 
+      to the values in memory at the addressed memory locations.  Non-idempotent memory locations can only be accessed when it is known 
+      the corresponding element load operation will not be restarted due to a trap or vector-length trimming.
+    schema:
+      type: string
+      enum: ["update_none", "custom"]
+  VECTOR_FF_SEG_PARTIAL_ACCESS:
```
ThinkOpenly: Suggestion:

```suggestion
  VECTOR_FF_SEG_EXCEPTION_PARTIAL_LOAD:
```

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+      For fault-only-first segment loads, if an exception is detected partway
+      through accessing the zeroth segment, the trap is taken.
+      If an exception is detected partway through accessing a subsequent segment,
+      `vl` is reduced to the index of that segment.
+      In both cases, it is implementation-defined whether a subset of the segment is
+      loaded.
+    schema:
+      type: string
+      enum: ["no_subsegment_loaded", "custom"]
+  VECTOR_LS_MISSALIGNED_EXCEPTION:
```
ThinkOpenly, line 110: Is this the same as `Sm.yaml`'s `MISALIGNED_LDST`?

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+    description: |
+      If an element accessed by a vector memory instruction is not naturally
+      aligned to the size of the element, either the element is transferred
+      successfully or an address-misaligned exception is raised on that
+      element.
+    schema:
+      type: string
+      enum: ["element_loaded_no_exception", "address-misaligned_exception"]
```
ThinkOpenly: If this parameter is retained, then I suggest it be made a boolean.

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+  VECTOR_LS_OVERWRITE_PAST_TRAP:
```
ThinkOpenly: This is for loads only, so...
```suggestion
  VECTOR_LOAD_PAST_TRAP:
```
?

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+    description: |
+      Load instructions may overwrite active destination vector register
+      group elements past the element index at which the trap is reported.
+    schema:
+      type: string
+      enum: ["preserve_inactive", "custom"]
```
ThinkOpenly: "inactive" is not appropriate here, as this is just about active elements.

Maybe this should just be a boolean.

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+  VECTOR_LS_SEG_FF_OVERLOAD:
```
ThinkOpenly: This is also for loads only.

However, I think this case is already covered by the "PARTIAL" parameter above.

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+      type: string
+      enum: ["no_overwrite", "custom"]
+  VECTOR_LS_SEG_PARTIAL_ACCESS:
+    description: |
+      If a trap occurs during
+      access to a segment, it is implementation-defined whether a subset
+      of the faulting segment's accesses are performed before the trap is taken.
+    schema:
+      type: string
+      enum: ["no_partial", "custom"]
```
ThinkOpenly: boolean?

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+  LEGAL_VSTART:
+    description: |
+      Implementations are permitted to raise illegal-instruction exceptions when
+      attempting to execute a vector instruction with a value of `vstart` that the
+      implementation can never produce when executing that same instruction with
+      the same `vtype` setting.
+    schema:
+      type: string
+      enum: ["1_stride", "2_stride", "4_stride", "custom"]
```
ThinkOpenly, line 147: I think this needs to comport more closely to the description. Maybe "EXCEPTION_ON_UNSUPPORTED_VSTART" or something like that? Then, it would just need to be a boolean.

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+  VECTOR_LS_WHOLEREG_MISSALIGNED_EXCEPTION:
+    description: |
+      Implementations are allowed to raise a misaligned address exception on
+      whole register loads and stores if the base address is not naturally
+      aligned to the larger of the size of the encoded EEW in bytes (EEW/8)
+      or the implementation's smallest supported SEW size in bytes
+      (SEW~MIN~/8).
+    schema:
+      type: string
+      enum: ["misaligned_exception", "permitted"]
```
ThinkOpenly: boolean?

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+    schema:
+      type: boolean
+  VECTOR_FF_NO_EXCEPTION_TRIM:
+    description: |
+      Even when an exception is not raised, implementations are permitted to process
+      fewer than `vl` elements and reduce `vl` accordingly, but if `vstart`=0 and
+      `vl`>0, then at least one element must be processed.
+    schema:
+      type: string
+      enum: ["no_trim", "custom"]
```
ThinkOpenly: boolean?

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+  vfredosum_nan
```
ThinkOpenly: accidentally left in?
```suggestion
```

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+  VFREDOSUM_NAN:
+    description: |
+      As a consequence of this definition, implementations need not propagate
```
ThinkOpenly, line 173: You might add the referenced definition for clarity:

```suggestion
      The reduction tree structure must be deterministic for a given value in vtype and vl.
      As a consequence of this definition, implementations need not propagate
```

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+  VECTOR_FF_NO_EXCEPTION_TRIM:
+    description: |
+      Even when an exception is not raised, implementations are permitted to process
+      fewer than `vl` elements and reduce `vl` accordingly, but if `vstart`=0 and
+      `vl`>0, then at least one element must be processed.
+    schema:
+      type: string
+      enum: ["no_trim", "custom"]
+  vfredosum_nan
+  VFREDOSUM_NAN:
```
ThinkOpenly: The description below is associated with `VFREDUSUM`:

```suggestion
  VFREDUSUM_NAN:
```

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+      As a consequence of this definition, implementations need not propagate
+      NaN payloads through the reduction tree when no elements are active. In
+      particular, if no elements are active and the scalar input is NaN,
+      implementations are permitted to canonicalize the NaN and, if the NaN is
+      signaling, set the invalid exception flag.  Implementations are alternatively
+      permitted to pass through the original NaN and set no exception flags, as with
+      `vfredosum`.
+    schema:
+      type: string
+      enum: ["no_change", "custom"]
```
ThinkOpenly, line 182: Could use more eyes on this one:
- Is the parameter just about `vfredusum` and "NaN" or should "no active elements" be part of the parameter name as well? In other words, is it specific to the case of "no active elements"?
- Does it need to be a parameter at all, given the description is in an "informative note" and not in the regular text?

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+  VSLIDE_OPTIMIZED_VALUES:
+    description: |
+      Implementations
+      may optimize certain OFFSET values for `vslideup` and `vslidedown`.
+      In particular, power-of-2 offsets may operate substantially faster
+      than other offsets.
+    schema:
+      type: string
+      enum: ["custom"]
```
ThinkOpenly: Since this is just a performance characterization, I don't think this needs a parameter.

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+  IMPRECISE_VECTOR_TRAP_SETTABLE:
+    description: |
+      Some profiles may choose to provide a privileged mode bit to select between precise and imprecise vector traps.
+    schema:
+      type: boolean
```
ThinkOpenly, line 212: Note the text that follows the paragraph containing the description:
> This mechanism is not defined in the current standard extensions.

Given that, it would need to be defined in a comprehensive overlay/addition, so I don't see a need for a parameter.
```suggestion
```

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,212 @@ versions:
...
+  FUSED_VSET_MICRO-OP:
+    description:
+      The primary motivation for the `vtype` CSR is to allow the
+      vector instruction set to fit into a 32-bit instruction encoding
+      space.  A separate `vset{i}vl{i}` instruction can be used to set `vl`
+      and/or `vtype` fields before execution of a vector instruction, and
+      implementations may choose to fuse these two instructions into a single
+      internal vector microop.
+    schema:
+      type: boolean
```
ThinkOpenly: This is a micro-architectural choice, invisible to the architecture, so there's no need for a parameter in the architectural description here.
```suggestion
```

spec/std/isa/ext/V.yaml
jacassidy: The issue below discusses this issue. The normative rule above simply states what is required to be implemented. For example, it doesnt state that SEW 64 LMUL 1/2 cannot be supported. Therefore in my mind I see it as up to the designer to allow for this implementation or not. Supporting vs not supporting these cases would produce a different result when executing the same program: that is why it should be a parameter in my mind.

[](https://github.com/riscv/riscv-isa-manual/issues/2359)

spec/std/isa/ext/V.yaml
jacassidy: I think this is a completely fair interpretation; I just thought there might be some value to even stating in a custom parameter "undeterministic." I think it would also be useful for a parameter to state that this behavior is never exercised, thats why I included "never_overwritten"

spec/std/isa/ext/V.yaml
jacassidy: In the testing framework we are developing for RISC-V ACT 4.0 we include SEW_MIN as a parameter. Even though it is not currently in a standard extension to use anything other than SEW_MIN=8, it is still a valid vector implementation to have a different SEW_MIN. As far as we can tell, SEW_MIN is completely arbitrary in non-standard extensions meaning it can be any SEW. For example, we interpret that a vector unit that only supports an SEW of 32 to be a valid implementation (SEW_MIN=32 and ELEN = 32).

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,208 @@ versions:
...
+      must be a power of 2.
+    schema:
+      type: integer
+      enum: [8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536]
+  VLEN:
+    description: |
+      The number of bits in a single vector register, _VLEN_ {ge} ELEN, which must be a power of 2, and must be no greater than 2^16^.
+    schema:
+      type: integer
+      enum: [8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536]
```
jordancarlin, line 30: This needs some kind of extra validation to ensure `VLEN` > `ELEN`.

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,208 @@ versions:
...
+      bit to hold `vill`.:
+    schema:
+      type: boolean
+  HW_MSTATUS_VS_DIRTY_UPDATE:
+    description: |
+      Implementations may also change `mstatus.VS` or `vsstatus.VS` from Initial or
+      Clean to Dirty at any time, even when there is no change in vector state.
+    schema:
+      type: string
+      enum: ["never", "precise", "imprecise", "custom"]
```
jordancarlin: This parameter is already defined at the end of the file.

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,208 @@ versions:
...
+  MSTATUS_VS_EXISTS:
+    description: |
+      Analogous to the way in which the floating-point unit is handled, the `mstatus.VS` field may exist even if `misa.V` is clear.
+    schema:
+      type: boolean
```
jordancarlin: This is already covered by the `MSTATUS_VS_LEGAL_VALUES` parameter.

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,208 @@ versions:
...
+    schema:
+      type: string
+      enum: ["never", "on_vlmax_change", "on_vill_set", "always", "custom"]
+  VECTOR_LS_INDEX_MAX_EEW:
+    description:
+      A profile may place an upper limit on the maximum supported index
+      EEW (e.g., only up to XLEN) smaller than ELEN.
+    schema:
+      type: integer
+      enum: [8, 16, 32, 64]
```
jordancarlin: Seems like the legal values need to go all the way up to ELEN, not stop at 64. This will mean extending the enum and then adding extra validation to ensure it it not greater than ELEN.

spec/std/isa/ext/V.yaml
jacassidy: This is a really good point I did not think of. I think in this case there should actually be 2 parameters, one to state if exceptions occur on unsupported VSTART values, and another to state which VSTART values are supported. With the current parameter definition I was trying to cover which VSTART values are supported by the implementation.

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,208 @@ versions:
...
+      __reserved__, but implementations can set `vill` if they do not
+      support these configurations.
+    schema:
+      type: boolean
+  VILL_IMPLICIT_ENCODING:
+    description: |
+      A small implementation supporting ELEN=32 requires only seven bits of state in `vtype`: two bits for `ma` and `ta`, two bits for `vsew[1:0]` and three bits for `vlmul[2:0]`.  The illegal value represented by `vill` can be internally encoded using the illegal 64-bit combination in `vsew[1:0]` without requiring an additional storage.
+      bit to hold `vill`.:
+    schema:
+      type: boolean
```
jordancarlin: This seems like a microarchitectural detail that is not architecturally visible. This is just about how the data is stored internally. When actually reading/writing the CSR there should be no visible difference.

spec/std/isa/ext/V.yaml
jacassidy: I think the most common case would be to either do both together or none at all, but I do not feel strongly about it being one parameter or two.

spec/std/isa/ext/V.yaml
jacassidy: do you think instead the options should be ["XLEN", "ELEN", "custom"]?

spec/std/isa/ext/V.yaml
jacassidy: Ill update the description to match below:

> If an element accessed by a vector memory instruction is not naturally aligned to the size of the element, either the element is transferred successfully or an address misaligned exception is raised on that element.
> 
> Support for misaligned vector memory accesses is independent of an implementation’s support for misaligned scalar memory accesses.

spec/std/isa/ext/V.yaml
jacassidy: Thanks for the catch! I thought it could be valuble to have a custom parameter to state how it overwrites active elements, if not then I think making it a boolean makes sense.

spec/std/isa/ext/V.yaml
jacassidy: if you're referring to "VECTOR_FF_SEG_PARTIAL_ACCESS," that parameter has to do specifically with the loading of a single segment, while this is referring to all the elements of the group. Interestingly I dont think this is covered by "VECTOR_LS_OVERWRITE_PAST_TRAP" as that parameter is constrained to only active elements. This statement (specifically referring to FF loads) states any "vector register group elements," not strictly active elements

spec/std/isa/ext/V.yaml
jacassidy: is it useful to state if a partial access is performed, how much of the partial access is performed? I might be missinterpreting this section.

spec/std/isa/ext/V.yaml
jacassidy: agreed

spec/std/isa/ext/V.yaml
jacassidy: yes my mistake

spec/std/isa/ext/V.yaml
jacassidy: does this also apply to VSSTATUS_VS_EXISTS, meaning should I instead make a parameter for legal values instead of one for if it exists or not.

spec/std/isa/ext/V.yaml
jacassidy: agreed

spec/std/isa/ext/V.yaml
ThinkOpenly: If I had to choose today, I'd pick the simple option: boolean. If somebody really cares about defining the (non-deterministic) behavior precisely, they can do so. ;-)

spec/std/isa/ext/V.yaml
ThinkOpenly: IMHO, it's non-deterministic, so boolean here is sufficient. (That still allows one, perhaps unwisely, to "rely" on the "no partial access" behavior for whatever nefarious purposes. ;-) )

spec/std/isa/ext/V.yaml
```diff
@@ -15,6 +15,172 @@ versions:
...
+      enum: [8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536]
+  ELEN:
+    description: |
+      The maximum size in bits of a vector element that any operation can produce or consume, _ELEN_ {ge} 8, which
+      must be a power of 2.
+    schema:
+      type: integer
+      enum: [8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536]
+    extra_validation: |
+      assert ELEN <= VLEN
```
jordancarlin: Minor nit, but since the `VLEN` description is the one that says `VLEN` must be greater than `ELEN`, the `extra_validation` should probably be attached to the `VLEN` parameter.

spec/std/isa/ext/V.yaml
ThinkOpenly: I don't think we're trying to describe non-standard implementations. Should a non-standard value be enforced? How? If it's not enforced, then is there value in defining it?

If we keep this parameterization, since the standard describes SEW_MIN as 8, then we should perhaps make that a default value.

spec/std/isa/ext/V.yaml
ThinkOpenly: My opinion is that there isn't value in trying to describe what is defined as non-deterministic. If the "agnostic" result is never overwritten, we already have a way to describe (and specify, by setting the respective agnostic bit to zero) that behavior. I don't think this should be parameterized.

spec/std/isa/ext/V.yaml
ThinkOpenly: I think either the name needs to be changed to reflect the expanded/dual scope, or two parameters. My preference would be the latter: two parameters.

spec/std/isa/ext/V.yaml
ThinkOpenly: If SEW=64, then ELEN=64, so LMUL=1/2 must be supported. Per the spec:
> For standard vector extensions with ELEN=64, fractional LMULs of 1/2, 1/4, and 1/8 must be supported.

I don't see a needed parameterization here, but maybe I'm missing something.

spec/std/isa/ext/V.yaml
ThinkOpenly: I was thinking
```
type: string
enum: [ "8", "16", "32", "64", "XLEN" ]
```
(That allows a configuration to change the XLEN without having to revisit this setting.)

spec/std/isa/ext/V.yaml
ThinkOpenly: > if you're referring to "VECTOR_FF_SEG_PARTIAL_ACCESS," that parameter has to do specifically with the loading of a single segment, while this is referring to all the elements of the group.

OK

>Interestingly I dont think this is covered by "VECTOR_LS_OVERWRITE_PAST_TRAP" as that parameter is constrained to only active elements.

Where is that parameter defined?

Maybe rename to "VECTOR_LOAD_SEG_FF_OVERWRITE_ELEMENTS_AFTER_FAULT". (ugh)

spec/std/isa/ext/V.yaml
ThinkOpenly: Valid VSTART values are not configurable.
> The vstart CSR is defined to have only enough writable bits to hold the largest element index (one less than the maximum VLMAX).

VSTART must be able to represent up to the maximum supported element number.
> The maximum vector length is obtained with the largest LMUL setting (8) and the
smallest SEW setting (8), so VLMAX_max = 8*VLEN/8 = VLEN. For example, for
VLEN=256, vstart would have 8 bits to represent indices from 0 through 255.

spec/std/isa/ext/V.yaml
jacassidy: I believe this line is refering to the fact that ELEN=64 and a standard extension implies an SEW_MIN of 8, meaning lmul 1/2, 1/4, 1/8 have to be supported LMUL. The line I quoted in the spec states that, (if ELEN =64 and SEW_MIN = 8) that for

SEW=64: fLMUL do not have to be supported
SEW=32: fMUL 1/2 has to be supported
SEW=16: fLMUL 1/2-1/4 have to be supported
SEW=8: fLMUL 1/2-1/8 have to be supported

My impression is this is leaving opportunity for fLMUL of 1/2 to be supported for SEW=64 and such if an ELEN of 128 is added. The spec does not state that one cannot implement fLMUL of 1/2 1/4 1/8 for all SEW, but does not state they have to. This causes different behavior on devices that chose to implement these lines vs not

spec/std/isa/ext/V.yaml
jacassidy: sorry forgot we renamed "VECTOR_LS_OVERWRITE_PAST_TRAP" to "VECTOR_LOAD_PAST_TRAP," if that makes sense I'm happy to resolve this comment

spec/std/isa/ext/V.yaml
jacassidy: I phrased my statement poorly, the CSR can always be set with any VSTART value, whether any vector instruction runs or trapps on a given VSTART value in the CSR is implementation dependent

spec/std/isa/ext/V.yaml
jacassidy: @jordancarlin

jacassidy:
@ThinkOpenly sorry I never addressed your initial comment, I added parameters for it now under "VFREDUSUM." Only a couple comments left!

PR #1251, opened 2025-11-15
feat: add pseudo instructions to read specific CSRs

cmd05:
Add pseudo instructions to read specific CSRs in `csrrs`. @ThinkOpenly

### DIFF
--- a/spec/std/isa/inst/Zicsr/csrrs.yaml
+++ b/spec/std/isa/inst/Zicsr/csrrs.yaml
@@ -33,6 +33,24 @@ access:
   vs: always
   vu: always
 pseudoinstructions:
+  - when: xs1 == 0 && csr == 0x001
+    to: frflags xd
+  - when: xs1 == 0 && csr == 0x002
+    to: frrm xd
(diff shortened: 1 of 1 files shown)


spec/std/isa/inst/Zicsr/csrrs.yaml
```diff
@@ -33,6 +33,24 @@ access:
...
+    to: rdcycle xd
+  - when: xs1 == 0 && csr == 0x0C01
+    to: rdtime xd
+  - when: xs1 == 0 && csr == 0x0C02
+    to: rdinstret xd
+  - when: xs1 == 0 && csr == 0x0C80
+    to: rdcycleh xd
+  - when: xs1 == 0 && csr == 0x0C81
+    to: rdtimeh xd
+  - when: xs1 == 0 && csr == 0x0C82
```
ThinkOpenly: Given that the `csr` field is 12 bits, could you lop off the high order nybble for all of the constants and make them 3 hex digits each?

ThinkOpenly (changes requested):

PR #1259, opened 2025-11-21
feat: adding RV64 riscv-tests suite in build and regression

henrikg-qc:
- RV64 tests for the ISS
 Added rv64-riscv-tests configuration 
 Updated test:riscv-tests target with ability to build and test RV64 configuration
- Corrected many instruction implementations to handle 64bit mode
- Corrected the page walker and address translation
- Added rv64 riscv-tests to the regression

### DIFF
--- a/.github/workflows/regress.yml
+++ b/.github/workflows/regress.yml
@@ -212,7 +212,7 @@ jobs:
         uses: ./.github/actions/singularity-setup
       - name: Run cpp unit tests
         run: ./do test:cpp_hart CONFIG=rv64 JOBS=4
-  regress-riscv-tests-32:
+  regress-riscv-tests:
     runs-on: ubuntu-latest
     env:
       SINGULARITY: 1


cfgs/rv64-riscv-tests.yaml
```diff
@@ -0,0 +1,129 @@
...
+  - [C, "2.0"]
+  - [M, "2.0"]
+  - [Zicsr, "2.0"]
+  - [Zicntr, "2.0"]
+  - [Smrnmi, "1.0"]
+  - [S, "1.11.0"]
+  - [Smpmp, "1.11.0"]
+  - [U, "1.0.0"]
+  - [Zifencei, "2.0.0"]
+  - [Sv32, "1.11.0"]
```
jordancarlin: Sv32 is rv32 only, so it definitely shouldn’t be part of this config. I’m surprised there is no error about an incompatible configuration because of this.

henrikg-qc: Done

jordancarlin: Great. @dhower-qc is this something that will be caught with the config validation in #891?

henrikg-qc: I'm going to resolve this conversation to be able to proceed with merge. The config validity and checks are not part pf this PR. #1260 references "validate all configs"

henrikg-qc:
> > @ThinkOpenly I replaced regress-riscv-tests with 2 new tests in the regression. It now seems to wait for the old test that is not ran anymore. Do you have an idea how to get around this?
> 
> Most likely, the issue is the list of required actions set in the GitHub repo settings. The exact names of the required actions need to be listed there. The way we got around this for `sail-riscv` is to have a dummy action that just depends on all of the previous actions, and then only require that action in the actual repo settings. Then, to change the required actions, anyone can modify the list of dependencies for that one required job. See https://github.com/riscv/sail-riscv/blob/659f5181d7991c957591b3e48e62807ae475b24b/.github/workflows/ci.yml#L105-L113.

Thanks, for this I can rename the 32 bit test regression to match the test name I replaced as they were the same, just slightly different way to call the tests using rake. I'm hoping this will get around it for now.

PR #1260, opened 2025-11-21
feat(ci): validate all config yamls in CI

jordancarlin:
Adds a new `do` target (`./do test:cfgs`) to run `udb validate cfg` on all config files in the `cfg` directory. This found issues in several existing config files, which are also corrected in this PR.

I also added the new target to the standard `regress.yml` GitHub Action CI job to make sure all of the config files stay valid.


jordancarlin:
Very strange. This is working locally, but the GitHub Actions run seems to indicate that it failed to find the `sudo` and `singularity` commands. I'm running `./.github/actions/singularity-setup`, which should install Singularity, and I have no idea how `sudo` would be missing.

PR #1262, opened 2025-11-27
fix(vector): handle rs1==x0 case in vsetvli

adingank-qualcomm:
The following comments refer to "the way we test" and may not be relevant to the PR itself.
- Added vsetvli_rs1_eq_zero test to verify that vl is set to vlmax when rs1 is x0 for cases of SEW=8,32,64 and LMUL=8.
- Also renamed the old vsetvli test to vsetvli_vl_lt_vlmax to reflect the intent better.

### DIFF
--- a/spec/std/isa/inst/V/vsetvli.yaml
+++ b/spec/std/isa/inst/V/vsetvli.yaml
@@ -68,7 +68,7 @@ operation(): |
           unpredictable("Implementations may choose a custom value for vl in the case AVL < (2*VLMAX), so long as ceil(AVL/2) <= vl <= VLMAX");
       }
   } else {
-     CSR[vl].VALUE = vlmax;
+      CSR[vl].VALUE = vlmax;
   }
 
   X[xd] = CSR[vl].VALUE;


spec/std/isa/inst/V/vsetvli.yaml
```diff
@@ -55,18 +55,20 @@ operation(): |
   XReg avl = X[xs1];
   XReg ceil_avl_over_two = (avl + 1) / 2;
 
-  if (avl < vlmax) {
-    CSR[vl].VALUE = avl;
+  if (xs1 == 0) {
+    CSR[vl].VALUE = vlmax;
+  } else if (avl < vlmax) {
+      CSR[vl].VALUE = avl;
```
henrikg-qc, line 61: Throughout the code I think indentation is 2 spaces. and tab = 2 spaces. This file is mixed. It looks like you have setup of tab = 3 spaces,  no functional difference but I thought I'd let you know..

Also note that if everything works on ajit/vset, and if it was based off henrikg/vector you can merge back to vector as henrikg/vector was = vector and then we merged the larger commit from ajit/vvadd there.

We can obsolete henrikg/vector and go back to vector for your development.

henrikg-qc (changes requested):

spec/std/isa/inst/V/vsetvli.yaml
adingank-qualcomm: Thanks, Henrik! My VSCode indentation is at 2 spaces for tab, but I must've added individual spaces (but I could see only one instance of it which I've fixed). 

For some reason, when I change the base branch to `vector` I get back to the 52 commits again!

spec/std/isa/inst/V/vsetvli.yaml
```diff
@@ -55,18 +55,20 @@ operation(): |
...
+      if (RVV_VL_WHEN_AVL_LT_DOUBLE_VLMAX == "ceil(AVL/2)") {
+        CSR[vl].VALUE = ceil_avl_over_two;
+      } else if (RVV_VL_WHEN_AVL_LT_DOUBLE_VLMAX == "VLMAX") {
+          CSR[vl].VALUE = vlmax;
+      } else {
+          unpredictable("Implementations may choose a custom value for vl in the case AVL < (2*VLMAX), so long as ceil(AVL/2) <= vl <= VLMAX");
+      }
   } else {
-    CSR[vl].VALUE = vlmax;
+      CSR[vl].VALUE = vlmax;
```
ThinkOpenly, line 71: Some whitespace changes crept in, FYA.

spec/std/isa/inst/V/vsetvli.yaml
```diff
@@ -55,18 +55,20 @@ operation(): |
   XReg avl = X[xs1];
   XReg ceil_avl_over_two = (avl + 1) / 2;
 
-  if (avl < vlmax) {
-    CSR[vl].VALUE = avl;
+  if (xs1 == 0) {
+    CSR[vl].VALUE = vlmax;
```
ThinkOpenly, line 59: Isn't this only when `xd != 0` ?

adingank-qualcomm: @ThinkOpenly Will fix in the new linked issue #1269

spec/std/isa/inst/V/vsetvli.yaml
adingank-qualcomm: @ThinkOpenly Will fix in #1270

spec/std/isa/inst/V/vsetvli.yaml
adingank-qualcomm: @ThinkOpenly Will fix in #1269

PR #1266, opened 2025-12-03
fix: add missing sign_extend to imm fields of instructions beq, c.beqz and c.bnez (fixes #1265)

### LINKED ISSUE #1265: beq c.beqz and c.bnez have missing sign_extend attribute for imm field
std/isa/inst/C/c.beqz.yaml
std/isa/inst/C/c.bnez.yaml
std/isa/inst/I/beq.yaml

**Expected value**
What do you think the value(s) should be?
sign_extend: true

Will send a fix PR soon.

hansfbaier:


hansfbaier:
> chore:update_golden_appendix

OK, added those mnemonics, which also had missing sign extensions and ran the chores.

ThinkOpenly (approved):

PR #1267, opened 2025-12-04
fix: setting GEM_HOME and GEM_PATH to allow repo under linux home dir

henrikg-qc:
Solving issue of not finding Gemfile when the repo dir is under your home directory

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -106,7 +106,7 @@ if [ -v GITHUB_ACTIONS ]; then
     echo "Running in a GitHub Action"
     CONTAINER_PATH=${ROOT}/.singularity/image.sif
     HOME_PATH=${GITHUB_WORKSPACE}
-    HOME_OPT="--home ${ROOT}/.home"
+    GEM_HOME=${ROOT}/.home/.gems && GEM_PATH=$GEM_HOME
     SINGULARITY_CACHE=--disable-cache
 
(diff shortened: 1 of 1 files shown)


bin/setup
```diff
@@ -106,7 +106,7 @@ if [ -v GITHUB_ACTIONS ]; then
     echo "Running in a GitHub Action"
     CONTAINER_PATH=${ROOT}/.singularity/image.sif
     HOME_PATH=${GITHUB_WORKSPACE}
-    HOME_OPT="--home ${ROOT}/.home"
+    GEM_HOME=${ROOT}/.home/.gems && GEM_PATH=$GEM_HOME
```
adingank-qualcomm, line 109: @henrikg-qc I found that this change to line 109 is not needed, only changing line 140 suffices. Have pushed it to #1268 but it's a much bigger PR so this one should go first!

henrikg-qc: I think we want to apply this for GITHUB_ACTIONS also. This was originally a workaround to use default install paths for ruby gems. Now the container has been updated what is needed and this should be consistent between singularity container and github actions environment. 
@dhower-qc , please advise

dhower-qc: This isn't needed because actions do not run under a home directory. That said, it's not going to hurt to change this to keep things more consistent.

dhower-qc (approved):

PR #1281, opened 2025-12-17
fix(vector): implement vsetvli model and tests

adingank-qualcomm:
- Updated model according to spec for the cases of `rs1==0` and for a `Reserved` value of VSEW
- Added 64 bit vector config with "F" extension to enable the UV TVM of `riscv-tests` 
- Filed Sail-RISC-V [Issue #1434 ](https://github.com/riscv/sail-riscv/issues/1434)

### DIFF
--- a/cfgs/rv64-vector.yaml
+++ b/cfgs/rv64-vector.yaml
@@ -21,6 +21,7 @@ implemented_extensions:
   - [Zifencei, "2.0.0"]
   - [Sv39, "1.11.0"]
   - [V, "1.0.0"]
+  - [F, "2.2"]
 
 params:
   MXLEN: 64
@@ -192,10 +193,8 @@ params:
   TRAP_ON_ECALL_FROM_S: true
   TRAP_ON_ECALL_FROM_U: true
   TRAP_ON_SFENCE_VMA_WHEN_SATP_MODE_IS_READ_ONLY: false
-  # MSTATUS_FS_WRITABLE: false
-  # MSTATUS_VS_WRITABLE: false
   MSTATUS_VS_LEGAL_VALUES: [0, 1, 2, 3]
-  MSTATUS_FS_LEGAL_VALUES: [0]
+  MSTATUS_FS_LEGAL_VALUES: [0, 1, 2, 3]
   MSTATUS_TVM_IMPLEMENTED: false
   NUM_PMP_ENTRIES: 16
   PMP_GRANULARITY: 12
@@ -209,3 +208,5 @@ params:
   FOLLOW_VTYPE_RESET_RECOMMENDATION: true
   VLEN: 128
   ELEN: 64
+  HW_MSTATUS_FS_DIRTY_UPDATE: never
+  MUTABLE_MISA_F: false
--- a/spec/std/isa/inst/V/vsetvli.yaml
+++ b/spec/std/isa/inst/V/vsetvli.yaml
@@ -57,25 +57,10 @@ operation(): |
       XReg avl = X[xs1];
       XReg ceil_avl_over_two = (avl + 1) / 2;
 
-      if (xs1 == 0 && xd != 0) {
+      if (xs1 == 0)
+      {
         CSR[vl].VALUE = vlmax;
-      }
-      if (xs1 == 0 && xd == 0) {
-        XReg vl = CSR[vl].VALUE;
-      }
-      # Sighting of a potential spec bug:
-      # Chapter 30, Section 6.2 says:
-      # "When rs1 is not x0, the AVL is an unsigned integer held in the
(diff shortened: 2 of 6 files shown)


cfgs/rv64-vector.yaml
```diff
@@ -192,9 +192,9 @@ params:
   TRAP_ON_ECALL_FROM_S: true
   TRAP_ON_ECALL_FROM_U: true
   TRAP_ON_SFENCE_VMA_WHEN_SATP_MODE_IS_READ_ONLY: false
-  MSTATUS_FS_WRITABLE: false
-  MSTATUS_VS_WRITABLE: false
-  MSTATUS_VS_LEGAL_VALUES: [0]
+  # MSTATUS_FS_WRITABLE: false
+  # MSTATUS_VS_WRITABLE: false
```
dhower-qc: @ThinkOpenly is going to ask if you need to keep these ;)

cfgs/rv64-vector.yaml
```diff
@@ -203,7 +203,6 @@ params:
   U_MODE_ENDIANNESS: little
   UXLEN: [64]
   MSTATEEN_ENVCFG_TYPE: rw
-  # HSTATEEN_ENVCFG_TYPE: rw
```
dhower-qc, line 206: and this

spec/std/isa/inst/V/vsetvli.yaml
```diff
@@ -33,45 +33,73 @@ operation(): |
       || ((new_vtype & 8'd0) != 0)    # reserved bits
       || (new_vtype[5] == 1)          # reserved vsew encoding
       || (new_vtype[2:0] == 3'b100)   # reserved vlmul encoding
-      || (xlen() == 32 && new_vtype[2:0] == 3'b101)) # reserved LMUL in RV32
+      # || (xlen() == 32 && new_vtype[2:0] == 3'b101) # reserved LMUL in RV32
```
dhower-qc, line 36: and this

spec/std/isa/inst/V/vsetvli.yaml
```diff
@@ -33,45 +33,73 @@ operation(): |
...
+      # Chapter 30, Section 6.2 says:
+      # "When rs1 is not x0, the AVL is an unsigned integer held in the
+      # x register specied by rs1, and the new vl value is also written
+      # to the x register specied by rd." (Below the table)
+      #
+      # Expected:
+      # Additional check for rd != 0 since for the case rd == x0 the new
+      # value should *not* be written because x0 is the zero register.
+      # The first row of the table should be split into two rows, one for
+      # rd == x0 and one for rd != x0.
```
dhower-qc: If you've sent an email to the unpriv list, might be good to add a link to the groups.io thread here

cfgs/rv64-vector.yaml
adingank-qualcomm: Agreed... And removed! :-)

spec/std/isa/inst/V/vsetvli.yaml
adingank-qualcomm: This is not a sighting because the zero register can be written to!
Removed the comments.

spec/std/isa/inst/V/vsetvli.yaml
adingank-qualcomm: I'll discuss this particular decision with @jmawet in the next iteration, since I couldn't find this behavior described in the spec....

ThinkOpenly: Do we have an issue open for this? (How do we not lose track of it?)

I wonder if the original limitation was added due to this text from the spec:
> When LMUL < SEWmin/ELEN, there is no guarantee an implementation would have
enough bits in the fractional vector register to store at least one element, as
VLEN=ELEN is a valid implementation choice. For example, with VLEN=ELEN=32,
and SEWmin=8, an LMUL of 1/8 would only provide four bits of storage in a vector
register.

...but that does not include a dependence on XLEN. hmm

spec/std/isa/inst/V/vsetvli.yaml
```diff
@@ -33,44 +33,56 @@ operation(): |
...
+                {
+                  CSR[vl].VALUE = vlmax;
+                } else
+                  {
+                    unpredictable("Implementations may choose a custom value for vl in the case AVL < (2*VLMAX), so long as ceil(AVL/2) <= vl <= VLMAX");
+                  }
+            } else
+              {
+                CSR[vl].VALUE = vlmax;
+              }
```
ThinkOpenly, line 83: Any particular reason the coding style was changed here?
(I prefer the previous style, myself. The indentation here is a bit odd, and things don't line up particularly well, IMHO.)

PR #1282, opened 2025-12-17
feat: generate `container.def` from `Dockerfile` using `spython`

jordancarlin:
Removes manually maintained `container.def` file and generates it as a pre-commit hook using `spython` from the `Dockerfile`.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -136,10 +136,14 @@ elif [ "${CONTAINER_TYPE}" == "docker" ] || [ "${CONTAINER_TYPE}" == "podman" ];
         GIT_PATH=$(git rev-parse --git-common-dir | tr -d '\n' | xargs dirname)
         CONTAINER_RUN_FLAGS="${CONTAINER_RUN_FLAGS} -v ${GIT_PATH}:${GIT_PATH}${SELINUX_LABEL}"
     fi
-    CONTAINER_RUN_FLAGS="${CONTAINER_RUN_FLAGS} --tmpfs /.cache --tmpfs /.npm"
+    CONTAINER_RUN_FLAGS="${CONTAINER_RUN_FLAGS} -e HOME"
+    CONTAINER_RUN_FLAGS="${CONTAINER_RUN_FLAGS} --tmpfs ${HOME}"
 
-    # run as the host user, and bind /etc/{passwd,group} to get the ids
-    CONTAINER_RUN_FLAGS="${CONTAINER_RUN_FLAGS} --user $(id -u):$(id -g) -v /etc/passwd:/etc/passwd:ro -v /etc/group:/etc/group:ro"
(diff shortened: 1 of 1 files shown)


jordancarlin:
The new pre-commit hook passes locally. I'll have to look into why it is hanging in CI.

bin/generate-container-def
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+ROOT=$(dirname $(dirname $(realpath ${BASH_SOURCE[0]})))
+source $ROOT/bin/setup
+
+echo "# This file is auto-generated by bin/generate-container-def" > container.def
+spython recipe .devcontainer/Dockerfile | sed '/%runscript/,$d' >> container.def
```
dhower-qc: This line isn't going to run in the container, meaning that spython needs to be installed on the host.

I think this works?:

```bash
$PYTHON -m spython recipe ...
```

bin/generate-container-def
```diff
@@ -0,0 +1,8 @@
...
+chmod 444 container.def
```
dhower-qc, line 21: Does this still work if you run generate-container-def after the file is marked read-only (do you need to remove it first?)

bin/generate-container-def
jordancarlin: That's what I had initially, but it gives the following error (which is why I changed it):
```
/opt/venv/bin/python3: No module named spython.__main__; 'spython' is a package and cannot be directly executed
```

I've modified the script to directly use the `${RUN}` variable set by the setup script to ensure it runs in the container.

bin/generate-container-def
jordancarlin: I've been using the Devcontainer/Podman container environment, and it looks like the user inside of that container is currently root, so read-only doesn't actually do anything, which is why I didn't notice this. The read-only files in #891 will probably run into the same issue.

I've updated the script (and tested it from the Singularity environment) to remove the file first, but we should probably look into running the Docker containers as a non-root user (in a separate PR).

dhower-qc: in #891, I tried to make sure I deleted the files before generation, so hopefully it's ok

jordancarlin: Sorry, to clarify, the issue is that if using the Docker/Podman/Devcontainer environment, setting the file to be read-only doesn't do anything because the user is root inside the container. So while it won't break anything, the benefit of making the files read-only is lost.

jordancarlin:
@dhower-qc It looks like you updated this PR but didn't rerun the `container.def` generation script.

dhower-qc:
> @dhower-qc It looks like you updated this PR but didn't rerun the `container.def` generation script.

Oops. I ran spython manually so it missed the warning comment.

jordancarlin:
@dhower-qc looks like you haven't pushed the latest changes that we discussed regrading heredocs.

PR #1289, opened 2025-12-27
fix: update c.nop encoding and assembly to support HINTs (#1177)

### LINKED ISSUE #1177: `c.nop` encoding is different than riscv-opcodes
`c.nop.yaml` has a match field with a fully constrained value of `0000000000000001`

**Expected value**
riscv-opcodes encodes `c.nop` as `000?00000?????01`. Note the wildcards.

It looks like the divergence is because of how hints are handled. From the spec:

Image

> The C.NOP code points with imm≠0 encode HINTs.

Discovered while comparing #1090 to the riscv-opcodes version of the header.

16srivarshitha:
## Description
This PR resolves the encoding divergence for c.nop between the unified database and riscv-opcodes, as reported in issue #1177.
## Changes
YAML (`c.nop.yaml`):

Updated the match pattern from the fully constrained value `0x0001` to `000-00000-----01`, using wildcards for the immediate field
This change enables the database to correctly recognize c.addi x0, x0, imm (where imm ≠ 0) as c.nop HINT variants

Generator (`template_helpers.rb`):

Enhanced assembly_fmt and assembly_fmt_args to support dynamic formatting based on the immediate value:

When imm = 0: displays as c.nop
When imm ≠ 0 (HINT variant): displays as c.nop <imm>

## Verification

- Confirmed the match field correctly enforces the rd = x0 constraint (bits 11-7)
- Verified the Ruby generator properly switches between assembly formats based on the immediate value

Fixes #1177


spec/std/isa/inst/C/c.nop.yaml
```diff
@@ -8,13 +8,23 @@ kind: instruction
 name: c.nop
 long_name: Non-operation
 description: |
+  C.NOP does not change any user-visible state, except for advancing the pc.
   C.NOP expands into `addi x0, x0, 0`.
+  C.NOP with non-zero immediate values are HINTs.
 definedBy:
+  anyOf:
+    - C
+    - Zca
```
ThinkOpenly: This is redundant with the "extension" attribute, below, so remove these lines.

spec/std/isa/inst/C/c.nop.yaml
```diff
@@ -8,13 +8,23 @@ kind: instruction
...
+assembly: "c.nop imm"
```
ThinkOpenly: The "assembly" attribute does not include the instruction mnemonic, so this should be changed.
```suggestion
assembly: imm
```
Also, this is redundant with the "assembly" attribute below. There can be only one.

Also, the "extension" below needs to be under the "definedBy" above, and this was inserted between them.

All that being said, this "imm" operand is _optional_, and UDB doesn't have a way to specify optional operands at present. Hopefully, this will come with #1527, but that might not be soon.

backends/cpp_hart_gen/lib/template_helpers.rb
```diff
@@ -13,6 +13,15 @@ def to_cxx
 end
 
 module Udb
+class Instruction
+  def assembly_fmt(xlen)
+    if name == "c.nop"
```
ThinkOpenly: I'd really like to avoid adding special cases for specific instructions in the Ruby code. I suspect this will be easier when operands can be explicitly marked "optional" in the YAML when #1527 is merged.

ThinkOpenly (changes requested):
Thank you for this contribution. There are a few issues with it for which I've added comments.

A final implementation is perhaps not possible until optional operands are supported.

If you wish to address the addressable comments, please do. This PR will probably need to stay open for some time, though, and then it will need substantial further work. Shall we leave it open until then, or would it be easier to close it and revisit the issue later?

backends/cpp_hart_gen/lib/template_helpers.rb
```diff
@@ -13,7 +13,7 @@ def to_cxx
 end
 
 module Udb
-  class Instruction
+class Instruction
```
ThinkOpenly, line 16: I think these changes are inadvertent, and need to be reverted?

PR #1292, opened 2025-12-31
feat(vector): Implemented vnsra instruction

latifbhatti:
**Description:**

This PR implements the VNSRA (Vector Narrow Shift Right Arithmetic) instruction. The implementation is complete with:

-  Vector masking logic.
- Tail-agnostic handling scheme.
- Implemented vector register grouping logic supporting all LMUL values as per the RISC-V V-extension specification (1, 2, 4, 8, 1/2, 1/4, 1/8).


cfgs/rv32-vector.yaml
```diff
@@ -54,7 +54,7 @@ params:
   REPORT_ENCODING_IN_MTVAL_ON_ILLEGAL_INSTRUCTION: true
   MTVAL_WIDTH: 32
   PMA_GRANULARITY: 12
-  PHYS_ADDR_WIDTH: 34
+  PHYS_ADDR_WIDTH: 32
```
ThinkOpenly, line 57: This seems quite unrelated, and incorrect. A reference from the "privileged" spec:
> The Sv32 page-based virtual-memory scheme described in Section 12.3 supports 34-bit physical addresses for RV32

dhower-qc:
We are almost ready to run riscv-vector-tests, so we are just waiting on that capability to test this instruction before this PR merges

PR #1297, opened 2026-01-06
fix(data): update `mseccfg` requirements

jordancarlin:
`mseccfg` was confusingly referred to as optional, but it really meant that it exists if and only if any extension that adds a field to it is implemented.

See https://github.com/riscv/riscv-isa-manual/issues/2549 and https://github.com/riscv/riscv-isa-manual/pull/2550 .


jordancarlin:
Looks like merging #891 closed all of the PRs built on top of that branch with no option to reopen them.

PR #1298, opened 2026-01-08
feat: memory map support from JSON 

henrikg-qc:
Includes memory map definition from JSON file by --mm or --memory-map option.
Example memmap.json under /cfgs
Without --memory-map option dense memory model is created based on ELF ALLOC sections range.


backends/cpp_hart_gen/cpp/src/iss.cpp
```diff
@@ -33,6 +39,35 @@ int parse_cmdline(int argc, char **argv, Options &options) {
...
+  if(!memmap.empty()) {
+    std::ifstream f(memmap);
+    json data = json::parse(f);
+    regions = data["regions"];
+
+    for (const auto& region : regions) {
+      auto type = region["type"];
+      if (type == "ram") {
+        std::string base = region["base"]["value"];
+        std::string size = region["size"]["value"];
```
jordancarlin, line 55: There should be documentation somewhere that only the base and size from the ram region are used and the rest of the memory map is currently ignored.

cfgs/memmap.json
```diff
@@ -0,0 +1,76 @@
+{
+  "regions": [
+    {
+      "type": "rom",
```
jordancarlin, line 4: The Sail version of this memory map (which this is based on) does not have a `type` key. What's the rationale for adding one?

henrikg-qc: To identify the memory type. RAM is what is used by the Dense memory model to place program and data

backends/cpp_hart_gen/cpp/src/iss.cpp
henrikg-qc: Problem is the definition of "currently". As the memory model get PMA implemented, they will be picked up from the provided JSON.
Do we have a release notes? probably not as we don't do releases. If there is a place that makes sense to put it we can.

cfgs/memmap.json
jordancarlin: The issue is that the RISC-V spec does not have a concept of a "ROM" region. All regions are either "main memory" or "io memory". Main memory regions support read and write of all access widths, are assumed to be idempotent, and have constraints on the memory ordering. Anything that does not meet all of those requirements is "io memory".

jordancarlin: We've been discussing this a bunch in https://github.com/riscv/sail-riscv/pull/1468#discussion_r2665802066.

henrikg-qc: That's fine, The rom type reference came from the SAIL example comment. 
We can change the type values we use for the dense memory model to be looking for "main memory" and other regions defined as "io memory" to follow the spec

PR #1302, opened 2026-01-08
fix(data): update `mseccfg` requirements

jordancarlin:
`mseccfg` was confusingly referred to as optional, but it really meant that it exists if and only if any extension that adds a field to it is implemented.

See https://github.com/riscv/riscv-isa-manual/issues/2549 and https://github.com/riscv/riscv-isa-manual/pull/2550.


spec/std/isa/csr/mseccfgh.yaml
```diff
@@ -16,6 +16,12 @@ definedBy:
   allOf:
     - xlen: 32
     - extension:
-        name: Sm
-        version: ">= 1.12"
+        allOf:
+          - name: Sm
+            version: ">= 1.12"
+          - anyOf:
+              - name: Zkr
```
dhower-qc, line 23: does mseccfg it 'exist' if one of these extensions is implemented, but none of them use the upper bits of mseccfg?

spec/std/isa/csr/mseccfg.yaml
```diff
@@ -14,6 +14,12 @@ length: 64
...
-    name: Sm
-    version: ">= 1.12"
+    allOf:
+      - name: Sm
+        version: ">= 1.12"
+      - anyOf:
+          - name: Zkr
+          - name: Smmpm
+          - name: Zicfilp
+          - name: Zicfilp
```
dhower-qc, line 23: this seems like a maintenance headache, but I don't have a solution for that.

spec/std/isa/csr/mseccfgh.yaml
jordancarlin: Good question. Asked in https://github.com/riscv/riscv-isa-manual/issues/2574.

spec/std/isa/csr/mseccfg.yaml
jordancarlin: Yeah. Not ideal, but seems unavoidable.

spec/std/isa/csr/mseccfgh.yaml
jordancarlin: Per https://github.com/riscv/riscv-isa-manual/issues/2574, if `mseccfg` is implemented, then `mseccfgh` is too no matter which bits are used.

dhower-qc (approved):

PR #1306, opened 2026-01-08
fix(ci): don't push docker image to Dockerhub in PRs

jordancarlin:
Build the Docker image once and then pass it between the different CI jobs as a GitHub artifact. This avoids uploading it to any registry during the PR process.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -247,4 +247,4 @@ jobs:
           path: "_site"
       - name: Deploy to GitHub Pages
         id: deployment
-        uses: actions/deploy-pages@v4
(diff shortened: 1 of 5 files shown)


jordancarlin:
@dhower-qc I saw you trying to use GHCR in #1301 to work around the issue with Dockerhub because PRs can't access secrets. Instead of trying to upload the image to GHCR (which might behave strangely depending on how individual users have their GitHub accounts set up), this PR directly passes the Docker image between to the different jobs using the native GitHub Action artifact approach. It means that the individual steps run with the Docker setup instead of the Devcontainer setup, but I don't think that is an issue.

jordancarlin:
@dhower-qc this approach is now passing CI from a fork.

PR #1314, opened 2026-01-11
ci: add Ruff linting for Python tools

### LINKED ISSUE #1235: add Python linting
We should probably add some kind of Python linting (I'm a fan of Ruff) to CI/pre-commit to help with these kinds of things in the future.

_Originally posted by @jordancarlin in https://github.com/riscv-software-src/riscv-unified-db/pull/1231#pullrequestreview-3429000621_

lucifer4330k:
## Summary
This PR adds Ruff linting to pre-commit and CI for Python code, resolving #1235.

## Changes
- Added Ruff pre-commit hooks (`ruff` and `ruff-format`) scoped to `tools/python/`
- Enabled additional lint rules in `pyproject.toml`:
  - `ANN` - type annotation rules
  - Plus existing: `F`, `E`, `W`, `I`, `UP`, `B`, `A`, `EXE`, `RUF`, `PTH`, `SIM`
- Fixed lint issues in `tools/python/profile_extensions.py` and `tools/python/udb.py`:
  - Added return type annotations
  - Added function parameter type hints
  - Used `pathlib` idioms (`Path.open()` instead of `open()`)
  - Flattened nested conditionals
- Added SPDX license headers to `pyproject.toml`

## Testing
- `pre-commit run ruff --all-files` passes ✅
- `ruff check tools/python` passes ✅

## Next Steps
The Ruff hooks are currently scoped to `tools/python/` to avoid breaking existing code in other directories. Coverage can be expanded incrementally by fixing or suppressing violations elsewhere.

Fixes #1235

#lfx


.pre-commit-config.yaml
```diff
@@ -104,6 +104,16 @@ repos:
         # TEMPORARILY DISABLE CLANG-FORMAT IN LIBHART
         # WIll RE-ENABLE WHEN NEXT PATCH COMES THROUGH
         exclude: backends/cpp_hart_gen
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.9.9
```
jordancarlin: This is a very old version of Ruff. The latest release is 0.14.11. There have been a number of significant changes, and it is likely that some of the Python code will not pass with the latest version of the linter without additional fixes.

.pre-commit-config.yaml
```diff
@@ -104,6 +104,16 @@ repos:
...
+    hooks:
+      - id: ruff
```
jordancarlin: The `ruff` id is deprecated in favor of `ruff-check` (to match `ruff-format`).

```suggestion
      - id: ruff-check
```

pyproject.toml
```diff
@@ -0,0 +1,49 @@
+# SPDX-FileCopyrightText: 2024-2025 Contributors to the RISCV UnifiedDB <https://github.com/riscv-software-src/riscv-unified-db>
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+[build-system]
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
```
jordancarlin: Why are you adding `setuptools` as a build backend? UDB does not currently build a Python package, so no build system should be required.

pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+[tool.ruff]
+line-length = 100
+target-version = "py312"
+exclude = [
+    ".git",
+    ".venv",
+    ".devcontainer",
+    "ext/rbi-central",
+    "tools/ruby-gems/*/ext/rbi-central",
+]
```
jordancarlin: Is specifying this necessary? Several of these are in the default Ruff ignore list and I don't think `rbi-central` has any Python files, so Ruff should skip it automatically.

pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+ignore = [
+    "E501",   # line too long (ruff has line-length setting, but doesn't check by default)
+    "ANN002", # missing type annotation for *args
+    "ANN003", # missing type annotation for **kwargs
+    "ANN401", # dynamically typed expressions are disallowed
+]
+
+[tool.ruff.lint.isort]
+known-first-party = ["udb", "idlc"]
```
jordancarlin: There is no Python module in the repo named `idlc` and I suspect Ruff can figure out that `udb` is first-party on its own since it is in the same directory as the other Python scripts. Can this be dropped?

.pre-commit-config.yaml
jordancarlin: If you are enabling the Ruff formatter (which I'm in favor of since it is significantly faster), then we should drop the Black formatter. Having two Python formatters competing (even if they do produce *almost* identical formats) is not a good idea.

jordancarlin: This can be replaced by the `UP` rules in Ruff.

jordancarlin (changes requested):
I haven't reviewed the list of rules that you enabled yet, but here are comments on the rest of the PR.

I'd also be in favor of enabling Ruff on the whole repo at once. There isn't that much Python code, so it shouldn't be an issue to switch it all over in one PR. I'm worried that having multiple linters/formatters at the same time while we are transitioning would be confusing.

.pre-commit-config.yaml
lucifer4330k: so , its good to use the previous versions na , or should i use the latest version ?

.pre-commit-config.yaml
jordancarlin: This was marked as resolved, but the latest version still uses `ruff` instead of `ruff-check`.

backends/generators/c_header/generate_encoding.py
```diff
@@ -3,23 +3,22 @@
...
-import yaml
 
 # Add parent directory to path to import generator.py
 parent_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 sys.path.append(parent_dir)
 
 # Import functions from generator.py
-from generator import (
-    load_instructions,
+from generator import (  # noqa: E402
```
jordancarlin: Why is the noqa necessary? E402 says `sys.path.append` is allowed between imports and shouldn’t throw a warning: https://docs.astral.sh/ruff/rules/module-import-not-at-top-of-file/.

pyproject.toml
```diff
@@ -0,0 +1,39 @@
+# SPDX-FileCopyrightText: 2024-2025 Contributors to the RISCV UnifiedDB <https://github.com/riscv-software-src/riscv-unified-db>
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+[tool.ruff]
+line-length = 100
```
jordancarlin, line 5: Black (the formatter we were using before) used a line length of 88. I'm fine with increasing it to 100 as you've done here, but the increased line length is responsible for most of the diff in the Python files. @dhower-qc @ThinkOpenly any opinions on this?

.pre-commit-config.yaml
lucifer4330k: I have modified it to ruff-check

pyproject.toml
```diff
@@ -0,0 +1,39 @@
...
+target-version = "py312"
+exclude = [
+    ".git",
+    ".venv",
+    ".devcontainer",
+    "ext",
+]
```
jordancarlin: Use `extend-exclude` so that we retain all of the default ignore patterns:

```suggestion
extend-exclude = [ "ext" ]
```

pyproject.toml
```diff
@@ -0,0 +1,39 @@
...
+    "B016",   # cannot raise a literal (legacy code)
+    "B904",   # raise from err (requires careful refactoring)
+    "SIM102", # nested if statements (readability preference)
+    "SIM108", # ternary operator (readability preference)
+    "SIM118", # use key in dict instead of key in dict.keys()
+    "RUF005", # consider list unpacking instead of concatenation
+]
+
+[tool.ruff.lint.pylint]
+max-args = 5
```
jordancarlin: Drop this. The Pylint rules are not even enabled.

pyproject.toml
```diff
@@ -0,0 +1,39 @@
...
+    "E",    # pycodestyle errors
+    "W",    # pycodestyle warnings
+    "I",    # isort - import related rules
+    "UP",   # pyupgrade - suggest upgrades for newer Python
+    "B",    # flake8-bugbear - bugbear rules
+    "A",    # flake8-builtins - shadow builtins
+    "EXE",  # flake8-executable - executable file shebangs
+    "RUF",  # ruff specific rules
+    "SIM",  # flake8-simplify - simplifications
+]
```
jordancarlin, line 21: There are quite a few other rules that I think we should activate, but I'm fine merging this as an initial version that just gets Ruff running and then we can add to it from here.

pyproject.toml
```diff
@@ -0,0 +1,39 @@
...
+ignore = [
+    "E501",   # line too long (handled by formatter)
+    "E711",   # comparison to None (too many to fix now)
+    "B016",   # cannot raise a literal (legacy code)
+    "B904",   # raise from err (requires careful refactoring)
+    "SIM102", # nested if statements (readability preference)
+    "SIM108", # ternary operator (readability preference)
+    "SIM118", # use key in dict instead of key in dict.keys()
+    "RUF005", # consider list unpacking instead of concatenation
+]
```
jordancarlin, line 31: We should work to eliminate most of these ignores (many of them are legitimate Python issues), but I'm ok deferring that to another PR.

pyproject.toml
dhower-qc: we use 100 for clang-format in the ISS. I think it's also set to 100 for rubocop, though we haven't been running that.

so, 100 makes sense to me

backends/generators/c_header/generate_encoding.py
lucifer4330k: i have removed noqa and i have added E402 in the ignore list .

jordancarlin: I'm still not clear why we need to ignore the rule at all. What is causing Ruff to fail on that rule? The Ruff documentation says that `sys.path.append` is allowed between imports.

lucifer4330k: The issue was that the original code had:
parent_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
sys.path.append(parent_dir)

The variable assignment was causing E402 to trigger, because Ruff only allows specific statements (like [sys.path.append()] between imports, not arbitrary variable assignments.

jordancarlin: All of the other Python files already had it inlined, so this is better for consistency anyway.

jordancarlin (approved):

requirements.txt
```diff
@@ -5,3 +5,4 @@ reuse==5.0.2
 ruamel.yaml==0.18.6
 mergedeep==1.3.4
 pytest==8.3.4
+ruff==0.14.11
```
jordancarlin, line 9: This needs to bump the `bin/.container_tag` version because it modifies the installed Python tools.

tools/ruby-gems/udb/python/yaml_resolver.py
```diff
@@ -1,31 +1,30 @@
...
 schemas = {}
 
-udb_root = lambda d: (
-    d if os.path.exists(os.path.join(d, "do")) else udb_root(os.path.dirname(d))
-)
+
+def udb_root(d):
+    return d if os.path.exists(os.path.join(d, "do")) else udb_root(os.path.dirname(d))
+
+
```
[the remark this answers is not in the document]
jordancarlin: @lucifer4330k looks like accepting this Copilot suggestion broke the ruff formatting check. You need to rerun pre-commit on this PR and then commit the results.

PR #1330, opened 2026-01-16
fix: sync `Gemfile.lock` `BUNDLED_WTH` version to match new Ruby version

jordancarlin:
Avoids the following new warning introduced in #1301:
```
Bundler 2.6.9 is running, but your lockfile was generated with 2.4.20. Installing Bundler 2.4.20 and restarting using that version.
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -369,4 +369,4 @@ RUBY VERSION
    ruby 3.2.3p157
 
 BUNDLED WITH
-   2.4.20
+   2.6.9


dhower-qc:
was this something you hand-coded? The way to fix this is to run `rake chore:update_deps`, which will, among other things, run `bundle update` on all the gems and the top-level.

jordancarlin:
This was done using `bundle update --bundler=2.6.9` to set the `BUNDLED_WITH` version back to the default version for the new Ruby. As far as I can tell, `./do chore:update_deps` will update all of the dependencies that Bundle installs, but not the version of Bundle itself. Bundle automatically downgrades itself when it detects a different `BUNDLED_WITH`, but I figured we should use the default version instead of having it automatically install an older version.

PR #1339, opened 2026-01-17
fix: resolve redundant method definitions and void context in lib/udbFix/redundant methods

pm-ju:
While auditing the codebase with `Rubocop` to improve repository health, I identified over 200 offenses. Upon manual review, I filtered out the cosmetic styling issues to isolate **two functional logic defects** that require immediate attention:

1.  **Duplicate Method Definition (`lib/udb/cfg_arch.rb`):**
    The method `convert_monospace_to_links` was defined twice (lines 1316 and 1477). The second definition silently overwrote the first (method shadowing), which creates ambiguity and potential maintenance bugs. I have removed the redundant definition to ensure deterministic behavior.

2.  **Void Context (`lib/udb/obj/csr.rb`):**
    The variable `@field_hash` was being evaluated in a void context (line 494), effectively executing dead code. I have removed this unneeded execution path.

## Verification
I ran the linter specifically on the affected files to confirm the logic issues are resolved:
```bash
bundle exec rubocop lib/udb/cfg_arch.rb lib/udb/obj/csr.rb

### CODE REVIEW

File: tools/ruby-gems/udb/lib/udb/obj/csr.rb [hunk 1]
```diff
@@ -491,7 +491,6 @@ def fields_for(effective_xlen)
 
     # @return [Hash<String,CsrField>] Hash of fields, indexed by field name
     def field_hash
-      @field_hash unless @field_hash.nil?
```
dhower-qc, line 494: This is indeed a bug, but the intent is that it `return @field_hash unless @field_hash.nil?` (it's caching the result)

dhower-qc (changes requested):

tools/ruby-gems/udb/lib/udb/obj/csr.rb
pm-ju: Ah, good catch! I assumed it was dead code, but memoization makes much more sense here. I'll update it to explicitly return the value.

PR #1342, opened 2026-01-19
feat: don't run as root in the Docker image

jordancarlin:
Running as root in the devcontainer meant that read-only files weren't actually read-only. This is also probably better from a security standpoint.


bin/.container-tag
```diff
@@ -1 +1 @@
-0.13
+0.14
```
dhower-qc, line 1: We have several PRs right now that bump the container version. Nothing wrong here, just want to merge carefully

dhower-qc:
see #1282, which changes setup to run as the host user with Docker (same as singularity default)

jordancarlin:
I don't think that will work for the devcontainer which runs fully inside the container.

PR #1347, opened 2026-01-21
feat: added Rust Code Generator for RISC-V Instructions and CSRs

yoursanonymous:
This PR adds a Rust code generator that transforms RISC-V instruction and CSR metadata into
clean, strongly-typed Rust definitions.

It outputs:
- Instruction enums
- Encoding metadata (opcode/funct3/funct7)
- CSR constants grouped by privilege level
- CSR enum with address lookup

Why this helps:
Maintaining RISC-V instruction tables by hand is error-prone and hard to keep in sync with the spec.
This generator automates the process and keeps Rust tooling aligned with official definitions.

Key features:
- Supports RV32/RV64/BOTH
- Optional extension filtering
- Privilege-level CSR grouping
- Simple CLI interface


dhower-qc (changes requested):
Thanks for adding a test. To make it more robust, can you also check in a golden output that we can check with regressions? This would be similar to the outputs we use for the instruction appendix (see `./do chore:update_golden_appendix`)

yoursanonymous:
Yes, I will add a golden reference output similar to the instruction appendix workflow and update the regression test accordingly.

PR #1365, opened 2026-01-23
feat: add Ssube extension definition

### LINKED ISSUE #1203: Add "Ssube" extension definition
Pretty difficult to find a definition. Perhaps this from the "Profiles & Bases & Extensions" spreadsheet:

ratification package: priv 1.11
description: sstatus.ube=1 Big-endian user-space.

(Sounds like fun! :-)

ishaan-arora-1:
## Summary

Add YAML definition for the Ssube (Big-endian User Mode) extension.

The Ssube extension indicates that the implementation supports big-endian user mode, meaning `sstatus.UBE` can be set to 1. When enabled, data accesses in U-mode use big-endian byte ordering. Instructions are always fetched as little-endian regardless of this setting.

This extension requires the `U_MODE_ENDIANNESS` parameter to be either `big` or `dynamic`.

## Test plan

- [x] YAML syntax valid
- [x] Validates against `ext_schema.json`
- [x] Pre-commit hooks pass

Closes #1203

### DIFF
--- a/spec/std/isa/ext/Ssube.yaml
+++ b/spec/std/isa/ext/Ssube.yaml
@@ -1,0 +1,25 @@
+# Copyright (c) Ishaan Arora
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../schemas/ext_schema.json
+
+$schema: "ext_schema.json#"
+kind: extension
+name: Ssube
+long_name: Big-endian User Mode
+description: |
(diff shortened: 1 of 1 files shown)


spec/std/isa/ext/Ssube.yaml
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
dhower-qc: Give yourself credit!

PR #1379, opened 2026-01-24
fix(ci): enable native multi-platform Docker builds for amd64 and arm64

### LINKED ISSUE #1378: No amd64 image uploaded to Docker Hub
Starting with #1282, there is only an arm64 image published to Docker Hub. The amd64 image is now missing. Before #1282 there was only an amd64 image, so we see to have reversed the situation instead of actually solving the issue.

vkprogrammer-001:
## Problem

Starting with #1282, only arm64 images were being published to Docker Hub, while amd64 images were missing. The issue was that both platform builds were running in parallel with the same tag, causing the last build to overwrite the first.

## Solution

This PR implements distributed native builds using the approach recommended by Docker (https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners) as suggested by @jordancarlin .

### Changes:
- Split `publish-docker-container` into two separate jobs:
  - `build-docker-container`: Builds natively on both platforms in parallel
  - `merge-docker-container`: Merges the builds into a single multi-platform manifest
- Each platform builds natively (no QEMU emulation) for better performance:
  - `ubuntu-24.04` → builds `linux/amd64`
  - `ubuntu-24.04-arm` → builds `linux/arm64`
- Uses digest-based push to prevent tag conflicts
- Digests are saved as artifacts and merged in the final step
- Added checkout step to merge job to read container tag file

### Benefits:
- ✅ Both amd64 and arm64 platforms available under single tag
- ✅ Native builds for better performance (no QEMU overhead)
- ✅ No tag conflicts or overwrites
- ✅ Users automatically get the correct platform when pulling

Fixes #1378
Related: #1282

### DIFF
--- a/bin/.functions.sh
+++ b/bin/.functions.sh
@@ -23,15 +23,15 @@ get_container_type() {
   elif [ -v SINGULARITY ]; then
   container_type=singularity
   else
-    echo -e "UDB tools run in a container. Docker, Podman, and Singularity/Apptainer are supported.\\n\\n1. Docker\\n2. Podman\\n3. Singularity\\n"
+    echo -e "UDB tools run in a container. Docker, Podman, and Singularity/Apptainer are supported.\\n\\n1. Docker\\n2. Podman\\n3. Singularity\\n" >&2
     while true; do
-      echo "Which would you like to use? (1/2/3) "
+      echo "Which would you like to use? (1/2/3) " >&2
       read -r ans
       case $ans in
           [1]* ) container_type=docker; break;;
           [2]* ) container_type=podman; break;;
           [3]* ) container_type=singularity; break;;
-          * ) echo -e "\\nPlease answer 1, 2, or 3.";;
+          * ) echo -e "\\nPlease answer 1, 2, or 3." >&2;;
       esac
     done
   fi
--- a/bin/setup
+++ b/bin/setup
@@ -42,6 +42,35 @@ if [ ! -f "${ROOT}/.container-type" ]; then
   echo "${CONTAINER_TYPE}" > "${ROOT}/.container-type"
 fi
 
+# Check if container type or tag has changed
+NEEDS_CLOBBER=false
+
+if [ -f "${ROOT}/.cache/previous-container-type" ]; then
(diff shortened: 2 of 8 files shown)


.github/workflows/container.yml
```diff
@@ -65,6 +66,14 @@ jobs:
           TAG=$(cat bin/.container-tag)
           echo "TAG=$TAG" >> $GITHUB_ENV
 
+      - name: Set platform
+        run: |
+          if [ "${{ matrix.runner }}" == "ubuntu-24.04" ]; then
+            echo "PLATFORM=linux/amd64" >> $GITHUB_ENV
+          else
+            echo "PLATFORM=linux/arm64" >> $GITHUB_ENV
+          fi
```
jordancarlin: Probably easier to modify the matrix to include these variables and then drop this step.

.github/workflows/container.yml
```diff
@@ -74,9 +83,65 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3
 
-      - name: Build and push
+      - name: Build and push by digest
+        id: build
         uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83 # v6
         with:
-          push: true
+          context: .
```
jordancarlin: Why is the context being changed?

.github/workflows/container.yml
```diff
@@ -74,9 +83,65 @@ jobs:
...
+      - name: Create manifest list and push
+        working-directory: /tmp/digests
+        run: |
+          docker buildx imagetools create \
+            --tag riscvintl/udb:${{ env.TAG }} \
+            $(printf 'riscvintl/udb@sha256:%s ' *)
+
+      - name: Inspect image
+        run: |
+          docker buildx imagetools inspect riscvintl/udb:${{ env.TAG }}
```
jordancarlin: Make sure to run pre-commit on this. It is missing a trailing newline.

PR #1398, opened 2026-01-27
feat: add mise for unified tool version management and fix Podman UID/GID issue 

### LINKED ISSUE #1394: Podman fails to pull pre-build image due to node_modules UID/GID
**Describe the bug**
When pulling the container image using Podman, the following error occurs:
```
Error: unable to copy from source docker://riscvintl/udb:0.15: copying system image from manifest list: writing blob: adding layer with blob "sha256:64f1ee69c61640b9f75907abdeaf8f7e3e160bfcd1c11139323f05b6ed25f309"/""/"sha256:301fbf64919459893db36f95c9d01cf5c28bdc115ac8321536efa0cb40012feb": unpacking failed (error: exit status 1; output: potentially insufficient UIDs or GIDs available in user namespace (requested 376884:5762 for /opt/node/node_modules/clone-stats/LICENSE.md): Check /etc/subuid and /etc/subgid if configured locally and run "podman system migrate": lchown /opt/node/node_modules/clone-stats/LICENSE.md: invalid argument)
```

The setup script falls back to building the image locally, but this slows initials runs down significantly.

**To Reproduce**
Run the following:
```sh
podman pull docker.io/riscvintl/udb:0.15
```

**Additional context**
Apparently older versions of npm preserve the UID/GID metadata from the build system, which is not always an available UID/GID on the system running the image. Docker ignores this, but Podman throws an error. This has supposedly been fixed in newer versions of npm (https://github.com/npm/cli/issues/5998#issuecomment-3535443506), so upgrading npm might be the cleanest solution. There are also some hacky workarounds that involve changing the file ownership.

vkprogrammer-001:
## Summary

Adds [mise](https://mise.jdx.dev) for unified tool version management (Ruby, Node.js, Python) and fixes the Podman UID/GID compatibility issue by upgrading to npm >= 10.2.0.

## Problem Statement

When pulling container images using Podman, users encountered the following error:

`Error: potentially insufficient UIDs or GIDs available in user namespace
(requested 376884:5762 for /opt/node/node_modules/clone-stats/LICENSE.md):
Check /etc/subuid and /etc/subgid: lchown: invalid argument`

This occurred because older versions of npm (< 10.2.0) preserved UID/GID metadata from the build system, which Podman validates against the user namespace. Docker ignores this, but Podman throws an error.

## Solution

- **Introduced mise** for managing tool versions via `.mise.toml`
- **Upgraded to Node.js 22.12.0** which includes npm 10.9.0 (contains fix from [npm/cli#5998](https://github.com/npm/cli/pull/5998))
- **Added `chown -R root:root`** for node_modules as extra safety measure
- **Improved maintainability** with unified version configuration

### New Files
- `.mise.toml` - Tool version configuration
- `.mise.local.toml.example` - Example local overrides
- `doc/mise-setup.adoc` - Comprehensive documentation
- `bin/validate-mise` - Configuration validation script
- `MISE-IMPLEMENTATION.md` - Implementation guide

### Modified Files
- `.devcontainer/Dockerfile` - Uses mise instead of manual tool installation
- `bin/.container-tag` - Bumped to `0.16`
- `container.def` - Regenerated for Singularity
- `.gitignore` - Added mise local config exclusions

### Tool Versions
- **Ruby**: 3.4.8 (with YJIT enabled)
- **Node.js**: 22.12.0
- **npm**: 10.9.0 ✅ Podman compatible
- **Python**: 3.12.12

## Benefits

✅ **Fixes Podman compatibility** - No more UID/GID errors  
✅ **Unified tool management** - Single `.mise.toml` configuration  
✅ **Easier maintenance** - No manual ruby-build compilation  
✅ **Better caching** - mise manages downloads efficiently  
✅ **Future-ready** - Easy to add more tools (Go, Rust, etc.)  
✅ **Local development** - `mise install` works outside containers  

### Container Build
```bash
docker build -t riscvintl/udb:0.16 -f .devcontainer/Dockerfile .
```

Fixes #1394

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,7 +1,13 @@
 FROM ubuntu:24.04
 LABEL org.opencontainers.image.source=https://github.com/riscv/riscv-unified-db
 
-ENV DEBIAN_FRONTEND=noninteractive
+ENV DEBIAN_FRONTEND=noninteractive \
+    MISE_DATA_DIR="/opt/mise" \
+    MISE_CONFIG_DIR="/opt/mise/config" \
+    MISE_CACHE_DIR="/opt/mise/cache" \
+    MISE_STATE_DIR="/opt/mise/state" \
+    MISE_INSTALL_PATH="/usr/local/bin/mise" \
+    MISE_YES=1
 
 WORKDIR /workspace
 
@@ -40,31 +46,27 @@ RUN apt-get update && \
   && apt-get autoremove -y \
   && rm -rf /var/lib/{apt,dpkg,cache,log}/*
 
-ENV MISE_DATA_DIR="/mise"
-ENV MISE_CONFIG_DIR="/mise"
-ENV MISE_CACHE_DIR="/mise/cache"
-# MISE_STATE_DIR needed to prevent permission errors in CI when mise tries to track configs
-# Without this, mise defaults to ~/.local/state which isn't writable by non-root CI users
-ENV MISE_STATE_DIR="/mise/state"
-ENV MISE_INSTALL_PATH="/usr/local/bin/mise"
-ENV MISE_YES=1
-
 RUN curl https://mise.run | sh
 
-COPY .mise.toml /opt/.mise.toml
-COPY .ruby-version /opt/.ruby-version
+COPY .mise.toml .ruby-version /workspace/
 
-WORKDIR /opt
-RUN mise trust /opt/.mise.toml && \
-    mise settings set trusted_config_paths '~' && \
-    mise install --yes
+# - 'mise install -y': Installs tools defined in .mise.toml
+RUN mise settings set trusted_config_paths '~' && \
+    mise trust /workspace/.mise.toml && \
+    mise install -y && \
+    mise prune -y
 
-RUN chmod -R 777 /mise
+# Set installed versions as global defaults (dynamically, no hardcoded versions!)
+# This creates latest symlinks that we can use in PATH
+RUN for tool in node python ruby; do \
+        version=$(ls /opt/mise/installs/$tool | head -1); \
+        mise use --global ${tool}@${version}; \
+    done && \
+    chmod -R 777 /opt/mise
 
-# Verify tools work after mise installation
-RUN cd /opt && \
-    mise env --shell bash | grep '^export PATH=' > /etc/profile.d/mise-path-temp.sh && \
-    bash -c 'source /etc/profile.d/mise-path-temp.sh && ruby --version && node --version && npm --version && python3 --version'
+# Set PATH using latest symlinks (avoids shim overhead and ARG_MAX limit)
+# mise use --global created /opt/mise/installs/{tool}/latest symlinks for us
+ENV PATH="/workspace/bin/container:/opt/node/node_modules/.bin:/opt/venv/bin:/opt/mise/installs/node/latest/bin:/opt/mise/installs/python/latest/bin:/opt/mise/installs/ruby/latest/bin:$PATH"
 
 # build/install eqntott
 RUN git clone --depth=1 https://github.com/TheProjecter/eqntott.git && \
@@ -95,17 +97,16 @@ RUN git clone https://github.com/jar-ben/mustool.git && \
 
 # Create Python virtual environment and install packages
 COPY requirements.txt /opt/requirements.txt
-RUN bash -l -c 'python3 -m venv /opt/venv && \
-  /opt/venv/bin/pip install --no-cache-dir -r /opt/requirements.txt && \
-  rm /opt/requirements.txt'
+RUN python -m venv /opt/venv && \
+  /opt/venv/bin/pip install -r /opt/requirements.txt && \
+  rm /opt/requirements.txt
 ENV PATH="/opt/venv/bin:$PATH"
 
 # Install npm packages globally in the container
 COPY package.json package-lock.json /opt/node/
-RUN bash -l -c 'cd /opt/node && \
-  npm ci --no-audit --no-fund && \
-  chown -R root:root /opt/node/node_modules && \
-  rm /opt/node/package.json /opt/node/package-lock.json'
+RUN cd /opt/node && \
+  npm ci && \
+  rm /opt/node/package.json /opt/node/package-lock.json
 ENV NODE_PATH="/opt/node/node_modules"
 ENV PATH="/opt/node/node_modules/.bin:$PATH"
 ENV IN_UDB_CONTAINER=true
@@ -126,19 +127,7 @@ COPY tools/ruby-gems/udb-gen/lib/gem_versions.rb /opt/tools/ruby-gems/udb-gen/li
 COPY tools/ruby-gems/udb-gen/lib/udb-gen/version.rb /opt/tools/ruby-gems/udb-gen/lib/udb-gen/
 COPY tools/ruby-gems/udb_helpers/udb_helpers.gemspec /opt/tools/ruby-gems/udb_helpers/
 COPY tools/ruby-gems/udb_helpers/lib/udb_helpers/version.rb /opt/tools/ruby-gems/udb_helpers/lib/udb_helpers/
-RUN bash -l -c 'bundle install --gemfile /opt/Gemfile'
-
-# Remove mise binary and unset MISE_* env vars to prevent workspace .mise.toml detection
-# Replace mise wrapper scripts (npm/npx) with direct symlinks to avoid "mise: command not found"
-# Tools remain accessible via the PATH extracted earlier to /etc/profile.d/mise-path.sh
-RUN rm -f /usr/local/bin/mise && \
-    rm -rf /mise/cache /mise/downloads && \
-    echo 'unset MISE_DATA_DIR MISE_CONFIG_DIR MISE_CACHE_DIR MISE_STATE_DIR MISE_INSTALL_PATH MISE_YES' >> /etc/bash.bashrc && \
-    for node_dir in /mise/installs/node/*/bin; do \
-        rm -f "$node_dir/npm" "$node_dir/npx" && \
-        ln -s ../lib/node_modules/npm/bin/npm-cli.js "$node_dir/npm" && \
-        ln -s ../lib/node_modules/npm/bin/npx-cli.js "$node_dir/npx"; \
-    done
+RUN bundle install --gemfile /opt/Gemfile
 
 # the binaries for gems need to be copied in
 RUN mkdir -p /workspace/bin/container
@@ -151,21 +140,4 @@ ENV PATH="/workspace/bin/container:$PATH"
 RUN echo ubuntu ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/ubuntu \
     && chmod 0440 /etc/sudoers.d/ubuntu
 
-# Set complete PATH as ENV for Singularity compatibility (doesn't support ENTRYPOINT)
-# We use mise's 'latest' symlinks for Node and Ruby to avoid hardcoding versions
-# Python uses /opt/venv (with packages) instead of mise Python (without packages)
-ENV PATH="/workspace/bin/container:/opt/node/node_modules/.bin:/opt/venv/bin:/opt/bin:/mise/installs/node/latest/bin:/mise/installs/ruby/latest/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
-
-# Create mise-path.sh with the final PATH for Docker ENTRYPOINT (matches ENV above)
-RUN echo "export PATH='/workspace/bin/container:/opt/node/node_modules/.bin:/opt/venv/bin:/opt/bin:/mise/installs/node/latest/bin:/mise/installs/ruby/latest/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin'" > /etc/profile.d/mise-path.sh && \
-    chmod +x /etc/profile.d/mise-path.sh
-
-# Create a wrapper script that sources profile and executes commands
-RUN echo '#!/bin/bash' > /entrypoint.sh && \
-    echo 'source /etc/profile.d/mise-path.sh' >> /entrypoint.sh && \
-    echo 'exec "$@"' >> /entrypoint.sh && \
-    chmod +x /entrypoint.sh
-
-# Set ENTRYPOINT to the wrapper script
-ENTRYPOINT ["/entrypoint.sh"]
-CMD ["/bin/bash"]
+USER ubuntu


.mise.toml
```diff
@@ -0,0 +1,14 @@
+# RISC-V Unified Database - Tool Version Management
+# https://mise.jdx.dev
+
+[tools]
+ruby = "3.4.8"
```
dhower-qc: since there is already `.ruby-version` (and we'd like to keep it), shouldn't this be:

```toml
[settings]
# tools can read the versions files used by other version managers
idiomatic_version_file_enable_tools = ['ruby']
```

.mise.local.toml.example
dhower-qc, line 1: I'm not sure we need this file.

container.def
dhower-qc, line 1: Something went wrong with the singularity def generation. It's most likely because the path of python moved:

```
program name = '/opt/venv/bin/python3'
```

take a look at

```bash
./bin/generate-container-def
```

MISE-IMPLEMENTATION.md
dhower-qc, line 1: we don't need to check this in

.devcontainer/Dockerfile
```diff
@@ -5,10 +5,12 @@ SHELL ["/bin/bash", "-c"]
 
 WORKDIR /workspace
 
-# please keep pkgs sorted
+# Install system dependencies
+# Note: nodejs, npm, python3 are now installed via mise for better version control
```
jordancarlin: Restore this comment to what it was before.

.devcontainer/Dockerfile
```diff
@@ -82,15 +105,17 @@ RUN git clone https://github.com/jar-ben/mustool.git && \
...
 COPY requirements.txt /opt/requirements.txt
-RUN /usr/bin/python3 -m venv /opt/venv && \
-  /opt/venv/bin/pip install -r /opt/requirements.txt && \
+RUN python3 -m venv /opt/venv && \
+  /opt/venv/bin/pip install --no-cache-dir -r /opt/requirements.txt && \
   rm /opt/requirements.txt
 ENV PATH="/opt/venv/bin:$PATH"
 
 # Install npm packages globally in the container
+# The newer npm from mise (>= 10.2.0) fixes Podman UID/GID issues (npm/cli#5998)
```
jordancarlin: Unnecessary comment.

.devcontainer/Dockerfile
```diff
@@ -24,34 +26,55 @@ RUN apt-get update && \
...
   \
   && apt-get clean autoclean \
   && apt-get autoremove -y \
   && rm -rf /var/lib/{apt,dpkg,cache,log}/*
 
-  # build / install ruby
+# Install mise for unified tool version management
+# This manages Ruby, Node.js, and Python versions
+RUN curl https://mise.run | sh
+ENV PATH="/root/.local/bin:$PATH"
```
jordancarlin: From https://mise.jdx.dev/mise-cookbook/docker.html, the recommended way of installing `mise` in a container is the following:

```
ENV MISE_DATA_DIR="/mise"
ENV MISE_CONFIG_DIR="/mise"
ENV MISE_CACHE_DIR="/mise/cache"
ENV MISE_INSTALL_PATH="/usr/local/bin/mise"
ENV PATH="/mise/shims:$PATH"
# ENV MISE_VERSION="..."

RUN curl https://mise.run | sh
```

.devcontainer/Dockerfile
```diff
@@ -24,34 +26,55 @@ RUN apt-get update && \
   less \
   libc6-dev-riscv64-cross \
   libelf-dev \
+  libffi-dev \
   libgmp-dev \
+  liblzma-dev \
   libnewlib-dev \
+  libreadline-dev \
+  libsqlite3-dev \
+  libssl-dev \
```
jordancarlin: Are all of these new libraries needed by `mise`? I don't see anything in the `mise` documentation about them being necessary.

bin/validate-mise
jordancarlin, line 1: This script seems unnecessary.

bin/generate-container-def
```diff
@@ -4,7 +4,14 @@
...
 source "$ROOT/bin/setup"
-SPYTHON=/opt/venv/bin/spython
+
+# When running in Docker, spython is installed in the venv
+# When running on host with mise, use mise's python
+if [ -f "/opt/venv/bin/spython" ]; then
+  SPYTHON=/opt/venv/bin/spython
+else
+  SPYTHON="python3 -m spython"
+fi
```
jordancarlin: In docker the venv should be activated, so I think this can be simplified to always use what is currently in the `else` case.

.mise.toml
```diff
@@ -0,0 +1,14 @@
+# RISC-V Unified Database - Tool Version Management
+# https://mise.jdx.dev
+
+[tools]
+node = "22.12.0"
```
jordancarlin: If we're upgrading node in this PR anyway, any reason not to switch to the latest LTS release? Currently that is node 24.

.devcontainer/Dockerfile
```diff
@@ -24,34 +26,55 @@ RUN apt-get update && \
...
 COPY .ruby-version /opt/.ruby-version
+
+# Install tools via mise (Ruby, Node.js, Python)
+WORKDIR /opt
+RUN mise trust /opt/.mise.toml && \
+    MISE_NODE_VERIFY=0 mise install --yes && \
+    mise reshim
+ENV PATH="/root/.local/share/mise/shims:$PATH"
+ENV MISE_DATA_DIR="/root/.local/share/mise"
 ENV RUBY_YJIT_ENABLE=1
```
jordancarlin: Is this still necessary now that it is set in `.mise.toml`?

.DS_Store
jordancarlin, line 1: This should not be checked in.

doc/mise-setup.adoc
jordancarlin, line 1: A lot of this document seems unnecessary. For example, there is no need to compare this to the previous approach.

.devcontainer/Dockerfile
```diff
@@ -82,15 +105,17 @@ RUN git clone https://github.com/jar-ben/mustool.git && \
 
 # Create Python virtual environment and install packages
 COPY requirements.txt /opt/requirements.txt
-RUN /usr/bin/python3 -m venv /opt/venv && \
-  /opt/venv/bin/pip install -r /opt/requirements.txt && \
+RUN python3 -m venv /opt/venv && \
+  /opt/venv/bin/pip install --no-cache-dir -r /opt/requirements.txt && \
```
jordancarlin: Eventually we should switch to managing Python using uv (through mise) for much faster installation and simplified venv management, but that can be addressed in a follow up PR. https://mise.jdx.dev/lang/python.html#mise-uv

.devcontainer/Dockerfile
vkprogrammer-001: thanks . done

.devcontainer/Dockerfile
```diff
@@ -44,14 +39,35 @@ RUN apt-get update && \
...
+ENV MISE_STATE_DIR="/mise/state"
+ENV MISE_INSTALL_PATH="/usr/local/bin/mise"
 ENV RUBY_YJIT_ENABLE=1
-RUN git clone --depth=1 https://github.com/rbenv/ruby-build.git && \
-    cd ruby-build && \
-    ./bin/ruby-build $(cat /opt/.ruby-version | tr -d "\n") /usr && \
-    cd .. && \
-    rm -rf ruby-build
+ENV MISE_YES=1
+ENV MISE_DISABLE_TRACKING=1
```
jordancarlin: What does `MISE_DISABLE_TRACKING` do? I don't see it in the mise docs.

.devcontainer/Dockerfile
```diff
@@ -44,14 +39,35 @@ RUN apt-get update && \
...
+
+RUN curl https://mise.run | sh
+
+COPY .mise.toml /opt/.mise.toml
+COPY .ruby-version /opt/.ruby-version
+
+WORKDIR /opt
+RUN mise trust /opt/.mise.toml && \
+    mise settings set trusted_config_paths '~' && \
+    MISE_NODE_VERIFY=0 mise install --yes && \
```
jordancarlin: Why is `MISE_NODE_VERIFY=0` needed? According to the docs, this setting verifies the downloaded assets using GPG. That seems like a good thing to leave on (which is the default).

.devcontainer/Dockerfile
```diff
@@ -44,14 +39,35 @@ RUN apt-get update && \
...
+    mise reshim
+
+RUN chmod -R 777 /mise
+
+# Add mise-installed tools directly to PATH
+ENV PATH="/mise/installs/ruby/3.4.8/bin:/mise/installs/node/24.13.0/bin:/mise/installs/python/3.12.12/bin:$PATH"
```
jordancarlin: Shouldn't this already be done by adding the shims to the PATH? Manually adding the tools to the PATH is brittle as versions change.

vkprogrammer-001: You're correct. But I changed from shims to hardcoded paths because we hit the "Argument list too long" error in CI ``(/usr/bin/env: 'bash': Argument list too long)``.

vkprogrammer-001: and on, making this change. It solved the problem.

.devcontainer/Dockerfile
vkprogrammer-001: I added this because the build was failing with GPG verification errors for Node.js downloads.

.devcontainer/Dockerfile
vkprogrammer-001: You're right, this is unnecessary as it doesn't solve anything. I added it in an attempt to fix the "Argument list too long" error in CI. I mistakenly believed it would disable the tracking feature writing to /mise/state/tracked-configs. My approach was incorrect.

.devcontainer/Dockerfile
jordancarlin: If there are actually GPG verification errors, I'd be curious to know what they are. Disabling the check seems like something we should really avoid if possible.

.devcontainer/Dockerfile
jordancarlin: Hmm. I'd still like to avoid manually adding them to the PATH like this. What about using something like `mise en`?

.devcontainer/Dockerfile
vkprogrammer-001: This is the actual error. I tried it locally, by removing the MISE_NODE_VERIFY=0 :

```
vipinkumar@vipins-MacBook-Air riscv-unified-db %  docker build -t riscvintl/udb:test -f .d
evcontainer/Dockerfile . 2>&1 | tail -50
#13 91.39 gpg: can't connect to the gpg-agent: Configuration error
#13 91.39 gpg: error running '/usr/bin/gpg-agent': probably not installed
#13 91.39 gpg: failed to start gpg-agent '/usr/bin/gpg-agent': Configuration error
#13 91.39 gpg: can't connect to the gpg-agent: Configuration error
#13 91.39 gpg: error running '/usr/bin/gpg-agent': probably not installed
#13 91.39 gpg: failed to start gpg-agent '/usr/bin/gpg-agent': Configuration error
#13 91.39 gpg: can't connect to the gpg-agent: Configuration error
#13 91.39 gpg: error running '/usr/bin/gpg-agent': probably not installed
#13 91.39 gpg: failed to start gpg-agent '/usr/bin/gpg-agent': Configuration error
#13 91.39 gpg: can't connect to the gpg-agent: Configuration error
#13 91.39 gpg: error running '/usr/bin/gpg-agent': probably not installed
#13 91.39 gpg: failed to start gpg-agent '/usr/bin/gpg-agent': Configuration error
#13 91.39 gpg: can't connect to the gpg-agent: Configuration error
#13 91.39 gpg: error running '/usr/bin/gpg-agent': probably not installed
#13 91.39 gpg: failed to start gpg-agent '/usr/bin/gpg-agent': Configuration error
#13 91.39 gpg: can't connect to the gpg-agent: Configuration error
#13 92.15 mise ruby@3.4.8      ==> Installing ruby-3.4.8...
#13 92.16 mise ruby@3.4.8      -> ./configure --prefix=/mise/installs/ruby/3.4.8 --with-openssl-dir=/mise/installs/ruby/3.4.8/openssl --enable-shared --with-ext=openssl,psych,+
#13 110.0 mise ruby@3.4.8      -> make -j 8
#13 189.4 mise ruby@3.4.8      -> make install
#13 221.0 mise ruby@3.4.8      ==> Installed ruby-3.4.8 to /mise/installs/ruby/3.4.8
#13 221.8 mise ruby@3.4.8    ✓ installed
#13 221.8 mise ERROR Failed to install core:node@24: gpg exited with non-zero status: exit code 2
#13 221.8 mise ERROR Run with --verbose or MISE_VERBOSE=1 for more information
#13 ERROR: process "/bin/bash -c mise trust /opt/.mise.toml &&     mise settings set trusted_config_paths '~' &&     mise install --yes &&     mise reshim" did not complete successfully: exit code: 1
------
 > [ 8/38] RUN mise trust /opt/.mise.toml &&     mise settings set trusted_config_paths '~' &&     mise install --yes &&     mise reshim:
91.39 gpg: failed to start gpg-agent '/usr/bin/gpg-agent': Configuration error
91.39 gpg: can't connect to the gpg-agent: Configuration error
92.15 mise ruby@3.4.8      ==> Installing ruby-3.4.8...
92.16 mise ruby@3.4.8      -> ./configure --prefix=/mise/installs/ruby/3.4.8 --with-openssl-dir=/mise/installs/ruby/3.4.8/openssl --enable-shared --with-ext=openssl,psych,+
110.0 mise ruby@3.4.8      -> make -j 8
189.4 mise ruby@3.4.8      -> make install
221.0 mise ruby@3.4.8      ==> Installed ruby-3.4.8 to /mise/installs/ruby/3.4.8
221.8 mise ruby@3.4.8    ✓ installed
221.8 mise ERROR Failed to install core:node@24: gpg exited with non-zero status: exit code 2
221.8 mise ERROR Run with --verbose or MISE_VERBOSE=1 for more information
------
Dockerfile:57
--------------------
  56 |     WORKDIR /opt
  57 | >>> RUN mise trust /opt/.mise.toml && \
  58 | >>>     mise settings set trusted_config_paths '~' && \
  59 | >>>     mise install --yes && \
  60 | >>>     mise reshim
  61 |     
--------------------
ERROR: failed to build: failed to solve: process "/bin/bash -c mise trust /opt/.mise.toml &&     mise settings set trusted_config_paths '~' &&     mise install --yes &&     mise reshim" did not complete successfully: exit code: 1

View build details: docker-desktop://dashboard/build/desktop-linux/desktop-linux/1x5wplzwefoocejlg1p737u2r

```

jordancarlin: That looks like gpg isn't installed in the container. Can we just install gpg? Add it to the apt list.

vkprogrammer-001: I have added the gpg-agent, and now I am not seeing any errors regarding it.

.devcontainer/Dockerfile
```diff
@@ -44,14 +40,38 @@ RUN apt-get update && \
...
-  # build / install ruby
-COPY .ruby-version /opt/.ruby-version
+ENV MISE_DATA_DIR="/mise"
+ENV MISE_CONFIG_DIR="/mise"
+ENV MISE_CACHE_DIR="/mise/cache"
+# MISE_STATE_DIR needed to prevent permission errors in CI when mise tries to track configs
+# Without this, mise defaults to ~/.local/state which isn't writable by non-root CI users
+ENV MISE_STATE_DIR="/mise/state"
+ENV MISE_INSTALL_PATH="/usr/local/bin/mise"
 ENV RUBY_YJIT_ENABLE=1
```
jordancarlin: Is this still needed? It is duplicated in `.mise.toml`.

vkprogrammer-001: It is removed in the latest changes

vkprogrammer-001:
Hi @jordancarlin, PTAL.

jordancarlin:
Closing now that #1590 is merged.

PR #1403, opened 2026-01-29
feat: vector parameters new format 

jacassidy:


spec/std/isa/param/IMPRECISE_VECTOR_TRAP_SETTABLE.yaml
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
dhower-qc: (in all files) This should be your copyright

spec/std/isa/param/IMPRECISE_VECTOR_TRAP_SETTABLE.yaml
```diff
@@ -0,0 +1,16 @@
...
+kind: parameter
+name: IMPRECISE_VECTOR_TRAP_SETTABLE
+description: |
+  Some profiles may choose to provide a privileged mode bit to select between precise and imprecise vector traps.
+long_name: TODO
+schema:
+  type: boolean
+definedBy:
+  extension:
+    name: V
```
dhower-qc: (for all) This actually needs to be `Zvl32b` instead of `V` since `V` is a superset of a basic vector implementation

spec/std/isa/param/IMPRECISE_VECTOR_TRAP_SETTABLE.yaml
jacassidy: Are you asking for a change? I believe this was in all files.

dhower-qc: Yes, since you wrote these the copyright should be assigned to you (or HMC, depending on your own legal guidance)

jacassidy: gotcha, should be all set. Thanks!

PR #1422, opened 2026-01-30
fix: updated tests build to work with latest riscv-tests commit 

henrikg-qc:


henrikg-qc:
Cannot add reviewers. I want to add Jordan Carlin and Ajit Dingankar in addition to Derek and Paul.

backends/cpp_hart_gen/tasks.rake
```diff
@@ -392,8 +392,14 @@ end
 task checkout_riscv_tests: "#{$root}/ext/riscv-tests/env/LICENSE"
 
 task build_riscv_tests: "checkout_riscv_tests" do
+  configs_name,build_name = configs_build_name
```
ThinkOpenly:
```suggestion
  configs_name, build_name = configs_build_name
```

henrikg-qc: Nice catch, "Rätt ska vara rätt"

PR #1431, opened 2026-01-30
feat(data): add Semihosting extension definition

### LINKED ISSUE #1393: Model `vmv.v.i`
In order to run a test such as `rv64vadd_vv-0` from the [vector tests](https://github.com/riscv-software-src/sail-riscv-tests/releases/download/2025-10-14/riscv-vector-tests-v128x64.tar.gz) in `sail-riscv-tests` we need to model the vector move instructions, especially `vmv.v.i` for that particular test.

ishaan-arora-1:
## Summary

Add the RISC-V Semihosting specification as the first standard non-ISA specification in `spec/std/non_isa/`.

## Changes

- Created `spec/std/non_isa/Semihosting.yaml` following the `non_isa_schema.json` format
- Uses structured prose with tagged statements (id, normative, text) as per the schema requirements

## Specification Details

The RISC-V Semihosting specification defines a binary interface that allows applications running in debug/simulation environments to access host system resources through a special instruction sequence:

```
slli x0, x0, 0x1f   # Entry NOP
ebreak              # Break to debugger  
srai x0, x0, 7      # Exit NOP
```

The specification includes:
- **Description**: Overview of semihosting concept and its design rationale
- **Breakpoint Instruction Sequence** (normative): Requirements for the 3-instruction sequence
- **Semihosting Parameters** (normative): Parameter passing via a0/a1 registers
- **Semihosting Services** (informative): Overview of available operations
- **References**: Links to official RISC-V and ARM semihosting specifications

## Notes

This is the first non-ISA specification being added to `spec/std/non_isa/`. Happy to iterate on the format based on feedback.

Closes #1393

### DIFF
--- a/spec/std/non_isa/Semihosting.yaml
+++ b/spec/std/non_isa/Semihosting.yaml
@@ -8,101 +8,84 @@ $schema: non_isa_schema.json#
 kind: non-isa specification
 name: Semihosting
 long_name: RISC-V Semihosting
-version: "0.95.0"
-description:
-  - id: non-isa-semihosting-overview
-    normative: false
-    text: |
-      Semihosting is a technique where an application running in a debug or
-      simulation environment can access elements of the system hosting the
-      debugger or simulator including console, file system, time, and other
-      functions. This allows for diagnostics, interaction, and measurement of
-      a target system without requiring significant infrastructure to exist
-      in that target environment.
-  - id: non-isa-semihosting-design
-    normative: false
-    text: |
-      The RISC-V semihosting specification adopts the design of the ARM semihosting
-      specification to minimize development effort. The services defined by the ARM
-      semihosting specification are portable across different architectures, and only
-      the mechanism of invoking a semihosting service (the semihosting binary interface)
-      is architecture-specific.
-  - id: non-isa-semihosting-scope
-    normative: false
-    text: |
-      This specification only defines the semihosting binary interface for RISC-V
-      platforms; all other aspects of semihosting are defined by the ARM semihosting
-      specification.
+version: "1.0.0"
+ratification_date: 2025-02
+description: |
+  Semihosting is a technique where an application running in a debug or
+  simulation environment can access elements of the system hosting the
+  debugger or simulator including console, file system, time, and other
+  functions. This allows for diagnostics, interaction, and measurement of
+  a target system without requiring significant infrastructure to exist
+  in that target environment.
+
+  The RISC-V semihosting specification adopts the design of the ARM semihosting
+  specification to minimize development effort. The services defined by the ARM
+  semihosting specification are portable across different architectures, and only
+  the mechanism of invoking a semihosting service (the semihosting binary interface)
+  is architecture-specific.
+
+  This specification only defines the semihosting binary interface for RISC-V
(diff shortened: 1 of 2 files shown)


spec/std/isa/ext/Semihosting.yaml
jordancarlin, line 1: If this is non-isa, it shouldn't be going in the `isa` directory. Presumably we will need a `spec/std/non-isa` directory (or something along those lines), but I don't think that infrastructure is in place yet.

ishaan-arora-1: @jordancarlin Alright i guess i'll close this pr for now?

jordancarlin: Let's see what @dhower-qc has to say about non-isa. It might be easy to set up, in which case this could be the first non-isa spec that we add.

ishaan-arora-1: Sure, sounds good, will be waiting @dhower-qc

dhower-qc:
@AFOliveira, thoughts?

ishaan-arora-1:
> All my requests are done, thanks.
> 
> Though, I am still not approcing this as I am unsure if the differences between normative and non-normative are correctly specified and would like to ask @BKhristenko to help in this regards.
> 
> Not directly tied to your PR, I don't think we have a way of building a doc of just one non-isa spec. That would be worthy and I may contribute to it, if we don't already. Seems like an important step as non-isa specs are currently held on their own, and we should at least allow that to happen in UDB.

awaiting feedback!! @BKhristenko

spec/std/non_isa/Semihosting.yaml
```diff
@@ -0,0 +1,121 @@
...
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../schemas/non_isa_schema.json
+
+---
+$schema: non_isa_schema.json#
+kind: non-isa specification
+name: Semihosting
+long_name: RISC-V Semihosting
+version: "0.95.0"
```
ThinkOpenly:
```suggestion
version: "1.0.0"
ratification_date: 2025-02
```

spec/std/non_isa/Semihosting.yaml
```diff
@@ -0,0 +1,121 @@
...
+description:
+  - id: non-isa-semihosting-overview
+    normative: false
+    text: |
```
ThinkOpenly: Hmm. I certainly appreciate the effort here -- nicely done. Unfortunately, "tagged_text" is deprecated, although we don't say so anywhere. (We actually need to remove it entirely, I think.)

Could you:
- Change `non_isa_schema.json`, replacing "tagged_text" with "spec_text". This is essentially just a (Asciidoc) string.
- Change all of the "description" content here to, well, (Asciidoc) text?

I presume from the other comments in this PR that the non-isa schema is not used anywhere else, so this change shouldn't break anything.

ishaan-arora-1: Thankyou for the feedback!! made the changes as requested. Let me know how things look now.

PR #1309, opened 2026-01-09
feat(vector): support basic 8-bit vector add from cherry-picked commits

adingank-qualcomm:
- Modeled vector load and store operations
- Fixed the range of destination vector register for the (previously modelled) add operation
- Manually wrote simple self-checking tests for vl8, vse8 and vadd.vv instructions
- Validated the above tests against the Sail RISC-V model
- Verified the UDB implementation by running the above tests on the generated ISS

PR #1467, opened 2026-02-01
feat(vector): support unit stride vector load/store

jmawet:
Fill instructions for unit stride vector load/store instructions. 

Based off the work from @adingank-qualcomm in #1309

### DIFF
--- a/spec/std/isa/inst/V/vle16.v.yaml
+++ b/spec/std/isa/inst/V/vle16.v.yaml
@@ -35,7 +35,7 @@ operation(): |
   }
 
   XReg virtual_address = X[xs1];
-  # "A simple in-order implementation can ignore the settings and simply execute all vector instructions using the undisturbed policy"
+  # Per RISC-V ISA: "A simple in-order implementation can ignore the settings and simply execute all vector instructions using the undisturbed policy"
   for (U32 i = 0; i < CSR[vl].VALUE; i++) {
     U32 start_bit_pos = (CSR[vstart].VALUE + i) * 16;
     U32 end_bit_pos = start_bit_pos + 15;
--- a/spec/std/isa/inst/V/vle8.v.yaml
+++ b/spec/std/isa/inst/V/vle8.v.yaml
@@ -35,7 +35,7 @@ operation(): |
   }
 
   XReg virtual_address = X[xs1];
-  # "A simple in-order implementation can ignore the settings and simply execute all vector instructions using the undisturbed policy"
+  # Per RISC-V ISA: "A simple in-order implementation can ignore the settings and simply execute all vector instructions using the undisturbed policy"
   for (U32 i = 0; i < CSR[vl].VALUE; i++) {
     U32 start_bit_pos = (CSR[vstart].VALUE + i) * 8;
     U32 end_bit_pos = start_bit_pos + 7;
--- a/spec/std/isa/inst/V/vle32.v.yaml
+++ b/spec/std/isa/inst/V/vle32.v.yaml
@@ -35,7 +35,7 @@ operation(): |
   }
 
   XReg virtual_address = X[xs1];
-  # "A simple in-order implementation can ignore the settings and simply execute all vector instructions using the undisturbed policy"
+  # Per RISC-V ISA: "A simple in-order implementation can ignore the settings and simply execute all vector instructions using the undisturbed policy"
   for (U32 i = 0; i < CSR[vl].VALUE; i++) {
     U32 start_bit_pos = (CSR[vstart].VALUE + i) * 32;
     U32 end_bit_pos = start_bit_pos + 31;
--- a/spec/std/isa/inst/V/vle64.v.yaml
+++ b/spec/std/isa/inst/V/vle64.v.yaml
@@ -35,7 +35,7 @@ operation(): |
   }
 
   XReg virtual_address = X[xs1];
-  # "A simple in-order implementation can ignore the settings and simply execute all vector instructions using the undisturbed policy"
+  # Per RISC-V ISA: "A simple in-order implementation can ignore the settings and simply execute all vector instructions using the undisturbed policy"
   for (U32 i = 0; i < CSR[vl].VALUE; i++) {
     U32 start_bit_pos = (CSR[vstart].VALUE + i) * 64;
     U32 end_bit_pos = start_bit_pos + 63;
--- a/spec/std/isa/inst/V/vse16.v.yaml
+++ b/spec/std/isa/inst/V/vse16.v.yaml
@@ -35,7 +35,7 @@ operation(): |
   }
 
   XReg virtual_address = X[xs1];
-  # "A simple in-order implementation can ignore the settings and simply execute all vector instructions using the undisturbed policy"
+  # Per RISC-V ISA: "A simple in-order implementation can ignore the settings and simply execute all vector instructions using the undisturbed policy"
   for (U32 i = 0; i < CSR[vl].VALUE; i++) {
     U32 start_bit_pos = (CSR[vstart].VALUE + i) * 16;
     U32 end_bit_pos = start_bit_pos + 15;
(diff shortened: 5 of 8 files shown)


jmawet:
Spent an hour trying to figure out the operation for these myself, then noticed @adingank-qualcomm already laid the groundwork in #1309 with vle/se8 lol
So I replicated his work and expanded to the other EEW values (16/32/64)
Still need to write tests, but y'all can double check my additions in the meantime. 

2 questions:
1. do we need to do anything with the vm (mask enable) value? I assume we need to at some point, it's just less important for the vadd testing?
2. once I write the tests, can someone remind me how to run them?

adingank-qualcomm:
@jmawet Thanks for your effort, and sorry it was a blind alley re parameterization!
Can you please open an issue for discussion with @dhower-qc ?

Re running the tests, I just ran 
`./do test:riscv_vector_tests CONFIG=rv64 BUILD_TYPE=DEBUG IGNOREUNDEFINED=YES JOBS=8` 
which worked (after I did `./do clean` but not sure if it's required). This runs the 4 old vector tests,
so I added the three new ones to the task in PR #1592

jmawet:
Opened issue for discussion: #1593, I will proceed with non-parameterized functions for now

I'll try running the tests again using the new command and let you know if it works or not.

spec/std/isa/inst/V/vle8.v.yaml
```diff
@@ -11,7 +11,7 @@ description: |
   Loads 8-bit elements from memory starting at address in xs1 register, using vector mask vm, and stores results in vector register vd. The effective element width is 8 bits regardless of SEW setting. Supports vector masking and follows vector tail policy.
 definedBy:
   extension:
-    name: Zvl32b
+    name: V
```
dhower-qc: This should stay as Zvl32b

ext/rbi-central
dhower-qc, line 1: was this intentional?

spec/std/isa/inst/V/vle8.v.yaml
jmawet: fixed

spec/std/isa/inst/V/vle16.v.yaml
```diff
@@ -30,6 +30,27 @@ access:
 data_independent_timing: false
 operation(): |
 
+  if (VECTOR_LS_MISALIGNED_LEGAL == false) {
```
ThinkOpenly: Since `VECTOR_LS_MISALIGNED_LEGAL` is boolean, this can be simplified (slightly).
```suggestion
  if (!VECTOR_LS_MISALIGNED_LEGAL) {
```

spec/std/isa/inst/V/vle16.v.yaml
```diff
@@ -30,6 +30,27 @@ access:
...
+    unpredictable("Implementations should raise address misaligned exception");
+  }
+
+  XReg virtual_address = X[xs1];
+  # "A simple in-order implementation can ignore the settings and simply execute all vector instructions using the undisturbed policy"
```
ThinkOpenly: Is this in quotes because it's a literal quote from the spec?

spec/std/isa/inst/V/vle16.v.yaml
```diff
@@ -30,6 +30,27 @@ access:
...
+  for (U32 i = 0; i < CSR[vl].VALUE; i++) {
+    U32 start_bit_pos = (CSR[vstart].VALUE + i) * 16;
+    U32 end_bit_pos = start_bit_pos + 15;
+    if (vm == 0) {
+      unpredictable("TODO: add masking support");
+    } else {
+      v[vd] = {
+        v[vd][VLEN-1:end_bit_pos + 1],
+        sext(read_memory<16>(virtual_address, $encoding), 16),
```
ThinkOpenly: Do we need to sign-extend a 16 bit value to 16 bits?

spec/std/isa/inst/V/vle16.v.yaml
jmawet: ok, wasn't sure if that syntax was supported in idl

spec/std/isa/inst/V/vle16.v.yaml
jmawet: Ajit wrote this
@adingank-qualcomm can you answer?

spec/std/isa/inst/V/vle16.v.yaml
jmawet: another bit I copied from Ajit's work on vle8
@adingank-qualcomm why are we sign extending here?

adingank-qualcomm: Sorry about the "senior moment" in copy-paste-edit cycle from another older instruction model! 
@jmawet Not sure how best to handle this, perhaps you can remove it as a part of the PR?

spec/std/isa/inst/V/vle16.v.yaml
adingank-qualcomm: @ThinkOpenly Yes, it's from a Note in [Section 30.3.4.3](https://docs.riscv.org/reference/isa/unpriv/v-st-ext.html#sec-agnostic). I'd appreciate any guidance for referring to a specific part of the spec in the code. This link is to the web version (unlike in the past when I had used a link to a document on the RVI Google Drive) but the section numbering may still change...

ThinkOpenly: Hmm. It's not normative text, to my understanding. Regardless, we don't have a convention for citations. I'd be content with something like:
> Per RISC-V ISA, "..."

If you wanted to put a version of some sort in there, you could, but I don't think it's necessary (and the current versioning of the spec is not the best).

I wouldn't spend a lot of time on it. I was just trying to understand why this comment's text was in quotation marks. Any sort of clarity there would be fine.

jmawet:
pulled latest changes from main and now the vle8 test is failing with some assertion I have not seen previously:
```
At /home/jdupac/git/riscv-unified-db/gen/cpp_hart_gen/rv64_Debug/include/udb/bits.hpp:1122:7: constexpr udb::_Bits<N, false> udb::_Bits<N, Signed>::extract(const MsbType<MsbN, false>&, const LsbType<LsbN, false>&) const [with MsbType = udb::_Bits; unsigned int MsbN = 32; LsbType = udb::_Bits; unsigned int LsbN = 1; unsigned int N = 128; bool Signed = false] :
   Assertion failed: (lsb.get() < N && msb.get() < N)
   Extract out of range
/usr/local/bundle/ruby/3.4.0/gems/rake-13.3.1/lib/rake/file_utils.rb:65:in 'block in FileUtils#create_shell_runner': Command failed with status (): [/home/jdupac/git/riscv-unified-db/gen/cpp_hart_gen/rv64_Debug/build/iss -m rv64 -c /home/jdupac/git/riscv-unified-db/cfgs/rv64-vector.yaml tests/isa/rv64uv-p-vle8] (RuntimeError)
```
any idea what's causing this? was there a recent change in the build/compiler?

spec/std/isa/inst/V/vle16.v.yaml
jmawet: removed the sign extend

spec/std/isa/inst/V/vle16.v.yaml
jmawet: added 
> per riscv isa "..."

ThinkOpenly:
>    Assertion failed: (lsb.get() < N && msb.get() < N)
>    Extract out of range

> any idea what's causing this? was there a recent change in the build/compiler?

That assert was recently modified, but the question would seem to be what values are `lsb` and `msb` here such that it would fail? The traceback includes:
> [...] unsigned int MsbN = 32; [...] unsigned int LsbN = 1; unsigned int N = 128 [...]

If those are the values going into the assert, the assert shouldn't fire.

Could there be an order of operations problem with a construct like "A < B && C < D"? The IDL documentation has "<" with precedence 9 and "&&" with precedence "14", so that should be fine.

PR #1527, opened 2026-02-02
feat: add support for fine-grained assembly representation

ThinkOpenly:
Currently, assembly syntax is represented as a simple string of comma-separated operands with a heuristic naming convention to indicate their respective type and purpose.

Make this more rigorous by adding a schema which supports:
- registers
- the registers' register file (GPR, FPR, VR, CSR, etc.)
- dereference syntax "(reg)"
- dereference+offset syntax "offset(reg)"
- immediates
- floating-point rounding mode and possible values
- FENCE scopes
- register lists (for POP/PUSH)

The new "operands" YAML field is currently optional and coexists with the existing "assembly" field. So, support can be added over time to both the YAML files and the infrastructure to support generation of actual assembly syntax where needed (documentation) until the "assembly" field is no longer needed.


ThinkOpenly:
Not terribly worried about this, but DO NOT MERGE.

I had to specify all required fields and could not depend on the "default" values being used when trying to get past `pre-commit run check-jsonschema`.

ThinkOpenly:
Notes on current state...
- I treat csr operands as coming from a "csr" regfile
- vector mask uses "possible_values: [0]" for "v0.t"
- I kinda brute-forced reg_list definitions, but am content with the implementation
- `cm.mvsa01` uses operands "r1s, r2s", which I find weird. I used operand description name "xdm" (X regfile, destination, "move") and "xsm" would be used for `cm.mva01s`, but I override them to "r1s" and "r2s", reluctantly.

Things that would be nice to address before merging:
- will that IDL "offset" syntax work? "operands[xs1].offset"
- add encoding and decoding tests for new "encdec" backend
- shall I implement "possible_values" for "csr" as "[0-4095]"? probably.
- add more descriptions to the YAML files? to operand definitions?
- operand description file names. This is subjective, and I'm not sure how important. Might be nice to document some conventions.

Can wait:
- `encode.py` and `decode.py` are very hacky/fragile. They need to invoke the IDL functions.
- need to validate "(reg)" operand definition "works" (xs-nooffset)
- need to better handle register names (ABI names)... should there be a parameter to `decode()`?
- what about CSR names as operand values?
- need real IDL testing for for `encode()`/`decode()`... some way for an assembler/disassembler to utilize IDL.
- support for pushing JSON schema "default" values into resolved arch
- support for identifying "implicit" operands (e.g. register pair)
- some sort of tie-in with "regfile" content instead of hardcoding regfile names?

henry-hsieh:
Copy some ideas from #1814 :

1. Implicit operand (with new implicit attribute), e.g., `c.addi4spn`
```yaml
$schema: "operand_schema.json#"
kind: operand
name: sp-implicit
data:
  $inherits: operand/xs.yaml#/data
  possible_values: [2]
  implicit: true
```
2. Name with multiple of actual registers, e.g., `c.ld` in RV32
- Introduce name to values mapping
- The name could be an array for ABI name swapping
- The values can be an array with all included values
```yaml
$schema: "operand_schema.json#"
kind: operand
name: xs-pair
data:
  type: reg_pair
  name: reg_pair_xs
  possible_values:
    - $inherits: operand/xs.yaml#/data
      name: ["x0", "zero"]
      values: [0, 1]
    - $inherits: operand/xs.yaml#/data
      name: ["x2", "sp"]
      values: [2, 3]
    - $inherits: operand/xs.yaml#/data
      name: ["x4", "tp"]
      values: [4, 5]
...
```
3. The instructions use the same operand for both source and destination, e.g., `c.and`
```yaml
$schema: "operand_schema.json#"
kind: operand
name: xsdc
data:
  $inherits: operand/xdc.yaml#/data
  source: true
```
4. The instruction only allow one possible value for an operand, e.g., `c.addi16sp`
```yaml
$schema: "operand_schema.json#"
kind: operand
name: xsd-sp-only
data:
  $inherits: operand/xs.yaml#/data
  possible_values: [2]
  destination: true
```

---

With Optimization 2, the reg_range can be integrated with GPR more naturally:
```yaml
$schema: "operand_schema.json#"
kind: operand
name: reg-range1-rv32
data:
  type: reg_range
  name: reg_range1
  possible_values:
    - $inherits: operand/xs.yaml#/data
      name: ["x8", "s0"]
      values: [8]
    - $inherits: operand/xs.yaml#/data
      name: ["x8-x9", "s0-s1"]
      values: [8, 9]
  optional: true
```
The problem of `vm` could be solved, too:
```yaml
$schema: "operand_schema.json#"
kind: operand
name: vm
data:
  $inherits: operand/vs.yaml#/data
  name: vm
  possible_values:
    - $inherits: operand/vs.yaml#/data
      name: ["v0.t"]
      values: [0]
  optional: true
```

henry-hsieh:
Small question for execution:

Do you consider move this PR to separate develop branch and let everyone can contribute to it? It would take quite some time to migrate every extension to new format. Moreover, this PR will be extremely large for review.

ThinkOpenly:
> Do you consider move this PR to separate develop branch and let everyone can contribute to it?

Fair question. I've been hoping to get a reasonably comprehensive schema in place as a "phase 1". I think it's close. "implicit" probably needs to be added first, although that could be a "phase 2". Thoughts?

Phases:
1. Add non-breaking support for enough _schema_ to encode and decode everything. Implement at least one example of every type of operand. Get general agreement with the approach.
2. Add "implicit" to give downstream more insight into dependencies/inputs and side-effects/outputs. Implement at least one example of every type of implicit operand. Get general agreement with the approach.
3. Implement for all instructions

Note that this PR could just cover phase 1, to get things moving.

> It would take quite some time to migrate every extension to new format.

It will be a fair amount of effort, but I'm really hoping that AI can pitch in a lot.

> Moreover, this PR will be extremely large for review.

I'm not sure the best way to mitigate that problem. One big gulp, or many bite-sized pieces. The effort will be roughly the same in aggregate. I'm in the process of creating some reasonably robust tests, although they do not utilize the IDL at all. I'd really like a way to invoke IDL functions from Python, but for now, I just did a very rough (hacky) equivalent translation of the IDL to Python.

henry-hsieh:
> > Do you consider move this PR to separate develop branch and let everyone can contribute to it?
> 
> Fair question. I've been hoping to get a reasonably comprehensive schema in place as a "phase 1". I think it's close. "implicit" probably needs to be added first, although that could be a "phase 2". Thoughts?
> 
> Phases:
> 
>     1. Add non-breaking support for enough _schema_ to encode and decode everything. Implement at least one example of every type of operand. Get general agreement with the approach.
> 
>     2. Add "implicit" to give downstream more insight into dependencies/inputs and side-effects/outputs. Implement at least one example of every type of implicit operand. Get general agreement with the approach.
> 
>     3. Implement for all instructions
> 

LGTM.

> 
> Note that this PR could just cover phase 1, to get things moving.
> 
> > It would take quite some time to migrate every extension to new format.
> 
> It will be a fair amount of effort, but I'm really hoping that AI can pitch in a lot.
> 
> > Moreover, this PR will be extremely large for review.
> 
> I'm not sure the best way to mitigate that problem. One big gulp, or many bite-sized pieces. The effort will be roughly the same in aggregate. I'm in the process of creating some reasonably robust tests, although they do not utilize the IDL at all. I'd really like a way to invoke IDL functions from Python, but for now, I just did a very rough (hacky) equivalent translation of the IDL to Python.

You gonna need to translate [grammar generated in Ruby](https://github.com/riscv/riscv-unified-db/blob/main/tools/ruby-gems/idlc/lib/idlc/idl.treetop) to the parser interpreter in Python like [Arpeggio](https://github.com/textX/Arpeggio). If I'm not wrong, both of them use PEG-based grammar. Then you visit the parser tree to get result. I'm not very familiar to parser, but this [article](https://tomassetti.me/parsing-in-python/) provides some directions to do this.

ThinkOpenly:
> You gonna need to translate grammar generated in Ruby

Or, call Ruby to do the work. There is some setup required as well, as "encode" needs access to the operands, and "decode" needs access to the fields.

This still needs thought and a fair amount of effort. :-/

henry-hsieh:
> > You gonna need to translate grammar generated in Ruby
> 
> Or, call Ruby to do the work. There is some setup required as well, as "encode" needs access to the operands, and "decode" needs access to the fields.
> 
> This still needs thought and a fair amount of effort. :-/

Sure. However, did you consider merge `encode()` and `decode()` into operand definition instead of putting it in variable declaration? It may reduce the pain of duplicate the same things among the instructions using shared operands. For example, you may duplicate following snippets in every C instruction using cregs:
```yaml
      encode(operands): |
        return reg2creg(operands[xd]);
      decode(): |
        return creg2reg(xd);
```

ThinkOpenly:
> did you consider merge `encode()` and `decode()` into operand definition instead of putting it in variable declaration? It may reduce the pain of duplicate the same things among the instructions using shared operands.

The most difficult example is Zcmp's PUSH/POP register lists. You might have an instruction like `cm.push x1,x8-x9,x18-x27,-160`. If the "operands" are considered to be the comma-separated elements, then how would each of the operands be encoded individually? What does "x1" encode to? or "x8-x9"? or "x18-x27"? In aggregate, they encode to "15", but there are no individual mappings.

ThinkOpenly:
@henry-hsieh 
> Implicit operand (with new implicit attribute), e.g., `c.addi4spn`

Checkout the changes in https://github.com/riscv/riscv-unified-db/pull/1527/changes/5f352f736968180c863f3cc6db75ad13a3f62922:
- implicit program counter
- implicit stack pointer
- implicit 2nd register of a (compressed) register pair

henry-hsieh:
> > did you consider merge `encode()` and `decode()` into operand definition instead of putting it in variable declaration? It may reduce the pain of duplicate the same things among the instructions using shared operands.
> 
> The most difficult example is Zcmp's PUSH/POP register lists. You might have an instruction like `cm.push x1,x8-x9,x18-x27,-160`. If the "operands" are considered to be the comma-separated elements, then how would each of the operands be encoded individually? What does "x1" encode to? or "x8-x9"? or "x18-x27"? In aggregate, they encode to "15", but there are no individual mappings.

I think in RISC-V most operands and encodings are one-to-one map. Except following:
1. Optional omitted operands: Floating rounding mode
2. Show or hide operands based on encoding: Vector vm register
3. Implicit operands that need to show in mnemonic: C extension sp-relative instructions

The Zcmp push/pop instructions are one-to-one map, actually.
In RISC-V [Zcmp spec](https://docs.riscv.org/reference/isa/unpriv/zc.html#Zcmp), the reg_list should be put in curly brackets. The modified example is: `cm.push {x1, x8-x9, x18-x27}, -160`. The problem can be narrowed down to how do we map a reglist (not single reglist component) and its encoding.

henry-hsieh:
> @henry-hsieh
> 
> > Implicit operand (with new implicit attribute), e.g., `c.addi4spn`
> 
> Checkout the changes in [5f352f7](https://github.com/riscv/riscv-unified-db/commit/5f352f736968180c863f3cc6db75ad13a3f62922):
> 
>     * implicit program counter
> 
>     * implicit stack pointer
> 
>     * implicit 2nd register of a (compressed) register pair

LGTM. Thanks for including implicit operand in this commit.

ThinkOpenly:
> 1. Optional omitted operands: Floating rounding mode

You must be referring to what is currently Chapter 27 of the RISC-V Assembly Programmer’s
Manual (https://github.com/riscv-non-isa/riscv-asm-manual), "Floating-point rounding modes":
> For floating-point instructions with a rounding mode field, the rounding mode can be specified by
adding an additional operand. e.g. fcvt.w.s with round-to-zero can be written as fcvt.w.s a0, fa0,
rtz. If unspecified, the default dyn rounding mode will be used.

We don't currently account for that. I confess I wasn't even aware. The current version of this PR should be able to handle that, but the specific optionality needs to be added.

> 2. Show or hide operands based on encoding: Vector vm register

This is already present here.

> 3. Implicit operands that need to show in mnemonic: C extension sp-relative instructions

This is also present.

> The Zcmp push/pop instructions are one-to-one map, actually. In RISC-V [Zcmp spec](https://docs.riscv.org/reference/isa/unpriv/zc.html#Zcmp), the reg_list should be put in curly brackets. The modified example is: `cm.push {x1, x8-x9, x18-x27}, -160`. The problem can be narrowed down to how do we map a reglist (not single reglist component) and its encoding.

Is it? The docs do seem to support that interpretation, although I completely looked through it. Presuming that is the case, I think the implementation could be easier than I have at present. That being said, are there any existing tools that support this? I couldn't get any syntax through the GNU Assembler (2.46) or Clang (22).

henry-hsieh:
> > 1. Optional omitted operands: Floating rounding mode
> 
> You must be referring to what is currently Chapter 27 of the RISC-V Assembly Programmer’s Manual (https://github.com/riscv-non-isa/riscv-asm-manual), "Floating-point rounding modes":
> 
> > For floating-point instructions with a rounding mode field, the rounding mode can be specified by
> > adding an additional operand. e.g. fcvt.w.s with round-to-zero can be written as fcvt.w.s a0, fa0,
> > rtz. If unspecified, the default dyn rounding mode will be used.
> 
> We don't currently account for that. I confess I wasn't even aware. The current version of this PR should be able to handle that, but the specific optionality needs to be added.

Yes, but I think it generally hard to find every assembly syntax in this RISC-V Assembly Programmer’s Manual. I hope we could have golden assembly syntax for every instruction in this database.

> 
> > 2. Show or hide operands based on encoding: Vector vm register
> 
> This is already present here.
> 
> > 3. Implicit operands that need to show in mnemonic: C extension sp-relative instructions
> 
> This is also present.
> 
> > The Zcmp push/pop instructions are one-to-one map, actually. In RISC-V [Zcmp spec](https://docs.riscv.org/reference/isa/unpriv/zc.html#Zcmp), the reg_list should be put in curly brackets. The modified example is: `cm.push {x1, x8-x9, x18-x27}, -160`. The problem can be narrowed down to how do we map a reglist (not single reglist component) and its encoding.
> 
> Is it? The docs do seem to support that interpretation, although I completely looked through it. Presuming that is the case, I think the implementation could be easier than I have at present. That being said, are there any existing tools that support this? I couldn't get any syntax through the GNU Assembler (2.46) or Clang (22).

Can you test `cm.push {x1, x8-x9, x18-x27}, -64` and `cm.push {ra, s0-s11}, -64`? I think `-160` is not valid stack_adj value. Moreover, you may need to specify `-mabi=ilp32 -march=rv32imc_zcmp` to let GCC recognize Zcmp extension. However, I still get following error:
```
riscv64-unknown-elf-gcc: fatal error: Cannot find suitable multilib set for '-march=rv32imc_zmmul_zca_zcmp'/'-mabi=ilp32'
compilation terminated.
```
The GCC version is 15.1.0, with GNU assembler 2.45.

But I can sure GCC rejects the syntax: `cm.push x1, x8-x9, x18-x27, -64`. It said: `` Error: illegal operands `cm.push x1,x8-x9,x18-x27,-64' ``

ThinkOpenly:
This AI-generated example worked for me:
```
.text
    .globl func
func:
    cm.push {ra, s0-s1}, -32
    cm.pop {ra, s0-s1}, 32
    ret
```
...using the command:
```
clang -c -march=rv64imac_zcmp -mabi=lp64 cm-push.s
```
(clang 22.0)
And objdump reproduced that syntax:
```
$ llvm-objdump -dr ./cm-push.o

./cm-push.o:    file format elf64-littleriscv

Disassembly of section .text:

0000000000000000 <func>:
       0: b862          cm.push {ra, s0-s1}, -32
       2: ba62          cm.pop  {ra, s0-s1}, 32
       4: 8082          ret
```
So, changes are needed to support that.

ThinkOpenly:
Oh, my. This has grown a bit, but I think it's ready for closer reviews. There are tests which actually utilize the IDL from Python, which is nice. `idl.py` is not a complete IDL interpreter, only what's necessary and sufficient. I don't think it would be difficult to keep extending that. "v0.t" is still weird and not handled well. There are other things, I'm sure. I need to take a breath and review the current state myself.

henry-hsieh:
Do we have any plan to merge this PR?

ThinkOpenly:
Odd. CI fails, but the same test, in a fresh clone, works for me.

PR #1537, opened 2026-02-04
fix(llvm): verify hint instructions against parent encoding

### LINKED ISSUE #1531: Tracking Hint/MOP Allocations
In the ABI group, we've come to the conclusion that MOP/HINT encodings cannot be reused by different extensions (at least, not in the standard/reserved encoding spaces - conflicting vendor extensions are a different problem). https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/474 (It's not clear to me if we've taken this conclusion back to the core ISA maintainers, which I should probably check)

We have updated the LLVM assembler to match this: https://github.com/llvm/llvm-project/pull/178609 - so the allocated hint/mop mnemonics are always available.

It would be good to track in the unified-db which MOP/HINT encodings have an allocated meaning.

I'm not sure the best way to do this, but we eventually want to model that assemblers always accept the allocated versions even if only Zimop/I is available.

**Additional context**

We currently verify the Unified DB against LLVM. The above LLVM PR is likely to break some of this testing when you update your git submodule.

krrish175-byte:
### Description
The LLVM verification script ([test_parsing.py](cci:7://file:///Users/krrishbiswas/Desktop/LFX/riscv-unified-db/ext/auto-inst/test_parsing.py:0:0-0:0)) was failing or incorrectly skipping hint instructions such as `lpad`. These instructions are defined in UnifiedDB but often do not appear as distinct, top-level entries in the LLVM `riscv.json` dump, or they are implemented as aliases/pseudo-instructions of a parent (e.g., `lpad` is a hint for `auipc`).

Attempting to verify them strictly against their parent's encoding would fail because the hint definition is more specific (e.g., `lpad` requires `rd=0` and `imm=0`, whereas `auipc` allows variable [rd](cci:1://file:///Users/krrishbiswas/Desktop/LFX/riscv-unified-db/tools/ruby-gems/udb/lib/udb/portfolio_design.rb:68:4-68:66) and `imm`).

### Changes
*   Modified [ext/auto-inst/parsing.py](cci:7://file:///Users/krrishbiswas/Desktop/LFX/riscv-unified-db/ext/auto-inst/parsing.py:0:0-0:0): 
    *   Extracts the [hints](cci:1://file:///Users/krrishbiswas/Desktop/LFX/riscv-unified-db/tools/ruby-gems/udb/lib/udb/obj/instruction.rb:1177:4-1180:7) field from YAML definitions.
    *   Updates [compare_yaml_json_encoding](cci:1://file:///Users/krrishbiswas/Desktop/LFX/riscv-unified-db/ext/auto-inst/parsing.py:147:0-255:22) to support an `allow_refinement` flag. This allows a YAML bit to be fixed (e.g., `0`) even if the LLVM JSON bit is variable, which is the correct behavior for hint/subset instructions.
*   Modified [ext/auto-inst/test_parsing.py](cci:7://file:///Users/krrishbiswas/Desktop/LFX/riscv-unified-db/ext/auto-inst/test_parsing.py:0:0-0:0):
    *   Builds a [hint_map](cci:1://file:///Users/krrishbiswas/Desktop/LFX/riscv-unified-db/ext/auto-inst/test_parsing.py:64:4-76:23) linking hint instructions to their parents (e.g., `lpad` -> `auipc`) during setup.
    *   Updates the test loop: if an instruction isn't found in the LLVM data, it checks if it is a known hint.
    *   If identified as a hint, it retrieves the parent's encoding from LLVM and verifies against that, enabling the `allow_refinement` check.

### Related Issues
Fixes #1531

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -119,7 +119,7 @@ namespace :test do
   # "Run the cross-validation against LLVM"
   task :llvm do
     begin
-      sh "/opt/venv/bin/python3 -m pytest tools/auto-inst/test_parsing.py -v"
+      sh "/opt/venv/bin/python3 -m pytest tools/python/auto-inst/test_parsing.py -v"
     rescue => e
       raise unless e.message.include?("status (5)") # don't fail on skipped tests
     end


ext/auto-inst/test_parsing.py
```diff
@@ -130,6 +145,17 @@ def test_instruction_encoding(self, instr_name):
...
+        allow_refinement = False
+        
+        if not json_key:
+            # Check if this instruction is a hint for another instruction
+            parent_name = self.hint_map.get(instr_name.lower())
+            if parent_name:
+                json_key = self._find_matching_instruction(parent_name)
+                if json_key:
+                    allow_refinement = True
+                    # print(f"Checking hint {instr_name} against parent {parent_name}")
```
AFOliveira: This seems like a lost print?

lenary:
I'm not entirely sure about the `allow_refinement` flag. Maybe I don't fully understand the assertions. I will explain my thoughts:

Let's focus on `auipc x0`/`lpad`, which is a simple case:
- There's YAML for `auipc` with encoding info
- There's YAML for `lpad` with encoding info that should be a subset of `auipc`'s YAML encoding information
- There's a TableGen "Instruction" definition for `auipc` (we get it in JSON)
- There's a Tablegen "InstAlias" definition for `lpad` (we get this in JSON)

We can already verify that `auipc`'s TableGen encoding info matches the `auipc` YAML, as I understand.

I think this PR is adding a check that `lpad`'s yaml info is more defined than the `auipc` info in Tablegen, which is good. This makes sense for the `allow_refinement`.

We should probably be checking there is some kind of connection in LLVM between lpad/auipc too. This connection would be via `InstAlias` data in TableGen. This can be a follow-up.


---

I see we don't actually have Zicfiss instruction definitions yet, which is where I was expecting breakage. The complexity there is that the `InstAlias` in LLVM goes in the other direction:
- There's a TableGen "Instruction" definition for `c.sspush`
- There's a Tablegen "InstAlias" for `c.mop.1` that refers to `c.sspush`

Another thing we're not checking is some form of completeness (in either direction), by which i mean "every instruction in unified-db appears in llvm" or vice versa. We're just checking the subset where we can discern that the two data sources overlap. This is fine (LLVM will always have a lot of vendor instructions, and may trail unified db on some standard extensions)

lenary (approved):
LGTM for the allow_refinement check. I don't contribute to unified db itself much, so i will defer to Alfonso/Derek for further comments/fixes.

AFOliveira (commented):
The PR is nice for me, but the CI needs to looked into, before formal approval, please just get that done and it's good for me.

AFOliveira (commented):
Looking again into it, the CI has this bug because of the changed path. It's hardcoded in the CI. Can you please change CI's path as well?

PR #1540, opened 2026-02-04
fix(csr): move HPM access control description to M-mode CSRs

### LINKED ISSUE #1533: Move M-mode HPM description content from Unpriv CSRs to M-mode CSRs
UDB description of HPM counters has information about M-mode version of these CSRs. This information should be moved into the UDB description for the M-mode HPM counters.

Shouldn't be describing M-mode behavior in Unpriv CSRs.

krrish175-byte:
### Motivation
The detailed description of privilege mode access controls (via `mcounteren`, `scounteren`, etc.) was originally placed in the Unprivileged HPM counter definitions (`hpmcounterN`). Since these controls are fundamentally about access *to* the counter (which is an alias of the M-mode counter) and involve M-mode CSRs (`mcounteren`), this information is better placed in the Machine-mode HPM counter definitions (`mhpmcounterN`) to avoid describing higher-privilege behavior in unprivileged CSRs.

### Changes
*   Moved the detailed access control tables and descriptions from:
    *   `spec/std/isa/csr/Zihpm/hpmcounterN.layout`
    *   `spec/std/isa/csr/Zihpm/hpmcounterNh.layout`
*   To:
    *   `spec/std/isa/csr/Zihpm/mhpmcounterN.layout`
    *   `spec/std/isa/csr/Zihpm/mhpmcounterNh.layout`
*   Updated unprivileged HPM counter descriptions to reference their M-mode counterparts for access control details.

### Related Issues
Closes #1533


dhower-qc (changes requested):
These affect generated files, so you'll need to run:

```bash
./do gen:arch
```

.pre-commit-config.yaml
```diff
@@ -20,7 +20,7 @@ repos:
       - id: end-of-file-fixer
         stages: [pre-commit]
         # Exclude reference and generated files
-        exclude: (cfgs/profile.*|.*\.golden\.adoc$|.*\.svg|amo.*\.yaml|container\.def|\.github/workflows/regress\.yml$)
```
ThinkOpenly: Why is this necessary?

ThinkOpenly:
> sorry for the late reply

no worries

> those .yaml files are generated by the ./do gen:arch script. I excluded them because the end-of-file-fixer hook conflicts with the script's output, causing it to fail CI when they get regenerated. 

Sounds like we need to fix `gen:arch`, if I understand correctly. (I'm not sure why these changes would provoke that. You're not changing anything near the end of any files.)

I checked out this branch and ran `pre-commit run --all-files`, and the end-of-file fixer did not complain. (?)

ext/rbi-central
ThinkOpenly, line 1: Do we need this included with this PR? I'm guessing not. Please remove if not needed.

tools/eclipse/udb-vscode/test-fixtures/badGrammar.udb
```diff
@@ -1,4 +1,4 @@
-kind: csr;
+ kind: csr;
```
ThinkOpenly: This should also be removed, I presume.

krrish175-byte:
> I still see `etc/rbi-central`

Fixed, my fork's main was behind upstream, so the submodule pointer was stale. Just reset it to match upstream/main

PR #1555, opened 2026-02-05
fix(data): add reserved_instruction check for AMO operations when misa.A is cleared

ishaan-arora-1:
Update the `operation()` methods for all Zaamo and Zabha instructions to check when the A extension is disabled at runtime.

- Add a `reserved_instruction` function in `globals.isa` that either traps with `IllegalInstruction` or enters an unpredictable state, depending on the `TRAP_ON_UNIMPLEMENTED_INSTRUCTION` parameter.
- Update Zaamo and Zabha layout templates to call `reserved_instruction()` when A extension is disabled at runtime.
- Regenerate all affected instruction YAML files (72 Zaamo + 72 Zabha)

### DIFF
--- a/spec/std/isa/inst/Zaamo/amoadd.SIZE.AQRL.layout
+++ b/spec/std/isa/inst/Zaamo/amoadd.SIZE.AQRL.layout
@@ -69,8 +69,8 @@ access:
   vs: always
   vu: always
 operation(): |
-  if (!implemented?(ExtensionName::<%= %w[b h].include?(size) ? "Zabha" : "A" %>) || (MISA_CSR_IMPLEMENTED && (CSR[misa].A == 1'b0))) {
-    raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
+  if (implemented?(ExtensionName::A) && MISA_CSR_IMPLEMENTED && CSR[misa].A == 1'b0) {
+    reserved_instruction($encoding);
   }
 
 <% if aq -%>
--- a/spec/std/isa/inst/Zaamo/amoadd.d.aq.yaml
+++ b/spec/std/isa/inst/Zaamo/amoadd.d.aq.yaml
@@ -39,8 +39,8 @@ access:
   vs: always
   vu: always
 operation(): |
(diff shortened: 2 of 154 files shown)


dhower-qc (commented):
Good catch. I don't think the spec explicitly states that Zabha can be disabled with misa.A, but it does seem to be implied since Zabha depends on Zaamo which is disabled.

Tracking this question here: https://lists.riscv.org/g/tech-privileged/message/2824

spec/std/isa/inst/Zaamo/amoadd.SIZE.AQRL.layout
```diff
@@ -69,7 +69,7 @@ access:
   vs: always
   vu: always
 operation(): |
-  if (!implemented?(ExtensionName::A) || (MISA_CSR_IMPLEMENTED && (CSR[misa].A == 1'b0))) {
+  if (!implemented?(ExtensionName::<%= %w[b h].include?(size) ? "Zabha" : "A" %>) || (MISA_CSR_IMPLEMENTED && (CSR[misa].A == 1'b0))) {
```
dhower-qc: After discussion in the UDB PR meeting, we determined this check really should be:

```
if (implemented?(ExtensionName::A) && MISA_CSR_IMPLEMENTED && CSR[misa].A == 1'b0) {
  reserved_instruction($encoding);
}
```

where `reserved_instruction` is a function in globals.isa (this is hand-written, probably won't compile):

```
function reserved_instruction {
  arguments Bits<INSTR_ENC_LENGTH> encoding
  body {
    if (TRAP_ON_UNIMPLEMENTED_INSTRUCTION) {
      raise(ExceptionCode::IllegalInstruction, encoding, mode());
    } else {
      unpredictable("Unimplemented instruction that does not trap");
    }
  }
}
```

dhower-qc (changes requested):

spec/std/isa/inst/Zaamo/amoadd.SIZE.AQRL.layout
ishaan-arora-1: Ive done the needfull @dhower-qc, let me know any feedback you have!!

PR #1565, opened 2026-02-07
fix(csr): csr reserved fields

### LINKED ISSUE #1186: CSR Reserved fields
Currently, UDB omits all Reserved/WPRI fields in CSRs as we suppose they are all reserved for the future. However, they have semantic behaviour and more than that, they have differences in semantic behaviour between some.

Let me explain:

WPRI - Software should ignore the values read from these fields, and should preserve the values held in these fields when writing values to other fields
RESERVED - software should write zeros and hardware implementations should return zeros

Therefore, if we plan to continue ommiting WPRI fields, we need to at least explicitly state the Reserved fields that are non-WPRI to avoid downstream tools that rely on this information to mistakenly intake this information as being the same behaviour.

For now, I have only caught reserved in MNSTATUS.

Reference ISA Manual clarification - https://github.com/riscv/riscv-isa-manual/issues/2318

amkr6207:
This PR addresses Issue #1186 by explicitly defining all reserved/WPRI fields in `mnstatus.yaml` as per the RISC-V Smrnmi specification.

**Changes:**
- Explicitly defined `Reserved_High` (Bits 63-13).
- Added `MNPELP` (Bit 9) from Zicfilp extension.
- Added `Reserved` fields for all gaps (Bits 0-2, 4-6, 8, 10).
- Fixed `Reserved` field locations to use valid `rv32`/`rv64` ranges.

**Note:** This is a pilot implementation to establish the correct pattern. Once approved, I will apply this pattern to the remaining CSR files in subsequent PRs.
Closes #1186


spec/std/isa/csr/Smrnmi/mnstatus.yaml
```diff
@@ -20,6 +20,13 @@ definedBy:
   extension:
     name: Smrnmi
 fields:
+  Reserved_High:
```
AFOliveira: I'm not super sure about this name for the reserved, locations... @ThinkOpenly @dhower-qc, any thoughts?

spec/std/isa/csr/Smrnmi/mnstatus.yaml
```diff
@@ -61,6 +68,29 @@ fields:
       } else {
         return true;
       }
+  Reserved_10:
+    location: 10
+    type: RO
+    reset_value: 0
+    description: Reserved (WPRI).
+
+  MNPELP:
```
AFOliveira: Does this mean we were also mssing a bit?

spec/std/isa/csr/Smrnmi/mnstatus.yaml
```diff
@@ -61,6 +68,29 @@ fields:
...
+    description: |
+      M-mode NMI Previous Expected Landing Pad state.
+      (Added by Zicfilp extension)
+    type: RW-H
+    reset_value: UNDEFINED_LEGAL
+    definedBy:
+      extension:
+        name: Zicfilp
+
+  Reserved_8:
```
AFOliveira: same

spec/std/isa/csr/Smrnmi/mnstatus.yaml
```diff
@@ -74,6 +104,13 @@ fields:
     definedBy:
       extension:
         name: H
+
+  Reserved_4_6:
```
AFOliveira: same

spec/std/isa/csr/Smrnmi/mnstatus.yaml
```diff
@@ -95,3 +132,9 @@ fields:
 
     type: RW-H
     reset_value: 0
+
+  Reserved_0_2:
```
AFOliveira: same

.gitignore
```diff
@@ -35,3 +35,7 @@ coverage
 
 # emacs backup files
 *~
+
```
AFOliveira: this changes shouldn't be here.

AFOliveira (changes requested):
Thanks for the contribution! This is new to UDB and may need some discussion. but if you're willing to wait and see, I think this is already along the lines of what to insert, I'm just not sure about the _how_. @ThinkOpenly and @dhower-qc have worked with CSRs inside UDB way more than me, so let's see what they think.

dhower-qc (commented):
The big-picture question on this is whether or not it's needed. What does explicitly naming the reserved fields add beyond just assuming that any unused bits are unnamed WPRI, as is the case now?

spec/std/isa/csr/Smrnmi/mnstatus.yaml
ThinkOpenly: We sure are. Good catch, @amkr6207! This change should really go in a separate PR. It's clearly not "csr reserved fields".

ThinkOpenly:
> > Talked about this in the PR meeting -- general direction is that we should treat missing bits as WPRI, and only document the reserved fields that are _not_ WPRI. Much less work!
> 
> That's what this PR does, right?

The guidance in the spec is:
> The other bits in `mnstatus` are reserved; software should write zeros and hardware implementations should return zeros.

I'm not sure I like that guidance, because it breaks forward compatibility. If I always write zeros to those bits, how can they ever be used in the future? Maybe we need to change the spec there. I think it should really be "the other bits in `mnstatus` are WPRI.`

ThinkOpenly:
FYI, I reopened https://github.com/riscv/riscv-isa-manual/issues/2318, because the words in the spec seem wrong to me.

ThinkOpenly:
> > FYI, I reopened [riscv/riscv-isa-manual#2318](https://github.com/riscv/riscv-isa-manual/issues/2318), because the words in the spec seem wrong to me.
> 
> I don't think it's the words. You seem to be disagreeing with the behavior, which is fine, but as of right now, I think the spec mandates what Greg described and I'm afraid that it has been implemented in real HW, so it seems to late to change it?

I understand your point. I've reopened the issue to get more clarity, for me at least. I will note that "should" != "must". So, that would make actual behavior implementation-dependent, and thus in need of a configuration parameter.

spec/std/isa/csr/Smrnmi/mnstatus.yaml
dhower-qc: Agreed, but I don't have anything that would make it better. Other than maybe having the names with numbers match the order of "location", e.g., Reserved_4_6 --> Reserved_6_4

spec/std/isa/csr/Smrnmi/mnstatus.yaml
amkr6207: Okay. I'll create a new PR for it.

spec/std/isa/csr/Smrnmi/mnstatus.yaml
amkr6207: Okay . I'll rename it to match the location order (  `Reserved_4_6`  --> `Reserved_6_4`  )

spec/std/isa/csr/Smrnmi/mnstatus.yaml
amkr6207: Hi @AFOliveira , @ThinkOpenly

I have raised a new PR #1687 updating MNPELP (bit 9). Please review. Thanks!

PR #1584, opened 2026-02-08
feat: add MMR support with shared HasFields module

AFOliveira:
Introduce memory-mapped registers (MMRs), accessed via physical addresses
rather than the CSR address space. MMRs reuse CSR field definitions but
have a physical_address instead of a 12-bit CSR address, and no privilege
mode or IDL integration.

Extract shared field management into a HasFields module included by both
Csr and Mmr, eliminating code duplication for field access, wavedrom
rendering, and config filtering.

### DIFF
--- a/spec/schemas/mmr_schema.json
+++ b/spec/schemas/mmr_schema.json
@@ -37,11 +37,11 @@
       "description": "One line description for the MMR"
     },
     "description": {
-      "$ref": "schema_defs.json#/$defs/spec_text",
+      "$ref": "schema_defs.json#/definitions/spec_text",
       "description": "Function of the register"
     },
     "definedBy": {
-      "$ref": "schema_defs.json#/$defs/requires_entry",
+      "$ref": "schema_defs.json#/definitions/condition",
       "description": "Extension(s) that define the MMR"
     },
     "physical_address": {
@@ -61,7 +61,7 @@
       "type": "object",
       "patternProperties": {
         "^[a-zA-Z].*$": {
-          "$ref": "csr_schema.json#/$defs/csr_field"
+          "$ref": "csr_schema.json#/definitions/csr_field"
         }
       },
       "description": "fields of this MMR",
--- a/tools/ruby-gems/udb/test/test_mmr_schema.rb
+++ b/tools/ruby-gems/udb/test/test_mmr_schema.rb
@@ -19,8 +19,12 @@ class TestMmrSchema < Minitest::Test
     @schemer = JSONSchemer.schema(
       schema_json,
       ref_resolver: proc { |uri|
-        path = SCHEMA_DIR / uri.path
-        JSON.parse(File.read(path))
+        path = SCHEMA_DIR / File.basename(uri.path)
+        data = JSON.parse(File.read(path))
+        # json_schemer 2.x enforces Draft-07 strictly: $defs is not recognised
+        # (Draft-07 uses "definitions"). Copy $defs so both keywords work.
+        data["definitions"] = data["$defs"] if data["$defs"] && !data.key?("definitions")
+        data
       }
     )
   end
@@ -37,7 +41,7 @@ class TestMmrSchema < Minitest::Test
       "description" => "A simple test register for unit testing.",
       "writable" => true,
       "physical_address" => 0x1000,
-      "definedBy" => { "name" => "Xexample" }
+      "definedBy" => { "extension" => { "name" => "Xexample" } }
     }
   end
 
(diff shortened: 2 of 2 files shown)


spec/schemas/mmr_schema.json
```diff
@@ -0,0 +1,84 @@
...
+
+      "properties": {
+        "$schema": {
+          "type": "string",
+          "format": "uri-reference",
+          "const": "mmr_schema.json#",
+          "description": "Path to schema, relative to <UDB ROOT>/schemas"
+        },
+        "kind": {
+          "type": "string",
```
ThinkOpenly: I don't believe you need a "type" for "const" objects, as it is self-evident.

spec/schemas/mmr_schema.json
```diff
@@ -0,0 +1,84 @@
...
+          "minimum": 0,
+          "description": "Physical memory address of the register"
+        },
+        "writable": {
+          "type": "boolean",
+          "description": "Whether or not the register can be written by software"
+        },
+        "length": {
+          "description": "Length, in bits, of the register. Must be a fixed integer (no XLEN-dependent lengths for MMIO).",
+          "enum": [32, 64]
```
ThinkOpenly: showing my ignorance, is it not possible to have 8, 16, 128? There are atomic stores for those lengths, of course.

AFOliveira: I don't think it's possible to have 128, but 8 and 16 are surely possible, thanks for the catch.

spec/schemas/mmr_schema.json
AFOliveira: fixed.

ThinkOpenly:
I imagine we'd eventually want to make MMRs "first-class" objects in the database, so that things like "./bin/udb list mmrs" work, but maybe we open an issue for that as follow-on work.

This seems like it's reached "MVP" to me. @dhower-qc? @jordancarlin?

spec/schemas/mmr_schema.json
```diff
@@ -0,0 +1,82 @@
...
+        "long_name": {
+          "type": "string",
+          "description": "One line description for the MMR"
+        },
+        "description": {
+          "$ref": "schema_defs.json#/$defs/spec_text",
+          "description": "Function of the register"
+        },
+        "definedBy": {
+          "$ref": "schema_defs.json#/$defs/requires_entry",
```
ThinkOpenly: After #891, I think we now want this:
```suggestion
          "$ref": "schema_defs.json#/$defs/condition",
```
...and requisite changes to the tests.

tools/ruby-gems/udb/test/test_mmr_schema.rb
```diff
@@ -0,0 +1,256 @@
...
+    {
+      "$schema" => "mmr_schema.json#",
+      "kind" => "mmr",
+      "name" => "example_reg0",
+      "long_name" => "Example Register 0",
+      "length" => 32,
+      "description" => "A simple test register for unit testing.",
+      "writable" => true,
+      "physical_address" => 0x1000,
+      "definedBy" => { "name" => "Xexample" }
```
ThinkOpenly:
```suggestion
      "definedBy" => { "extension" => { "name" => "Xexample" } }
```

tools/ruby-gems/udb/test/test_mmr_schema.rb
```diff
@@ -0,0 +1,256 @@
...
+    assert_valid(data, "MMR with large physical_address")
+  end
+
+  # SHOULD PASS: valid MMR with definedBy as allOf-style
+  def test_valid_mmr_defined_by_complex
+    data = valid_mmr.merge(
+      "definedBy" => {
+        "allOf" => [
+          { "name" => "Xext1" },
+          { "name" => "Xext2" }
```
ThinkOpenly:
```suggestion
          { "extension" => { "name" => "Xext1" } },
          { "extension" => { "name" => "Xext2" } }
```

AFOliveira:
@ThinkOpenly I think it's solved, I also added your suggestions. Thanks!

tools/ruby-gems/udb/test/test_mmr_schema.rb
AFOliveira: Fixed.

tools/ruby-gems/udb/test/test_mmr_schema.rb
AFOliveira: done

spec/schemas/mmr_schema.json
AFOliveira: done

PR #1589, opened 2026-02-09
fix: add missing argument to sext call in DecodeVariable#extract

lucifer4330k:
Fixed a bug in tools/ruby-gems/udb/lib/udb/obj/instruction.rb where the sext function call generated for sign-extended decode variables was missing the second argument (the bit width/sign bit position). This aligns the generated IDL code with the sext function definition in util.idl.


dhower-qc (changes requested):
You'll need to update the golden appendix:

then commit and push. That will fix CI

.gitignore
```diff
@@ -35,3 +35,4 @@ coverage
 
 # emacs backup files
 *~
+vendor/bundle
```
ThinkOpenly: This is unrelated to the PR. Please remove it from this PR. If it is needed in general, please open a separate PR.

ThinkOpenly (changes requested):

tools/ruby-gems/udb/test/test_instruction.rb
```diff
@@ -0,0 +1,64 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
```
dhower-qc: This should be your own copyright.

jordancarlin (approved):

PR #1590, opened 2026-02-09
feat: drop Singularity support and use `mise` and `uv` for package management

### LINKED ISSUE #1394: Podman fails to pull pre-build image due to node_modules UID/GID

jordancarlin:
Overhauls tool/container management:
- Drops support for Singularity/Apptainer
- Update Python infrastructure to use modern `pyproject.toml` and install packages using `uv`. `uv` is much faster and creates venvs automatically as needed. This will make a future native option much easier.
- Use `mise` to manage `python`, `node`, and `ruby` and install a few other tools. This gives us control over all of the versions and will make a future native option easier.

This is now working locally and in CI.

supersedes #1398
fixes #1394

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -5,7 +5,6 @@ on:
       - main
   workflow_dispatch:
 jobs:
-
   build-docker-container:
     env:
       DOCKER: 1
@@ -20,7 +19,7 @@ jobs:
     runs-on: ${{ matrix.runner }}
     steps:
       - name: Clone Github Repo Action
-        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Get container tag
         run: |
@@ -45,7 +44,7 @@ jobs:
 
       - name: Build and push by digest
         id: build
-        uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83 # v6.18.0
+        uses: docker/build-push-action@601a80b39c9405e50806ae38af30926f9d957c47 # v6.19.1
         with:
           file: ./.devcontainer/Dockerfile
           platforms: ${{ matrix.platform }}
@@ -74,7 +73,7 @@ jobs:
       - build-docker-container
     steps:
       - name: Clone Github Repo Action
-        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Download digests
         uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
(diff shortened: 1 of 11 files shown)


jordancarlin:
@dhower-qc @ThinkOpenly this is a big change, but it would be great to try getting it reviewed and merged quickly because the version of node in the old container is breaking the container for many users of act4.

.ruby-version
dhower-qc, line 1: We want to keep this file since it's used by rbenv (and picked up by, e.g., the RubyLSP vscode extension).

I think mise can be told to look at this file instead of writing the ruby version directly in mise.toml

dhower-qc (changes requested):

.ruby-version
jordancarlin: Normally it can and that's what I tried at first, but unfortunately it causes trust issues in the Docker container. We aren't using rbenv anymore, so I'm not sure that it matters.

jordancarlin: @dhower-qc Normally it can and that's what I tried at first, but unfortunately it causes trust issues in the Docker container. We aren't using rbenv anymore, so I'm not sure that it matters.

dhower-qc: Are the trust issues resolvable? Removing this would break the Ruby vscode extensions

jordancarlin: I can try to resolve them, but it looks like the Ruby LSP extension supports `mise` natively: https://shopify.github.io/ruby-lsp/#version-manager-integrations.

dhower-qc: ah, I rescind my concern then

jordancarlin:
This is now working with docker, podman, and devcontainer. It is failing for currently unclear reasons with Singularity. Is it worth me figuring out why or are we good to go forward with dropping support for Singularity as discussed?

dhower-qc:
let me ping inside QC to see if everyone has switched over

jordancarlin:
This should be good to go now.

PR #1598, opened 2026-02-10
Move unknownliteral test to unit test 

lucifer4330k:
Moved the test code for UnknownLiteral which was executing at load time in tools/ruby-gems/idlc/lib/idlc/ast.rb to a proper unit test file tools/ruby-gems/idlc/test/test_unknown_literal.rb. This ensures that the test is run with the rest of the suite and doesn't clutter the production code.

### DIFF
--- a/tools/ruby-gems/udb/lib/udb/obj/instruction.rb
+++ b/tools/ruby-gems/udb/lib/udb/obj/instruction.rb
@@ -851,7 +851,7 @@ module Udb
           else
             ops[0]
           end
-        ops = "sext(#{ops}, #{size_in_encoding})" if sext?
+        ops = "sext(#{ops}, #{size})" if sext?
         ops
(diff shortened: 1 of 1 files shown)


tools/ruby-gems/udb/lib/udb/obj/instruction.rb
```diff
@@ -851,7 +851,7 @@ def extract
           else
             ops[0]
           end
-        ops = "sext(#{ops})" if sext?
+        ops = "sext(#{ops}, #{size})" if sext?
```
dhower-qc, line 854: This looks like it belongs in another PR, and is causing CI failures

lucifer4330k: i have created a new pr for this issue

PR #1611, opened 2026-02-12
fix(data): Sv32 and Sv39 require S

jordancarlin:


jordancarlin:
That's a good question and might be worth an issue on riscv-isa-manual. The extension is not mentioned in the list of extensions/versions in the preface and does not have a version number in the sub-chapter name.

PR #1617, opened 2026-02-12
Create autofix.yml

dhower-qc:
Trying out autofix

### DIFF
--- a/container.def
+++ b/container.def
@@ -5,6 +5,8 @@ Bootstrap: docker
 From: ubuntu:24.04
 Stage: spython-base
 
+
+
 %files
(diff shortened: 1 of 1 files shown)


.github/workflows/autofix.yml
```diff
@@ -0,0 +1,22 @@
+name: autofix.ci  # needed to securely identify the workflow
+
+on:
+  pull_request:
+  push:
+    branches: [ "main" ]
```
jordancarlin, line 6: I'd rather we limit this to running on PRs only. I don't think we want it trying to push things to main.

PR #1624, opened 2026-02-13
feat: Renovate improvements

jordancarlin:
- Switch to json5 so we can include comments
- Don't rebase branches outside of scheduled window
- Another attempt at only updating LLVM weekly
- Extend default github-actions manager instead of using a custom manager for CI templates
- Reenable Dockerfile updates now that Singularity is gone
- Add .mise.toml and Gemfile to list of files that should bump the container tag

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,4 +5,4 @@
 tools/eclipse @adingank-qualcomm @dhower-qc
 
 # Renovate owned by Jordan Carlin
-renovate.json @jordancarlin
+renovate.json5 @jordancarlin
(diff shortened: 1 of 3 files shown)


ThinkOpenly:
We may want to hold off on adding comments to JSON. I'm not sure how widely adopted that non-backwards-compatible change is. (Although, maybe not much downstream dependence on the JSON here.)

jordancarlin:
I don’t think anything else uses the renovate config, so this is unlikely to be an issue. For now, I think adding the documentation is more valuable than keeping pure json, but we can always remove the comments down the road if necessary.

PR #1669, opened 2026-02-16
feat(data): add operation() IDL for Zawrs wrs.nto and wrs.sto instructions

ishaan-arora-1:
## Summary

- Add the missing `operation()` IDL implementations for both Zawrs wait-on-reservation-set instructions
- Add `wrs_nto` and `wrs_sto` builtin functions in `builtin_functions.idl`, following the same pattern as the existing `wfi` builtin

### `wrs.nto` (Wait-on-Reservation-Set with No Timeout)
- Implements TW/VTW trap checks similar to WFI, with the key spec difference: **does NOT trap in U-mode when TW=0** (unlike WFI which always traps in U-mode)
- When `mstatus.TW=1`, any mode other than M traps with `IllegalInstruction`
- When H extension is present, `mstatus.TW=0`, and `hstatus.VTW=1`, VS/VU mode traps with `VirtualInstruction`
- Delegates actual stall behavior to `wrs_nto()` builtin

### `wrs.sto` (Wait-on-Reservation-Set with Short Timeout)
- No TW/VTW checks needed — the built-in implementation-defined short timeout guarantees forward progress
- Delegates stall behavior to `wrs_sto()` builtin

## References

- [Zawrs Spec v1.01](https://docs.riscv.org/reference/isa/unpriv/zawrs.html)

## Test plan

- [x] All YAML files pass JSON schema validation (pre-commit `check-jsonschema`)
- [x] IDL compiles successfully (`./do chore:update_golden_appendix` completes without error)
- [x] All pre-commit checks pass

### DIFF
--- a/backends/cpp_hart_gen/cpp/include/udb/hart.hpp
+++ b/backends/cpp_hart_gen/cpp/include/udb/hart.hpp
@@ -134,6 +134,14 @@ namespace udb {
       //throw WfiException();
     }
 
+    void wrs_nto() {
+      // no-op: a valid implementation per the Zawrs spec
+    }
+
+    void wrs_sto() {
+      // no-op: a valid implementation per the Zawrs spec
+    }
+
     void pause() {
       //Do nothing for now
       //throw PauseException();


spec/std/isa/inst/Zawrs/wrs.nto.yaml
```diff
@@ -53,3 +53,25 @@ access:
...
 operation(): |
+  # wrs.nto follows TW/VTW trap rules similar to WFI, but unlike WFI it
+  # does NOT cause an illegal instruction exception in U-mode when TW=0.
+
+  # When TW=1, any mode other than M traps with IllegalInstruction
+  if ((CSR[misa].S == 1) && (CSR[mstatus].TW == 1'b1)) {
+    if (mode() != PrivilegeMode::M) {
+      raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+    }
+  }
```
ThinkOpenly, line 64: I know this was merged already, but I'm puzzled in comparing this code to the spec:
> When the `TW` (Timeout Wait) bit in `mstatus` is set and `WRS.NTO` is executed in any privilege mode other than M mode, and it does not complete within an implementation-specific bounded time limit, the `WRS.NTO` instruction will cause an illegal-instruction exception.

The code does not hew to that precisely. Why is `misa.S` checked?

spec/std/isa/inst/Zawrs/wrs.nto.yaml
```diff
@@ -53,3 +53,25 @@ access:
...
+
+  # When H extension is present, TW=0, and VTW=1, VS/VU mode traps
+  # with VirtualInstruction. Unlike WFI, VU does NOT trap when VTW=0.
+  if (CSR[misa].H == 1) {
+    if ((CSR[mstatus].TW == 1'b0) && (CSR[hstatus].VTW == 1'b1)) {
+      if ((mode() == PrivilegeMode::VS) || (mode() == PrivilegeMode::VU)) {
+        raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
+      }
+    }
+  }
```
ThinkOpenly, line 74: Similarly, is this code trying to represent this from the spec:
> When executing in VS or VU mode, if the `VTW` bit is set in `hstatus`, the `TW` bit in `mstatus` is clear, and the `WRS.NTO` does not complete within an implementation-specific bounded time limit, the `WRS.NTO` instruction will cause a virtual-instruction exception.

Because it doesn't seem like a match to me.

spec/std/isa/inst/Zawrs/wrs.nto.yaml
ishaan-arora-1: The mstatus.TW field is defined as part of the S extension, without S, TW is hardwired to 0 or doesn't exist. The misa.S guard ensures we only check TW when it's meaningful. This follows the same pattern as the WFI instruction.

spec/std/isa/inst/Zawrs/wrs.nto.yaml
ishaan-arora-1: Yes, this code represents that spec text. Breaking it down: CSR[misa].H == 1 guards that hstatus/VTW exists, then CSR[mstatus].TW == 1'b0 checks "TW is clear", CSR[hstatus].VTW == 1'b1 checks "VTW is set", and mode() == VS || VU checks the privilege mode matching all the conditions in the spec passage. The misa.H guard follows the same pattern as WFI. Is there a specific part that looks off?

spec/std/isa/inst/Zawrs/wrs.nto.yaml
ThinkOpenly: OK. And what about the "bounded time limit" _before_ raising an exception?

spec/std/isa/inst/Zawrs/wrs.nto.yaml
ThinkOpenly: It seems like all of the conditions are covered except "...and the `WRS.NTO` does not complete within an implementation-specific bounded time limit". I see the wait appears below, but the checks must all be true before the exception is raised, and we haven't attempted the wait yet.

PR #1672, opened 2026-02-18
move Docker image loading into reusable workflow for CI

jordancarlin:


tools/test/regress-gh-template.yaml
```diff
@@ -54,14 +54,6 @@ jobs:
...
           name: docker_image
           path: ${{ runner.temp }}/docker_image.tar
-  never-runs:
-    # this just exists so that renovate can keep the download-artifact pointer up-to-date
-    # we use it for regressions later
-    runs-on: ubuntu-latest
-    if: "false"
-    steps:
-      - name: Use download-artifact
-        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7
```
dhower-qc, line 64: Is this supposed to be here?

PR #1674, opened 2026-02-18
Correct and complete asserts in _Bits::extract

ThinkOpenly:
When extracting a contiguous string of bits from a `_Bits` object of N bits, we need to ensure 4 conditions for the (presumably unsigned) start and end bit positions (also presuming bit indexes are numbered in decreasing order (N-1)..0):
1. start >= end    # negative range
2. end < N         # end out of range
3. start < N       # start out of range
4. start - end < N # range is too large

The asserts in _Bits::extract are currently:
- start >= end
- end <= N

Of note:
- The second is an incorrect version of (2): off by 1.
- (3) is missing.
- (4) can be derived from (1) and (3), so does not need to be explicitly expressed.

Fix the incorrect condition (2) and add the missing one (3).

### DIFF
--- a/backends/cpp_hart_gen/cpp/include/udb/bits.hpp
+++ b/backends/cpp_hart_gen/cpp/include/udb/bits.hpp
@@ -1119,7 +1119,7 @@ namespace udb {
     >
     constexpr _Bits<N, false> extract(const MsbType<MsbN, false>& msb, const LsbType<LsbN, false>& lsb) const {
       udb_assert(msb >= lsb, "Negative range is not allowed");
-      udb_assert(lsb.get() <= N, "Extract out of range");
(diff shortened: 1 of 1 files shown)


dhower-qc (approved):

ThinkOpenly:
CI failed in the merge queue:
```
> Run actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131
  with:
    name: docker_image
    path: /home/runner/work/_temp
    merge-multiple: false
    repository: riscv/riscv-unified-db
    run-id: 22186521122
  env:
    DOCKER: 1
    DEPLOY_DIR: _site
Downloading single artifact
Error: Unable to download artifact(s): Failed to ListArtifacts: Received non-retryable error: Failed request: (403) Forbidden: Error from intermediary with HTTP status code 403 "Forbidden"
```

PR #1696, opened 2026-02-26
Add `cert_normative_rules` to extension requirements 

### LINKED ISSUE #1689: Add extension dependency norm rule references
@allenjbaum has suggested that we tag normative rules to capture extension dependencies since they are normative rules. 

Can this be added into the extensions YAML schema to add a reference to the normative rule that defines extension dependencies?

CC: @mmhus

krrish175-byte:
Resolves #1689.

This PR adds the ability to tag normative rules to track extension dependencies, as suggested in #1689.

**Changes:**
- Modified `extension_requirement` in [spec/schemas/schema_defs.json](cci:7://file:///Users/krrishbiswas/Desktop/LFX/riscv-unified-db/spec/schemas/schema_defs.json:0:0-0:0) to include an optional `cert_normative_rules` array. 
- Added an optional `note` string field to allow arbitrary context to be recorded explaining why a dependency exists.

By adding `cert_normative_rules`, extensions can now reference the specific normative rule ID that mandates the dependency directly in their `requirements` block.

### DIFF
--- a/spec/schemas/schema_defs.json
+++ b/spec/schemas/schema_defs.json
@@ -1,8 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-
   "title": "Common patterns used by all schemas",
-
   "$defs": {
     "$source": {
       "type": "string",
@@ -34,7 +32,10 @@
     },
     "field_location": {
       "oneOf": [
-        { "type": "integer", "description": "Location of a single bit" },
+        {
+          "type": "integer",
+          "description": "Location of a single bit"
+        },
         {
           "type": "string",
           "pattern": "^[0-9]+-[0-9]+$",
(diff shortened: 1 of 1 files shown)


spec/schemas/schema_defs.json
```diff
@@ -34,7 +32,10 @@
     },
     "field_location": {
       "oneOf": [
-        { "type": "integer", "description": "Location of a single bit" },
```
james-ball-qualcomm, line 37: Wondering why you are making this change and all the changes splitting enum arrays into multiple lines. Doesn't seem required or related.

james-ball-qualcomm (changes requested):
It's too hard to find the relevant changes given all the unrelated (and unnecessary) formatting changes.

PR #1702, opened 2026-02-28
fix: use $HOME instead of $H for XDG_CACHE_HOME fallback

latifbhatti:
The script previously used **${H}/.cache** as fallback for **XDG_CACHE_HOME**,
but **$H** is not a standard environment variable. Changed to use **$HOME**,
which is the standard environment variable for user home directory.
This prevents the script from defaulting to '**/.cache**' when neither
variable is set, avoiding permission issues.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -194,7 +194,7 @@ setup_z3() {
 
   local xdg_cache
   if [ "$CONTAINER_TYPE" == "native" ]; then
-    xdg_cache="${XDG_CACHE_HOME:-${H}/.cache}"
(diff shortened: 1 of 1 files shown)


ThinkOpenly (changes requested):
Thank you for this contribution!

A couple of comments/requests:
- There are many other changes included in this PR, possibly from another PR or branch? Could you make it so that this PR has just the one change?
- The specific change `s/H/HOME/` is also in #1615, but frankly shouldn't be included in 1615. If you can get this fixed up soon, we can merge this and remove it from 1615.

PR #1703, opened 2026-03-01
Fix sspush/sspop

ThinkOpenly:
`sspush` and `sspopchk` are each single mnemonics that take a very restricted set of operand values.

Also, improve descriptions.


ThinkOpenly:
@dhower-qc, I think the CI tests are failing because the encodings for `sspopchk` and `ssrdp` overlap _if_ the exclusions (`not` attributes) are not accounted for.

`ssrdp` has:
```
  match: 11001101110000000100-----1110011
  variables:
    - name: xd
      location: 11-7
      not: 0
```
`sspopchk` has:
```
  match: "110011011100-----100000001110011"
  variables:
    - name: xs1
      location: 19-15
      not: [ 0, 2, 3, 4, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31 ]
```
... so they each have a hard value of zero where the other has a variable that cannot be 0.

Is there a better way to do this, or do we need to enhance the decode tree implementation?

ThinkOpenly:
Closing in lieu of #2330

PR #1722, opened 2026-03-05
feat: GDBServer support for ISS

henrikg-qc:
This change includes:

- GDB Server support in ISS
  Enable debug support with -g switch.
  Halt ISS before running program with --halt.
  Target port configurable with -p, or --gdbport, switch. Default port 2159.
- VSCode RISC-V application debug launch target
  Debug RISC-V application with stepping, breakpoints, register and memory view in VSCode.
- Standalone GDB debugging
  Target register description for RV64 available under riscv-unified-db/backends/cpp_hart_gen/gdb.

### DIFF
--- a/spec/schemas/csr_schema.json
+++ b/spec/schemas/csr_schema.json
@@ -129,7 +129,7 @@
 
       "allOf": [
         {
-          "$comment": "Location is required, but it can either be a static value or a function",
+          "$comment": "Location is required, but it can either be a single XLEN-independent value or a value for each XLEN",
           "oneOf": [
             {
               "required": ["location"]


backends/cpp_hart_gen/cpp/include/udb/hart.hpp
```diff
@@ -73,10 +92,8 @@ namespace udb {
     // get the next instruction encoding
     virtual uint64_t fetch() = 0;
 
-    PrivilegeMode mode() const { return m_current_priv_mode; }
-    void set_mode(const PrivilegeMode& next_mode) {
-      m_current_priv_mode = next_mode;
-    }
+    virtual PrivilegeMode GetMode() = 0;
+    virtual void SetMode(const PrivilegeMode& next_mode) = 0;
```
dhower-qc: any reason these use CamelCase instead of snake_case like the other methods?

backends/cpp_hart_gen/cpp/src/iss.cpp
```diff
@@ -5,116 +5,629 @@
...
+#include "udb/NotificationHandler.hpp"
+
+#define RISCV_REG_GPR_FIRST 0
+#define RISCV_REG_GPR_LAST  0x1f
+#define RISCV_REG_PC        0x20
+#define RISCV_REG_FPR_FIRST 0x21
+#define RISCV_REG_FPR_LAST  0x40
+#define RISCV_REG_FPR_LAST  0x40
+#define RISCV_REG_CSR_FIRST 0x41
+#define RISCV_REG_CSR_LAST  0x1040
```
dhower-qc, line 26: are these gdb definitions or something we made up?

backends/cpp_hart_gen/cpp/src/iss.cpp
```diff
@@ -5,116 +5,629 @@
...
+
+// RISCV/GDB register ranges
+constexpr int REG_GPR_FIRST = 0;
+constexpr int REG_GPR_LAST = 0x1F;
+constexpr int REG_PC = 0x20;
+constexpr int REG_FPR_FIRST = 0x21;
+constexpr int REG_FPR_LAST = 0x40;
+constexpr int REG_CSR_FIRST = 0x41;
+constexpr int REG_CSR_LAST = 0x1040;
+constexpr int REG_PRIV = 0x1041;
```
dhower-qc: why do we have macro and global variable definitions for these constants?

backends/cpp_hart_gen/cpp/src/iss.cpp
henrikg-qc: GDB definitions

backends/cpp_hart_gen/cpp/src/iss.cpp
henrikg-qc: ?...haha I'll remove those

backends/cpp_hart_gen/cpp/include/udb/hart.hpp
henrikg-qc: Changed to _set_mode and _get_mode as of the IDL definitions of the same name

backends/cpp_hart_gen/cpp/src/iss.cpp
henrikg-qc: Removed

backends/cpp_hart_gen/templates/csrs_impl.hxx.erb
```diff
@@ -90,7 +90,7 @@ void <%= name_of(:csr_field, cfg_arch, csr.name, field.name) %><SocType>::reset(
 #define __UDB_FUNC_CALL m_parent->
 #define __UDB_CONSTEXPR_FUNC_CALL <%= name_of(:hart, cfg_arch) %><SocType>::
 #define __UDB_XLEN m_parent->xlen().to_defined()
-#define __UDB_ENCODING Bits<<%= name_of(:hart, cfg_arch) %><SocType>::__MAX_INST_ENCODING_SIZE>{m_parent->m_cur_inst->encoding()}
+#define __UDB_ENCODING Bits<<%= name_of(:hart, cfg_arch) %><SocType>::__MAX_INST_ENCODING_SIZE>{m_parent->m_cur_inst ? m_parent->m_cur_inst->encoding() : 0x23 }
```
ThinkOpenly: What's the magic "0x23"?

backends/cpp_hart_gen/gdb/riscv-rv64.xml
```diff
@@ -0,0 +1,337 @@
...
+-->
+<target version="1.0">
+  <architecture>riscv:rv64</architecture>
+
+  <!-- ================================================================
+       Integer registers  -  regnum 0-31 (x0-x31) + regnum 32 (pc)
+       ================================================================ -->
+  <feature name="org.gnu.gdb.riscv.cpu">
+    <!-- x0 / zero -->
+    <reg name="zero" bitsize="64" type="int"      group="general" regnum="0"/>
```
ThinkOpenly, line 27: Sure would be nice to use UDB to generate this content. ;-) See `spec/std/isa/register_file/X.yaml`.

ThinkOpenly (commented):
Style police report varied uses of where the opening brace of a thing goes, sometimes on the same line:
```
thing {
```
sometimes on the next line:
```
thing
{
```
NBD, as we don't have a project style guide. Just noting if you wanted to be more consistent. (I prefer the former, to conserve vertical space.)

backends/cpp_hart_gen/templates/csrs_impl.hxx.erb
henrikg-qc: NOP = ADDI, x0, x0, 0
I can add a #define to clarify

henrikg-qc: This is a fix to a problem introduced by some reactor at some time where m_cur_inst ended up never getting set.
Without the check the ISS would SEGFAULT every time an unimplemented CSR was read.
m_cur_inst has since been used, but it will still be nullptr before the programs starts running. The debugger can still read CSR at this point, so the check is needed,

backends/cpp_hart_gen/gdb/riscv-rv64.xml
henrikg-qc: Yes, I already mentioned this to Derek. This is just a RV64 template I used to test GDB with. AI generated of course (with corrections as it failed to close some XML elements). 
I can create an issue for generating a target xmlRegister file for the configuration, but that s not included in this PR.

backends/cpp_hart_gen/templates/csrs_impl.hxx.erb
henrikg-qc: Fixed with #define...and I had the encoding wrong too.

backends/cpp_hart_gen/gdb/riscv-rv64.xml
henrikg-qc: Created #1728 to implement target description in XML format

PR #1776, opened 2026-03-31
Fix `noneOf` condition handling

### LINKED ISSUE #1775: Bare S mode with TRAP_ON_SFENCE_VMA_WHEN_SATP_MODE_IS_READ_ONLY
**Describe the bug**
I am not sure if this a bug.
I am trying to test with act4 in Bare S mode with sfence.vma not supported.
So I enable S without any Svxx (and regardless of Svbare, but here without Svbare) and set TRAP_ON_SFENCE_VMA_WHEN_SATP_MODE_IS_READ_ONLY=true.
I get an error.

**To Reproduce**
running act4 with the container image 0.40.

(I kept only the part of the condition that is false)
```
  * Parameter is not defined by this config: 'TRAP_ON_SFENCE_VMA_WHEN_SATP_MODE_IS_READ_ONLY'.
... (`Sv32=1.11.0`{false} || `Sv32=1.12.0`{false} || `Sv32=1.13.0`{false} || `Sv39=1.11.0`{false} || `Sv39=1.12.0`{false} || `Sv39>=1.13`{false}) && ...
```

jordancarlin:
`noneOf` was treated as OR instead of NOR when in an `ExtensionCondition`. This updates the behavior to match ParamCondition.

`TRAP_ON_SFENCE_VMA_WHEN_SATP_MODE_IS_READ_ONLY` is removed from several configs as a result of this because it is no longer defined for those configs.

Fixes #1775


tools/ruby-gems/udb/lib/udb/condition.rb
```diff
@@ -2007,7 +2007,10 @@ def to_logic_tree_helper(yaml)
         elsif yaml.key?("anyOf")
           LogicNode.new(LogicNodeType::Or, yaml["anyOf"].map { |node| to_logic_tree_helper(node) })
         elsif yaml.key?("noneOf")
-          LogicNode.new(LogicNodeType::Or, yaml["noneOf"].map { |node| to_logic_tree_helper(node) })
+          LogicNode.new(
+            LogicNodeType::Not,
+            [LogicNode.new(LogicNodeType::Or, yaml["noneOf"].map { |node| to_logic_tree_helper(node) })]
+          )
```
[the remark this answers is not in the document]
jordancarlin: Tests added.

cfgs/example_rv64_with_overlay.yaml
```diff
@@ -606,7 +606,6 @@ params:
   TRAP_ON_ECALL_FROM_M: true
   TRAP_ON_ECALL_FROM_U: true
   TRAP_ON_EBREAK: true
-  TRAP_ON_SFENCE_VMA_WHEN_SATP_MODE_IS_READ_ONLY: false
```
ThinkOpenly, line 609: Why are these config changes here?

jordancarlin: That parameter is not supposed to be defined for this config. It was previously incorrectly required because of the or vs nor mixup. It now fails validation if that parameter is set.

dhower-qc (approved):

ThinkOpenly (approved):

PR #1780, opened 2026-04-07
feat: add backend for an instruction table

thecharlesjenkins:
Add a generator for a single-file instruction table that can be used by downstream
projects to generate their own content specific to RISC-V instructions.

```
./bin/generate inst-table [--cfg=CFG] [-o output_file_name]
```

(One use-case is for the Linux kernel, for use in generating instruction headers.
See https://lore.kernel.org/linux-riscv/20260407-riscv_insn_table-v1-0-54b4736a1e77@gmail.com/T/#t.)

Signed-off-by: Charlie Jenkins <thecharlesjenkins@gmail.com>
Co-developed-by: Paul A. Clarke <paul.clarke@qti.qualcomm.com>


ThinkOpenly:
> To be able to commit to Linux, I am marking the generated file as GPL-2.0-only and Reuse requires adding the License file to this repo for that.

I don't believe generated content intended for a GPL project needs to licensed as GPL at creation. It does need to be GPL-compatible, and there is a good description of many licenses and their compatibility with GPL at https://www.gnu.org/licenses/license-list.en.html. As an example, you'll find "MIT license" not only compatible with GPL, but also already present here.

> organize all of the requested instructions into a table modeled after the syscall table in Linux

What are "requested instructions"? Put differently, what exactly are you generating? Are you intending to replace existing content, or generate new content? How is it used in Linux? (Is this for instruction single-step?)

ThinkOpenly:
> Oh interesting, if it's preferred to have a different license I can do that to.

To me, it seems like the "easy button" is to use the same license for content generated from other content as the original content. The YAML files here are BSD-based (described as compatible with GPL at the link above as "The Clear BSD License").

> Here is the patch series to Linux I just sent out to use this table https://lore.kernel.org/linux-riscv/20260407-riscv_insn_table-v1-0-54b4736a1e77@gmail.com/T/#t. Instruction single-step is part of it but there are a ton of places that use manually generated instructions that I swap over to using this table in the series.

That seems very, very cool! I'd love to look at the script that performs the generation, but it doesn't appear that patch (01/16) made it to the mailing list archives. :-/

backends/inst_table/lib/udb/generate_inst_table.rb
```diff
@@ -0,0 +1,157 @@
...
+require 'fileutils'
+require 'optparse'
+require 'udb/resolver'
+
+module Udb
+  module GenerateInstTable
+    module_function
+
+    def parse_range_str(range)
+      if range.to_s.include?('...')
```
ThinkOpenly: (I'm far from a Ruby expert.) Is this looking for the _literal_ `...`? That does not exist in the database that I see.

backends/inst_table/lib/udb/generate_inst_table.rb
```diff
@@ -0,0 +1,157 @@
...
+        a,b = range.to_s.split('...')
+        lo = a.to_i
+        hi = b.to_i - 1
+      elsif range.to_s.include?('..')
```
ThinkOpenly: same

backends/inst_table/lib/udb/generate_inst_table.rb
```diff
@@ -0,0 +1,157 @@
...
+        fields = []
+
+        enc_obj = inst.encoding(base)
+        fixed = enc_obj.opcode_fields.map do |fo|
+          lo, hi = parse_range_str(fo.range.to_s)
+          "#{fo.name}<#{lo}"
+        end
+
+        fields.append(fixed.join("|"))
+
```
ThinkOpenly: extra blank line here. intended?

backends/inst_table/lib/udb/generate_inst_table.rb
```diff
@@ -0,0 +1,157 @@
...
+          unless d.excludes.empty?
+            field += "!#{d.excludes.join('!')}"
+          end
+
+          if d.left_shift > 0
+            field += "<#{d.left_shift}"
+          end
+
+          field+="=#{d.location.to_s}"
```
ThinkOpenly:
```suggestion
          field += "=#{d.location.to_s}"
```

backends/inst_table/lib/udb/generate_inst_table.rb
```diff
@@ -0,0 +1,157 @@
...
+
+          fields.append(field)
+        end
+
+        return fields
+    end
+
+    def generate(root:, cfg: '_', out_path: 'out/inst_table.txt')
```
ThinkOpenly: I'd prefer the `out_path` not have a random default. I suggest either making this a required parameter, or (preferable) write to stdout as default.

backends/inst_table/lib/udb/generate_inst_table.rb
```diff
@@ -0,0 +1,157 @@
...
+      resolver = Udb::Resolver.new(root)
+      cfg_arch = resolver.cfg_arch_for(cfg)
+
+      lines = []
+      cfg_arch.instructions.each do |inst|
+        enc_32 = get_encoding(inst, 32)
+        enc_64 = get_encoding(inst, 64)
+
+        # Skip if instruction not supported by either 32 or 64-bit
+        next if enc_32.empty? && enc_64.empty?
```
ThinkOpenly: should we also emit some sort of message, maybe a warning or "info"?

backends/inst_table/lib/udb/generate_inst_table.rb
```diff
@@ -0,0 +1,157 @@
...
+        end
+
+        if enc_64.any?
+          lines << ([inst.name, "64"] + enc_64).join(' ')
+          next
+        end
+      end
+
+      header = <<EOM
+# SPDX-License-Identifier: GPL-2.0-only
```
ThinkOpenly: Discussed in the issue conversation, but I think the easier path here is to use "BSD-3-Clause-Clear".

backends/inst_table/lib/udb/generate_inst_table.rb
```diff
@@ -0,0 +1,157 @@
...
+#                             a "constraint" is an integer value that is invalid for this variable
+# VARIABLE_LIST               a variable sized list of all variables in the instruction definition
+#                             in the form of name[~][<num][!num...]=(high[-low])|...
+#                             symbols after the name represent different modifiers:
+#                                 ~ sign extension, can only appear once
+#                                 < left shift by 'num' amount on extraction, can only appear once
+#                                 ! mark 'num' as an invalid input for this variable, any number may appear
+EOM
+
+      FileUtils.mkdir_p File.dirname(out_path)
```
ThinkOpenly: I prefer we don't create new directories. If the user needs the output somewhere special, the user should create that directory beforehand. This also seemed to fail for me, unexpectedly:
```
/mise/installs/ruby/3.4.8/lib/ruby/3.4.0/fileutils.rb:403:in 'Dir.mkdir': File exists @ dir_s_mkdir - [...]/out (Errno::EEXIST)
```
[edited for brevity, removing the irrelevant full path]
...that's from the first time I ran it, so the `out` subdirectory did not exist. (Strange.)

backends/inst_table/lib/udb/generate_inst_table.rb
```diff
@@ -0,0 +1,157 @@
...
+      File.write(out_path, header + lines.join("\n") + "\n")
+      out_path
```
ThinkOpenly: (Again, not a Ruby expert...)
Is this line necessary?

backends/inst_table/lib/udb/generate_inst_table.rb
```diff
@@ -0,0 +1,157 @@
...
+    end
+
+    def run(argv = ARGV)
+      options = { out: 'out/inst_table.txt', cfg: '_', root: Dir.pwd }
```
ThinkOpenly: As above, I think the default should be stdout.

backends/inst_table/bin/generate_inst_table
ThinkOpenly, line 1: I couldn't get this to run in my environment. (Probably me, but other things run.)
We should probably tie this into `do` or (preferred) `bin/generate`. I found `./do gen:inst_table` sorta works, but doesn't pass command line arguments through properly.

Are you willing to address this in this PR?

backends/inst_table/lib/udb/generate_inst_table.rb
```diff
@@ -0,0 +1,157 @@
...
+      optp = OptionParser.new do |opts|
+        opts.banner = 'Usage: generate_inst_table [options]'
+        opts.on('-o', '--out PATH', 'Output file') { |v| options[:out] = v }
+        opts.on('-c', '--cfg NAME', 'Config name') { |v| options[:cfg] = v }
```
ThinkOpenly: Is there a way to put a default value for an option here? (Maybe not, which seems odd.)

backends/inst_table/tasks.rake
```diff
@@ -0,0 +1,21 @@
+# Rake tasks to generate the flat instruction table using the udb ruby helper
+namespace :gen do
+  desc "Generate instruction flat table using udb gem helper"
+  task :inst_table do
+    out = ENV['OUT'] || ($root.join('out', 'inst_table.backend.txt').to_s)
```
ThinkOpenly: can we just generate it in the current directory?

tools/ruby-gems/udb/lib/udb/obj/instruction.rb
```diff
@@ -794,7 +794,7 @@ def initialize(name, field_data)
         @left_shift = field_data["left_shift"].nil? ? 0 : field_data["left_shift"]
         @sext = field_data["sign_extend"].nil? ? false : field_data["sign_extend"]
         @alias = field_data["alias"].nil? ? nil : field_data["alias"]
-        @location = field_data["location"]
+        @location = field_data["location"].to_s
```
ThinkOpenly, line 794: Is this redundant with the `d.location.to_s` in `generate_inst_table.rb`, above?

ThinkOpenly (changes requested):
One thing we try to do for backends that generate consistent output is to add a simple test that compares newly generated output with a "golden" reference. This can be a bit of a pain to maintain, but I expect we'll have "autofix" fix things up automatically in the near future. This helps if someone, for example, changes the scripts here but thinks the changes are innocuous when they aren't.

backends/inst_table/lib/udb/generate_inst_table.rb
thecharlesjenkins: I thought there were some ranges that used '...' and some that used '..', but now that I look I only see '..' ranges. I'll get rid of this and split along '..'

backends/inst_table/lib/udb/generate_inst_table.rb
thecharlesjenkins: This is matching on '..' and only '..' so  I can simplify this

tools/ruby-gems/udb/lib/udb/obj/instruction.rb
thecharlesjenkins: Yes it does look redundant. I am also not a ruby expert, but it looks like a type error is thrown if I get rid of this one, but I can get rid of the one in generate_inst_table.rb.

backends/inst_table/lib/udb/generate_inst_table.rb
thecharlesjenkins: Sounds good, will change

backends/inst_table/bin/generate_inst_table
thecharlesjenkins: Oh cool, that script looks useful! I have been messing with this for a while and it seems like a lot of the infrastructure has changed so I'll use that new infrastructure. That will hopefully fix all of the file creation problems and such.

backends/inst_table/lib/udb/generate_inst_table.rb
thecharlesjenkins: No it's not, thank you

backends/inst_table/lib/udb/generate_inst_table.rb
thecharlesjenkins: Okay, I have made it only generate a file

backends/inst_table/lib/udb/generate_inst_table.rb
thecharlesjenkins: Sure

backends/inst_table/lib/udb/generate_inst_table.rb
thecharlesjenkins: Sounds good

backends/inst_table/lib/udb/generate_inst_table.rb
thecharlesjenkins: Thanks!

backends/inst_table/lib/udb/generate_inst_table.rb
thecharlesjenkins: Oops, thanks

backends/inst_table/lib/udb/generate_inst_table.rb
ThinkOpenly: I don't see any instances where `location` contains any `.`.

thecharlesjenkins: The yaml has '-' but that gets converted into '..'. It looks like it's possible to use the ruby range functions so I can just call range.first instead

ThinkOpenly:
> I am not sure ... how I would go about making a test that wouldn't break each time a new instruction is added.

It's supposed to break when new content is added. We're trying to catch unintentionally added content, and for intentionally added content, to _see_ it is processed as expected. With no representation of the output of the script, there's nothing to test against and script modifications go untested.

> I am not sure how to add tests in the new generator infrastructure

I'm thinking more like the testing done in `backends/instruction_appendix` or `tools/python` (`profile_extensions`). Nothing fancy.

You could take it further, of course:
- Add the kernel generation script (`insn_tbl.sh`) here as well and use it to generate `insn_gen.h` here just to make sure that script continues working forever.
- Maybe instead of adding the script, live pull the latest version of that script from the kernel tree instead of storing a copy here. (Or get "renovate" to keep it up to date here automatically, once we get "renovate" working satisfactorily... if that's within renovate's scope.)
- Compile something meaningful that uses the generated `insn_gen.h` to verify `inst_tbl.sh` is generating valid content. "Something meaningful" could be as sophisticated as desired, including more pieces directly from the kernel.

All of that is a little risky if paths or file names change in the kernel, of course.

backends/inst_table/lib/udb/generate_inst_table.rb
ThinkOpenly: I see. I was not aware of that conversion (printing out the values of `range`):
```
Generating instruction table (CFG=_) -> /.../out/inst_table.backend.txt
25..31
12..14
0...7
[...]
```

tools/ruby-gems/udb-gen/Gemfile
ThinkOpenly, line 1: remove this change, please

tools/ruby-gems/udb-gen/Rakefile
ThinkOpenly, line 1: remove this change, too

Rakefile
```diff
@@ -262,6 +262,53 @@ namespace :test do
 
     puts "All IDL passed type checking"
   end
+
+  task :inst_table_gen do
```
ThinkOpenly: This whole stanza should probably not be in the main `Rakefile`, but closer to the generator itself, and included here. We have a mechanism already set up well above that automatically pulls in `tools/*/tasks.rake`. So, I suggest creating a new `tools/ruby-gems/tasks.rake` and move this content there.

Rakefile
```diff
@@ -262,6 +262,53 @@ namespace :test do
...
+    require_relative "tools/ruby-gems/udb-gen/lib/udb-gen/generators/inst_table/table_builder"
+    golden_path = "tests/data/golden/test_table.txt"
+    cfg = "_"
```
ThinkOpenly: This should allow other configs, at the user's discretion. You can find a use of `CFG` environment variable for allowing that use-case above.

Rakefile
```diff
@@ -262,6 +262,53 @@ namespace :test do
...
+          system("diff -u #{golden_path} #{actual_path}")
+
+          raise "Golden file mismatch for #{cfg}"
+        end
+      else
+        puts "Golden file not found. Creating #{golden_path}..."
+        FileUtils.mkdir_p(File.dirname(golden_path))
+        File.write(golden_path, actual_output)
+        puts "Done."
+      end
```
ThinkOpenly: This is carrying a big load. Let's split this into three different tasks:
- gen:inst_table
  Just generates the table (optionally based on a specified configuration).
  Used in local testing and by the next two tasks.
- test:inst_table
  Verify that a newly generated (via gen:inst_table) table is identical to the golden reference.
  Mostly used by CI and maybe pre-commit.
- chore:update_inst_table
  Generate a new inst_table and makes it the golden reference.
  Used locally to complete a PR, maybe pre-commit as well, and ideally autofix.

bin/chore
```diff
@@ -444,13 +444,6 @@ do_ruby_type_def() {
   cd "${UDB_ROOT}" || exit 1
   "${UDB_ROOT}"/bin/bundle exec --gemfile "${UDB_ROOT}"/Gemfile tapioca gems --all
   "${UDB_ROOT}"/bin/bundle exec --gemfile "${UDB_ROOT}"/Gemfile tapioca dsl
-
-  # tapioca annotations requires the ext/rbi-central submodule to be initialized.
-  if [ ! -f "${UDB_ROOT}/ext/rbi-central/index.json" ]; then
-    echo "Initializing ext/rbi-central submodule..."
-    git -C "${UDB_ROOT}" submodule update --init ext/rbi-central
-  fi
-  "${UDB_ROOT}"/bin/bundle exec --gemfile "${UDB_ROOT}"/Gemfile tapioca annotations
```
ThinkOpenly, line 489: @dhower-qc and/or @jordancarlin, thoughts here? @thecharlesjenkins explained this change in the commit message https://github.com/riscv/riscv-unified-db/pull/1780/changes/a3f4b4f2263727bfa0eba0355cee6752dbf1bb85:
> Tapioca annotations in rbi-central are community maintained and are not
always correct. Remove them from the chore script to avoid them being
forcefully added by the CI. This was making it impossible to add "mocha"
as a dependency because the rbi-central types are conflicting with the
sorbet-typed versions.

ThinkOpenly (changes requested):

ThinkOpenly:
Whoa. I guess I forgot to actually submit comments from an earlier review session. sigh.

Rakefile
thecharlesjenkins: I don't quite see how this fits together. The tasks.rake was used run the "old" style of the backends, but not to test them. The "testing" for these old backends is simply the CI running the gen command on the backends and seeing if it crashes. I probably should have split the commit up, but in this last commit I also made the `tools/ruby-gems/udb-gen/test` directory that leverages the minitest framework that I saw is used in other places in the repo and I put my unit test cases for this new backend there.

However, this test case is not a unit test and is instead an integration test so I don't think it should live inside of the generator itself. It is testing the integration between the configs at the root of the directory, the UDB module, and udb-gen module. This block appears to currently the only place where integration tests are written, so if that is not desired I can construct a new place to put integration tests.

Rakefile
thecharlesjenkins: This is for a test case, why should CFG be configurable?

Rakefile
thecharlesjenkins: This block is purely for test cases (it's in the :test namespace). If a user wants to just generate a table they are expected to use `bin/generate`, so it's really only the second two that are coupled and I can add a chore to update the new golden reference.

Rakefile
thecharlesjenkins: On second thought, I'll add this integration test into udb-gen, it will keep things organized

bin/chore
ThinkOpenly: We discussed this at the UDB PR meeting today. The thought is that this may have crossed paths with some Ruby Gemfile changes. Could you restore this code, rebase, then see if any related issues are no longer present?

Rakefile
ThinkOpenly: Yep, I missed the context. Resolving...

Rakefile
ThinkOpenly: Yep, I missed the context here, too.

tools/ruby-gems/udb-gen/test/integration/fixtures/inst_table/expected.txt.license
```diff
@@ -0,0 +1,2 @@
+# SPDX-FileCopyrightText: 2025 Contributors to the RISCV UnifiedDB <https://github.com/riscv/riscv-unified-db>
```
ThinkOpenly: You are welcome and encouraged to put your own copyright here, if you wish.

Regardless, let's change this to "2026".

tools/ruby-gems/udb-gen/test/integration/test_helper.rb
```diff
@@ -0,0 +1,24 @@
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+# Copyright (c) Charlie Jenkins
```
ThinkOpenly: If I understand things correctly:
```suggestion
SPDX-FileCopyrightText: # Copyright (c) Charlie Jenkins
```

ThinkOpenly: Here and elsewhere, of course.

tools/ruby-gems/udb-gen/test/integration/test_inst_table.rb
```diff
@@ -0,0 +1,41 @@
...
+module UdbGen
+  class InstTableTest < Minitest::Test
+    include Udb
+
+    def setup
+      @gen_dir = Dir.mktmpdir
+      @resolver = Udb::Resolver.new(
+        Udb.repo_root,
+        gen_path_override: Pathname.new(@gen_dir)
+      )
```
ThinkOpenly, line 21: Do we delete this temporary directory somewhere?

tools/ruby-gems/tasks.rake
```diff
@@ -212,3 +212,30 @@ namespace :release do
...
+      Dir.chdir($root / "tools/ruby-gems/udb-gen") do
+        sh "ruby -Ilib:test test/integration/run.rb"
+      end
+    end
+  end
+end
+
+namespace :chore do
+  namespace :udb_gen do
+    task :update_fixtures do
```
ThinkOpenly, line 235: Add a description here, please.

bin/chore
thecharlesjenkins: This is purely related to mocha and cannot be impacted by any other Gemfile changes. "annotations" come from rbi-central which is community maintained and contains incorrect type definitions for mocha. The officially maintained and CI-validated type definitions come from `tapioca`. There is an effort to make rbi-central a good replacement for `tapioca`, but it is not there yet for mocha. I admittedly don't know much about Ruby so if you are able to find more accurate information please let me know. Testing this can be done by running `./bin/chore gen ruby-type-def` then `./do test:sorbet`, as is done in the CI.

thecharlesjenkins: This should fix it I think https://github.com/Shopify/rbi-central/pull/395 if you would like to wait for that to merge

tools/ruby-gems/udb-gen/test/integration/fixtures/inst_table/expected.txt.license
thecharlesjenkins: Changed, thanks!

tools/ruby-gems/tasks.rake
thecharlesjenkins: Added

tools/ruby-gems/udb-gen/test/integration/test_inst_table.rb
thecharlesjenkins: Oops, I missed the teardown, thank you!

tools/ruby-gems/udb-gen/test/integration/test_helper.rb
thecharlesjenkins: Changed!

bin/chore
thecharlesjenkins: This has now been merged and so this doesn't show up anymore. I removed the commit that got rid of annotations.

ThinkOpenly:
Would you consider a change such that the `-o` argument is optional and if not specified, output goes to stdout? Here's a suggested patch, which I might be able to push myself, but don't want to presume you're ok with the concept nor this implementation.

[insn_table-stdout.patch](https://github.com/user-attachments/files/28431953/insn_table-stdout.patch)

ThinkOpenly:
_sorry, hit the wrong button_

tools/ruby-gems/udb-gen/lib/udb-gen/generators/inst_table/table_builder.rb
```diff
@@ -0,0 +1,123 @@
...
+        header = <<EOM
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+# GENERATED WITH https://github.com/riscv-software-src/riscv-unified-db
+# "#{command}"
+#
+# Each line of the instruction table should have the following format:
+# NAME BASE FIXED_BITS [VARIABLE_LIST]
+# NAME                        instruction name
+# BASE                        instruction base size (common[,(32|64)])
```
ThinkOpenly, line 108: Last thing: could you expand on the definition of "BASE"? The meaning of a value like "common,32", as used above, is not obvious. I presume this means something like "this definition has BOTH RV32 and RV64 encodings, which are different, and THIS is the RV32 encoding" ?

thecharlesjenkins: Good point. I added a detailed description. Not related, but I also added the test cases to the CI

ThinkOpenly:
Original PR description:
> This new backend that can be generated with "bundle exec rake gen:inst_table" is designed to organize all of the requested instructions into a table modeled after the syscall table in Linux. The usecase for this is to commit the generated table into Linux which will use the table to generate the necessary instruction headers.
> 
> To be able to commit to Linux, I am marking the generated file as GPL-2.0-only and Reuse requires adding the License file to this repo for that.

ThinkOpenly (approved):
Thanks for sticking with this, @thecharlesjenkins !

PR #1791, opened 2026-04-15
LFX - Phase 4: LLM Extraction Pipeline

ishaan-arora-1:
## Summary

- Add `extract.py` — automated LLM extraction pipeline for identifying architectural parameters in the RISC-V specification
- Features token-aware rate limiting, exponential backoff for API throttling, source file skipping for non-parameter content (19 files), and pilot/run/merge/status CLI modes
- Includes v1 extraction results from Anthropic Claude across 59 spec chunks (208 unique parameters, ~$3.60 API cost)

### Key capabilities
- **Pilot mode**: Test extraction on `machine.adoc` chunks only (prompt validation)
- **Run mode**: Full extraction across all spec chunks with configurable delay and retry
- **Rate limiter**: Token-bucket algorithm tracking usage over 60s windows with 0.75 safety factor (stays within Anthropic's 30K input tokens/min tier)
- **Skip list**: 19 boilerplate `.adoc` files (bibliography, index, rationale, etc.) automatically excluded

### Results structure
- Per-chunk JSON results in `results/claude-sonnet-4/chunk_NNN.json`
- Merged `all_results_claude-sonnet-4.json` with all extracted parameters
- Each result includes: parameters found, confidence scores, classifications, token usage, and latency

## Test plan
- [x] Pilot run on `machine.adoc` chunks validates prompt quality
- [x] Full extraction completes all 59 chunks without errors
- [x] Rate limiter prevents 429 errors during sustained extraction
- [x] Merge produces valid combined results file
- [x] Status command reports accurate progress
- [x] Pre-commit hooks pass (ruff, SPDX headers)

---

## Update — review change + v1 re-run

Pushed the prompt change in `extract.py` — added a line right above the existing "include line numbers" instruction telling the model to treat lines before `content_start_line` as overlap context only and not re-extract a parameter whose defining sentence lives entirely in that overlap region.

While I was re-running the full v1 extraction I had to land three small infra fixes on `extract.py` in a separate commit (`fix(phase4): make extract.py runnable on Python 3.13 and tunable`):

- `from datetime import UTC, datetime` — the script was calling `datetime.now(datetime.UTC)` which only works as a chained attribute in a different import pattern. On Python 3.13 with the existing `from datetime import datetime`, it raises `AttributeError`. Switched to importing `UTC` explicitly.
- Made `RATE_LIMIT_TOKENS_PER_MIN` configurable via a `RATE_LIMIT_TPM` env var (default unchanged at 30,000). I bumped it to 200,000 for the re-run so the full extraction took ~20 minutes instead of ~2 hours.
- Made `max_tokens` configurable via a `MAX_OUTPUT_TOKENS` env var (default unchanged at 8,192). One large CSR-heavy chunk (the first machine.adoc chunk) was getting its JSON response truncated; 16,384 fixed it.

All three are env-driven with the original values as defaults, so nothing changes for the existing usage.

**V1 re-run results (claude-sonnet-4):**

| Metric                | Value |
|-----------------------|-------|
| Chunks processed      | 60    |
| Errors                | 0     |
| Raw parameters found  | 225   |
| Input tokens          | ~919K |
| Output tokens         | ~62K  |

Per-chunk JSON results and the merged `all_results_claude-sonnet-4.json` are refreshed.


param_extraction/scripts/extract.py
```diff
@@ -0,0 +1,942 @@
...
+
+def _format_chunk(chunk_text: str, meta: dict) -> str:
+    lines = [
+        "## Specification Text to Analyze",
+        "",
+        f"File: `{meta['source_file']}`",
+        f"Lines: {meta['start_line']}-{meta['end_line']}",
+        "",
+        "Analyze the following specification text and extract all architectural parameters.",
+        f"Include line numbers relative to the original file (starting from line {meta['start_line']}).",
```
miguelcsx, line 233:
```suggestion
        "Analyze the following specification text and extract all architectural parameters.",
        f"Treat lines before {meta['content_start_line']} as overlap context only; avoid re-extracting a parameter when its defining sentence appears entirely in that overlap region.",
        f"Include line numbers relative to the original file (starting from line {meta['start_line']}).",
```

PR #1792, opened 2026-04-15
LFX - Phase 5: Analysis, Alignment & Metrics

ishaan-arora-1:
## Summary

- Add `analyze.py` — comprehensive analysis pipeline for evaluating LLM extraction results against UDB ground truth
- Implements deduplication (handling cross-chunk duplicates with confidence-based selection), multi-strategy alignment (exact match, one-to-many mappings, concept groups, fuzzy name matching), and detailed metrics computation
- Generates discrepancy reports categorizing differences as naming mismatches, class disagreements, recall misses, new discoveries, and hallucination suspects

### V1 Evaluation Results
| Metric | Value |
|--------|-------|
| UDB parameters | 185 |
| LLM params (deduped) | 215 |
| Raw recall | **60.0%** |
| Adjusted recall | **62.7%** |
| Classification accuracy | 67.9% |
| New params discovered | 153 |

### Key analysis capabilities
- **Deduplication**: Cross-chunk parameter deduplication preferring in-content-region matches over overlap-region duplicates
- **Curated one-to-many groups**: Hand-reviewed allowlist (`data/one_to_many_groups.json`) of UDB parameter groups where many per-exception / per-register / per-extension variants are defined by a single spec sentence (e.g. `REPORT_VA_IN_MTVAL_ON_*` has 10 UDB members but is defined once in the spec). Every group carries a `justification` field pointing at the text it covers, so the matching is reviewable rather than algorithmic.
- **Stem / prefix matching**: Catches close-but-not-exact pairs that share a long common token stem (e.g. LLM `REPORT_ENCODING_IN_MTVAL_ON_X` vs UDB `REPORT_ENCODING_IN_VSTVAL_ON_X`), which strict Jaccard would miss.
- **Concept groups**: Single-member shared-concept groups for recall calculation.
- **Per-class recall**: Breakdown of recall by parameter classification (NORM_DIRECT, NORM_CSR_WARL, etc.)
- **Confusion matrix**: Classification accuracy analysis across all parameter classes

## Test plan
- [x] Deduplication correctly resolves cross-chunk duplicates
- [x] Alignment matches UDB parameters via exact, fuzzy, stem, concept-group, and curated one-to-many strategies
- [x] Metrics computation produces valid recall and accuracy figures
- [x] Discrepancy CSV categorizes all differences correctly
- [x] Pre-commit hooks pass (ruff, SPDX headers)

---

## Update — review changes + analyze re-run

Pushed two changes to `analyze.py`:

1. Restricted concept-group alignment to groups with exactly one UDB member (`len(group_members) == 1`). The old code marked **every** member of a multi-member group as aligned whenever any LLM finding matched the group's keywords, which inflated recall by claiming alignment for parameters the LLM never actually produced.
2. Broadened the `already_aligned_llm` filter from `match_type in ("exact",)` to `match_type != "none"`, so LLM findings already matched via fuzzy / one-to-many / concept-group don't get re-considered for fuzzy matching.

Re-ran the analysis on the refreshed v1 extraction. With the strict fix in place, raw recall measured 36.8% — the honest number after removing the inflated multi-member matches.

---

## Update 2 — curated one-to-many groups + stem matching

After the strict concept-group fix, a lot of legitimate alignments were being dropped: the spec describes things like "mtval reports the faulting virtual address on these exceptions" **once**, but UDB splits that across 10 separate `REPORT_VA_IN_MTVAL_ON_*` parameters. With the strict fix, finding the concept counted as zero hits; with the old loose code, any keyword overlap counted as 10 hits. Neither was right.

I added a curated allowlist (`param_extraction/data/one_to_many_groups.json`) of these multi-variant groups. Each entry has:
- the UDB prefix it covers,
- the keywords that signal the concept,
- a per-group minimum match score,
- and a written `justification` pointing at the spec text the group represents.

When the LLM's name + excerpt clearly match the concept, every group member is counted as aligned. There are currently 8 such groups, all reviewed by hand:

- `REPORT_VA_IN_MTVAL_ON_*` / `..._STVAL_ON_*` / `..._VSTVAL_ON_*` (per-exception VA reporting, 10 variants each)
- `REPORT_GPA_IN_*` (guest-physical-address reporting, 5 variants)
- `REPORT_ENCODING_IN_*` (illegal-instruction encoding in tval, 4 variants)
- `REPORT_CAUSE_IN_*` (software-check cause reporting, 6 variants)
- `TINST_VALUE_ON_*` (transformed-instruction reporting, 16 variants)
- `TRAP_ON_ECALL_FROM_*` (ECALL trap per privilege mode, 4 variants)

I also added a **stem/prefix matching** pass for the tail cases — names that share most of their tokens with a UDB entry but differ by one register-name swap (e.g. LLM `..._MTVAL_..._ILLEGAL_INSTRUCTION` vs UDB `..._VSTVAL_..._ILLEGAL_INSTRUCTION`).

### V1 metrics, before vs after the analysis lift

| Metric | Strict-only | With curated groups + stem |
|---|---|---|
| Raw recall | 36.8% | **60.0%** |
| Adjusted recall | 38.4% | **62.7%** |
| Classification accuracy | 67.9% | 67.9% |
| LLM params (deduped) | 215 | 215 |
| New params discovered | 157 | 153 |

The lift is purely analysis-side — none of the underlying LLM extractions changed. We just stopped counting a single spec sentence as "10 missed parameters."


param_extraction/scripts/analyze.py
```diff
@@ -0,0 +1,763 @@
...
+                        udb_class=udb_info.get("classification"),
+                        class_match=None,
+                    )
+                )
+                matched_udb.add(member)
+
+    # Pass 4: Fuzzy name matching for remaining unmatched
+    already_aligned_llm = {a.llm_name for a in alignments if a.match_type in ("exact",)}
+    unmatched_udb = udb_names - matched_udb
+    unmatched_llm = [p for p in deduped if p["parameter_name"] not in already_aligned_llm]
```
miguelcsx:
```suggestion
    already_aligned_llm = {a.llm_name for a in alignments if a.match_type != "none"}
    unmatched_udb = udb_names - matched_udb
    unmatched_llm = [p for p in deduped if p["parameter_name"] not in already_aligned_llm]
```

param_extraction/scripts/analyze.py
```diff
@@ -0,0 +1,763 @@
...
+                        llm_name=best_llm_name,
+                        udb_name=member,
+                        match_type="concept_group",
+                        match_score=round(best_score, 3),
+                        llm_class=llm_by_name.get(best_llm_name, {}).get("class", ""),
+                        udb_class=udb_info.get("classification"),
+                        class_match=None,
+                    )
+                )
+                matched_udb.add(member)
```
miguelcsx:
```suggestion
        if best_llm_name and best_score >= 0.4 and len(group_members) == 1:
            member = group_members[0]
            udb_info = udb_by_name[member]
            alignments.append(
                AlignmentEntry(
                    llm_name=best_llm_name,
                    udb_name=member,
                    match_type="concept_group",
                    match_score=round(best_score, 3),
                    llm_class=llm_by_name.get(best_llm_name, {}).get("class", ""),
                    udb_class=udb_info.get("classification"),
                    class_match=None,
                )
            )
            matched_udb.add(member)
```

PR #1793, opened 2026-04-15
LFX - Phase 6: Prompt Refinement & V2 Results

ishaan-arora-1:
## Summary

- Refine LLM prompts based on Phase 5 gap analysis, targeting 49 recoverable UDB recall misses
- Add v2 system prompt with classification disambiguation rules and 7 commonly missed parameter pattern categories (counter/HPM, VM modes, tval reporting, alignment, implementation values, conditional SC failure, stateen control)
- Add 4 new positive few-shot examples targeting previously missed parameter types
- Implement prompt versioning support (`PROMPT_VERSION` env var) in `run_prompt.py` and `extract.py` for side-by-side v1/v2 comparison

### V1 vs V2 Comparison (final, after analysis improvements)

| Metric                  | V1     | V2       | Delta     |
|-------------------------|--------|----------|-----------|
| Parameters found        | 225    | **361**  | +60%      |
| Deduped unique          | 215    | **346**  | +61%      |
| Raw recall              | 60.0%  | **69.7%** | +9.7 pp  |
| Adjusted recall         | 62.7%  | **72.9%** | +10.2 pp |
| Classification accuracy | 67.9%  | **88.4%** | +20.5 pp |
| New params discovered   | 153    | **256**  | +67%      |
| Per-class NORM_DIRECT recall | 47%  | **83%**  | +36 pp |
| Per-class NORM_CSR_RW recall | 41%  | **63%**  | +22 pp |
| Per-class NORM_CSR_WARL recall | 25%  | **50%**  | +25 pp |

### What changed in v2 prompts
- **System prompt additions**: Classification disambiguation section clarifying NORM_CSR_WARL vs NORM_CSR_RW vs NORM_DIRECT boundaries; "Commonly Missed Parameter Patterns" section with 7 specific categories and indicators
- **New examples**: `COUNTINHIBIT_EN` (counter inhibit), `GSTAGE_MODE_BARE` (VM mode support), `REPORT_ENCODING_IN_MTVAL_ON_ILLEGAL_INSTRUCTION` (tval reporting), `LRSC_FAIL_ON_NON_EXACT_LRSC` (LR/SC conditional failure)
- **Versioning**: Results stored in `results/v2/` directory, prompts in `prompts/v2/`

## Test plan
- [x] V2 extraction completes all chunks without errors
- [x] Classification accuracy improves over v1 (88.4% vs 67.9%)
- [x] Prompt versioning correctly isolates v1 and v2 results
- [x] Pre-commit hooks pass (ruff, SPDX headers, formatting)

---

## Update — review change + v2 re-run

Mirrored the Phase 4 prompt change in the v2 pipeline by adding the overlap-context instruction to `run_prompt.py`:

```
Use overlap text only as context; avoid re-extracting a parameter when
its defining sentence appears entirely in a previous chunk.
```

Then re-ran the v2 extraction end-to-end and the analysis on top of it.

---

## Update 2 — analysis lift (curated one-to-many + stem matching)

Carried forward the Phase 5 changes: a curated one-to-many group allowlist (`param_extraction/data/one_to_many_groups.json`) plus stem/prefix matching in `analyze.py`. The previous v2 recall numbers were undercounting any case where the spec described many UDB variants in a single sentence (e.g. mtval/stval/vstval VA reporting, transformed-instruction reporting, ECALL-per-mode). With curated groups, each such sentence now legitimately aligns to the full set of UDB variants it describes.

**V2 metrics, before vs after the analysis lift:**

| Metric | Strict-only | With curated groups + stem |
|---|---|---|
| Raw recall | 46.5% | **69.7%** |
| Adjusted recall | 48.6% | **72.9%** |
| Classification accuracy | 88.4% | 88.4% |
| Deduped LLM params | 346 | 346 |
| Per-class NORM_DIRECT | 42% | **83%** |

None of the LLM extractions changed — only the alignment recognises real semantic equivalence between a spec sentence and its UDB variants. Classification accuracy is unchanged because nothing in the underlying class assignments moved.

The v2 prompt's "Classification Disambiguation" section and "Commonly Missed Parameter Patterns" remain the headline win for accuracy (88.4% vs 67.9%). The lift here just makes the recall number an honest reflection of what the v2 extraction actually covers.


param_extraction/scripts/run_prompt.py
```diff
@@ -0,0 +1,459 @@
...
+    lines = [
+        "## Specification Text to Analyze",
+        "",
+        f"File: `{meta['file']}`",
+        f"Lines: {meta['start_line']}-{meta['end_line']} "
+        f"(chunk {meta['chunk_index']}/{meta['total_chunks']})",
+        "",
+        "Analyze the following specification text and extract all architectural parameters.",
+        "Include line numbers relative to the original file (starting from "
+        f"line {meta['start_line']}).",
```
miguelcsx, line 141:
```suggestion
        "Analyze the following specification text and extract all architectural parameters.",
        "Use overlap text only as context; avoid re-extracting a parameter when its defining sentence appears entirely in a previous chunk.",
        "Include line numbers relative to the original file (starting from "
        f"line {meta['start_line']}).",
```

PR #1794, opened 2026-04-16
 fix(ci): repair GitHub Pages deployment workflow

### LINKED ISSUE #1787: GitHub Pages Deployment failing
Recent merges to main have been failing when deploying to the GitHub Pages site. See https://github.com/riscv/riscv-unified-db/deployments/github-pages.

amkr6207:
Root cause:
- `.github/workflows/pages.yml` had an unmatched `fi` in the "Download and publish schema files" shell block.

Fix:
- Removed the stray `fi` so the step executes correctly.

Closes #1787

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -251,7 +251,6 @@ jobs:
               mkdir -p "_site/schemas/${schema_name}/${version}"
               gh release download "${tag}" --dir "_site/schemas/${schema_name}/${version}" --pattern "*.json" 2>/dev/null || true
             done
-          fi
           # Write a JSON index of all available schema versions and files
           ruby tools/scripts/gen_schema_index.rb _site/schemas _site/schemas/index.json
 


amkr6207:
@jordancarlin,  PTAL. Thanks!

jordancarlin (approved):

jordancarlin:
I thought we had a shell linter enabled. Surprised it didn't catch this. Maybe it isn't running on GitHub Actions jobs (we definitely should run it on them though if possible).

amkr6207:
> I thought we had a shell linter enabled. Surprised it didn't catch this. Maybe it isn't running on GitHub Actions jobs (we definitely should run it on them though if possible).

Makes sense. It looks like our current shellcheck setup doesn’t lint inline `run:` bash in workflow YAML, which is likely why this slipped through. I can open a small follow-up to add `actionlint` for workflow checks.

PR #1804, opened 2026-04-26
Programmable RISC-V IDE (PRIDE)

brayjmendoza:
The final project of the 2025-2026 Harvey Mudd College clinic team for Qualcomm. Involve an Xtext DSL (with a .udb file extension) to create a IDE that  can parse YAML files of specifications and includes features like syntax-highlighting, error messages, cross referencing, etc. that align with the json schemas. This also includes a python script to convert between .udb to .yaml files and a shell script to generate a language server (which is then used for a VSCode extension)


brayjmendoza:
PR moved to #1833

PR #1808, opened 2026-04-28
Add STANDARD_SM_SUPPORTED parameter

jordancarlin:
Discussed during today's UDB SIG meeting.

Since `Sm` is always a required extension for complete configs in UDB, this parameter is needed for ACT4 so that it knows whether the machine mode CSRs exist.


dhower-qc (commented):
To truly implement this, should every M-mode CSR be defined by CONFORIMING_SM_SUPPORTED == true?

jordancarlin:
Parameter name updated from `CONFORMING_SM_SUPPORTED` to `STANDARD_SM_SUPPORTED` per discussion in CSC. @dhower-qc were you going to look into adding support for this to the IDL? This is going to be needed imminently for ACT4, so it would be good to get this done sooner rather than later.

jordancarlin:
@dhower-qc would you be opposed to merging this as a parameter that does nothing to the UDB ISS for now (and documenting that somewhere) and then extend it later? It would be great to get this in for ACT4.

dhower-qc (approved):

ThinkOpenly:
What's needed here?

jordancarlin:
I'll revisit this one in a few weeks. There is some discussion in the CSC about changing this terminology again. We also never decided how to integrate this into the ISS (most likely it would just print an error and terminate).

________________________________

PR #1809, opened 2026-04-28
Update pre-commit to use prek priorities for parallel execution

jordancarlin:


dhower-qc (commented):
What is the purpose of this (how does it improve prek)?

jordancarlin:
It allows prek to run non-conflicting hooks in parallel, which speeds things up when running all of the hooks.

dhower-qc (approved):

PR #1814, opened 2026-04-30
feat: optimize C extensions

henry-hsieh:
Add 2 new attributes to instruction schema:
1. `add`: Used to adjust operand value by adding an integer to binary value. Useful for describing x8-x15 registers in C extensions.
2. `implicit`: Used to hint that the instruction is using an operand without explicitly allocating a field in the encoding. Useful for describing sp-/ra-relative instructions in C extensions.

Though `implicit` is not used by current riscv-unified-db, it may be useful for the library relied on it.

Other notable changes:

1. `inst_schema.json` version is bumped to v0.3
2. Because `sign_extend` default to false, I have added it to C instructions required signed-extend.
3. Fix several C instructions missing `alias` for `rd'/rs1'` operands.
4. Update pretty_print that if `not` array size is more than its reversion, use it instead.

### DIFF
--- a/spec/schemas/inst_schema.json
+++ b/spec/schemas/inst_schema.json
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "v0.2",
+  "$id": "v0.3",
   "$defs": {
     "fully_resolved_opcodes": {
       "type": "object",
@@ -267,6 +267,15 @@
               "default": 0,
               "description": "Amount the field should be left shifted before use (e.g., for opcode[5:3], left_shift is 3)"
             },
+            "add": {
+              "type": "integer",
+              "default": 0,
+              "description": "Amount the field should be added before use, e.g., source register is ranged between 8 to 15, add is 8"
+            },
+            "implicit": {
+              "type": "integer",
+              "description": "Implicitly use an operand with the value"
+            },
             "alias": {
               "type": "string",
               "description": "Alias of the field. Used when a field can represent multiple things, e.g., when a source register is also the destination register"
@@ -286,7 +295,14 @@
               "description": "Specific value(s) that are not permitted for this field."
             }
           },
-          "required": ["name", "location"],
+          "oneOf": [
+            {
+              "required": ["name", "location"]
+            },
+            {
+              "required": ["name", "implicit"]
+            }
+          ],
           "additionalProperties": false
         }
       ]
--- a/spec/std/isa/inst/C/c.add.yaml
+++ b/spec/std/isa/inst/C/c.add.yaml
@@ -23,6 +23,7 @@ encoding:
     - name: xd
       location: 11-7
       not: 0
+      alias: xs1
 access:
   s: always
   u: always
--- a/spec/std/isa/inst/C/c.addi.yaml
+++ b/spec/std/isa/inst/C/c.addi.yaml
@@ -22,9 +22,11 @@ encoding:
     - name: imm
       location: 12|6-2
       not: 0
+      sign_extend: true
     - name: xd
       location: 11-7
       not: 0
+      alias: xs1
 access:
   s: always
   u: always
--- a/spec/std/isa/inst/C/c.addi16sp.yaml
+++ b/spec/std/isa/inst/C/c.addi16sp.yaml
@@ -23,6 +23,44 @@ encoding:
       location: 12|4-3|5|2|6
       left_shift: 4
       not: 0
+      sign_extend: true
+    - name: xd
+      location: 11-7
+      not:
+        [
+          0,
+          1,
+          3,
+          4,
+          5,
+          6,
+          7,
+          8,
+          9,
+          10,
+          11,
+          12,
+          13,
+          14,
+          15,
+          16,
+          17,
+          18,
+          19,
+          20,
+          21,
+          22,
+          23,
+          24,
+          25,
+          26,
+          27,
+          28,
+          29,
+          30,
+          31,
+        ]
+      alias: xs1
 access:
   s: always
   u: always
--- a/spec/std/isa/inst/C/c.beqz.yaml
+++ b/spec/std/isa/inst/C/c.beqz.yaml
@@ -23,6 +23,9 @@ encoding:
       sign_extend: true
     - name: xs1
       location: 9-7
+      add: 8
+    - name: xs2
+      implicit: 0
 access:
   s: always
   u: always
--- a/spec/std/isa/inst/C/c.ldsp.yaml
+++ b/spec/std/isa/inst/C/c.ldsp.yaml
@@ -34,7 +34,9 @@ encoding:
         left_shift: 3
       - name: xd
         location: 11-7
-        not: [0, 1, 3, 5, 7]
+        not: [0, 1, 3, 5, 7, 9, 11, 13, 15, 17, 19, 21, 23, 25, 27, 29, 31]
+      - name: xs1
+        implicit: 2
   RV64:
     match: 011-----------10
     variables:
@@ -43,6 +45,8 @@ encoding:
         left_shift: 3
       - name: xd
         location: 11-7
+      - name: xs1
+        implicit: 2
 access:
   s: always
   u: always
--- a/spec/std/isa/inst/C/c.li.yaml
+++ b/spec/std/isa/inst/C/c.li.yaml
@@ -20,9 +20,12 @@ encoding:
   variables:
     - name: imm
       location: 12|6-2
+      sign_extend: true
     - name: xd
       location: 11-7
       not: 0
+    - name: xs1
+      implicit: 0
 access:
   s: always
   u: always
--- a/spec/std/isa/inst/C/c.mv.yaml
+++ b/spec/std/isa/inst/C/c.mv.yaml
@@ -9,7 +9,7 @@ name: c.mv
 long_name: Move Register
 description: |
   C.MV (move register) performs copy of the data in register xs2 to register xd
-  C.MV expands to addi xd, x0, xs2.
+  C.MV expands to add xd, x0, xs2.
 definedBy:
   extension:
     name: Zca
@@ -20,6 +20,8 @@ encoding:
     - name: xd
       location: 11-7
       not: 0
+    - name: xs1
+      implicit: 0
     - name: xs2
       location: 6-2
       not: 0
--- a/spec/std/isa/inst/C/c.sdsp.yaml
+++ b/spec/std/isa/inst/C/c.sdsp.yaml
@@ -27,10 +27,12 @@ encoding:
     variables:
       - name: xs2
         location: 6-2
-        not: [1, 3, 5, 7]
+        not: [1, 3, 5, 7, 9, 11, 13, 15, 17, 19, 21, 23, 25, 27, 29, 31]
       - name: imm
         location: 9-7|12-10
         left_shift: 3
+      - name: xs1
+        implicit: 2
   RV64:
     match: 111-----------10
     variables:
@@ -39,6 +41,8 @@ encoding:
       - name: imm
         location: 9-7|12-10
         left_shift: 3
+      - name: xs1
+        implicit: 2
 access:
   s: always
   u: always
--- a/tools/ruby-gems/udb/lib/udb/obj/instruction.rb
+++ b/tools/ruby-gems/udb/lib/udb/obj/instruction.rb
@@ -422,6 +422,8 @@ module Udb
       if effective_xlen.nil?
         if defined_in_base?(32)
           encoding(32).decode_variables.each do |d|
+            next if d.size.zero?
+
             qualifiers = [:const]
             qualifiers << :signed if d.sext?
             width = d.size
@@ -432,6 +434,8 @@ module Udb
         end
         if defined_in_base?(64)
           encoding(64).decode_variables.each do |d|
+            next if d.size.zero?
+
             qualifiers = [:const]
             qualifiers << :signed if d.sext?
             width = d.size
@@ -457,6 +461,8 @@ module Udb
           Idl::Var.new("__effective_xlen", Idl::Type.new(:bits, width: 7), effective_xlen)
         )
         encoding(effective_xlen).decode_variables.each do |d|
+          next if d.size.zero?
+
           qualifiers = [:const]
           qualifiers << :signed if d.sext?
           width = d.size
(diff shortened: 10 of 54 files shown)


dhower-qc:
This has some commonalities with #1532. @ThinkOpenly, take a look and see what you think about any overlap.

henry-hsieh:
Didn't notice about the ongoing refactor in #1527. Feel free to close this one.
After short dig in, I think the `add` field is covered by new `encode` / `decode` field.
But I think `implicit` is not covered.

spec/schemas/inst_schema.json
```diff
@@ -267,6 +267,15 @@
               "default": 0,
               "description": "Amount the field should be left shifted before use (e.g., for opcode[5:3], left_shift is 3)"
             },
+            "add": {
+              "type": "integer",
+              "default": 0,
+              "description": "Amount the field should be added before use, e.g., source register is ranged between 8 to 15, add is 8"
+            },
```
ThinkOpenly, line 274: IMHO, while this is useful in isolation, it presents some challenges by only solving one arithmetic issue, and leaves others (like additional operators and "precedence") unresolved. In #1527, I'm trying to lump all of the operand<->field transformations into new, general `encode()` and `decode()` methods. These are IDL, which allows most any transformation, but also complicates their use. Would these methods accommodate your use-case well enough?

spec/schemas/inst_schema.json
```diff
@@ -267,6 +267,15 @@
...
+            "implicit": {
+              "type": "integer",
+              "description": "Implicitly use an operand with the value"
+            },
```
ThinkOpenly, line 278: I think we do need something like this. You've applied it to opcode fields, but the variable fields are by definition, variable, and "implicit" by definition is not really variable, so it's not a great fit.

The _implicit_ content seems to be more closely associated with operands (which is the approach in #1527): the _input_ to the instruction.

It is a bit subjective.

Either way, I think it needs to be generalized to at least cover:
- stack pointer (as you do here)
- program counter (which can't be represented as an "integer" as is done here)
- the unmentioned register in register pair usage (this is awkward because it depends on one of the non-implicit operands, so we'll need a way to represent that). I guess this case is sort of implicit _and_ variable.

spec/std/isa/inst/C/c.add.yaml
```diff
@@ -23,6 +23,7 @@ encoding:
     - name: xd
       location: 11-7
       not: 0
+      alias: xs1
```
ThinkOpenly, line 26: I didn't even know UDB had this feature. Are you adding this just to match the ISA manual? Do you have a use-case for it? Just curious. It seems pretty harmless, so no objections.

spec/std/isa/inst/C/c.addi.yaml
```diff
@@ -22,9 +22,11 @@ encoding:
     - name: imm
       location: 12|6-2
       not: 0
+      sign_extend: true
```
ThinkOpenly, line 25: These `sign_extend` fixes should probably go in their own PR, since they are not subjective and we could get them merged much faster.

spec/std/isa/inst/C/c.addi16sp.yaml
```diff
@@ -23,6 +23,44 @@ encoding:
       location: 12|4-3|5|2|6
       left_shift: 4
       not: 0
+      sign_extend: true
+    - name: xd
+      location: 11-7
```
ThinkOpenly, line 28: I'm guessing this fails to validate properly, as this new "variable" is hardcoded to 2 ("00010") in the "match" string just above. Try running `./do gen:resolved_arch`.

An open question, perhaps: where is the syntax defined for `c.addi16sp`? Is is `c.addi16sp sp, imm` or `c.addi16sp imm`? (Why would "sp" need to be an actual operand when it's attached to the mnemonic?

spec/std/isa/inst/C/c.li.yaml
```diff
@@ -20,9 +20,12 @@ encoding:
   variables:
     - name: imm
       location: 12|6-2
+      sign_extend: true
     - name: xd
       location: 11-7
       not: 0
+    - name: xs1
+      implicit: 0
```
ThinkOpenly: Please remove this, as there is no `xs1` for `c.li`.

spec/std/isa/inst/C/c.beqz.yaml
```diff
@@ -23,6 +23,9 @@ encoding:
       sign_extend: true
     - name: xs1
       location: 9-7
+      add: 8
+    - name: xs2
+      implicit: 0
```
ThinkOpenly, line 28: `c.beqz` does not have an `xs2`. Are you trying to line up for a theoretical expansion to the base instruction, `beq`? That seems unnecessary here to me.

There are a number of changes like this.

spec/std/isa/inst/C/c.ldsp.yaml
```diff
@@ -34,7 +34,9 @@ encoding:
         left_shift: 3
       - name: xd
         location: 11-7
-        not: [0, 1, 3, 5, 7]
+        not: [0, 1, 3, 5, 7, 9, 11, 13, 15, 17, 19, 21, 23, 25, 27, 29, 31]
```
ThinkOpenly, line 37: Good catch! This is just a bug fix and should probably go in a separate PR.

spec/std/isa/inst/C/c.mv.yaml
```diff
@@ -9,7 +9,7 @@ name: c.mv
 long_name: Move Register
 description: |
   C.MV (move register) performs copy of the data in register xs2 to register xd
-  C.MV expands to addi xd, x0, xs2.
+  C.MV expands to add xd, x0, xs2.
```
ThinkOpenly, line 12: Another bug fix for a separate PR.

spec/std/isa/inst/C/c.sdsp.yaml
```diff
@@ -27,10 +27,12 @@ encoding:
     variables:
       - name: xs2
         location: 6-2
-        not: [1, 3, 5, 7]
+        not: [1, 3, 5, 7, 9, 11, 13, 15, 17, 19, 21, 23, 25, 27, 29, 31]
```
ThinkOpenly, line 30: bug fix

tools/ruby-gems/udb/lib/udb/obj/instruction.rb
```diff
@@ -422,6 +422,8 @@ def fill_symtab(effective_xlen, ast)
       if effective_xlen.nil?
         if defined_in_base?(32)
           encoding(32).decode_variables.each do |d|
+            next if d.size.zero?
```
ThinkOpenly, line 425: I'm not a Ruby expert... is this ignoring variables which are "implicit"?

tools/ruby-gems/udb/lib/udb/obj/instruction.rb
```diff
@@ -661,9 +667,14 @@ class DecodeVariable
...
       attr_reader :encoding_fields
 
-      sig { returns(String) }
+      sig { returns(T.nilable(String)) }
       attr_reader :location
 
+      # implicitly use an operand with the value
+      attr_reader :implicit
+
+      def implicit? = @implicit != 0
```
ThinkOpenly, line 676: You have an use of `implicit: 0`, in `c.beqz` above. What does this do?

tools/ruby-gems/udb/lib/udb/obj/instruction.rb
```diff
@@ -763,6 +786,8 @@ def inst_range_to_var_range(r)
 
       # array of constituent encoding fields
       def grouped_encoding_fields
+        return [] if @encoding_fields.empty?
+
```
ThinkOpenly, line 790: is this needed because of the changes in this PR?

ThinkOpenly (changes requested):
Thank you for submitting this PR!

I suggest creating a separate PR with the bug fixes. That's an easy objective review which I expect would be merged quickly.

Aliases are probably fine and could go in a separate PR to get them reviewed/merged relatively quickly.

I have reservations for "add" described in a review comment.

I think "implicit" is probably needed, but associating it with operands instead of opcode fields is perhaps a better approach, and maybe it should be carried in #1527.

spec/schemas/inst_schema.json
henry-hsieh: True. `encode()` and `decode()` should be enough for this. With these 2 functions, I think `left_shift()` could be merged into them, too.

spec/schemas/inst_schema.json
henry-hsieh: I'm fine with moving it to the fields other than variable. To integrate with new operand field, I have some ideas:
1. Implicit operand (with new implicit attribute)
```yaml
$schema: "operand_schema.json#"
kind: operand
name: sp-implicit
data:
  $inherits: operand/xs.yaml#/data
  possible_values: [2]
  implicit: true
```
2. Name with multiple of actual registers
- Introduce name to values mapping
- The name could be an array for ABI name swapping
- The values can be an array with all included values
```yaml
$schema: "operand_schema.json#"
kind: operand
name: xs-pair
data:
  type: reg_pair
  name: reg_pair_xs
  possible_values:
    - $inherits: operand/xs.yaml#/data
      name: ["x0", "zero"]
      values: [0, 1]
    - $inherits: operand/xs.yaml#/data
      name: ["x2", "sp"]
      values: [2, 3]
    - $inherits: operand/xs.yaml#/data
      name: ["x4", "tp"]
      values: [4, 5]
...
```

spec/std/isa/inst/C/c.add.yaml
henry-hsieh: Yes, just to match ISA manual. But I think it also beneficial for someone who wants to use this database for compiler scheduling or RTL decoder generation. For these purposes, you must know these operands are used for both source and destination. I think with new operand system. We could just create a new xsdc operand for this purpose.
```yaml
$schema: "operand_schema.json#"
kind: operand
name: xsdc
data:
  $inherits: operand/xdc.yaml#/data
  source: true
```

spec/std/isa/inst/C/c.addi.yaml
henry-hsieh: Sure. I'll file another PR for this in ~~a few days~~.

Edit: A few minutes actually :sweat_smile:

spec/std/isa/inst/C/c.addi16sp.yaml
henry-hsieh: I forget to change mask "00010" to "-----". I think two approaches are valid:
1. Keep mask as is, but add implicit operand to this instruction 
2. Change mask to "-----" and add a new operand which only has 1 possible value, i.e., 2.

I don't think `c.addi16sp` is defined in any official document. I think we use this syntax because official GCC is using it: riscv-collab/riscv-gnu-toolchain#372

spec/std/isa/inst/C/c.beqz.yaml
henry-hsieh: Yes, I just want to align the theoretical expansion to the base instruction. But it's totally fine not to do this.

spec/std/isa/inst/C/c.ldsp.yaml
henry-hsieh: OK. I'll file another PR for this.

tools/ruby-gems/udb/lib/udb/obj/instruction.rb
henry-hsieh: Yes, otherwise the execution of Ruby script will fail. If the `implicit` will be added to operands instead of fields, I think this change should be discarded.

tools/ruby-gems/udb/lib/udb/obj/instruction.rb
henry-hsieh: I want to use this field to skip references to variable with implicit attribute. However, I eventually used checks like `if @encoding_fields.empty?` and `if !field_data["location"].nil?`. These lines should be removed.

tools/ruby-gems/udb/lib/udb/obj/instruction.rb
henry-hsieh: Yes. `@encoding_fields` may be empty for implicit variables. If we have consensuses that `implicit` should be add in `operand` instead of `variable`, all the changes in this Ruby file can be discarded.

spec/schemas/inst_schema.json
ThinkOpenly: > I think `left_shift()` could be merged into them, too.

True.

spec/schemas/inst_schema.json
ThinkOpenly: Good ideas. Feel free to post them in #1527. (Or, I'll copy them there if you wish.)

spec/std/isa/inst/C/c.add.yaml
ThinkOpenly: Agree. I didn't yet model a source/destination register, but I think the capability is already there (as in your example).

spec/std/isa/inst/C/c.addi16sp.yaml
ThinkOpenly: Oh. Thanks for digging up that (very sad) reference. 😕 Not a choice I would've made.

ThinkOpenly: This is a strange case, indeed. Making the field a "variable with only one possible value" seems a little odd. Making "sp" an implicit operand when it is explicitly there also seems odd. This instruction is odd.

I could be swayed either way, but prefer the "variable with one possible value", if you want to update the mask.

henry-hsieh: I prefer the "variable with one possible value", too. But I want to use new operand system to do it instead of handcoding 0-31 exclude 2.

spec/schemas/inst_schema.json
henry-hsieh: Thanks! I'll copy them to the PR.

spec/std/isa/inst/C/c.beqz.yaml
ThinkOpenly: Let's not. I think it would be confusing. There is already enough information to get the right values for all of the operands in the base instruction.

spec/std/isa/inst/C/c.addi16sp.yaml
ThinkOpenly: I think we should do both the field and the operand. We do want to properly constrain the field so that simple decode won't match (if the purpose of the decode was to _execute_ rather than disassemble, for example).

henry-hsieh:
If everything is addressed, I would like to close this PR and wait for the new operand system. What do you think?

ThinkOpenly:
> If everything is addressed, I would like to close this PR and wait for the new operand system. What do you think?

How did you want to handle the operand alias names that you added here?

henry-hsieh:
> > If everything is addressed, I would like to close this PR and wait for the new operand system. What do you think?
> 
> How did you want to handle the operand alias names that you added here?

I think add an operand like this is enough.

https://github.com/riscv/riscv-unified-db/pull/1814#discussion_r3202313307

PR #1815, opened 2026-04-30
add Svukte ISA extension

ingallsj:
copy from https://github.com/riscv/riscv-isa-manual/pull/1564

### DIFF
--- a/spec/std/isa/csr/hstatus.yaml
+++ b/spec/std/isa/csr/hstatus.yaml
@@ -1,5 +1,5 @@
-# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
-# SPDX-License-Identifier: BSD-3-Clause-Clear
+# Copyright (c) RISC-V International
+# SPDX-License-Identifier: CC-BY-4.0
 
 # yaml-language-server: $schema=../../../schemas/csr_schema.json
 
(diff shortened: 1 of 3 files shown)


spec/std/isa/csr/hstatus.yaml
```diff
@@ -1,5 +1,5 @@
-# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
-# SPDX-License-Identifier: BSD-3-Clause-Clear
+# Copyright (c) RISC-V International
+# SPDX-License-Identifier: CC-BY-4.0
```
dhower-qc: We shouldn't change the copyright of the entire file. If you want to change the copyright/license of this contribution, you can use inline reuse-compatible comments:

https://reuse.software/faq/#partial-license

spec/std/isa/csr/senvcfg.yaml
```diff
@@ -1,5 +1,5 @@
-# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
-# SPDX-License-Identifier: BSD-3-Clause-Clear
+# Copyright (c) RISC-V International
+# SPDX-License-Identifier: CC-BY-4.0
```
dhower-qc: same here

dhower-qc (changes requested):

dhower-qc (approved):

spec/std/isa/csr/senvcfg.yaml
```diff
@@ -124,6 +124,24 @@ fields:
...
+      privilege mode U or VU are Svukte-qualified, as described in <<sec:svukte>>.
+      If Svukte is not implemented, UKTE is read-only zero.
+
+    definedBy:
+      extension:
+        name: Svukte
+    type(): |
+      return (implemented?(ExtensionName::Svukte)) ? CsrFieldType::RW : CsrFieldType::RO;
+    reset_value(): |
+      return (implemented?(ExtensionName::Svukte)) ? UNDEFINED_LEGAL : 0;
```
ThinkOpenly, line 144: Late to the party here, but I think there is no need to test for Svukte here, since the bit is defined by that extension.

```suggestion
    type: RW
    reset_value: UNDEFINED_LEGAL
```

ingallsj: You're right: bits for other extensions do not test for whether the extension is implemented. https://github.com/riscv/riscv-unified-db/pull/1822

PR #1826, opened 2026-05-05
fix(clean): skip riscv-tests cleanup when unavailable (#1718)

### LINKED ISSUE #1718: "do clean" complains about missing "gcc"
```
$ ./do clean
make: Entering directory 'tests/isa'
make -C ./../../ext/riscv-tests/isa XLEN=64 clean
make[1]: Entering directory 'ext/riscv-tests/isa'
Makefile:58: *** couldn't find gcc.  Stop.
make[1]: Leaving directory 'ext/riscv-tests/isa'
make: *** [Makefile:85: clean] Error 2
make: Leaving directory 'tests/isa'
make: Entering directory 'tests/isa'
make -C ./../../ext/riscv-tests/isa XLEN=32 clean
make[1]: Entering directory 'ext/riscv-tests/isa'
Makefile:58: *** couldn't find gcc.  Stop.
make[1]: Leaving directory 'ext/riscv-tests/isa'
make: *** [Makefile:85: clean] Error 2
make: Leaving directory 'tests/isa'
```

amkr6207:
Closes #1718

`./do clean` should not fail when `riscv-tests` dependencies are unavailable.

This change updates `tests/isa/Makefile` clean behavior to:
- skip `ext/riscv-tests/isa` cleanup if the submodule is not initialized
- skip `ext/riscv-tests/isa` cleanup if `RISCV_GCC` is not found
- still clean local `tests/isa` artifacts

### DIFF
--- a/tests/isa/Makefile
+++ b/tests/isa/Makefile
@@ -85,5 +85,13 @@ riscv_tests:
 # Clean up
 
 clean:
-	$(MAKE) -C $(riscv_tests_dir)/isa XLEN=$(XLEN) clean
+	@if [ -d "$(riscv_tests_dir)/isa" ] && [ -f "$(riscv_tests_dir)/isa/Makefile" ]; then \
+		if command -v "$(RISCV_GCC)" > /dev/null 2>&1; then \
+			$(MAKE) -C $(riscv_tests_dir)/isa XLEN=$(XLEN) clean; \
+		else \
+			echo "Skipping $(riscv_tests_dir)/isa clean: $(RISCV_GCC) not found"; \
+		fi; \
+	else \
+		echo "Skipping $(riscv_tests_dir)/isa clean: submodule not initialized"; \
+	fi
 	rm -rf $(junk)
--- a/tests/isa/Makefile
+++ b/tests/isa/Makefile
@@ -85,13 +85,7 @@ riscv_tests:
 # Clean up
 
 clean:
-	@if [ -d "$(riscv_tests_dir)/isa" ] && [ -f "$(riscv_tests_dir)/isa/Makefile" ]; then \
-		if command -v "$(RISCV_GCC)" > /dev/null 2>&1; then \
-			$(MAKE) -C $(riscv_tests_dir)/isa XLEN=$(XLEN) clean; \
-		else \
-			:; \
-		fi; \
-	else \
-		:; \
+	@if [ -f "$(riscv_tests_dir)/isa/Makefile" ] && command -v "$(RISCV_GCC)" > /dev/null 2>&1; then \
+		$(MAKE) -C $(riscv_tests_dir)/isa XLEN=$(XLEN) clean; \
 	fi
 	rm -rf $(junk)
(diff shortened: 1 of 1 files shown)


tests/isa/Makefile
```diff
@@ -85,5 +85,13 @@ riscv_tests:
 # Clean up
 
 clean:
-	$(MAKE) -C $(riscv_tests_dir)/isa XLEN=$(XLEN) clean
+	@if [ -d "$(riscv_tests_dir)/isa" ] && [ -f "$(riscv_tests_dir)/isa/Makefile" ]; then \
+		if command -v "$(RISCV_GCC)" > /dev/null 2>&1; then \
```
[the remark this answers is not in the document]
ThinkOpenly: It might be better to use something like is done in the `COMPILER_SUPPORTS` macro elsewhere in the file. However, it would be nice to get @henrikg-qc's opinion, as he shows up as the "blamee" for the `make` command emitting the errors. :-)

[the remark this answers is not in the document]
henrikg-qc: This is the correct way to do this. Looks over complicated at first but since the riscv-tests/isa/Makefile got updated with more checking it requires '$(RISCV_PREFIX)gcc' (default to 'riscv$(XLEN)-unknown-elf-gcc', or it will error out.

The only change I'd like to see is to quiet the output for the failing conditions. That way clean also has a clean output without warnings. Default target should fail if the subrepo is not there or $(RISCV_GCC) is not installed.

tests/isa/Makefile
```diff
@@ -85,5 +85,13 @@ riscv_tests:
...
+			$(MAKE) -C $(riscv_tests_dir)/isa XLEN=$(XLEN) clean; \
+		else \
+			echo "Skipping $(riscv_tests_dir)/isa clean: $(RISCV_GCC) not found"; \
```
henrikg-qc: I would skip the output in both else cases for cleaner output (no pun intended)

amkr6207: @henrikg-qc 

Updated. I removed output in both `else` cases so `clean` stays quiet.

henrikg-qc (approved):

tests/isa/Makefile
ThinkOpenly: Let's combine these into a single conditional...

tests/isa/Makefile
```diff
@@ -85,5 +85,13 @@ riscv_tests:
...
 clean:
-	$(MAKE) -C $(riscv_tests_dir)/isa XLEN=$(XLEN) clean
+	@if [ -d "$(riscv_tests_dir)/isa" ] && [ -f "$(riscv_tests_dir)/isa/Makefile" ]; then \
+		if command -v "$(RISCV_GCC)" > /dev/null 2>&1; then \
+			$(MAKE) -C $(riscv_tests_dir)/isa XLEN=$(XLEN) clean; \
+		else \
+			:; \
+		fi; \
+	else \
+		:; \
```
ThinkOpenly: ...and remove these empty `else` statements.

tests/isa/Makefile
amkr6207: Done, I combined those checks into a single conditional.

tests/isa/Makefile
amkr6207: Done, I removed the empty else blocks.

ThinkOpenly (approved):

PR #1827, opened 2026-05-05
Add explicit PC metadata for instructions (#1825)

### LINKED ISSUE #1825: Add a means to indicate instruction references/changes PC
> We should probably make this an instruction attribute. It's probably derivable by scanning the "operation()" IDL, but that's unnecessarily fraught. It could be an attribute of the "offset" field of the opcode... maybe not, since instructions like JALR access the PC, but not as in the mode of PC+offset. It could be a "implicit" operand... we should open an issue for discussion. This is somewhat related to #1527. 

 _Originally posted by @ThinkOpenly in [#1802](https://github.com/riscv/riscv-unified-db/pull/1802/changes#r3182466800)_

amkr6207:
Fixes #1825.

Add explicit instruction-level metadata for whether an instruction references and/or changes the PC, rather than inferring that from `operation()` IDL.

This change:
- adds optional `pc.references` / `pc.changes` fields to the instruction schema
- makes the metadata available through the Ruby instruction model
- adds a small test for the new metadata
- documents the new metadata
- annotates representative instructions:
  - `auipc`
  - `jalr`
  - `mret`

### DIFF
--- a/doc/data-templates.adoc
+++ b/doc/data-templates.adoc
@@ -97,6 +97,9 @@ access:
   vs: always   # VS-mode access
   vu: always   # VU-mode access
 data_independent_timing: true  # For cryptographic extensions
+pc:
+  references: true # Optional: instruction semantics reference the current PC
+  changes: false   # Optional: instruction semantics update the PC
 operation(): |
   # IDL code implementing the instruction
   X[xd] = X[xs1] + $signed(imm);
@@ -109,6 +112,11 @@ Common variable properties:
 * `location` - Bit range (e.g., `31-20`, `19-15`, or `31|7|30-25|11-8` for non-contiguous)
 * `sign_extend` - Set to `true` for signed immediates
 * `left_shift` - Shift value left by N bits (common for branch offsets)
+
+Optional instruction metadata:
+
+* `pc.references` - Set to `true` when the instruction semantics explicitly reference the current PC
+* `pc.changes` - Set to `true` when the instruction semantics explicitly update the PC
 
 === Pseudoinstructions
 
--- a/doc/data-templates.adoc
+++ b/doc/data-templates.adoc
@@ -97,7 +97,7 @@ access:
   vs: always   # VS-mode access
   vu: always   # VU-mode access
 data_independent_timing: true  # For cryptographic extensions
-pc:
+pc: # program counter
   references: true # Optional: instruction semantics reference the current PC
   changes: false   # Optional: instruction semantics update the PC
 operation(): |
--- a/spec/schemas/inst_schema.json
+++ b/spec/schemas/inst_schema.json
@@ -395,6 +395,23 @@
       "description": "Whether or not the instruction must execute with data-independent timing when the Zkt extension is supported",
       "default": false
     },
+    "pc": {
+      "type": "object",
+      "description": "Explicit metadata describing whether the instruction reads and/or writes the program counter",
+      "properties": {
+        "references": {
+          "type": "boolean",
+          "default": false,
+          "description": "Whether the instruction semantics reference the current program counter"
+        },
+        "changes": {
+          "type": "boolean",
+          "default": false,
+          "description": "Whether the instruction semantics update the program counter"
+        }
+      },
+      "additionalProperties": false
+    },
     "pseudoinstructions": {
       "description": "Variations of this instruction that form a pseudoinstruction",
       "type": "array",
--- a/spec/schemas/inst_schema.json
+++ b/spec/schemas/inst_schema.json
@@ -397,7 +397,7 @@
     },
     "pc": {
       "type": "object",
-      "description": "Explicit metadata describing whether the instruction reads and/or writes the program counter",
+      "description": "Whether the instruction reads and/or writes the program counter",
       "properties": {
         "references": {
           "type": "boolean",
--- a/spec/std/isa/inst/I/auipc.yaml
+++ b/spec/std/isa/inst/I/auipc.yaml
@@ -28,4 +28,6 @@ access:
   vs: always
   vu: always
 data_independent_timing: true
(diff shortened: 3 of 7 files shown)


doc/data-templates.adoc
```diff
@@ -97,6 +97,9 @@ access:
   vs: always   # VS-mode access
   vu: always   # VU-mode access
 data_independent_timing: true  # For cryptographic extensions
+pc:
```
ThinkOpenly:
```suggestion
pc: # program counter
```

doc/data-templates.adoc
```diff
@@ -97,6 +97,9 @@ access:
...
+  references: true # Optional: instruction semantics reference the current PC
+  changes: false   # Optional: instruction semantics update the PC
```
ThinkOpenly, line 102: Subjective, but these might be better as `read`/`write`:
```suggestion
  reads: true   # Optional: instruction semantics reference the PC
  writes: false # Optional: instruction semantics update the PC
```
If adopted here, these changes would be needed throughout this PR, of course.

spec/schemas/inst_schema.json
```diff
@@ -395,6 +395,23 @@
       "description": "Whether or not the instruction must execute with data-independent timing when the Zkt extension is supported",
       "default": false
     },
+    "pc": {
+      "type": "object",
+      "description": "Explicit metadata describing whether the instruction reads and/or writes the program counter",
```
ThinkOpenly: For consistency with other definitions here:
```suggestion
      "description": "Whether the instruction reads and/or writes the program counter",
```

ThinkOpenly (commented):
Neat! I think this is a good addition. @dhower-qc ? @lenary ?

If adopted, we need to annotate many other instructions, of course:
```
$ grep -l -E -i 'jump\(|jump_halfword\(|\$pc' $(find gen/resolved_spec/_/inst -type f)
gen/resolved_spec/_/inst/Smrnmi/mnret.yaml
gen/resolved_spec/_/inst/Zcmt/cm.jalt.yaml
gen/resolved_spec/_/inst/Zcmt/cm.jt.yaml
gen/resolved_spec/_/inst/Zcmp/cm.popret.yaml
gen/resolved_spec/_/inst/Zcmp/cm.popretz.yaml
gen/resolved_spec/_/inst/C/c.ebreak.yaml
gen/resolved_spec/_/inst/C/c.j.yaml
gen/resolved_spec/_/inst/C/c.bnez.yaml
gen/resolved_spec/_/inst/C/c.jr.yaml
gen/resolved_spec/_/inst/C/c.jal.yaml
gen/resolved_spec/_/inst/C/c.beqz.yaml
gen/resolved_spec/_/inst/C/c.jalr.yaml
gen/resolved_spec/_/inst/S/sret.yaml
gen/resolved_spec/_/inst/I/bne.yaml
gen/resolved_spec/_/inst/I/bge.yaml
gen/resolved_spec/_/inst/I/bgeu.yaml
gen/resolved_spec/_/inst/I/jalr.yaml
gen/resolved_spec/_/inst/I/blt.yaml
gen/resolved_spec/_/inst/I/beq.yaml
gen/resolved_spec/_/inst/I/ebreak.yaml
gen/resolved_spec/_/inst/I/jal.yaml
gen/resolved_spec/_/inst/I/auipc.yaml
gen/resolved_spec/_/inst/I/bltu.yaml
gen/resolved_spec/_/inst/I/mret.yaml
```

ThinkOpenly (changes requested):

ThinkOpenly:
I'm also not sure how a really weird instruction like `qc.cm.ilut` should be annotated, since as part of its semantics, it _executes_ other instructions which may or may not reference/update the PC.

ThinkOpenly:
> I'm also not sure how a really weird instruction like `qc.cm.ilut` should be annotated, since as part of its semantics, it _executes_ other instructions which may or may not reference/update the PC.

Maybe nevermind on that concern. That instruction does not permit executing PC-relative instructions .

ThinkOpenly:
Another way of implementing this would be adding an implicit operand to the instructions. The operand definitions are emerging in #1527. One nice thing that's already present there is the concept of "source" (read) and "destination" (write). Not yet present there, but probably needs to be is "implicit" operands. Program counter is certainly one. Stack pointer is another.

dhower-qc (changes requested):
Not opposed to this, but would like to see a regression test that ensures IDL and the YAML annotation agree.

spec/schemas/inst_schema.json
amkr6207: Updated. Thanks!

doc/data-templates.adoc
amkr6207: Updated. Thanks!

amkr6207:
@ThinkOpenly @dhower-qc 

Addressed the requested changes.

I added the regression test, the doc comment, and the schema wording update. I left `references` / `changes` unchanged since the `reads` / `writes` suggestion seemed optional.

Thanks!

amkr6207:
@ThinkOpenly, @dhower-qc 

Please review the PR. Thanks!

amkr6207:
@ThinkOpenly, @dhower-qc 

PTAL. Thanks!

ThinkOpenly:
@amkr6207, could you take a look at the latest in #1527? In there,
- `operand_schema.json` defines the schema for including the pc as an operand, including source, destination, and implicit attributes.
- `pc.yaml` describes an implicit source program counter operand.
- `bne.yaml` shows a use of an implicit pc operand which is both source and destination.

(An open question, perhaps, is whether an implicit opcode field is also needed. That would be slightly redundant with the operand definition, but it might be helpful for use-cases like a decoder which may not care about assembly operands.)

PR #1833, opened 2026-05-14
Programmable RISC-V IDE (PRIDE)

brayjmendoza:
The final project of the 2025-2026 Harvey Mudd College clinic team for Qualcomm. Involve an Xtext DSL (with a .udb file extension) to create a IDE that can parse YAML files of specifications and includes features like syntax-highlighting, error messages, cross referencing, etc. that align with the json schemas. This also includes a python script to convert between .udb to .yaml files and a shell script to generate a language server (which is then used for a VSCode extension)


brayjmendoza:
IDL support in PRIDE relies on idlc/ having both the Gemfile and Gemfile.lock, I made an issue about it at https://github.com/niluo-shiqi/riscv-unified-db-hmc-clinic-team-/issues/16

tools/scripts/language-server-script/regen-udb-ls.bat
```diff
@@ -0,0 +1,136 @@
+:: SPDX-FileCopyrightText: 2026 Harvey Mudd Clinic Team
+:: SPDX-License-Identifier: BSD-3-Clause-Clear
+
+@echo off
+setlocal EnableDelayedExpansion
+
+:: ─────────────────────────────────────────────
+:: regen-udb-ls.bat  (Windows)
+:: Rebuilds udb-ls-all.jar and copies it into the
+:: udb-vscode extension's server folder.
```
ThinkOpenly, line 10: Could you elaborate on the purpose of these scripts? (Think of your audience as someone who stumbles upon the script with no other knowledge.)

tools/python/convertudb.py
```diff
@@ -0,0 +1,534 @@
...
+        file.writelines(output_lines)
+
+
+if __name__ == "__main__":
+    # Get CLI arguments
+    if len(sys.argv) == 2:
+        INPUT_FILE = sys.argv[1]  # .udb file or .yml file to convert
+    else:
+        print("USAGE: python convertudb.py [INPUT]")
+        print("where INPUT is either a .udb file or a .yaml file to convert to the other format.")
```
ThinkOpenly: Maybe add a bit more explanation here as to the purpose of the script.

tools/scripts/language-server-script/regen-udb-ls.bat
ThinkOpenly: Also, maybe we should add some help text if invoke with a "help" option like "-h" or "--help" (or whatever equivalent on Windows).

tools/eclipse/dev/org.xtext.udb.parent/org.xtext.udb.ide/.project
```diff
@@ -30,11 +25,27 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.m2e.core.maven2Builder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
```
ThinkOpenly, line 32: Did this stanza need to move? In general, it is preferred to not include unnecessary changes.

tools/eclipse/dev/org.xtext.udb.parent/org.xtext.udb.ide/.project
```diff
@@ -30,11 +25,27 @@
...
 		<nature>org.eclipse.m2e.core.maven2Nature</nature>
 		<nature>org.eclipse.xtext.ui.shared.xtextNature</nature>
 		<nature>org.eclipse.jdt.core.javanature</nature>
 		<nature>org.eclipse.pde.PluginNature</nature>
 	</natures>
+	<filteredResources>
+		<filter>
+			<id>1777165792840</id>
+			<name></name>
+			<type>30</type>
```
ThinkOpenly: (Showing my ignorance...) What are these magic numbers?

tools/eclipse/dev/org.xtext.udb.parent/org.xtext.udb.ide/pom.xml
```diff
@@ -172,6 +172,40 @@
...
+			            </goals>
+			            <configuration>
+			                <target>
+			                    <!-- Copy and rename jar to udb-ls-all.jar -->
+			                    <copy todir="${project.basedir}/../../../udb-vscode/server"
+			                          flatten="true">
+			                        <fileset dir="${project.build.directory}">
+			                            <include name="*-ls.jar"/>
+			                        </fileset>
+			                        <mapper type="glob" from="*-ls.jar" to="udb-ls-all.jar"/>
```
ThinkOpenly, line 194: Do we need to use wildcards here?

tools/eclipse/dev/org.xtext.udb.parent/org.xtext.udb.ide/pom.xml
```diff
@@ -172,6 +172,40 @@
...
+			                    </copy>
+			                    <!-- Copy idlc from ruby-gems -->
+			                    <copy todir="${project.basedir}/../../../udb-vscode/server/idlc">
+			                        <fileset dir="${project.basedir}/../../../../ruby-gems/idlc"/>
+			                    </copy>
+			                    <!-- Copy vendor from jruby module -->
+			                    <copy todir="${project.basedir}/../../../udb-vscode/server/vendor">
+			                        <fileset dir="${project.basedir}/../org.xtext.udb.jruby/vendor"/>
+			                    </copy>
```
ThinkOpenly, line 203: Why are these copies needed?

tools/eclipse/dev/org.xtext.udb.parent/org.xtext.udb.tests/src/org/xtext/example/udb/tests/UdbParsingTest.xtend
```diff
@@ -116,12 +122,67 @@ class UdbParsingTest {
 		Assertions.assertEquals("RW-RH", sattype);
 
 		// testing alias
-		var vxalias = vxrm.getAlias().getAlias().getAlias().get(0);
-		var satalias = vxsat.getAlias().getAlias().getAlias().get(0);
+		var vxalias = vxrm.getAlias().getAliasName().getName();
+		var satalias = vxsat.getAlias().getAliasName().getName();
 		Assertions.assertEquals("vxrm.VALUE[1:0]", vxalias);
 		Assertions.assertEquals("vxsat.VALUE[0]", satalias);
```
ThinkOpenly: if we don't absolutely need this whitespace, please remove it.
```suggestion
```

tools/eclipse/dev/org.xtext.udb.parent/org.xtext.udb.tests/src/org/xtext/example/udb/tests/UdbParsingTest.xtend
```diff
@@ -116,12 +122,67 @@ class UdbParsingTest {
...
+		// requirements testing
+		var reqs = ext.getRequirements();
+		System.out.println(reqs.getRequirements());
+
+
+
+	}
+
+
+
```
ThinkOpenly: remove whitespace here, too
```suggestion
	}
```

tools/eclipse/dev/org.xtext.udb.parent/org.xtext.udb/src/org/xtext/example/udb/naming/UdbQualifiedNameProvider.java
```diff
@@ -0,0 +1,101 @@
...
+
+public class UdbQualifiedNameProvider extends DefaultDeclarativeQualifiedNameProvider {
+
+	public QualifiedName qualifiedName(CsrName csrName) {
+		if (csrName != null && csrName.getName() != null) {
+			return QualifiedName.create(csrName.getName());
+		}
+		return null;
+	}
+
```
ThinkOpenly, line 27: extra whitespace here, too. please check globally for this PR content.

tools/eclipse/dev/org.xtext.udb.parent/org.xtext.udb/.antlr-generator-3.2.0-patch.jar
ThinkOpenly, line 1: what's this?

tools/eclipse/run/testudb/src/csr/vcsr.udb
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
```
ThinkOpenly, line 3: Maybe add comments here describing the purpose of this file?

tools/eclipse/udb-vscode/server/udb-ls-all.jar
ThinkOpenly, line 1: Do we need this binary file in the repo? (Maybe "yes", but I have to ask. Binary files are generally discouraged, especially if they are executable code.)

tools/eclipse/udb-vscode/syntaxes/udb.tmLanguage.json
```diff
@@ -14,11 +15,15 @@
...
...e|address|virtual_address|indirect_address|indirect_slot|writable|priv_mode|length|description|definedBy|fields|type|alias|reset_value|location|location_rv32|location_rv64|affectedBy|xlen|company|versions|version|state|url|text_url|ratification_date|contributors|email|requirements|changes|doc_license|license|assembly|format|inherits|opcodes|funct7|funct3|display_name|value|access|access_detail|data_independent_timing|operation|operation_ast|encoding|match|variables|base|source|introduction|marketing_name|naming_scheme|processor_kind|register_length|register_class|registers|abi_mnemonics|caller_saved|callee_saved|roles|when|arch_read|arch_write|manual|marketing_version|uses_isa_manual|isa_manual_tree|volumes|chapters|authors|extensions|presence|parent_of|child_of|release|recommendations|extra_notes|mandatory_extensions|non_mandatory_extensions|prohibited_extensions|implemented_extensions|additional_extensions|arch_overlay|params|sections|references|hints|sign_extend|left_shift|pseudoinstructions|branch|num|index|size|equal|notEqual|greaterThan|lessThan|greaterThanOrEqual|lessThanOrEqual|includes|reason|cert_normative_rules|doc_links|rvi_jira_issue|id|register_file|schema|opcode)\\b"
```
ThinkOpenly, line 18: In some ways, it would be nice to build this string from a more readable list/array.

tools/eclipse/dev/org.xtext.udb.parent/org.xtext.udb/.antlr-generator-3.2.0-patch.jar
brayjmendoza: used for generating the language server I believe, not necessary to keep in the repo as it gets regenerated whenever the language server needs to be updated

tools/eclipse/run/testudb/src/csr/vcsr.udb
brayjmendoza: It's purpose is just to serve as an example file of the DSL, specifically for the CSR schema. Thoughts on just mentioning that in the documentation instead? (`tools/eclipse/README.md`)

tools/eclipse/udb-vscode/server/udb-ls-all.jar
brayjmendoza: not needed, but this PR does remove it

tools/eclipse/dev/org.xtext.udb.parent/org.xtext.udb.ide/.project
niluo-shiqi: Good call, I will put it back!

tools/eclipse/dev/org.xtext.udb.parent/org.xtext.udb.ide/.project
niluo-shiqi: The id is a unique identifier Eclipse auto-assigns, and type=30 is a bitmask (EXCLUDE_ALL + FILES + FOLDERS + INHERITABLE), so it excludes matching files and folders recursively. Auto-created by the Eclipse JDT language server to skip node_modules/.git during indexing. No effect on the build or extension, thanks for the catch, happy to drop it. It regenerates locally on its own if needed.

tools/eclipse/dev/org.xtext.udb.parent/org.xtext.udb.ide/pom.xml
niluo-shiqi: The wildcard was just to dodge hardcoding the version so it would still be able to find the generated language server jar file even if the name of the file changes, but you are right, a wildcard isn't clean and might cause problems later on. I will replace the search and rename with a single file copy with ${project.build.finalName}-ls.jar so its more explicit.

tools/eclipse/dev/org.xtext.udb.parent/org.xtext.udb.ide/pom.xml
niluo-shiqi: These copies are needed because idlc and vendor are the Ruby sources the embedded JRuby parser needs at runtime for IDL validation. They have to be bundled inside the extension's server/ folder because it ships as a .vsix, once a user installs it, ruby-gems/ and the jruby module aren't on their machine. Without them, IDL validation would not work in the vscode extension, so we chose to copy them in at build time.

tools/eclipse/dev/org.xtext.udb.parent/org.xtext.udb/src/org/xtext/example/udb/naming/UdbQualifiedNameProvider.java
niluo-shiqi: Thank you for catching this! I will check globally

tools/eclipse/udb-vscode/syntaxes/udb.tmLanguage.json
niluo-shiqi: I agree, that huge string isn't a great solution. It would be better to have a small build step that generates the JSON from keyword arrays, but since it's self-contained and this PR's already gotten pretty big, would it be alright if I tackled it as a follow-up and opened an issue for it?

tools/scripts/language-server-script/regen-udb-ls.bat
niluo-shiqi: Good idea on both. Expanded the header comment to explain what the script builds and why, and added -h/--help (plus /? for Windows) to the .bat. Also mirrored the help into the .sh so both platforms match, and tested the .sh on macOS.

niluo-shiqi (commented):
Thank you for the thorough review!

PR #1834, opened 2026-05-14
fix(ci): use correct gem name in gem bump PR body

jordancarlin:
The gem-bump workflow's PR body listed every bumped gem as "lib" because dirname was only applied twice to the version file path, stopping at tools/ruby-gems/<gem>/lib instead of <gem>. Add the missing third dirname level so basename returns the gem name.


jordancarlin:
@dhower-qc Once this is merged, can we create a new release of the gems so that ACT4 can use the config specific generated C/SVH headers?

dhower-qc (approved):

Issue #1173, opened 2025-10-06
Output glitches in main HTML output, as of 2025 Oct 01


nwf:
Using https://riscv-software-src.github.io/riscv-unified-db/ as of 5d6c397acee0e2ff4e22fcf1958b562437117b4d, a few things seem unwell.  I'm afraid I don't know the tooling in this repository well enough to point at why these might have happened, but hopefully the report that they have is useful to someone.

1. https://riscv-software-src.github.io/riscv-unified-db/isa_explorer/inst_table.html links to https://riscv-software-src.github.io/riscv-unified-db/manual/html/isa/isa_20240411/insts/c.ld, which is a 404, as is https://riscv-software-src.github.io/riscv-unified-db/manual/html/isa/isa_20240411/insts/c.lw .

2. The "defining extensions" look to be getting rendered as text rather than as structured material.  Consider https://riscv-software-src.github.io/riscv-unified-db/manual/html/isa/isa_20240411/insts/c.fld, for example: 

```
<div class="sect1">
<h2 id="_defining_extension"><a class="anchor" href="#_defining_extension"></a>Defining extension</h2>
<div class="sectionbody">
<div class="paragraph">
<p>anyOf:
* allOf:
<strong> C, version &gt;= C@2.0.0
</strong> D, version &gt;= D@2.2.0
* Zcd, version &gt;= Zcd@1.0.0</p>
</div>
</div>
</div>
```
The result looks unpleasnt:
<img width="846" height="92" alt="Image" src="https://github.com/user-attachments/assets/6e7dd527-c299-45d5-be04-205d0fc86a13" />

ThinkOpenly:
> Using https://riscv-software-src.github.io/riscv-unified-db/ as of [5d6c397](https://github.com/riscv-software-src/riscv-unified-db/commit/5d6c397acee0e2ff4e22fcf1958b562437117b4d), a few things seem unwell.

Are you saying it looked fine before that commit hash, or is that just a point of reference?

ThinkOpenly:
A lot of the instructions are 404.

nwf:
> Are you saying it looked fine before that commit hash, or is that just a point of reference?

Just as point of reference; I have not gone wandering through history.

aclfe:
@dhower-qc I'll send a PR for this. There's two parts to address, I'll address the 404 url issue first

PR #1841, opened 2026-05-19
fix: fixing 404 links (part 1)

aclfe:
Fixes 404 errors of #1173 

Dotted names (e.g. c.ld) must use the sanitized name (c_ld) in the file path so Antora can resolve the xref and GitHub Pages can serve the resulting HTML page. Memoized per Rake application

Tried locally: 
image

### DIFF
--- a/tools/ruby-gems/udb-gen/lib/udb-gen/generators/manual/tasks.rake
+++ b/tools/ruby-gems/udb-gen/lib/udb-gen/generators/manual/tasks.rake
@@ -6,6 +6,14 @@
 
 require "digest"
 
+def sanitized_inst_lookup
+  Rake.application.instance_variable_get(:@sanitized_inst_lookup) ||
+    Rake.application.instance_variable_set(
+      :@sanitized_inst_lookup,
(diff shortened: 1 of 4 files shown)


tools/ruby-gems/udb-gen/lib/udb-gen/generators/manual/tasks.rake
```diff
@@ -6,6 +6,14 @@
 
 require "digest"
 
+def sanitized_inst_lookup
```
aclfe, line 9: reverse lookup

dhower-qc:
Thanks for the contribution!  What is (part 2) going to be?

aclfe:
> Thanks for the contribution! What is (part 2) going to be?

@dhower-qc 
Original issue mentioned two issues: 
- 404 link issues (this pr)
- "defining extensions" malformed rendering (part 2)

As of now, changes are made to address 404. Second part we can address defining extensions. 
Both issues are different stacks, seems cleaner to isolate changes into different parts.

dhower-qc (approved):

Issue #1731, opened 2026-03-11
WFI and WRS need to wait before exception when mstatus.TW=1


ThinkOpenly:
> When TW=1, then if WFI is executed in any less-privileged mode, and it does not
complete within an implementation-specific, bounded time limit, the WFI instruction causes an illegal instruction exception. An implementation may have WFI always raise an illegal-instruction exception in
less-privileged modes when TW=1, even if there are pending globally-disabled interrupts when the
instruction is executed

and

> When the TW (Timeout Wait) bit in mstatus is set and WRS.NTO is executed in any privilege mode other than
M mode, and it does not complete within an implementation-specific bounded time limit, the WRS.NTO instruction will cause an illegal-instruction exception.
> When executing in VS or VU mode, if the VTW bit is set in hstatus, the TW bit in mstatus is clear, and the
WRS.NTO does not complete within an implementation-specific bounded time limit, the WRS.NTO instruction
will cause a virtual-instruction exception.

_Originally posted by @ThinkOpenly in https://github.com/riscv/riscv-unified-db/pull/1669#discussion_r2844452564_

The current implementations check the conditions and may raise an exception _before waiting_. This is not correct.

iakshkhurana:
Can i work on this?

PR #1843, opened 2026-05-20
fix(WFI/WRS): fixing raise an exception before waiting issue #1731

aclfe:
Addresses https://github.com/riscv/riscv-unified-db/issues/1731 as stated originally in the issue:

> A possible implementation is to pass required architectural state into the builtin(s) as new parameters, so that the builtins can perform a wait for a configuration-specific (parameterized) time and provide a new return value indicating whether the wait was "successful", or an exception should be thrown.

### DIFF
--- a/backends/cpp_hart_gen/cpp/include/udb/hart.hpp
+++ b/backends/cpp_hart_gen/cpp/include/udb/hart.hpp
@@ -124,12 +124,18 @@ namespace udb {
       throw AbortInstruction();
     }
 
-    void wfi() {
-      throw WfiException();
+    bool wfi(bool tw_trap, bool vtw_trap) {
+      //No timeout handling yet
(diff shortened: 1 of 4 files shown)


aclfe:
Tested changes with `╰─$ mise exec -- bundle exec ./do test:idl CFG=_`

dhower-qc (changes requested):
This will also require a change to the implementation of the builtin in trusim. @henrikg-qc

henrikg-qc:
> This will also require a change to the implementation of the builtin in trusim. @henrikg-qc

I will talke a look and push updates to this PR if allowed. Otherwise I will synch with @aclfe

aclfe:
image

@henrikg-qc  Already ticked, should be good to go :)

PR #1860, opened 2026-06-04
Fix DocLink error handling bugs (#1744)

### LINKED ISSUE #1744: Issues from Copilot in `doc_link.rb`
Originally posted by Copilot in #1743 

> `initialize` raises an ArgumentError using `#{data.class}`, but `data` is not defined in this scope (will raise `NameError` when `dst_link` is not a String). Also, the `nil` check happens after the type check, so `dst_link=nil` won't hit the intended "Missing documentation link" error. Use `dst_link.class` in the message and check `dst_link.nil?` before the type check (or otherwise structure the validation so both cases produce the intended errors).

> YARD types here say `db_obj` is a `String`, but this method calls `db_obj.name` and `db_obj.kind`, so it’s some kind of database object (e.g., `DatabaseObject`). Please update the `@param db_obj [...]` type to match the actual expected type to keep the docs accurate.
```suggestion
    # @param db_obj [Object] Database object
```

Felix-Gong:
## Summary

- Fix undefined variable `data` → `dst_link` in type check error message
- Reorder nil check before type check to provide more specific error
- Fix YARD type annotation `[String]` → `[Object]` for db_obj parameter

Closes #1744

### DIFF
--- a/tools/ruby-gems/udb/lib/udb/doc_link.rb
+++ b/tools/ruby-gems/udb/lib/udb/doc_link.rb
@@ -33,12 +33,11 @@
 module Udb
   class DocLink
     # @param dst_link [String] The documentation link provided in the YAML
-    # @param db_obj [String] Database object
+    # @param db_obj [Object] Database object
     def initialize(dst_link, db_obj)
-      raise ArgumentError, "Need String but was passed a #{data.class}" unless dst_link.is_a?(String)
+      raise ArgumentError, "Missing documentation link for #{db_obj.name} of kind #{db_obj.kind}" if dst_link.nil?
+      raise ArgumentError, "Need String but was passed a #{dst_link.class}" unless dst_link.is_a?(String)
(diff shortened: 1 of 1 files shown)


AFOliveira (approved):
lgtm, as long as CI passes. Thanks for contributing

Felix-Gong:
> lgtm, as long as CI passes. Thanks for contributing

@AFOliveira The Codecov patch coverage failure is on 2 defensive raise ArgumentError lines in doc_link.rb. Note that doc_link.rb has no existing test coverage at all — this is not introduced by this PR. Would you like me to add tests for doc_link.rb, or can we merge as-is?

ThinkOpenly (approved):

ThinkOpenly:
> Note that doc_link.rb has no existing test coverage at all — this is not introduced by this PR.

Agreed. While coverage patches are certainly welcome, there is no need to hold up this PR.

PR #1861, opened 2026-06-04
Split all instructions

wmat:
Breaks all_instructions.adoc into four files, unpriv_rv32.adoc, unpriv_rv64.adoc, priv_rv32.adoc, and priv_rv64.adoc.  These are then synced to riscv-isa-manual.  Note that this is currently manual, however, could be automated once the ISA manual changes are accepted.

### DIFF
--- a/backends/instructions_appendix/split_instructions.rb
+++ b/backends/instructions_appendix/split_instructions.rb
@@ -155,19 +155,24 @@ def increment_heading_levels(text)
   text.gsub(/^(=+) /) { "#{$1}= " }
 end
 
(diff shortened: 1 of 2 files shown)


wmat:
Just want to add context to this PR.  Note that it's output is 4 adoc files for use in riscv-isa-manual.  These are unpriv and priv indexes.  For reference see https://github.com/riscv/riscv-isa-manual/pull/3178

Also, this is one solution for the instruction per page indexes within the ISA manuals. @aswaterman is currently researching another method for this, so this PR could await the final decision on a solution.  I will update here once that occurs.

PR #1862, opened 2026-06-05
Fix shell quoting for paths with spaces in toolchain container (#1709)

### LINKED ISSUE #1709: Shell Quoting Failure with Spaces in Paths
**Describe the bug**
When the riscv-unified-db repository path contains spaces, UDB fails with a misleading error message. The root cause is improper shell quoting in the `bin/setup` and `bin/udb` scripts, which causes unquoted variable expansion to split arguments at whitespace characters when invoking Podman container commands.

**To Reproduce**
Run any udb command

```
Using podman environment
Image not found locally. Attempting to pull from registry...
Successfully pulled image from registry.
Error: repository name must be lowercase
UDB configuration validation failed.
```
**Additional context**
Refactoring `bin/udb` and `bin/setup` should fix it.

Felix-Gong:
## Summary

- Convert `TOOLCHAIN_RUN` from string to array in `_setup_toolchain_run()` to properly handle paths containing spaces
- Update all bin/ wrapper scripts to use `"${TOOLCHAIN_RUN[@]}"` for correct array expansion
- Properly quote all path variables in volume mounts

Closes #1709

### DIFF
--- a/.toolchain/.build-elfutils.sh
+++ b/.toolchain/.build-elfutils.sh
@@ -25,7 +25,7 @@ if [ "${UDB_TOOLCHAIN_CONTAINER:-0}" = "1" ]; then
   source "${ROOT}/../bin/.toolchain.sh"
   cd "$BINARY_DIR"
   _setup_toolchain_run
-  exec $TOOLCHAIN_RUN bash -c '
+  exec "${TOOLCHAIN_RUN[@]}" bash -c '
     set -e
     cd "$1"
     "$2/configure" \
--- a/bin/.toolchain.sh
+++ b/bin/.toolchain.sh
@@ -48,29 +48,33 @@ _setup_toolchain_run() {
   fi
 
   local selinux_label=""
-  local user_flags=""
   if [[ "$runtime" == *podman* ]]; then
     selinux_label=":z"
-    user_flags="--userns=keep-id"
+  fi
+
+  TOOLCHAIN_RUN=("$runtime" "run" "--rm")
+  if [ -t 1 ] && [ -t 0 ]; then
+    TOOLCHAIN_RUN+=("-it")
+  fi
+  TOOLCHAIN_RUN+=("-v" "${ROOT}:${ROOT}${selinux_label}")
+
+  if [[ "$runtime" == *podman* ]]; then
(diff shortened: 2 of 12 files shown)


ThinkOpenly:
Thank you for this contribution! These changes certainly help with commands that depend on the toolchain container. Before:
```
$ bin/riscv64-unknown-elf-objdump -dr blah
docker: invalid reference format

Run 'docker run --help' for more information
```
After:
```
$ bin/riscv64-unknown-elf-objdump -dr blah
riscv64-unknown-elf-objdump: 'blah': No such file
```

However, the changes do not fully address the issue reported in #1709:
```
$ bin/udb list csrs
ℹ info    Merged architecture files written to .../with space/gen/spec/_
bundler: failed to load command: udb (.../.local/share/gem/ruby/3.4.0/bin/udb)
.../with space/tools/ruby-gems/udb/lib/udb/yaml/yaml_resolver.rb:1035:in 'Udb::Yaml::Resolver#validate_against_schema': Schema validation failed for exception_code/Breakpoint.yaml: (RuntimeError)
  - /$source: format
[...traceback...]
```

Felix-Gong:
@ThinkOpenly  Thanks for checking this. I reproduced the remaining failure with a checkout path containing spaces:

```text
/tmp/udb path with space
```

Before the update, `bin/udb list csrs` failed with the same schema validation error:

```text
Schema validation failed for exception_code/Breakpoint.yaml:
  - /$source: format
```

I updated the PR to handle this case by escaping `$source` only in the JSON object used for schema validation. The resolved YAML still stores `$source` as the original filesystem path, so existing code that treats `$source` as a path is unchanged.

I also added a regression test that validates an input directory containing spaces against a schema requiring `$source` to be a `uri-reference`.

Validation:

```text
bin/udb list csrs
```

now succeeds from `/tmp/udb path with space`.

Additional checks:

```text
test_schema_validation_accepts_source_paths_with_spaces:
1 runs, 2 assertions, 0 failures, 0 errors, 0 skips

tools/ruby-gems/udb/test/test_yaml_resolver.rb:
14 runs, 462291 assertions, 0 failures, 0 errors, 0 skips

git diff --check: passed
```

ThinkOpenly (approved):

PR #1869, opened 2026-06-11
fix(inst_table): shift from beginning of segment

thecharlesjenkins:
Shifting from the end of the segment causes the segment to overflow. The shifting is supposed to be from the front to get all of the bits to properly align.


radimkrcmar (approved):
I don't understand why this is a fix rather than a design change, yet the new format seems better, because the usual expectation about left bit shift is that
`0b111 < 4 -> 0b1110000` (the shifted value begins at fifth bit) and not
`0b111 < 4 ->   0b11100` (the shifted value ends at fifth bit).

Probably the most important part is to reach a decision before any external projects start consuming the format -- Linux is already looking forward to depending on this.

ThinkOpenly (approved):

PR #1878, opened 2026-06-24
fix: add Supm extension dependency on Ssnpm or Smnpm

### LINKED ISSUE #1768: Supm relies on Ssnpm or Smnpm
According to the Linux discussion [1] and code, extension Supm relies on Ssnpm or Smnpm. But I can't find this in the database.

[1] https://lore.kernel.org/linux-riscv/f1b541f4-6d34-4080-b322-f097c4e2f48d@sifive.com/

Felix-Gong:
## What

Add `requirements` to `Supm.yaml` specifying that Supm depends on either `Ssnpm` or `Smnpm`.

## Why

Fixes #1768. According to the RISC-V spec and Linux kernel discussion:
- When S-mode is implemented, Supm requires Ssnpm (pointer masking for U-mode)
- When only M-mode is implemented, Supm requires Smnpm (pointer masking for U-mode)

## Changes

```yaml
requirements:
  extension:
    anyOf:
      - name: Ssnpm
      - name: Smnpm
```

### DIFF
--- a/spec/std/isa/ext/Supm.yaml
+++ b/spec/std/isa/ext/Supm.yaml
@@ -17,6 +17,11 @@ description: |
   and not the associated CSR controls that exist at a higher privilege level (i.e., in the execution environment).
 
 type: privileged
+requirements:
+  extension:
+    anyOf:
+      - name: Ssnpm
+      - name: Smnpm
 versions:
   - version: "1.0.0"
     state: ratified


Felix-Gong:
After re-analyzing the issue, I believe this is not actually a bug. 

The key insight from the spec:
> Supm describes an execution environment but has **no bearing on hardware implementations**.

- **Supm/Sspm** are execution environment extensions (what the EE provides)
- **Ssnpm/Smnpm** are hardware extensions (what hardware implements)

These are different abstraction layers. Supm doesn't 'require' Ssnpm/Smnpm - it's a separate concern. The issue reporter may have confused execution environment extensions with hardware extensions.

Closing this PR.

PR #1879, opened 2026-06-26
fix: accept bare version strings in implemented_extensions

Felix-Gong:
Align `implemented_extensions.version` handling between the config JSON schema and the Ruby config parser.

Both forms are now accepted for fully configured architecture configs:

  - Bare exact versions, e.g. `"2.1"`
  - Exact `=` versions, e.g. `"= 2.1"`

Other comparison operators such as `">= 2.1"` remain invalid for `implemented_extensions`, since these entries represent exact implemented extension versions rather than version requirements.


jordancarlin (commented):
With this change, the Ruby accepts versions with or without the comparison operators, while the schema still requires the version without. This is definitely an improvement, but we should bring both of those into alignment with each other. @ThinkOpenly, any opinions on dropping support for the comparison operators in the Ruby vs adding them to the schema? I think I’d lean towards dropping them from the Ruby, even though it is a breaking change.

ThinkOpenly:
How was this issue discovered, or what new data provokes it? (Why haven't we encountered this before?)

jordancarlin:
I’ve been running into it for months and just never got around to fixing it. The Ruby parser and the JSON have different expectations for the data, but it looks like the schema is not validated for config files, so as long as you follow what Ruby expects, everything “works.” This results in the error squiggles under all of the extensions in the config files, though, because you aren’t following the schema.

Felix-Gong:
@jordancarlin @ThinkOpenly  Thanks, I agree with aligning the Ruby parser with the schema here.

  I updated the PR to drop comparison-operator support for `implemented_extensions` hash entries. The parser now treats implemented extension versions as exact `VersionSpec`
  values instead of requirement strings, so values like `"2.1"` are accepted while values like `"= 2.1"` are rejected.

  I also updated the existing `qc_iu` config to use bare implemented-extension versions so the in-tree config follows the same schema/Ruby behavior.

  The mismatch was hidden because normal config processing appears to follow the Ruby parser behavior rather than enforcing the JSON schema on these config files. Existing configs could therefore keep working when they used Ruby-style requirement strings such as `"= 2.1"`.

  The issue shows up in schema-aware tooling/IDE validation: the schema expects a bare `rvi_version`, while the old Ruby parser expected a requirement string. This PR resolves that by making both the schema and Ruby parser use bare exact versions for `implemented_extensions`.

  Validation:
  - `test_implemented_extensions_hash_versions_match_schema`: 1 run, 2 assertions, 0 failures
  - `tools/ruby-gems/udb/test/test_cfg_arch.rb`: 11 runs, 436 assertions, 0 failures, 0 errors, 0 skips
  - `git diff --check`: passed

jordancarlin:
@Felix-Gong thanks for working on this. Sorry to make you do more work, but we discussed this during the UDB meeting this week and decided that both the schema and Ruby should support both forms. We don’t want to break existing configurations that already use =. They should accept either a bare version or a version with an =, but not any other comparison operator. Would you be able to update this PR accordingly?

Felix-Gong:
@jordancarlin  Thanks for the update. I updated the PR to support both exact-version forms in both places:

- Schema now accepts bare versions like `"2.1"` and exact `=` versions like `"= 2.1"` for `implemented_extensions.version`.
- Ruby parsing accepts both forms and normalizes them internally to the bare version string.
- Other comparison operators such as `">= 2.1"` are still rejected by both schema validation and Ruby parsing.
- I restored `qc_iu.yaml` to use the existing `= ...` form so this keeps coverage for backward compatibility with existing configs.

Validation:
- `test_implemented_extensions_hash_versions_match_schema`: 1 run, 9 assertions, 0 failures
- `tools/ruby-gems/udb/test/test_cfg_arch.rb`: 11 runs, 443 assertions, 0 failures, 0 errors, 0 skips
- `git diff --check`: passed

jordancarlin (approved):

PR #1880, opened 2026-06-27
Fix REUSE Compliance

norwd:
At the time of writing, the REUSE badge is indicating non-compliance.

[REUSE status](https://api.reuse.software/info/github.com/riscv/riscv-unified-db)

This is because the `cfg_header_base.rb` file contains the `SPDX-License-Identifier` expression in a string, the formatting of which is being detected as saying that the remainder of that file is licensed under `BSD-3-Clause-Clear",` (note the trailing quotation mark and comma), which is not a valid licence.

The solution to this is to wrap that line in the `REUSE-Ignore{Start/End}` tags.

Screenshot 2026-06-27 at 8 58 56 PM


jordancarlin (approved):
Thanks for this!

PR #1884, opened 2026-06-30
Resolve return-type warnings in IDL functions and jvt.BASE sw_write

### LINKED ISSUE #1883: fix: resolve -Wreturn-type warnings in IDL functions and jvt CSR sw_write
## Problem

Several IDL functions and the `jvt` CSR `sw_write` block are missing explicit
`return` statements on all code paths. This causes `-Wreturn-type` compiler
warnings (and undefined behavior under optimizing compilers) when the IDL is
compiled into the generated C++ ISS hart model.

Affected locations:
- `spec/std/isa/csr/jvt.yaml`: `sw_write` block missing `return 0` after `unimplemented_csr()`
- `spec/std/isa/isa/builtin_functions.idl`: `read_physical_memory` missing `return 0` in unreachable `assert(false)` branch
- `spec/std/isa/isa/globals.isa`: multiple functions (`compatible_mode?`, `mpv`, `tinst_value_for_guest_page_fault`, `tinst_transform`, `gstage_page_walk`, `stage1_page_walk`, `canonical_vaddr?`, `canonical_gpaddr?`, `read_memory`, `write_memory`) missing explicit returns on all paths
- `spec/std/isa/isa/interrupts.idl`: `highest_priority_interrupt` missing `unreachable()` after terminal `assert(false)`

## Fix

Add explicit `return` statements (or `unreachable()` for provably-unreachable paths) to all affected functions so every control-flow path has a defined exit.

A corresponding PR is attached.

jpak-qc:
Closes #1883

## Summary

Several IDL functions and the \`jvt\` CSR \`sw_write\` block were missing explicit
\`return\` statements on all control-flow paths, causing \`-Wreturn-type\` warnings
when the IDL is compiled into the generated C++ ISS hart model.

### \`jvt.yaml\` sw_write

The \`sw_write\` block calls \`unimplemented_csr()\` to signal an access violation,
but \`unimplemented_csr()\` does not terminate execution in the IDL type system.
The compiler sees a non-void function with a path that falls off the end after
the call, producing a \`-Wreturn-type\` warning. Added \`return 0\` immediately after
the call to give every path an explicit exit.

### IDL functions (globals.isa, builtin_functions.idl, interrupts.idl)

Multiple functions had similar gaps: branches ending in \`assert(false)\`,
\`unpredictable()\`, or \`raise()\` with no following \`return\`, and top-level
fall-through at the end of function bodies. All three are terminal calls —
\`assert(false)\` and \`unpredictable()\` are \`[[noreturn]]\` in the generated C++;
\`raise()\` always terminates execution by either calling \`unpredictable()\` or
redirecting \`\$pc\` to the trap handler. Added explicit \`return\` statements (or
\`unreachable()\` for provably-unreachable paths) to all affected functions.


henrikg-qc (approved):

spec/std/isa/isa/globals.isa
```diff
@@ -1631,6 +1635,7 @@ function tinst_value_for_guest_page_fault {
         }
       }
     }
+    return 0;
```
ThinkOpenly, line 1638: Do we really need this one? Every path to this point has a `return`, by my reckoning.

spec/std/isa/isa/globals.isa
```diff
@@ -2024,6 +2031,7 @@ function gstage_page_walk {
         ppn = pte[PA_SIZE-3:10] << 12;
       }
     }
+    return result;
```
ThinkOpenly: I presume this is unreachable code? `result` has never been initialized at this point. `result` is declared well above and only given a value just before a `return` above. We should return a known value here, perhaps 0?

spec/std/isa/isa/globals.isa
```diff
@@ -2658,8 +2669,10 @@ function read_memory {
         return result;
       } else if (MISALIGNED_SPLIT_STRATEGY == "custom") {
         unpredictable("An implementation is free to break a misaligned access any way, leading to unpredictable behavior when any part of the misaligned access causes an exception");
+        return 0;
       }
     }
+    return 0;
```
ThinkOpenly, line 2679: This would only be reached if `raise` is not a "doesn't return" function. This would be akin to `assert`. Is that the case? (Probably.)

If so, please update the PR description to include `raise` alongside `assert` and `unpredictable`.

spec/std/isa/isa/interrupts.idl
```diff
@@ -259,6 +259,7 @@ function highest_priority_interrupt {
       }
     }
     assert(false, "There is no valid interrupt");
+    unreachable();
```
ThinkOpenly, line 262: In `builtin_functions.idl`, above, you used `unreachable()` after an `assert()`. Is there a reason these two cases are handled differently? If not, let's be consistent. (I prefer `unreachable()`.)

ThinkOpenly (changes requested):
Thanks for taking this on, @jpak-qc ! I've a few minor comments...

spec/std/isa/isa/globals.isa
ThinkOpenly: Maybe we should use `unreachable()` here, too, if appropriate.

spec/std/isa/isa/globals.isa
jpak-qc: THere is two fall through paths the compiler should see: one when _for_final_vs_pte_ is false and then _REPORT_GPA_IN_TVAL_ON_INTERMEDIATE_GUEST_PAGE_FAULT_ is also false, the else branch exists with no return; second if op is neither Fetch, RD, WR, or ReadModifyWrite, the outer if/else if chain falls through. Actually third as well, when for_final_vs_pte is false REPORT_GPA_IN_TVAL_ON_INTERMEDIATE_GUEST_PAGE_FAULT is true, but VSXLEN is neither 32 nor 64 , so the if/else if also falls through. So I would argue the outer return 0; is needed to cover all three. Would like to know your opinion with this additional context.

spec/std/isa/isa/globals.isa
jpak-qc: You're right! I'll change it to unreachable()

spec/std/isa/isa/globals.isa
jpak-qc: Yep! I'll update the PR description to include `raise ` alongside them.

spec/std/isa/isa/globals.isa
jpak-qc: Loop either finds a leaf PTE and returns `result` from the inside, or raises an exception which never returns. The fall through can never be reached in the correct execution.

spec/std/isa/isa/interrupts.idl
jpak-qc: I believe this was addressed already although not reflected yet on an earlier version before the branch got updated!

spec/std/isa/isa/globals.isa
ThinkOpenly: Yes, indeed you are correct. My bad. Thanks for the explanation!

spec/std/isa/isa/globals.isa
```diff
@@ -1180,6 +1181,7 @@ function mpv {
       return (MXLEN == 32) ? CSR[mstatush].MPV : CSR[mstatus].MPV;
     } else {
       assert(false, "TODO");
+      return 0;
```
ThinkOpenly:
```suggestion
      unreachable();
```

spec/std/isa/isa/globals.isa
```diff
@@ -1658,9 +1663,11 @@ function tinst_transform {
         return {encoding[31:20], addr_offset, encoding[14:0]};
       } else {
         assert(false, "Bad transform");
+        return 0;
```
ThinkOpenly:
```suggestion
        unreachable();
```

spec/std/isa/isa/globals.isa
```diff
@@ -1658,9 +1663,11 @@ function tinst_transform {
...
       }
     } else {
       assert(false, "TODO: compressed instruction");
+      return 0;
```
ThinkOpenly:
```suggestion
      unreachable();
```

spec/std/isa/isa/globals.isa
```diff
@@ -3059,6 +3072,7 @@ function write_memory {
         }
       } else if (MISALIGNED_SPLIT_STRATEGY == "custom") {
         unpredictable("An implementation is free to break a misaligned access any way, leading to unpredictable behavior when any part of the misaligned access causes an exception");
+        return;
       }
     }
```
ThinkOpenly, line 3081: Do we need another `return` here, for the fall-through case? If not, do we need the `return` under `unpredictable()` above?

spec/std/isa/isa/globals.isa
ThinkOpenly:
```suggestion
    unreachable();
```

ThinkOpenly, line 1798:
```suggestion
      assert(false, "Unhandled exception type");
      unreachable();
```

ThinkOpenly, line 1984:
```suggestion
              assert(false, "Unexpected PTESIZE");
              unreachable();
```

ThinkOpenly, line 2310:
```suggestion
              assert(false, "Unexpected PTESIZE");
              unreachable();
```

ThinkOpenly, line 2465:
```suggestion
      assert(false, "Unexpected SatpMode");
      unreachable();
```

ThinkOpenly, line 2795:
```suggestion
      assert(false, "Unexpected LRSC_RESERVATION_STRATEGY");
      unreachable();
```

ThinkOpenly (changes requested):
In an effort to be consistent, I think either:
- the compiler will know that `assert(false)` is a "doesn't return" case, and we don't need an explicit `unreachable()`, OR
- the compiler doesn't know, and we need `unreachable()` after all `assert(false)` statements.

My comments here presume the latter.

jpak-qc:
> In an effort to be consistent, I think either:
> 
> * the compiler will know that `assert(false)` is a "doesn't return" case, and we don't need an explicit `unreachable()`, OR
> * the compiler doesn't know, and we need `unreachable()` after all `assert(false)` statements.
> 
> My comments here presume the latter.

Hey Paul, thanks for the feedback much appreciated. So I had some agents sift through the compiler, i've validated that your scenario 2 is correct. 

- IDL compiler's `always_terminates?` only seems to recognize `raise `not `assert`. 
- Code generator already emits `__builtin_unreachable()` after every` assert(false, ...) `in gen c++.
- The udb_assert macro is not noreturn even though `__udb_assert_fail() ` is the ternary expansionwill lose that attribute.

Since I agree with all changes I've accepted them all. I'll also remember to be more consistent with control flow termination patterns across the whole file, not just specific sites for future reference. Thanks!

spec/std/isa/isa/globals.isa
jpak-qc: Good catch, I mistakenly treated it the same as the non-void functions we were fixing without checking that` write_memory `is void. `write_memory ` is void so no return is needed anymore. I'll remove the return statement under `unpredictable()`!

ThinkOpenly:
> * Code generator already emits `__builtin_unreachable()` after every`assert(false, ...)`in gen c++.

Do I interpret that correctly in that we don't _need_ to add `unreachable()` after every `assert(false)`?

> * The udb_assert macro is not noreturn even though `__udb_assert_fail() ` is the ternary expansionwill lose that attribute.

I'm unable to parse that. In the most abstract sense, `assert` will _conditionally_ not return, so it can't be marked `noreturn` (because it returns if the given expression evaluates to true).

> Since I agree with all changes I've accepted them all. I'll also remember to be more consistent with control flow termination patterns across the whole file, not just specific sites for future reference.

OK, so the convention will be to add `unreachable()` after _every_ `assert(false)`. That's probably the best for readability.

ThinkOpenly (approved):

jpak-qc:
> > * Code generator already emits `__builtin_unreachable()` after every`assert(false, ...)`in gen c++.
> 
> Do I interpret that correctly in that we don't _need_ to add `unreachable()` after every `assert(false)`?
> 
> > * The udb_assert macro is not noreturn even though `__udb_assert_fail() ` is the ternary expansionwill lose that attribute.
> 
> I'm unable to parse that. In the most abstract sense, `assert` will _conditionally_ not return, so it can't be marked `noreturn` (because it returns if the given expression evaluates to true).
> 
> > Since I agree with all changes I've accepted them all. I'll also remember to be more consistent with control flow termination patterns across the whole file, not just specific sites for future reference.
> 
> OK, so the convention will be to add `unreachable()` after _every_ `assert(false)`. That's probably the best for readability.

Agreed, `assert `in general can't be noreturn because of when its true. However, just to clarify I believe and meant that `assert(false) ` even at that call site where we know it will never return, the compiler still can't tell because the noreturn attribute of __udb_assert_fail gets swallowed by the ternary expansion.

But, yeah IMO I believe adding unreachable() would still help for readability like you said. Thanks for working with me on this!

PR #1677, opened 2026-02-19
fix(data): use LHS bit-slice assignment in vmv.v.i operation

ishaan-arora-1:
## Summary

- Fixes the ISS assertion failure (`Assertion failed: (msb >= lsb)`) in `vmv.v.i` by adding boundary guards to the concatenation-based element update.
- The original code produced invalid bit-slice ranges when the element touched bit 0 (`v[vd][start_bit_pos-1:0]` with `start_bit_pos=0`) or bit `VLEN-1` (`v[vd][VLEN-1:end_bit_pos+1]` with `end_bit_pos=VLEN-1`).
- The direct LHS bit-slice assignment (`v[vd][msb:lsb] = val`) suggested by @ThinkOpenly in #1639 cannot be used because the IDL parser does not support nested array access on the LHS, and the C++ backend's `bit_insert` fails with runtime-unknown `VLEN`.

Fixes #1618

## Test plan

- [x] `gen:arch` passes — IDL compiles successfully
- [x] `test:instruction_appendix` passes — golden output matches
- [x] All pre-commit checks pass
- [ ] CI `regress-riscv-tests-32` and `regress-riscv-tests-64` should now pass

Issue #1618, opened 2026-02-12
Focus test for vmv_v_i still asserts on ISS but passes on Sail now


adingank-qualcomm:
@ishaan-arora-1 fixed issue #1599 by correcting the hex arguments to the `vmv_v_i` instructions in [`vmv_v_i.S`](https://github.com/riscv/riscv-unified-db/blob/main/tests/isa/rv64uv/vmv_v_i.S).

After recompilation the test now passes on Sail, but the ISS still gives the same assertion error. 

@jmawet Please note (and skip the `vmv_v_i` test for your development. You may want to consult with @henrikg-qc  re the ISS assertion.

- Sail (Binary Release 0.9): Snippet for `test_1`
>$ sail_riscv_sim --trace-instr --config /opt/sail-riscv/share/sail-riscv/config/rv64d_v128_e64.json tests/isa/rv64uv-p-vmv_v_i
...
[77] [U]: 0x00000000800001A4 (0x00100193) addi x3, x0, 0x1                          test_1+0
[78] [U]: 0x00000000800001A8 (0xC1817057) vsetivli x0, 0x2, e64, m1, tu, mu         test_1+4
[79] [U]: 0x00000000800001AC (0x5E02B0D7) vmv.v.i v1, 0x5                           test_1+8
[80] [U]: 0x00000000800001B0 (0x421022D7) vmv.x.s x5, v1                            test_1+12
[81] [U]: 0x00000000800001B4 (0x00500393) addi x7, x0, 0x5                          test_1+16
[82] [U]: 0x00000000800001B8 (0x06729463) bne x5, x7, 0x68                          test_1+20
...
SUCCESS

- ISS
> $ ./do test:riscv_vector_tests CONFIG=rv64 BUILD_TYPE=DEBUG IGNOREUNDEFINED=YES JOBS=8
...
/home/adingank/git/forks/riscv-unified-db/gen/cpp_hart_gen/rv64_Debug/build/iss -m rv64 -c /home/adingank/git/forks/riscv-unified-db/cfgs/rv64-vector.yaml tests/isa/rv64uv-p-vmv_v_i
...
PC 800001a0 mret
PC 800001a4 addi x3, x0, 1
R x0 0
R= x3 1
PC 800001a8 vsetivli x0, 2, 24
R= x0 0
PC 800001ac vmv.v.i 1, 5
[...]

PR #1886, opened 2026-07-01
fix(cpp_hart_gen): vector register writes silently dropped to 0 (runtime bit_insert)

ishaan-arora-1:
Vector register writes on the ISS were silently dropping to `0`. For example `vle8.v v2, (x5)` (unit-stride load of 8-bit elements from the address in `x5` into `v2`) left `v2 = 0` instead of the loaded bytes; `vmv.v.i v1, 5` (broadcast the sign-extended immediate `5` into every element of `v1`) wrote the wrong value; and `vadd.vv vd, vs2, vs1` failed the same way. Scalar loads like `lb`/`lw` were fine, which is what made it puzzling at first.

Caused by #1813, which changed vector register width from compile-time-known to the weaker runtime-known.

Concretely, once `V` became a runtime-width register file, every `V[vd][msb:lsb] = value` write started going through the runtime overload of `bit_insert()` in `util.hpp`, which builds its field mask like:

```c++
BitsClass<T + 1, Signed>{1_b} << (msb - lsb + 1)
```

The problem stems from `{1_b}` having a runtime width of 1: for runtime-width Bits, a non-widening left shift by an amount `>=` the operand's width returns `0` (the bits fall off the end), so `1 << 8` becomes `0` and the whole mask collapses to `0`. `BitsClass<T>{value} << lsb` collapses the same way. With a zero mask the write is a no-op. Fixed-width `X` registers are `PossiblyUnknownBits<MXLEN>`, so their shift width is the compile-time template width and never collapses — which is why scalar worked and vector didn't.

The fix materializes the `1` and the value at the register's own (runtime) width before shifting, so nothing collapses:

```c++
BitsClass<T, Signed> zero  = target ^ target;   // zero, same width as the register
BitsClass<T, Signed> one   = zero | 1_b;        // 1,     same width
BitsClass<T, Signed> field = zero | value;      // value, same width
BitsClass<T, Signed> mask  = ((one << (msb - lsb + 1)) - one) << lsb;
target = (target & ~mask) | ((field << lsb) & mask);
```

This works for both the fixed-width and runtime-width instantiations, so `X` behaves exactly as before.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backends/cpp_hart_gen/cpp/include/udb/util.hpp
+++ b/backends/cpp_hart_gen/cpp/include/udb/util.hpp
@@ -48,8 +48,17 @@ namespace udb {
   >
   void bit_insert(BitsClass<T, Signed> &target, const MsbType &msb, const LsbType &lsb,
                   const ValueType &value) {
-    BitsClass<T, Signed> mask = ((BitsClass<T + 1, Signed>{1_b} << (msb - lsb + _Bits<1, false>{1_b})) - 1_b) << lsb;
-    target = (target & ~mask) | ((BitsClass<T, Signed>{value} << lsb) & mask);
+    // Derive the "1" and the value from `target` so that they carry the register's
+    // full (runtime) width. For runtime-width Bits (e.g. vector registers wider than 64
+    // bits) a non-widening left shift by an amount >= the operand's width yields 0; building
+    // the mask from a 1-bit literal (BitsClass<T+1>{1_b}, width 1) would therefore collapse
+    // the mask — and likewise `{value} << lsb` would collapse — to 0. The final assignment
+    // re-masks `target` to its own width, so intermediate widths are harmless.
+    BitsClass<T, Signed> zero = target ^ target;
+    BitsClass<T, Signed> one = zero | _Bits<1, false>{1_b};
+    BitsClass<T, Signed> field = zero | value;
+    BitsClass<T, Signed> mask = ((one << (msb - lsb + _Bits<1, false>{1_b})) - one) << lsb;
+    target = (target & ~mask) | ((field << lsb) & mask);
   }
 
   template <unsigned FirstExtendedBit, unsigned ResultWidth,


ThinkOpenly:
> What this fixes

You can remove all of these headers.

> `vle8.v` left `v2 = 0`

Can you add more context here? Give the full instruction including operands, so we know why "v2" is significant.

> `vmv.v.i` wrote garbage, `vadd.vv` failed

Here, too.

> A bit of history so this isn't confused with #1618: that issue (the old `vmv.v.i` `msb >= lsb` assertion) was already fixed back in February by #1677. This is a _different_, later regression — it showed up after #1813 ("Use regfile YAMLs for F, V") moved the vector registers to a runtime-width register file. Ajit and Henrik hit it recently while re-running the vector tests.

You could probably shorten this to something like "Caused by #1813, which changed vector register width from compile-time-known to weaker runtime-known."

> The problem is the seed: `{1_b}` gives a runtime Bits of width **1**.

I wouldn't call this a seed, personally. "The problem stems from `{1_b}` having a runtime width of 1."

> And for runtime-width Bits, a non-widening left shift by an amount `>=` the operand's width just returns `0` (bits are shifted off the end).

Another way to look at the problem is that the shift should be widening. Can we use a widening shift here?

> Seed the `1` and the value from `target` itself, so they carry the register's full (runtime) width instead of width 1:
> 
> ```c++
> BitsClass<T, Signed> zero  = target ^ target;
> BitsClass<T, Signed> one   = zero | 1_b;
> BitsClass<T, Signed> field = zero | value;
> ```

These seem hacky to me, but I'm not very familiar with how the BitsClass works.

> ```c++
> BitsClass<T, Signed> mask  = ((one << (msb - lsb + 1)) - one) << lsb;
> target = (target & ~mask) | ((field << lsb) & mask);
> ```
> 
> The final assignment re-masks `target` to its own width (`operator=` + `apply_mask`), so any intermediate width growth is harmless.

Where would "width growth" come from?

> ## Verification
> Built the generated rv64 ISS (`rv64-vector`, VLEN=128) and ran the vector tests before/after:
> 
> * **Before:** `vle8.v` → `v2 = 0`, `lw x2 = ffffffffffff0000`; `vmv_v_i` and `vadd.vv` FAIL.
> * **After:** `vle8`, `vmv_v_i`, `vadd.vv`, `vsetivli`, `vsetvl`, `vsetvli_rs1_eq_zero`, `vsetvli_vl_lt_vlmax` all SUCCESS; `lw x2 = ffffffffffff2d2c` (the expected value).
> * Spot-checked 12 scalar `rv64ui`/`rv64um` tests — all pass, no regression on the `X` path.
> 
> Also confirmed it in isolation with a tiny `bit_insert` test: runtime-width writes fail before the change and pass after, while the fixed-width case is unchanged.

Put this content in a follow-on comment instead of the PR description so it doesn't end up in the commit message, please.

> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

This can stay in the PR description, but remove the emoji, please.

I'd like @henrikg-qc to review (I presume he's more familiar with this code. :-) )

ishaan-arora-1:
> > What this fixes
> 
> You can remove all of these headers.
> 
> > `vle8.v` left `v2 = 0`
> 
> Can you add more context here? Give the full instruction including operands, so we know why "v2" is significant.
> 
> > `vmv.v.i` wrote garbage, `vadd.vv` failed
> 
> Here, too.
> 
> > A bit of history so this isn't confused with #1618: that issue (the old `vmv.v.i` `msb >= lsb` assertion) was already fixed back in February by #1677. This is a _different_, later regression — it showed up after #1813 ("Use regfile YAMLs for F, V") moved the vector registers to a runtime-width register file. Ajit and Henrik hit it recently while re-running the vector tests.
> 
> You could probably shorten this to something like "Caused by #1813, which changed vector register width from compile-time-known to weaker runtime-known."
> 
> > The problem is the seed: `{1_b}` gives a runtime Bits of width **1**.
> 
> I wouldn't call this a seed, personally. "The problem stems from `{1_b}` having a runtime width of 1."
> 
> > And for runtime-width Bits, a non-widening left shift by an amount `>=` the operand's width just returns `0` (bits are shifted off the end).
> 
> Another way to look at the problem is that the shift should be widening. Can we use a widening shift here?
> 
> > Seed the `1` and the value from `target` itself, so they carry the register's full (runtime) width instead of width 1:
> > ```c++
> > BitsClass<T, Signed> zero  = target ^ target;
> > BitsClass<T, Signed> one   = zero | 1_b;
> > BitsClass<T, Signed> field = zero | value;
> > ```
> 
> These seem hacky to me, but I'm not very familiar with how the BitsClass works.
> 
> > ```c++
> > BitsClass<T, Signed> mask  = ((one << (msb - lsb + 1)) - one) << lsb;
> > target = (target & ~mask) | ((field << lsb) & mask);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The final assignment re-masks `target` to its own width (`operator=` + `apply_mask`), so any intermediate width growth is harmless.
> 
> Where would "width growth" come from?
> 
> > ## Verification
> > Built the generated rv64 ISS (`rv64-vector`, VLEN=128) and ran the vector tests before/after:
> > 
> > * **Before:** `vle8.v` → `v2 = 0`, `lw x2 = ffffffffffff0000`; `vmv_v_i` and `vadd.vv` FAIL.
> > * **After:** `vle8`, `vmv_v_i`, `vadd.vv`, `vsetivli`, `vsetvl`, `vsetvli_rs1_eq_zero`, `vsetvli_vl_lt_vlmax` all SUCCESS; `lw x2 = ffffffffffff2d2c` (the expected value).
> > * Spot-checked 12 scalar `rv64ui`/`rv64um` tests — all pass, no regression on the `X` path.
> > 
> > Also confirmed it in isolation with a tiny `bit_insert` test: runtime-width writes fail before the change and pass after, while the fixed-width case is unchanged.
> 
> Put this content in a follow-on comment instead of the PR description so it doesn't end up in the commit message, please.
> 
> > 🤖 Generated with [Claude Code](https://claude.com/claude-code)
> 
> This can stay in the PR description, but remove the emoji, please.
> 
> I'd like Henrik Gustafsson (@henrikg-qc) to review (I presume he's more familiar with this code. :-) )

Thanks Paul, that all makes sense. I've updated the description along those lines.

On the widening shift idea, I gave it a try. It does fix the collapse, but it ends up wiping the register's upper bits. The mask you get from the widening shift is only as wide as the field, so when you invert it for target & ~mask it doesn't cover the rest of the register and those bits get cleared. I checked with a small test: if I set v[127:64] to 0xAAAA... and then write 0x2C into v[7:0], the widening version zeros out the top half, while the version in the PR leaves it alone. So the mask really needs to be the full register width. 

On the constructs looking hacky, that's fair. All I'm doing is getting the 1 and the value to the register's width before shifting. target ^ target is just a zero at that width, and the OR promotes them to it. I looked for a cleaner way to make a same-width zero on the Bits type and couldn't find one, but if you or Henrik know a nicer idiom I'm glad to switch to it.

On the width growth, that comes from the subtraction, which returns a value one bit wider. It doesn't matter in the end because the final assignment to target goes back through the register's own width and re-masks, so the extra bit gets dropped.

Happy to have @henrikg-qc  look it over too.

ishaan-arora-1:
Verification

I built the generated rv64 ISS (config rv64-vector, VLEN=128) and ran the vector tests before and after the change:

- Before: `vle8.v` left `v2 = 0`, and `lw x2` read back `ffffffffffff0000`. The `vle8` and `vmv_v_i` tests both failed.
- After: `vle8`, `vmv_v_i`, `vadd.vv`, `vsetivli`, `vsetvl`, `vsetvli_rs1_eq_zero`, and `vsetvli_vl_lt_vlmax` all pass, and `lw x2` reads `ffffffffffff2d2c`, which is the value the test expects.
- I also ran 12 scalar rv64ui/rv64um tests as a sanity check and they all still pass, so the fixed-width X path is unchanged.

I also tested `bit_insert` on its own to confirm both the runtime-width and fixed-width cases behave, including that writing a low element leaves the register's upper bits intact.

henrikg-qc (commented):
This works, just note that there is support for widening logical left shifts in the BITS type.
The mask could be built with a widening shift

henrikg-qc (approved):

PR #1889, opened 2026-07-06
fix(isa): reset vstart to zero after vle8.v and vse8.v

### LINKED ISSUE #1888: fix(isa): reset vstart to zero after vle8.v and vse8.v
## Problem

`vle8.v` and `vse8.v` are missing `CSR[vstart].VALUE = 0` at the end of their `operation()` bodies.

All 6 other implemented vector instructions already include this reset:
- `vadd.vv`, `vadd.vx`, `vadd.vi`, `vsetvli`, `vsetivli`, `vsetvl`

This is required by the spec: `spec/std/isa/csr/V/vstart.yaml` states:
> "All vector instructions are defined to reset the vstart CSR to zero at the end of execution"

## Fix

Add `CSR[vstart].VALUE = 0;` as the final line of the `operation()` body in both `spec/std/isa/inst/V/vle8.v.yaml` and `spec/std/isa/inst/V/vse8.v.yaml`.

## Why the fix is safe

The reset is placed after the element loop. If `read_memory` or `write_memory` raises an exception mid-loop, the instruction aborts and never reaches the reset line — `vstart` stays at element `i` for trap resumption, which is correct per spec. The reset only runs on clean completion.

## Testing

- `./do test:schema` ✓
- `./do test:idl CFG=_` ✓
- `./bin/regress --tag smoke` — 17/17 ✓

A PR is attached.

jpak-qc:
Closes #1888

`vle8.v` and `vse8.v` are missing `CSR[vstart].VALUE = 0` at the end of their `operation()` bodies

This behavior is required by spec (normative rule "vstart_update"):
> All vector instructions [...] reset the vstart CSR to zero at the end of execution.

The reset is placed after the element loop. If `read_memory` or `write_memory` raises an exception mid-loop, the instruction aborts and never reaches the reset line — `vstart` stays at element `i` for trap resumption, which is correct. The reset only runs on clean completion (normative rule "vstart_unmodified").

### DIFF
--- a/spec/std/isa/inst/V/vle8.v.yaml
+++ b/spec/std/isa/inst/V/vle8.v.yaml
@@ -36,3 +36,4 @@ operation(): |
     U32 end_bit_pos = start_bit_pos + 7;
     V[vd][end_bit_pos:start_bit_pos] = read_memory(8, virtual_address + i, $encoding)[7:0];
   }
+  CSR[vstart].VALUE = 0;
--- a/spec/std/isa/inst/V/vse8.v.yaml
+++ b/spec/std/isa/inst/V/vse8.v.yaml
@@ -36,3 +36,4 @@ operation(): |
     U32 end_bit_pos = start_bit_pos + 7;
     write_memory(8, virtual_address + i, V[vs3][end_bit_pos:start_bit_pos], $encoding);
   }
+  CSR[vstart].VALUE = 0;


ThinkOpenly:
_original PR description (because I'm about to edit it)_ ...
Closes #1888

## Summary

- `vle8.v` and `vse8.v` were missing `CSR[vstart].VALUE = 0` at the end of their `operation()` bodies
- All 6 other implemented vector instructions already include this reset
- Required by spec: `spec/std/isa/csr/V/vstart.yaml` states "All vector instructions are defined to reset the vstart CSR to zero at the end of execution"

## Why the fix is safe

The reset is placed after the element loop. If `read_memory` or `write_memory` raises an exception mid-loop, the instruction aborts and never reaches the reset line — `vstart` stays at element `i` for trap resumption, which is correct. The reset only runs on clean completion.

## Test plan

- [x] `./do test:schema` ✓
- [x] `./do test:idl CFG=_` ✓
- [x] `./bin/regress --tag smoke` — 17/17 ✓

ThinkOpenly (approved):
Thanks, @jpak-qc !

PR #1891, opened 2026-07-09
Update many dependencies

jordancarlin:
Before reenabling Renovate, manually update many dependencies to avoid a flood of PRs. Also fixes deprecation warnings raised by the newer packages.


jordancarlin:
It is not clear to me why code coverage claims to be dropping by more than 6% when this is not materially changing any code. @ThinkOpenly if you are ok with it, I would be in favor of merging this as is and we can look into resolving whatever is going on with code coverage in a later PR.

AI suspects it is because our code coverage tool recently started supporting branch coverage, so it appears as though coverage dropped because it is looking at a wider range of coverage.

PR #1932, opened 2026-07-12
fix(z3): implement anyOf for integer parameter constraints

### LINKED ISSUE #1890: anyOf/oneOf/noneOf/if-then-else unimplemented in Z3 parameter constraints
Udb::Z3ParameterTerm's integer constraint builder (constrain_int in tools/ruby-gems/udb/lib/udb/z3.rb) does not support the anyOf JSON Schema keyword. Only allOf is implemented for combinator keywords anyOf, oneOf, noneOf, and if/then/else all raise a RuntimeError instead of being translated into Z3 constraints. The same gap exists for boolean, string, and array constraints in the same file,only allOf is implemented there too.

On branch main,In a Ruby console (or test) with the udb gem loaded, construct a parameter schema that uses anyOf for an integer type, e.g 
schema = {
     "anyOf" => [
       { "const" => 1 },
       { "const" => 2 }
     ]
   }
   solver = Udb::Z3Solver.new
   Udb::Z3ParameterTerm.new("my_param", solver, schema)

This raises immediately instead of constraining the term to be 1 or 2.

Text output
RuntimeError: TODO: anyOf not yet implemented for integer constraints
from tools/ruby-gems/udb/lib/udb/z3.rb:393 

Additional context
The allOf keyword is already fully implemented and working for integer, boolean, string, and array constraints (see the surrounding code in the same method). anyOf/oneOf/noneOf/if simply raise "TODO: ... not yet implemented" in each of the four type-specific constraint builders:

integer: z3.rb:392-406
boolean: z3.rb:466,470,474,478
string: z3.rb:529,533,537,541
array: z3.rb:612,616,620,624

Any extension/parameter schema that legitimately uses one of these keywords (rather than only allOf) currently crashes the solver instead of resolving

I'd like to fix this, starting with anyOf for integer constraints mirroring the existing allOf pattern with a Z3 disjunction and have a working fix + tests.

Shaheen614:
This PR Implements 'anyOf' support in 'Z3ParameterTerm.constrain_int' ('tools/ruby-gems/udb/lib/udb/z3.rb'), mirroring the existing 'allOf' pattern. Each subschema under 'anyOf' is translated into its own constraint(s), conjoined if there's more than one (e.g. a subschema with both 'minimum' and 'maximum', then all subschemas are OR'd together so the term is valid if it satisfies at least one of them.

'oneOf', 'noneOf', 'if'/'then'/'else', and the boolean/string/array variants are still unimplemented,flagged in the linked issue as follow-up work, out of scope for this PR to keep it reviewable.

**Testing**

Added three tests to 'test/test_z3_parameter_constraints.rb':
- 'test_constrain_int_with_anyof_const' exact-value disjunction (only 1 or 2 allowed)
- 'test_constrain_int_with_anyof_ranges' range disjunction, including a value in the gap between ranges that must be unsatisfiable
- 'test_constrain_int_with_anyof_and_other_constraints 'anyOf' combined with a sibling 'minimum', confirming conjunction with surrounding constraints still holds

All existing tests in 'test_z3_parameter_constraints.rb', 'test_z3_extensions.rb', and 'test_z3_finite_array.rb' still pass.

Ran rubocop on both changed files,clean except one pre-existing, unrelated offense (Style/ClassMethodsDefinitions at line number 1087) that predates this change and is outside the scope of this PR.

Closes #1890


AlgoArtist06:
Superseded by #2019, which now implements `anyOf` for integer, boolean, and string through a single shared `any_of_constraint` helper. @ThinkOpenly preferred that implementation in [this comment](https://github.com/riscv/riscv-unified-db/pull/2019#issuecomment-5078917188), and keeping both would have left two different idioms for the same disjunction in one file.

The three integer tests from this PR (const, ranges, and `anyOf` with a sibling `minimum`) were carried over intact. #2019 additionally fixes a nested-`allOf` leak on the integer path that this implementation did not address: `constrain_int` recursed into `allOf` with the default `assert: true`, so a nested conjunction inside an `anyOf` branch escaped into the solver and made that branch mandatory.

Closing in favour of #2019. Note that `Closes #1890` here was too broad anyway, since the issue also covers arrays, `oneOf`, `noneOf`, and `if`/`then`/`else`.

PR #1949, opened 2026-07-13
Remove duplicate version information for UDB dependencies and allow Renovate to manage them

jordancarlin:
z3/eqntott/espresso/must had their versions hardcoded in multiple locations. This updates the scripts to use the `*_VERSION` file as the sole source of truth for the version. It also enables Renovate to manage the version going forward for automatic updates.


ThinkOpenly (approved):
I skimmed the changes admittedly, but seems like a nice improvement!

PR #1962, opened 2026-07-13
Revert attempt to start using new instruction encoding schema

abkshaqster:
## Summary

This change restores the legacy `encoding` representation for the affected ISA instruction YAML files under `spec/std/isa`.

The previous migration replaced the `encoding` block with the newer `format` representation. This PR restores the original `encoding` representation while preserving all subsequent updates

The intent is to restore the original encoding format without reverting unrelated improvements introduced after the migration.

## Scope

- Replaced `format` with `encoding` for the affected instruction YAML files.
- Kept the current operand naming (`xd`, `xs1`, `xs2`) and existing instruction behavior unchanged.
- No functional changes to instruction semantics are intended.

## Notes

This is a targeted restoration of the encoding representation only and is not a full revert of the original migration.

### DIFF
--- a/spec/std/isa/inst/B/andn.yaml
+++ b/spec/std/isa/inst/B/andn.yaml
@@ -16,17 +16,15 @@ definedBy:
       - name: Zbb
       - name: Zbkb
 assembly: xd, xs1, xs2
-format:
-  $inherits:
-    - inst_subtype/R/R-x.yaml#/data
-  opcodes:
-    funct7:
-      display_name: ANDN
-      value: 0b0100000
-    funct3:
-      display_name: ANDN
-      value: 0b111
-    opcode: { $inherits: inst_opcode/OP.yaml#/data }
+encoding:
+  match: 0100000----------111-----0110011
+  variables:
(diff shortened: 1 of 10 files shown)


jordancarlin:
- Please update the PR title. This restores the encoding, not the assembly format.
- Should we consider removing all of the YAML files related to the "format" approach that we are no longer using? @ThinkOpenly any thoughts on switching over to it later vs keeping the plain encoding long term?

ThinkOpenly:
> Should we consider removing all of the YAML files related to the "format" approach that we are no longer using? Paul Clarke (@ThinkOpenly) any thoughts on switching over to it later vs keeping the plain encoding long term?

I'd prefer to leave them for now. We need to do _something_ for instruction formats eventually, maybe something close to what was attempted earlier. Not ideal, I know.

jordancarlin (approved):

ThinkOpenly (approved):

ThinkOpenly:
I'm not waiting for codecov to return for this PR. Merging...

PR #1967, opened 2026-07-13
restore default values for parameters lost in PR #891

### LINKED ISSUE #704: Add "recommended" values for extension parameters

AnshulPatil2005:
Summary

`MSTATEEN_CONTEXT_TYPE` and `HSTATEEN_CONTEXT_TYPE` are missing the `default: rw`
value that was originally added in #926 to resolve #704. It was silently dropped
when parameters were migrated from inline extension YAML into standalone
`spec/std/isa/param/*.yaml` files 

This restores it.


ThinkOpenly:
Looks like there were a few more `default` values dropped by #891 as well. Would you like to include those here?
- `PRECISE_SYNCHRONOUS_EXCEPTIONS` (`true`)
- `TRAP_ON_SFENCE_VMA_WHEN_SATP_MODE_IS_READ_ONLY` (`false`)
- `TRAP_ON_ECALL_FROM_M` (`true`)

AnshulPatil2005:
> Looks like there were a few more `default` values dropped by #891 as well. Would you like to include those here?
> 
> * `PRECISE_SYNCHRONOUS_EXCEPTIONS` (`true`)
> * `TRAP_ON_SFENCE_VMA_WHEN_SATP_MODE_IS_READ_ONLY` (`false`)
> * `TRAP_ON_ECALL_FROM_M` (`true`)

added all 3. While I was at it, I also ran a full scan across every commit that ever touched a default: value under spec/std/isa/param/ and spec/std/isa/ext/, to make sure there weren't others from the same #891 migration. Turned up 3 more with the identical issue: TRAP_ON_EBREAK (true), MTVEC_BASE_ALIGNMENT_DIRECT (4), MTVEC_BASE_ALIGNMENT_VECTORED (4), added those too.

ThinkOpenly (approved):

PR #1831, opened 2026-05-13
LFX - Phase 7: Final Parameter Spreadsheet

ishaan-arora-1:
## Summary

- Add `generate_spreadsheet.py` — consolidates the V2 deduplicated LLM extraction (Phase 6, #1793) with the UDB ground truth (Phase 1, #1765) and the LLM↔UDB alignment (Phase 5/6, #1792, #1793) into a single authoritative spreadsheet of every confirmed architectural parameter
- Produces `parameters.csv` (programmatic use) and `parameters.xlsx` (presentation / review) with the 9 columns specified in #1753, plus a human-readable `parameters_stats.txt` summary
- Adds `openpyxl` as a project dependency for XLSX output

Closes #1753. Builds on #1765, #1766, #1783, #1791, #1792, #1793.

## Columns

| Column | Source |
|---|---|
| `adoc_file` | LLM result `_source_file` |
| `line_number` | LLM result `line_number` |
| `excerpt` | Verbatim spec text |
| `parameter_name` | UDB name when aligned, else normalized ALL_CAPS_WITH_UNDERSCORES |
| `named` | `yes` if already in UDB, `no` otherwise |
| `class` | LLM classification (`NORM_DIRECT`, `NORM_CSR_WARL`, `NORM_CSR_RW`, `SW_RULE`, `NON_ISA`, …) |
| `value_type` | `binary`, `enum`, `range`, `set`, `bitmask`, `value` |
| `confidence` | `high` / `medium` (low filtered out per Phase 8 plan) |
| `notes` | Reviewer notes (seeded with alignment path for non-exact matches) |

## Resolution Rules

For each deduplicated V2 result row:
1. `existing_udb_name` set & in `udb_param_names.txt` → use UDB name, `named=yes`
2. Alignment exact match → UDB name, `named=yes`
3. Alignment `one_to_many` / `concept_group` / `fuzzy_name` → UDB name, `named=yes`, `notes` seeded with the alignment path for reviewer context
4. Otherwise → normalized LLM name, `named=no`, `notes='newly discovered'`

## Results

| Metric | Value |
|---|---|
| Total rows (confidence ≥ medium) | **330** |
| Already named in UDB | 97 |
| Newly discovered | 233 |
| Spec files touched | 43 |

### Classification breakdown

| Class | Count |
|---|---|
| `NORM_DIRECT` | 208 |
| `NORM_CSR_RW` | 63 |
| `NORM_CSR_WARL` | 52 |
| `NON_ISA` | 5 |
[...]

PR #1968, opened 2026-07-13
add Zbs reserved-encoding parameters for RV32

AnshulPatil2005:
Adds 4 new architectural parameters for the Zbs single-bit instructions
(`bclri`, `bexti`, `binvi`, `bseti`): on RV32, the `shamt[5]=1` encoding
of each is reserved, and whether executing it traps as illegal or
produces undefined behavior wasn't previously captured as UDB data.

- `BCLRI_SHAMT_RSV_RV32`
- `BEXTI_SHAMT_RSV_RV32`
- `BINVI_SHAMT_RSV_RV32`
- `BSETI_SHAMT_RSV_RV32`

Each follows the existing `TRAP_ON_RESERVED_INSTRUCTION` pattern
(boolean: true = IllegalInstruction exception, false = UNDEFINED_LEGAL).

These came from the LFX parameter-extraction project's Phase 7
spreadsheet (#1831), which flagged them as newly-discovered parameters
with no existing YAML. I verified each excerpt directly against current
spec text rather than trusting the spreadsheet as-is

This is 4 of 259 newly-discovered parameters the pipeline flagged
across the spec (250 of which are independently confirmed the
excerpt was successfully located and tagged in spec text by the
pipeline's own tooling, not just self-reported LLM confidence)


AnshulPatil2005:
Happy to keep working through the rest of the confirmed list (250 more across the spec) if useful.

let me know.

jordancarlin:
Shouldn't this be captured by the existing `TRAP_ON_RESERVED_INSTRUCTION` parameter? Having unique parameters for every instruction that has reserved fields feels excessive and will make the config files very verbose.

AnshulPatil2005:
> Shouldn't this be captured by the existing `TRAP_ON_RESERVED_INSTRUCTION` parameter? Having unique parameters for every instruction that has reserved fields feels excessive and will make the config files very verbose.

Fair point, though I don't think TRAP_ON_RESERVED_INSTRUCTION technically fits that one's about a completely unassigned opcode, this is a reserved field value in an instruction that's fully defined. UDB draws that same distinction elsewhere (e.g. VILL_SET_ON_RESERVED_VTYPE).

That said, one parameter per instruction is overkill the identical also applies to rori. I'll consolidate into a single shared parameter covering all of them.

AnshulPatil2005:
@jordancarlin can you re-review the changes

jordancarlin:
I still think this parameter is unnecessary. `TRAP_ON_RESERVED_INSTRUCTION` doesn't say anything about only applying to completely unassigned opcode spaces. My interpretation is that it applies to any instruction that is reserved when considering the full encoding of the instruction. `VILL_SET_ON_RESERVED_VTYPE` is different because the value can come from a register instead of the instruction encoding. Unless somebody actually implements a core where other instructions trap and these don't, I would be hesitant to add this because I think it will lead to dozens of additional parameters.

PR #1971, opened 2026-07-14
Import AGENTS.md in CLAUDE.md

jordancarlin:
Claude Code is the only AI agent (that I know of) that does not read AGENTS.md.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,1 +1,5 @@
-AGENTS.md
+<!--
+Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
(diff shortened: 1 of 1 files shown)


ThinkOpenly:
> Claude Code is the only AI agent (that I know of) that does not read AGENTS.md.

Should we create a symlink, or an actual file `CLAUDE.md` that imports `AGENTS.md`:
```
@AGENTS.md
```
Either probably works in anything supported here. Windows doesn't support symlinks (to my knowledge).

jordancarlin:
Forgot about Windows and the symlink issue. I have been running everything through Linux. Happy to switch to import instead if you think that would be more robust.

ThinkOpenly:
> Happy to switch to import instead if you think that would be more robust.

If the result is the same, let's go with import.

jordancarlin:
Updated.

ThinkOpenly (approved):

PR #1973, opened 2026-07-14
Port Pages schema and GitHub pages publishing to Python

### LINKED ISSUE #1787: GitHub Pages Deployment failing

jordancarlin:
Replace the Ruby/ERB tooling for GitHub Pages and schema publishing with Python scripts. The Pages landing page now uses a tokenized HTML template instead of ERB, and both the schema index generation and release publishing workflows use Python.

Enhancements over the old Ruby versions:
- Escape branch, commit, schema, and version values before inserting them into generated HTML.
- URL-quote schema link path components so unusual schema names or versions produce valid links.
- Validate template tokens so missing replacements fail generation instead of publishing unresolved placeholders.
- Add configurable deploy, template, and output paths for local and CI use.
- Keep schema release publishing structured around explicit helpers while preserving existing behavior for creating per-schema-version releases and skipping unchanged assets.

The existing schema workflow seems to have failed CI every time it ran (I couldn't find a passing case in the logs and there are no published schema) and that was causing the GitHub Pages deploy workflow to fail for the past few months. They were failing because of some Ruby tools/Gem related issue, so rather than trying to fix the Ruby environment (which I don't know anything about), I ported them to Python. Lots of AI help with this one.

Fixes #1787


tools/scripts/gen_pages_index.py
```diff
@@ -0,0 +1,119 @@
...
+
+    schema_index = json.loads(schema_index_path.read_text(encoding="utf-8"))
+    schemas = schema_index.get("schemas", {})
+    if not isinstance(schemas, dict) or not schemas:
+        return "          <p><em>No schema versions published yet.</em></p>\n"
+
+    html_lines: list[str] = []
+    for schema_name, versions in sorted(schemas.items()):
+        if not isinstance(versions, list):
+            continue
```
ThinkOpenly: Is this something that should be reported as a warning or error?

tools/scripts/gen_pages_index.py
```diff
@@ -0,0 +1,119 @@
...
+
+        version_names = sorted({str(version) for version in versions}, reverse=True)
+        if not version_names:
+            continue
```
ThinkOpenly: same

tools/scripts/gen_schema_index.py
```diff
@@ -0,0 +1,71 @@
...
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("schemas_dir", type=Path)
+    parser.add_argument("output_json", type=Path)
+    args = parser.parse_args()
+    return args.schemas_dir, args.output_json
+
+
+def load_schemas(index_path: Path) -> SchemaIndex:
+    if not index_path.is_file():
+        return {}
```
ThinkOpenly, line 30: same

tools/scripts/gen_schema_index.py
```diff
@@ -0,0 +1,71 @@
...
+
+    data: dict[str, SchemaIndex] = json.loads(index_path.read_text(encoding="utf-8"))
+    return data["schemas"]
+
+
+def find_schemas(schemas_dir: Path) -> SchemaIndex:
+    if not schemas_dir.is_dir():
+        return {}
```
ThinkOpenly, line 38: same

ThinkOpenly (changes requested):

tools/scripts/gen_pages_index.py
jordancarlin: Yes. Updated.

tools/scripts/gen_pages_index.py
jordancarlin: Yes. Updated.

tools/scripts/gen_schema_index.py
jordancarlin: I think we should leave this one. This allows the GitHub Pages site to be published even if there are no schemas currently released (which is the case right now).

tools/scripts/gen_schema_index.py
jordancarlin: I think we should leave this one. This allows the GitHub Pages site to be published even if there are no schemas currently released (which is the case right now).

ThinkOpenly (approved):

PR #1974, opened 2026-07-14
Add directed fp unit tests, and fix bugs found

drhower:
Migrated from #1811 (opened under the wrong GitHub account). Closing that one in favor of this PR; addressing the remaining unresolved Copilot review comments from #1811 here.


backends/cpp_hart_gen/CMakeLists.txt
```diff
@@ -101,6 +101,15 @@ FetchContent_Declare(
...
 FetchContent_MakeAvailable(Catch2)
 
+# nlohmann_json is header-only; fetched here so its headers are available to
+# hart (transitively needed by nlohmann_json_schema_validator headers).
+FetchContent_Declare(
+  nlohmann_json
+  GIT_REPOSITORY https://github.com/nlohmann/json.git
+  GIT_TAG        v3.11.3
+)
+FetchContent_MakeAvailable(nlohmann_json)
```
jordancarlin, line 111: If this is included transitively already, do we need to add it here? This will now compile it twice as is.

backends/cpp_hart_gen/CMakeLists.txt
```diff
@@ -116,6 +125,39 @@ FetchContent_Declare(
...
+
+ExternalProject_Get_Property(berkeley_softfloat SOURCE_DIR)
+file(MAKE_DIRECTORY "${SOURCE_DIR}/source/include") # Prevent CMake error before it is cloned
+
+add_library(libsoftfloat_static STATIC IMPORTED)
+set_target_properties(libsoftfloat_static PROPERTIES
+  IMPORTED_LOCATION             "${SOURCE_DIR}/build/${SOFTFLOAT_BUILD_SPEC}/softfloat.a"
+  INTERFACE_INCLUDE_DIRECTORIES "${SOURCE_DIR}/source/include"
+)
+add_dependencies(libsoftfloat_static berkeley_softfloat)
```
jordancarlin, line 159: If this is only used for testing, it should be declared as a dependency of the main build.

jordancarlin (changes requested):
I only reviewed the build system so far. I think we should switch to https://github.com/cover-float/cover-float for our test vectors eventually.

backends/cpp_hart_gen/CMakeLists.txt
henrikg-qc: This becomes a makefile target that is not included in the build:iss rake target. 
It is included in build:cpp_hart target

If we are to change these tests building also consider the others called from the rakefile
 Dir.chdir "#{CPP_HART_GEN_DST}/#{build_name}/build" do
      **sh "make -j #{$jobs} test_bits_directed"
      sh "make -j #{$jobs} test_bits_random"
      sh "make -j #{$jobs} test_softfloat_fp"
      sh "make -j #{$jobs} test_regfile"
      sh "ctest -T coverage -T test"**
    end

henrikg-qc (approved):

jordancarlin (approved):
Deferring the build changes for now.

PR #1980, opened 2026-07-15
fix: correct InterruptCode type references in exception_code.rb

### LINKED ISSUE #1979: InterruptCode#<=> and InterruptCode#hash reference ExceptionCode instead of InterruptCode
In tools/ruby-gems/udb/lib/udb/obj/exception_code.rb, the InterruptCode class has two copy-paste errors where it references ExceptionCode instead of itself:

- **Line 85** (InterruptCode#<=>):
  `return nil unless T.cast(other, Object).is_a?(ExceptionCode)`
  Should be `is_a?(InterruptCode)`

- **Line 96** (InterruptCode#hash):
  `def hash = [ExceptionCode, num].hash`
  Should be `[InterruptCode, num].hash`

**Impact:** InterruptCode#<=> always returns nil when comparing two InterruptCode objects, breaking sorting. InterruptCode#hash collides with ExceptionCode objects sharing the same num value.

**Root cause:** Class was copy-pasted from ExceptionCode (lines 30-62) and two references were not updated.

**Proposed fix:** Change ExceptionCode to InterruptCode on lines 85 and 96.

AdeshDeshmukh:
Fix two copy-paste errors in InterruptCode where ExceptionCode was
used instead of InterruptCode.

Closes #1979

### DIFF
--- a/tools/ruby-gems/udb/lib/udb/obj/exception_code.rb
+++ b/tools/ruby-gems/udb/lib/udb/obj/exception_code.rb
@@ -82,9 +82,9 @@ module Udb
 
     sig { override.params(other: BasicObject).returns(T.nilable(Integer)) }
     def <=>(other)
-      return nil unless T.cast(other, Object).is_a?(ExceptionCode)
+      return nil unless T.cast(other, Object).is_a?(InterruptCode)
 
-      num <=> T.cast(other, ExceptionCode).num
+      num <=> T.cast(other, InterruptCode).num
     end
 
     sig { override.params(other: BasicObject).returns(T::Boolean) }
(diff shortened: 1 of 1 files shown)


ThinkOpenly (approved):

ThinkOpenly:
Can you add a "Signed-off-by" line, please? DCO was made mandatory in this repository a few weeks ago, which is a good thing, but does require a bit of effort for these commits which didn't make it in beforehand.

ThinkOpenly (changes requested):
Could you add a DCO ("Signed-off-by") to your commit here and force-push?

AdeshDeshmukh:
@ThinkOpenly Done!

ThinkOpenly (approved):

PR #1981, opened 2026-07-15
docs: improve installation verification instructions

2024itb047samata:
## Summary

Improve the installation verification section in `README.adoc` by adding guidance on verifying the development environment after running `bin/setup`.

## Changes

- Add a dedicated "Verify the installation" section.
- Document how contributors can verify that the setup completed successfully by running the wrapper scripts in `bin/`.
- Clarify what contributors should do if the verification commands fail.

## Why

While setting up the project, I noticed that the README explains how to install the development environment using `bin/setup`, but it does not explicitly describe how contributors can verify that the setup completed successfully. This change adds a simple verification step to improve the onboarding experience for new contributors.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -91,18 +91,16 @@ When it finishes, it will tell you how to activate mise in your shell profile so
 mise-managed tools are available in future terminal sessions, and remind you to use the
 wrappers in `bin/` when invoking any tool.
 
-==== Verify the installation
+=== Verify the installation
 
-After `bin/setup` completes, verify that the development environment has been configured successfully by running the available wrapper scripts from the `bin/` directory:
+After `bin/setup` completes, verify that the development environment has been configured successfully by running:
 
 [source,bash]
 ----
-./bin/regress -h
-./bin/generate -h
-./bin/chore -h
+./bin/doctor
 ----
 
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -91,6 +91,18 @@ When it finishes, it will tell you how to activate mise in your shell profile so
...
+After `bin/setup` completes, verify that the development environment has been configured successfully by running the available wrapper scripts from the `bin/` directory:
+
+[source,bash]
+----
+./bin/regress -h
+./bin/generate -h
+./bin/chore -h
+----
+
+Each command should display its help message without errors. If any command fails, re-run `bin/setup` and ensure that `mise` has been activated in your shell as instructed by the setup script.
```
jordancarlin: There is a `./bin/doctor` script that verifies everything is installed correctly. That is probably what we should be directing users to run.

jordancarlin (changes requested):

README.adoc
2024itb047samata: Thanks for the suggestion! That makes sense. I'll update the README to recommend ./bin/doctor as the verification step instead.

jordancarlin (approved):

PR #1983, opened 2026-07-15
fix: match C.NOP encoding with wildcards for imm fields

### LINKED ISSUE #1177: `c.nop` encoding is different than riscv-opcodes

shivamsingh-007:
## Summary

The C.NOP encoding match was fully constrained to \`0000000000000001\`, which
matches only the specific case where \`imm=0\`. Per the RISC-V specification,
C.NOP code points with \`imm \\u2260 0\` encode HINTs, so the encoding match
must wildcard the immediate bits to accept any immediate value.

## What changed

- \`spec/std/isa/inst/C/c.nop.yaml\`: Changed match from fully-constrained
  \`0000000000000001\` to \`000-00000-----01\` with a variable \`imm\` field at
  bit location \`12|6-2\` (sign-extended). This matches the encoding used by
  \`riscv-opcodes\`: \`000?00000?????01\`.
- Updated generated inst\\_table fixture to reflect the new encoding format.

## Why this approach

This encoding is a strict subset of the C.ADDI encoding with \`xd=x0\`. The
distinction between NOP (imm=0) and HINTs (imm\\u2260 0) is handled at the
disassembly/decoding layer, not at the encoding match level — matching
riscv-opcodes' approach.

## Testing

- Schema validation: all files validate against their JSON schema
- Encoding conflict check: PASSED (no conflicts)
- Profile config consistency: PASSED
- Smoke test suite: 17/17 passed
- Verified against riscv-opcodes encoding table

Fixes #1177


spec/std/isa/inst/C/c.nop.yaml
jordancarlin, line 1: This needs a pseudoinstruction added so that when the immediate is zero the immediate is not included in the assembly.

jordancarlin (changes requested):

shivamsingh-007:
Addressed: added pseudoinstruction so that when imm == 0, the assembly shows no operand (canonical NOP).

spec/std/isa/inst/C/c.nop.yaml
```diff
@@ -8,13 +8,23 @@ kind: instruction
...
 encoding:
-  match: "0000000000000001"
+  match: 000-00000-----01
+  variables:
+    - name: imm
+      location: 12|6-2
+      sign_extend: true
+pseudoinstructions:
+  - when: imm == 0
+    to: ""
```
jordancarlin: pseudoinstructions need the full string in the `to` field.

```suggestion
pseudoinstructions:
  - when: imm == 0
    to: "c.nop"
```

jordancarlin (changes requested):

shivamsingh-007:
Thanks @jordancarlin. Already fixed in commit 61fae6a5 - to: '' changed to to: c.nop. Resolving thread.

jordancarlin (approved):

shivamsingh-007:
CI/CD: golden file
regress-gen-instruction-appendix was failing because the instruction appendix golden file still had the old c.nop encoding (no imm operand). The encoding changes in this PR (wildcard imm fields) mean the generated output now shows \c.nop imm\ with decode variables.

Updated tests/golden/all_instructions.golden.adoc to match the current generated output:

Assembly: c.nop → c.nop imm
Encoding: now includes imm fields
Description: now shows full C.NOP semantics (HINTs, no state change)
Decode Variables: now has imm = sext({encoding[12], encoding[6:2]}, 6)

PR #1985, opened 2026-07-15
Update bin/doctor and bin/install_hooks to be bash 3.2 compatible

### LINKED ISSUE #1984: fix: bin/doctor and bin/install_hooks fail on macOS with system bash (3.2)
## Environment
- macOS (Apple Silicon / arm64)
- bash 3.2.57 (macOS system default at `/bin/bash`)

## Problem

Running `bin/doctor` on macOS with the system bash produces an immediate parse error before any checks can run:

```
bin/doctor: line 112: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
bin/doctor: line 274: conditional binary operator expected
bin/doctor: line 274: syntax error near `GITHUB_ACTIONS'
bin/doctor: line 274: `  if [[ -v GITHUB_ACTIONS ]]; then'
```

Root cause: `bin/doctor` uses `#!/bin/bash`, which on macOS resolves to bash 3.2. The script uses:
- `declare -A` (associative arrays, requires bash 4.0+) at line 112
- `[[ -v VAR ]]` (variable-set test, requires bash 4.2+) at line 274

`bin/install_hooks` has the same `[[ -v GITHUB_ACTIONS ]]` pattern and also fails silently on bash 3.2.

## Steps to Reproduce

On a stock macOS machine (no Homebrew bash installed):
```
/bin/bash bin/doctor
/bin/bash bin/install_hooks
```

## Expected Behavior

Either the scripts work correctly, or they emit a clear error message directing the user to install a newer bash.

## Context

This repo already has a `bin/bash` wrapper (routing through mise-managed bash) and commit 297bfa81 ("use env to select bash") previously fixed this exact issue for 12 other `bin/` scripts. The same pattern should apply here.

krrishverma1805-web:
Replace `declare -A` (bash 4.0+) and `[[ -v VAR ]]` (bash 4.2+) with Bash 3.2 compatible versions so `bin/doctor` and `bin/install_hooks` work with the native macOS Bash 3.2.

Closes #1984

### DIFF
--- a/bin/doctor
+++ b/bin/doctor
@@ -1,23 +1,5 @@
-#!/usr/bin/env bash
+#!/bin/bash
 # bin/doctor — check that the development environment is correctly set up.
-
-# Require bash 4.2+ for associative arrays (declare -A) and the [[ -v VAR ]]
-# variable-set test used throughout this script.
-# macOS ships bash 3.2 due to GPLv2 licensing constraints. Install a newer bash:
-#   brew install bash
-# Then re-run with: /opt/homebrew/bin/bash bin/doctor
-# Or add /opt/homebrew/bin to the front of your PATH.
-if [[ "${BASH_VERSINFO[0]}" -lt 4 ]] || \
-   { [[ "${BASH_VERSINFO[0]}" -eq 4 ]] && [[ "${BASH_VERSINFO[1]}" -lt 2 ]]; }; then
-  echo "ERROR: bin/doctor requires bash 4.2 or later." >&2
-  echo "  Your bash: ${BASH_VERSION} ($(command -v bash))" >&2
-  echo "" >&2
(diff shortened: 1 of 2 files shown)


jordancarlin (changes requested):
Rather than requiring a newer version of bash, can we modify the scripts to be compatible with Bash 3.2?

krrishverma1805-web:
Thanks for the feedback, @jordancarlin! I've updated the PR to remove the bash version check entirely and rewrote both scripts to be fully compatible with Bash 3.2 (replacing `declare -A` with a simple conditional mapping, and `[[ -v ... ]]` with `[[ -n ... ]]`). This preserves the original `#!/bin/bash` shebangs and works out-of-the-box on older bash versions.

jordancarlin (approved):

PR #1970, opened 2026-07-14
Port regress workflow generator to Python

jordancarlin:
Replace the Ruby regress workflow generator with a Python ruamel.yaml round-trip implementation so generated GitHub Actions workflows preserve template comments, including Renovate action version hints. Also removes extra quoting added to make Psych work.

Update chore, pre-commit, and docs references to use the Python generator, simplify the workflow template now that Psych quoting workarounds are gone, regenerate regress.yml, and allow Renovate to scan the generated workflow so manual updates are no longer needed.

PR #1987, opened 2026-07-15
add tree-sitter-based IDL formatter with pre-commit and CI integration

drhower:
- Add `bin/idl-format` using Topiary CLI for canonical IDL formatting
- Add pre-commit hook and CI job to enforce IDL formatting
- Add IDL syntax highlighting for highlight.js and Prism (docs)
- Add IDL style guide documentation
- Apply formatter to existing IDL/YAML spec files
- Wire up Rust/Cargo tooling via mise for Topiary dependency

---

This is a re-opening of #1842 from a different fork, with merge conflicts against `main` resolved (the branch had drifted, most notably conflicting with #1875's bitmanip off-by-one fixes and #1970's regress workflow generator port to Python). #1842 will be closed in favor of this PR.


.github/workflows/autofix.yaml
```diff
@@ -16,5 +16,10 @@ jobs:
           fetch-depth: 0
       - name: Set up mise environment
         uses: "./.github/actions/mise-setup"
+      - name: Cache tree-sitter-idl npm packages
+        uses: actions/cache@v4
+        with:
+          path: tools/node/tree-sitter-idl/node_modules
+          key: tree-sitter-idl-npm-${{ runner.os }}-${{ hashFiles('tools/node/tree-sitter-idl/package-lock.json') }}
```
jordancarlin: Why is `autofix.yaml` the only workflow that is caching these? Should it be part of the mise-setup action instead so all workflows get the caching?

.github/workflows/release_udb_deps.yml
jordancarlin, line 20: The paths that automatically trigger this workflow should be updated to cover changes to `tree-sitter-idl`.

jordancarlin: This needs to be updated to include tree-sitter-idl.

.github/workflows/release_udb_deps.yml
```diff
@@ -196,3 +197,40 @@ jobs:
           else
             ./bin/chore update must -n
           fi
+
+  build-tree-sitter-idl:
+    if: inputs.tool == 'tree-sitter-idl' || inputs.tool == 'all'
```
jordancarlin: This should use `contains(fromJson(needs.detect-tools.outputs.tools), 'tree-sitter-idl')` to be consistent with all of the other builds.

backends/cpp_hart_gen/lib/gen_cpp.rb
```diff
@@ -300,15 +300,6 @@ def gen_cpp(symtab, indent = 0, indent_spaces: 2)
...
   end
 
-  class MultiVariableAssignmentAst < AstNode
-    sig { override.params(symtab: SymbolTable, indent: Integer, indent_spaces: Integer).returns(String) }
-    def gen_cpp(symtab, indent = 2, indent_spaces: 2)
-      lhs = "std::tie(#{variables.map { |v| v.gen_cpp(symtab, 0, indent_spaces:) }.join(', ')})"
-      rhs = function_call.gen_cpp(symtab, 0, indent_spaces:)
-      "#{' ' * indent}#{lhs} = #{rhs}"
-    end
-  end
```
jordancarlin, line 310: Why is this changing cpp_hart_gen? How does this relate to tree-sitter?

tools/node/tree-sitter-idl/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+# SPDX-FileCopyrightText: Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+[package]
+name = "tree-sitter-idl"
+description = "Idl grammar for tree-sitter"
+version = "0.0.1"
+license = "MIT"
```
jordancarlin: This says license MIT but the file itself says BSD-3-Clause-Clear. Seems inconsistent.

tools/node/tree-sitter-idl/pyproject.toml
jordancarlin, line 1: Why does this directory have a Python project setup, a Rust project setup, a Node project setup, and a Swift project setup?

tools/node/tree-sitter-idl/TS_IDL_VERSION
jordancarlin, line 1: This version is manually encoded in all of the other project files in this directory. How do we ensure they are kept in sync?

tools/ruby-gems/idlc/lib/idlc.rb
```diff
@@ -78,17 +40,98 @@ def unset_pb
...
+      end
+    end
+
     def compile_file(path, source_mapper = nil)
       path_key = path.realpath.to_s
 
-      m = T.let(@@parse_cache[path_key], T.nilable(IsaSyntaxNode))
+      # Cache now stores IsaAst directly (built by TsAstBuilder).
+      # Reads without the lock are safe under MRI; writes are locked.
+      cached = T.let(@@parse_cache[path_key], T.nilable(IsaAst))
```
jordancarlin: Not familiar enough with this code to judge whether or not this matters, but my AI review said the following:

compile_file now caches and returns a shared mutable IsaAst
- The old parser cached IsaSyntaxNode and called to_ast on each compile, producing a fresh AST. This PR caches the built IsaAst directly, then mutates it during include expansion and later callers freeze/type-check it against config-specific symbol tables.
- Because @@parse_cache is class-wide, multiple CfgArch/compiler instances can now share the same mutable/frozen global AST across configurations. That risks stale symbols, already-expanded includes, and cross-config behavior.
- Suggested fix: cache source/tree/syntax only and build a fresh AST per compile_file, or deep-clone before include expansion/return. Add a regression test that compiles the same .isa twice and asserts independent AST objects/state.

bin/idl-format
```diff
@@ -0,0 +1,326 @@
...
+          path = "#{GRAMMAR_PATH}",
+        },
+      },
+    },
+  }
+NCL
+TOPIARY_CONFIG_FILE.flush
+
+TOPIARY_ENV = { "TOPIARY_LANGUAGE_DIR" => QUERY_DIR }.freeze
+TOPIARY_CMD = ["topiary", "--configuration", TOPIARY_CONFIG_FILE.path].freeze
```
jordancarlin: This should not be calling bare `topiary`. We should add a shim to the `bin` directory that goes through `mise exec -- ...`.

tools/ruby-gems/idlc/idlc.gemspec
jordancarlin, line 1: 3. Published gems appear to omit the formatter query/reflow assets
     - tools/ruby-gems/idlc/idlc.gemspec:23-40, tools/ruby-gems/tasks.rake:64-72, tools/ruby-gems/idlc/lib/idlc/formatter.rb:36-55
     - Formatter looks for lib/idlc/queries and lib/idlc/idl-reflow inside the installed gem, but those directories are not present under tools/ruby-gems/idlc/lib/idlc, and release:idlc:prepare only stages parser C sources/headers/version.
     - Result: released idlc/udb-gen installs will silently skip formatting/reflow even though templates now call Idl::Formatter.
     - Suggested fix: stage tools/node/tree-sitter-idl/queries and tools/node/idl-reflow into the idlc gem release dir and add a gem-install test that Idl::Formatter.format actually formats in an installed-gem context.

bin/setup
```diff
@@ -146,7 +146,33 @@ fi
...
+  Darwin) _TS_IDL_SOEXT=dylib ;;
+  *)      _TS_IDL_SOEXT=so ;;
+esac
+_TS_IDL_LIB="$ROOT/tools/node/tree-sitter-idl/libtree-sitter-idl.${_TS_IDL_SOEXT}"
+
+if [[ -f "$_TS_IDL_LIB" ]]; then
+  _ok "IDL tree-sitter parser already built"
+else
+  _run "Building libtree-sitter-idl.${_TS_IDL_SOEXT}..."
+  if ! make -C "$ROOT/tools/node/tree-sitter-idl" "libtree-sitter-idl.${_TS_IDL_SOEXT}"; then
```
jordancarlin: Should use the `bin/make` shim.

bin/chore
```diff
@@ -280,17 +319,17 @@ do_gen_regress() {
...
+do_gen_idl_highlight() {
+  echo "Generating IDL highlight artifacts from tree-sitter grammar"
   local out="$UDB_ROOT/tools/vscode/idl/syntaxes/idl.tmLanguage.json"
-  echo "Generating tools/vscode/idl/syntaxes/idl.tmLanguage.json"
   if [ $fail_on_change -eq 1 ]; then
     local sha_before
     sha_before=$(sha256sum "$out")
   fi
-  node "$UDB_ROOT/tools/node/idl-grammar-gen/index.js"
+  node "$UDB_ROOT/tools/node/idl-gen/index.js"
```
jordancarlin: Should use the `bin/node` shim.

bin/idl-format
```diff
@@ -0,0 +1,326 @@
...
+  else
+    ARGV.dup
+  end
+
+idl_files, yaml_files = files.partition { |f| f.match?(/\.(idl|isa)\z/) }
+
+any_changed = false
+
+# -------------------------------------------------------------------------
+# Format standalone .idl / .isa files
```
jordancarlin, line 90: `bin/idl-format -f` does not mirror normal formatting for standalone `.idl`/`.isa`: check mode only runs Topiary, while normal mode also runs the reflow pass (`bin/idl-format:93-130`). This can let CI pass even though `bin/idl-format` would still rewrite files.

jordancarlin (changes requested):
Here is a very high level review. I didn't review most of it in detail.

PR #1994, opened 2026-07-16
add in param ZAWRS_NTO_IS_NOP

Teemooooooooo:
Added in parameter ZAWRS_NTO_IS_NOP, which can configure to behave as a no-op instruction.

### DIFF
--- a/spec/std/isa/inst/Zawrs/wrs.nto.yaml
+++ b/spec/std/isa/inst/Zawrs/wrs.nto.yaml
@@ -73,5 +73,7 @@ operation(): |
     }
   }
 
-  # stall while the reservation set is valid and no interrupt is pending
-  wrs_nto();
+  if (!ZAWRS_NTO_IS_NOP) {
+    # stall while the reservation set is valid and no interrupt is pending
+    wrs_nto();
+  }
--- a/spec/std/isa/param/ZAWRS_NTO_IS_NOP.yaml
+++ b/spec/std/isa/param/ZAWRS_NTO_IS_NOP.yaml
@@ -1,0 +1,19 @@
+# Copyright (c) 2025-26 Harvey Mudd College
+# Author: Ellen Yu (ellyu@hmc.edu)
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../schemas/param_schema.json
+
+$schema: param_schema.json#
+kind: parameter
+name: ZAWRS_NTO_IS_NOP
+description: |
+  When true, WRS.NTO completes immediately, behaving as a no-op, rather than
+  stalling the hart.
+
+long_name: TODO
+schema:
+  type: boolean
+definedBy:
+  extension:
+    name: Zawrs


jordancarlin:
@henrikg-qc can you look into what needs to be done to wire this up to the ISS?

henrikg-qc:
Trusim implements wrs.nto as no-op today. For completeness I'll add contitional on ZAWRS_NTO_IS_NOP.

jordancarlin (approved):

ThinkOpenly:
Is there somewhere in the spec that says that an implementation can implement `WRS.NTO` as a no-op? I believe it was implemented as a callback to facilitate any implementation defined behavior, and that's basically what Trusim already did. So, why do we need this parameter?

jordancarlin:
I think the connection to the spec is the following chain:

From `wrs.nto`:
> WRS.NTO and WRS.STO instructions follow the rules of the WFI instruction for resuming execution on a locally enabled pending interrupt.

From `wfi`:
> Implementations are permitted to resume execution for any reason, even if an enabled interrupt has not become pending. Hence, a legal implementation is to simply implement the WFI instruction as a NOP.

Beyond the pure spec connection, sail-riscv already has this as a config option and the ACTs need to know whether or not `wrs.nto` acts as a nop for test generation purposes. While it is not strictly needed for Trusim, the UDB parameters are used beyond Trusim and those use cases need this parameter.

ThinkOpenly:
> > Implementations are permitted to resume execution for any reason, even if an enabled interrupt has not become pending. Hence, a legal implementation is to simply implement the WFI instruction as a NOP.
> 
> Beyond the pure spec connection, sail-riscv already has this as a config option and the ACTs need to know whether or not `wrs.nto` acts as a nop for test generation purposes. While it is not strictly needed for Trusim, the UDB parameters are used beyond Trusim and those use cases need this parameter.

Not urgent, but how does whatever WFI does impact the ACTs? How is `WFI==nop` testable? (I'm curious.) It seems to be in the category of "any reason", and I can't see how you'd test for that.

jordancarlin:
@Teemooooooooo would be best suited to answer that.

Teemooooooooo:
This parameter is not so much about testing the behavior of WFI and Zawrs instructions when they are no-ops (In fact, I don't think `WFI==nop` is tested since it is not required). 

The wfi parameter is needed because this instruction is tested with a timer interrupt that is set up with a delay so that it is fired during the execution of WFI instruction if WFI is not nop. The parameter and Sail configuration are used to ensure that Sail matches DUT behavior. 

This parameter (ZAWRS_NTO_IS_NOP) is more about ensuring that the resuming behavior (such as when illegal instruction exception or interrupts fire) are tested when `WRS.NTO !=nop`. For example, the spec states, "When the TW (Timeout Wait) bit in mstatus is set and WRS.NTO is executed in any privilege mode other than M mode, and it does not complete within an implementation-specific bounded time limit, the WRS.NTO instruction will cause an illegal-instruction exception." and "While stalled, an implementation is permitted to occasionally terminate the stall and complete execution for any reason." 
The test assumes that if ZAWRS_NTO_IS_NOP is false, illegal instruction exception will trigger eventually even if it terminates early for some cases. The test will repeat until illegal instruction exception triggers to test the timeout behavior. 
When `WRS.NTO ==nop`, no illegal instruction exception will happen since the execution completes immidiately.

ThinkOpenly:
> The test assumes that if ZAWRS_NTO_IS_NOP is false, illegal instruction exception will trigger eventually even if it terminates early for some cases. The test will repeat until illegal instruction exception triggers to test the timeout behavior.

Whoa. That seems exceedingly fragile. The phrase "for any reason" is infinitely broad.

What you really seek is something like "can be interrupted" (inclusive) versus "is nop" (exclusive).

PR #2009, opened 2026-07-17
Pmlen supported values

### LINKED ISSUE #2008: PMLEN definition
The PMLEN parameter is not aligned well with the spec.  The parameter specifies an arbitrary integer.
The spec indicates that PMLEN of 7 and/or 16 are supported.  Seems like the parameter should indicate
which of these values are supported.

https://github.com/riscv/riscv-unified-db/blob/main/spec/std/isa/param/PMLEN.yaml

AnshulPatil2005:
fixes #2008 
changed the schema from an unconstrained integer to an array of the values this implementation supports — type: array, items: {enum: [7, 16]}, minItems: 1, maxItems: 2, uniqueItems: true. I modeled this directly on the existing MTVEC_MODES.yaml pattern (same shape: "which of these standard options does this implementation support")


AnshulPatil2005:
@davidharrishmc Hi David opened a fix for this in #2008. Changed PMLEN to an array of supported values ([7, 16], 1-2 items, modeled on MTVEC_MODES), so an implementation can support one or both.

Wanted to check this matches what you meant by "indicate which of these values are supported" I read it as "an implementation may support one or both of the two standard options.

spec/std/isa/param/PMLEN.yaml
```diff
@@ -8,10 +8,32 @@ kind: parameter
...
+      PMLEN=XLEN-57 (i.e., PMLEN=7 on RV64)
+    16::
+      PMLEN=XLEN-48 (i.e., PMLEN=16 on RV64)
+
+  The current standard only supports these values (`zpm.adoc`, tag
+  `norm:pmlen_supported_values`, plus the mandatory PMM=00 disabled
+  setting); a setting is reserved to potentially support other nonzero
+  values in future standards. If only one value is given, `PMM` is
+  assumed to be read-only with that value. Otherwise, `PMM` is
+  read-write and software selects between the given values.
```
jordancarlin: Drop this note about future expansion.

jordancarlin (changes requested):
I think we need to split this into 3 separate parameters. One corresponding to Smmpm, one for Smnpm, and one for Ssnpm. The spec does not say that all three need to support the same set of values.

AnshulPatil2005:
> I think we need to split this into 3 separate parameters. One corresponding to Smmpm, one for Smnpm, and one for Ssnpm. The spec does not say that all three need to support the same set of values.

makes sense, Thanks

spec/std/isa/param/PMLEN_SMMPM.yaml
```diff
@@ -0,0 +1,39 @@
...
+      PMLEN=XLEN-57 (i.e., PMLEN=7 on RV64)
+    16::
+      PMLEN=XLEN-48 (i.e., PMLEN=16 on RV64)
+
+  The current standard only supports these values (`zpm.adoc`, tag
+  `norm:pmlen_supported_values`, plus the mandatory PMM=00 disabled
+  setting). If only one value is given, `PMM` is assumed to be
+  read-only with that value. Otherwise, `PMM` is read-write and
+  software selects between the given values.
+long_name: TODO
```
ThinkOpenly:
```suggestion
long_name: Pointer mask length for M mode
```

spec/std/isa/param/PMLEN_SMNPM.yaml
```diff
@@ -0,0 +1,40 @@
...
+      PMLEN=XLEN-57 (i.e., PMLEN=7 on RV64)
+    16::
+      PMLEN=XLEN-48 (i.e., PMLEN=16 on RV64)
+
+  The current standard only supports these values (`zpm.adoc`, tag
+  `norm:pmlen_supported_values`, plus the mandatory PMM=00 disabled
+  setting). If only one value is given, `PMM` is assumed to be
+  read-only with that value. Otherwise, `PMM` is read-write and
+  software selects between the given values.
+long_name: TODO
```
ThinkOpenly:
```suggestion
long_name: Pointer mask length for next mode below M mode
```

spec/std/isa/param/PMLEN_SSNPM.yaml
```diff
@@ -0,0 +1,40 @@
...
+      PMLEN=XLEN-57 (i.e., PMLEN=7 on RV64)
+    16::
+      PMLEN=XLEN-48 (i.e., PMLEN=16 on RV64)
+
+  The current standard only supports these values (`zpm.adoc`, tag
+  `norm:pmlen_supported_values`, plus the mandatory PMM=00 disabled
+  setting). If only one value is given, `PMM` is assumed to be
+  read-only with that value. Otherwise, `PMM` is read-write and
+  software selects between the given values.
+long_name: TODO
```
ThinkOpenly:
```suggestion
long_name: Pointer mask length for next mode below S-mode
```

ThinkOpenly (changes requested):

spec/std/isa/param/PMLEN_SMMPM.yaml
```diff
@@ -0,0 +1,39 @@
...
+    7::
+      PMLEN=XLEN-57 (i.e., PMLEN=7 on RV64)
+    16::
+      PMLEN=XLEN-48 (i.e., PMLEN=16 on RV64)
+
+  The current standard only supports these values (`zpm.adoc`, tag
+  `norm:pmlen_supported_values`, plus the mandatory PMM=00 disabled
+  setting). If only one value is given, `PMM` is assumed to be
+  read-only with that value. Otherwise, `PMM` is read-write and
+  software selects between the given values.
```
jordancarlin: This is not consistent with the implementation below. It says here that PMM=00 (disabled) is mandatory, but the parameter currently allows you to specify only 7 or only 16. If 0 is mandatory, maybe it should not be a valid value for the array? Just have 7 and 16 as the supported values and remove the sentence saying if only one value is given, PMM is read-only. Applies to all 3 of the parameters.

spec/std/isa/param/PMLEN_SMMPM.yaml
```diff
@@ -0,0 +1,39 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../schemas/param_schema.json
+
+$schema: param_schema.json#
+kind: parameter
+name: PMLEN_SMMPM
```
jordancarlin: What about naming these parameters `SUPPORTED_PMLEN_S**PM` instead? The current name makes it seem like they indicate what PMLEN itself is, but PMLEN is changeable at runtime and these really indicate what the legal values for it are.

jordancarlin (changes requested):

spec/std/isa/param/PMLEN_SMMPM.yaml
AnshulPatil2005: Yea, That makes sense

spec/std/isa/param/PMLEN_SMMPM.yaml
AnshulPatil2005: will do that

spec/std/isa/param/SUPPORTED_PMLEN_SMMPM.yaml
```diff
@@ -0,0 +1,36 @@
...
+  here.
+long_name: Pointer mask length for M mode
+schema:
+  type: array
+  items:
+    type: integer
+    enum: [7, 16]
+  minItems: 1
+  maxItems: 2
+  uniqueItems: true
```
ThinkOpenly, line 33: Another approach, instead of dropping the ability to specify 0 is to force it to be included, as was done with `HPM_COUNTER_EN`, like:
```suggestion
  type: array
  items:
    - const: 0
  additionalItems:
    type: integer
    enum: [7, 16]
  minItems: 1
  maxItems: 3
  uniqueItems: true
```

spec/std/isa/param/SUPPORTED_PMLEN_SMNPM.yaml
```diff
@@ -0,0 +1,38 @@
...
+  PMLEN=0) is mandatory and always available, so it is not enumerated
+  here.
+long_name: Pointer mask length for next mode below M mode
+schema:
+  type: array
+  items:
+    type: integer
+    enum: [7, 16]
+  minItems: 1
+  maxItems: 2
```
ThinkOpenly:
```suggestion
    - const: 0
  additionalItems:
    type: integer
    enum: [7, 16]
  minItems: 1
  maxItems: 3
```

spec/std/isa/param/SUPPORTED_PMLEN_SSNPM.yaml
```diff
@@ -0,0 +1,38 @@
...
+  PMLEN=0) is mandatory and always available, so it is not enumerated
+  here.
+long_name: Pointer mask length for next mode below S-mode
+schema:
+  type: array
+  items:
+    type: integer
+    enum: [7, 16]
+  minItems: 1
+  maxItems: 2
```
ThinkOpenly:
```suggestion
    - const: 0
  additionalItems:
    type: integer
    enum: [7, 16]
  minItems: 1
  maxItems: 3
```

spec/std/isa/param/SUPPORTED_PMLEN_SMMPM.yaml
```diff
@@ -0,0 +1,36 @@
...
+$schema: param_schema.json#
+kind: parameter
+name: SUPPORTED_PMLEN_SMMPM
+description: |
+  The `PMLEN` values supported by the `Smmpm` pointer masking facility
+  for M-mode, selected via the `PMM` field in `mseccfg`. `PMLEN` is the
+  number of high-order bits of an address that are masked.
+
+  Options:
+
```
ThinkOpenly, line 15:
```suggestion

    0::
      PMLEN=0 (pointer masking is disabled)
```

spec/std/isa/param/SUPPORTED_PMLEN_SMMPM.yaml
```diff
@@ -0,0 +1,36 @@
...
+    7::
+      PMLEN=XLEN-57 (i.e., PMLEN=7 on RV64)
+    16::
+      PMLEN=XLEN-48 (i.e., PMLEN=16 on RV64)
+
+  The current standard only supports these values (`zpm.adoc`, tag
+  `norm:pmlen_supported_values`). `PMM`=00 (pointer masking disabled,
+  PMLEN=0) is mandatory and always available, so it is not enumerated
+  here.
```
ThinkOpenly:
```suggestion
  PMLEN=0) is mandatory and always available, so it is mandatory
  here.
```

spec/std/isa/param/SUPPORTED_PMLEN_SMNPM.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+    7::
+      PMLEN=XLEN-57 (i.e., PMLEN=7 on RV64)
+    16::
+      PMLEN=XLEN-48 (i.e., PMLEN=16 on RV64)
+
+  The current standard only supports these values (`zpm.adoc`, tag
+  `norm:pmlen_supported_values`). `PMM`=00 (pointer masking disabled,
+  PMLEN=0) is mandatory and always available, so it is not enumerated
+  here.
```
ThinkOpenly:
```suggestion
  PMLEN=0) is mandatory and always available, so it is mandatory
  here.
```

spec/std/isa/param/SUPPORTED_PMLEN_SSNPM.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+    7::
+      PMLEN=XLEN-57 (i.e., PMLEN=7 on RV64)
+    16::
+      PMLEN=XLEN-48 (i.e., PMLEN=16 on RV64)
+
+  The current standard only supports these values (`zpm.adoc`, tag
+  `norm:pmlen_supported_values`). `PMM`=00 (pointer masking disabled,
+  PMLEN=0) is mandatory and always available, so it is not enumerated
+  here.
```
ThinkOpenly:
```suggestion
  PMLEN=0) is mandatory and always available, so it is mandatory
  here.
```

spec/std/isa/param/SUPPORTED_PMLEN_SMNPM.yaml
```diff
@@ -0,0 +1,38 @@
...
+  The `PMLEN` values supported by the `Smnpm` pointer masking facility
+  for the next privilege mode below M-mode (S/HS if S-mode is
+  implemented, or U-mode otherwise), selected via the `PMM` field in
+  `menvcfg`. `PMLEN` is the number of high-order bits of an address
+  that are masked.
+
+  Options:
+
+    7::
+      PMLEN=XLEN-57 (i.e., PMLEN=7 on RV64)
```
ThinkOpenly, line 21:
```suggestion
    0::
      PMLEN=0 (pointer masking is disabled)
    7::
      PMLEN=XLEN-57 (i.e., PMLEN=7 on RV64)
```

spec/std/isa/param/SUPPORTED_PMLEN_SSNPM.yaml
```diff
@@ -0,0 +1,38 @@
...
+  The `PMLEN` values supported by the `Ssnpm` pointer masking facility
+  for the next privilege mode below S-mode (U-mode, and VS-/VU-modes if
+  the H extension is present), selected via the `PMM` field in
+  `senvcfg`. `PMLEN` is the number of high-order bits of an address
+  that are masked.
+
+  Options:
+
+    7::
+      PMLEN=XLEN-57 (i.e., PMLEN=7 on RV64)
```
ThinkOpenly, line 21:
```suggestion
    0::
      PMLEN=0 (pointer masking is disabled)
    7::
      PMLEN=XLEN-57 (i.e., PMLEN=7 on RV64)
```

spec/std/isa/param/SUPPORTED_PMLEN_SSNPM.yaml
```diff
@@ -0,0 +1,42 @@
...
+schema:
+  type: array
+  items:
+    - const: 0
+  additionalItems:
+    type: integer
+    enum: [7, 16]
+  minItems: 1
+  maxItems: 3
+  uniqueItems: true
```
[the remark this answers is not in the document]
ThinkOpenly: Description updated.

spec/std/isa/param/SUPPORTED_PMLEN_SMMPM.yaml
```diff
@@ -0,0 +1,40 @@
...
+schema:
+  type: array
+  items:
+    - const: 0
+  additionalItems:
+    type: integer
+    enum: [7, 16]
+  minItems: 1
+  maxItems: 3
+  uniqueItems: true
```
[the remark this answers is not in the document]
ThinkOpenly: Description updated.

ThinkOpenly (approved):

PR #2019, opened 2026-07-18
fix(z3): support anyOf for integer, boolean, and string parameters

AlgoArtist06:
## Summary

- implement JSON Schema `anyOf` for integer, boolean, and string Z3 parameter constraints
- route all three through one shared `any_of_constraint` helper (conjoin within a branch, disjoin across branches)
- fix a latent leak where a nested `allOf` recursed with the default `assert: true`, pushing branch-local assertions straight into the solver
- add satisfiability coverage for integer const/range/sibling-constraint branches, boolean alternatives, and string const/enum branches

Supersedes #1932, which implemented the integer case only. This PR now covers the whole scalar half of #1890; array constraints and the other combinators (`oneOf`, `noneOf`, `if`/`then`/`else`) remain out of scope.

### The nested `allOf` leak

`constrain_int` and `constrain_bool` both recursed into their `allOf` subschemas without forwarding `assert: false`. At top level that is harmless, since the caller asserts the collected array anyway. Inside an `anyOf` branch it is not: the nested assertions land in the solver unconditionally, so a branch that should have been one alternative of a disjunction becomes mandatory.

`test_constrain_int_with_anyof_nested_allof` pins this. With the `assert: false` forwarding reverted, it fails.

Part of #1890

## Testing

- `./bin/regress -n regress-udb-unit-test --matrix=test=z3_parameter_constraints` - 42 runs, 64 assertions, 0 failures
- `./bin/regress -n regress-udb-unit-test --matrix=test=z3` - 38 runs, 55 assertions, 0 failures
- `./bin/regress -n regress-udb-unit-test --matrix=test=z3_finite_array` - 37 runs, 44 assertions, 0 failures
- `./bin/regress -n regress-sorbet` - no errors
- `./bin/prek run --files tools/ruby-gems/udb/lib/udb/z3.rb tools/ruby-gems/udb/test/test_z3_parameter_constraints.rb` - passed
- `./bin/regress --all` - attempted locally; remaining failures are existing macOS setup limitations, chiefly the Linux-only `must`/`eqntott` binaries, plus cascading schema/submodule/docs host failures. The changed Z3 suites pass independently.

### DIFF
--- a/tools/ruby-gems/udb/lib/udb/z3.rb
+++ b/tools/ruby-gems/udb/lib/udb/z3.rb
@@ -433,6 +433,7 @@ module Udb
     # Handles JSON schema keywords:
     # - const: exact boolean value
     # - allOf: conjunction of subschemas
+    # - anyOf: disjunction of subschemas
     #
     # @param solver [Z3Solver] The solver to add assertions to
     # @param term [Z3::BoolExpr] The Z3 boolean term to constrain
@@ -458,12 +459,15 @@ module Udb
 
       if schema_hsh.key?("allOf")
         schema_hsh.fetch("allOf").each do |h|
-          assertions += constrain_bool(solver, term, h)
+          assertions += constrain_bool(solver, term, h, assert: false)
         end
       end
 
       if schema_hsh.key?("anyOf")
-        raise "TODO: anyOf not yet implemented for boolean constraints"
+        branches = schema_hsh.fetch("anyOf").map do |h|
+          constrain_bool(solver, term, h, assert: false)
+        end
+        assertions << any_of_constraint(branches)
       end
 
       if schema_hsh.key?("oneOf")
@@ -492,6 +496,7 @@ module Udb
     # Handles JSON schema keywords:
     # - const: exact string value (compared via hash)
     # - enum: one of several string values (compared via hash)
+    # - anyOf: disjunction of subschemas
     #
     # @param solver [Z3Solver] The solver to add assertions to
     # @param term [Z3::IntExpr] The Z3 integer term representing the string hash
@@ -526,7 +531,10 @@ module Udb
       end
 
       if schema_hsh.key?("anyOf")
-        raise "TODO: anyOf not yet implemented for string constraints"
+        branches = schema_hsh.fetch("anyOf").map do |h|
+          constrain_string(solver, term, h, assert: false)
+        end
+        assertions << any_of_constraint(branches)
       end
 
       if schema_hsh.key?("oneOf")
@@ -546,6 +554,15 @@ module Udb
       end
       assertions
     end
+
+    sig { params(branches: T::Array[T::Array[Z3::BoolExpr]]).returns(Z3::BoolExpr) }
+    def self.any_of_constraint(branches)
+      expressions = branches.map do |branch|
+        branch.reduce(Z3.True) { |expression, assertion| expression & assertion }
+      end
+      expressions.reduce(Z3.False) { |expression, branch| expression | branch }
+    end
+    private_class_method :any_of_constraint
 
     # Extract array constraints from JSON schema
     #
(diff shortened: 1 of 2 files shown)


tools/ruby-gems/udb/lib/udb/z3.rb
```diff
@@ -433,6 +433,7 @@ def self.constrain_int(solver, term, schema_hsh, name: nil, assert: true)
     # Handles JSON schema keywords:
     # - const: exact boolean value
     # - allOf: conjunction of subschemas
+    # - anyOf: disjunction of subschemas
```
ThinkOpenly: I don't think "disjunction" is a precisely correct description for "anyOf". Maybe "inclusive disjunction"? (I worry that this definition is less well understood than "anyOf" itself, but so be it.)

tools/ruby-gems/udb/lib/udb/z3.rb
```diff
@@ -492,6 +496,7 @@ def self.constrain_bool(solver, term, schema_hsh, name: nil, assert: true)
     # Handles JSON schema keywords:
     # - const: exact string value (compared via hash)
     # - enum: one of several string values (compared via hash)
+    # - anyOf: disjunction of subschemas
```
ThinkOpenly: same

ThinkOpenly (commented):
Just a few teeny-tiny nits.

tools/ruby-gems/udb/lib/udb/z3.rb
AlgoArtist06: Agreed, and the imprecision matters here specifically because `oneOf` (the exclusive disjunction) is handled a few lines below. Fixed in 8dd59fc5.

On the worry that the term is less well understood than `anyOf` itself: I appended a plain-English gloss so the line defines itself, rather than trading one piece of jargon for another.

```ruby
# - anyOf: inclusive disjunction of subschemas (at least one must hold)
```

While in here I also refreshed the `Z3ParameterTerm` class summary, which still listed `anyOf` under "Not yet supported" and described it as something that "would require disjunctive constraints". That stopped being true with this PR. Happy to drop that hunk if you would rather keep the diff to the nits.

tools/ruby-gems/udb/lib/udb/z3.rb
AlgoArtist06: Same change applied to `constrain_string` in 8dd59fc5.

ThinkOpenly:
Wait, isn't this is a duplicate of #1932? If I had to choose, though, I'd probably choose this implementation. Thoughts?

AlgoArtist06:
Not quite duplicates, they were disjoint halves. #1932 only touched `constrain_int`; this PR only touched `constrain_bool` and `constrain_string`. Same keyword, same file, no overlapping lines. Fair confusion though, given they share a title verb and an issue, and that split was my mistake for filing them separately in the first place.

You picking this implementation settles it, so I have folded the integer case in here (846dc222) rather than leaving two idioms in one file. #2019 now covers all three scalar types through the single `any_of_constraint` helper.

One thing worth calling out, because it is the reason consolidating was better than just rebasing #1932 on top:

Both `constrain_int` and `constrain_bool` recursed into their `allOf` subschemas without forwarding `assert: false`:

```ruby
assertions += constrain_int(solver, term, h)   # default assert: true
```

At top level that is harmless, the caller asserts the collected array anyway. Inside an `anyOf` branch it is not. The nested assertions go into the solver unconditionally, so a branch that should be one alternative of a disjunction becomes mandatory. A schema like

```ruby
"anyOf" => [
  { "allOf" => [{ "minimum" => 10 }, { "maximum" => 20 }] },
  { "minimum" => 100, "maximum" => 110 }
]
```

would wrongly reject 105. The bool path got this fix in the original commit here; the int path needed the same one, and #1932's inline version did not have it. `test_constrain_int_with_anyof_nested_allof` pins the behaviour, it fails if the `assert: false` forwarding is reverted.

Also carried over #1932's three integer tests (const, ranges, anyOf-with-sibling-minimum). 42 runs, 64 assertions, 0 failures; sorbet clean.

Still out of scope and still raising `TODO`: `anyOf` for arrays, plus `oneOf`, `noneOf`, and `if`/`then`/`else`. Happy to take those as follow-ups against #1890.

ThinkOpenly (approved):

PR #2027, opened 2026-07-19
Fix: Extension#<=> compared operands in reverse order

### LINKED ISSUE #2026: `Extension#<=>` compares in reverse order, breaking `sort`/`min`/`max`/`<`/`>` on Extension objects
**File:** `tools/ruby-gems/udb/lib/udb/obj/extension.rb`, line 404

**Current code:**
```ruby
def <=>(other_ext)
  return nil unless other_ext.is_a?(Extension)
  other_ext.name <=> name
end
```

**Problem:**
This compares `other_ext` against `self` instead of `self` against `other_ext`, so the result is inverted. Since `Extension` includes `Comparable`, this silently reverses `<`, `>`, `.between?`, and any bare `.sort`/`.min`/`.max` called on a collection of `Extension` objects.

Sibling classes in the same file implement it the standard way:
- `ExtensionVersion#<=>` (line 1173): `@name <=> other.name`
- `ExtensionRequirement#<=>` (line 1745): `@name <=> other.name`

`Extension` is the only one reversed.

**Steps to reproduce:**
```bash
bin/bundle exec irb -Itools/ruby-gems/udb/lib -r udb/resolver
```
```ruby
resolver = Udb::Resolver.new(Pathname.new(Dir.pwd))
cfg_arch = resolver.cfg_arch_for("_")
exts = cfg_arch.extensions.select { |e| ["I", "M", "A", "C", "D"].include?(e.name) }

exts.sort_by(&:name).map(&:name)
# => ["A", "C", "D", "I", "M"]   (correct)

exts.sort.map(&:name)
# => ["M", "I", "D", "C", "A"]   (reversed — uses the buggy <=>)
```

**Expected:** `exts.sort.map(&:name)` should match `exts.sort_by(&:name).map(&:name)`.

**Actual:** Order is reversed.

**Suggested fix:**
```ruby
def <=>(other_ext)
  return nil unless other_ext.is_a?(Extension)
  name <=> other_ext.name
end
```

**Notes:**
- Current call sites in the codebase (e.g. `isa_nav.adoc.erb`, `table_builder.rb`) all use `sort_by(&:name)` or an explicit block, so they bypass this bug — but any future code relying on bare `.sort`, `<`, `>`, or `.between?` on `Extension` objects will silently get wrong results.
- No existing test coverage catches this (no RSpec/Minitest for `Extension#<=>`).
- Introduced in #403 (Mar 2025); appears to be an unintentional operand-order slip, not deliberate.

Maanvi212006:
`Extension#<=>` in `tools/ruby-gems/udb/lib/udb/obj/extension.rb` compares `other_ext.name <=> name` instead of `name <=> other_ext.name`, inverting the result.

Since `Extension` includes `Comparable`, this silently reversed `<`, `>`, `.between?`, and any bare `.sort`/`.min`/`.max` on `Extension` objects. Sibling classes (`ExtensionVersion`, `ExtensionRequirement`) in the same file already use the correct operand order.

Invert the comparison to match expected use.

Fixes #2026


ThinkOpenly:
[ed. original PR description before editing]
## What this PR does
`Extension#<=>` in `tools/ruby-gems/udb/lib/udb/obj/extension.rb` (line 404) compared `other_ext.name <=> name` instead of `name <=> other_ext.name`, inverting the result.

Since `Extension` includes `Comparable`, this silently reversed `<`, `>`, `.between?`, and any bare `.sort`/`.min`/`.max` on `Extension` objects. Sibling classes (`ExtensionVersion`, `ExtensionRequirement`) in the same file already use the correct operand order.

Fixes #2026 

## Testing
```ruby
resolver = Udb::Resolver.new(Pathname.new(Dir.pwd))
cfg_arch = resolver.cfg_arch_for("_")
exts = cfg_arch.extensions.select { |e| ["I","M","A","C","D"].include?(e.name) }

exts.sort.map(&:name)
# before: ["M","I","D","C","A"]  (wrong)
# after:  ["A","C","D","I","M"]  (matches sort_by(&:name))
```

Ran full suite: `bin/bundle exec rake test:udb:unit` → 3583 runs, 22571 assertions, 0 failures, 0 errors.

Maanvi212006:
It looks like CI is currently failing during the environment setup step, before any project code is built or tested.

The common failure is:

```text
Failed to install aqua:j178/prek@0.4.9
GitHub artifact attestations verification failed:
GitHub API returned 503 Service Unavailable
"trust-metadata-api service unavailable"
```

I checked multiple recent PRs and they appear to be hitting the same error, so this doesn't seem to be specific to this PR. It looks like a transient GitHub artifact attestation/API issue causing mise install to fail.

ThinkOpenly (approved):

PR #2028, opened 2026-07-19
fix(udb): correct copy-paste errors in InterruptCode#<=> and #hash

### LINKED ISSUE #1979: InterruptCode#<=> and InterruptCode#hash reference ExceptionCode instead of InterruptCode

krrishverma1805-web:
## Problem

`InterruptCode` in `tools/ruby-gems/udb/lib/udb/obj/exception_code.rb` was copy-pasted from `ExceptionCode` but two class references were not updated, introducing two silent bugs:

**1. `InterruptCode#<=>` always returns `nil`**

```ruby
# Before (wrong)
def <=>(other)
  return nil unless T.cast(other, Object).is_a?(ExceptionCode)  # always false for InterruptCode
  num <=> T.cast(other, ExceptionCode).num
end
```

Comparing two `InterruptCode` objects always returns `nil` because an `InterruptCode` instance is never `is_a?(ExceptionCode)`. Any `sort`, `min`, `max`, or `Enumerable` operation using the base-class comparator raises `ArgumentError: comparison failed`.

**2. `InterruptCode#hash` collides with `ExceptionCode`**

```ruby
# Before (wrong)  
def hash = [ExceptionCode, num].hash
```

An `InterruptCode` with `num = 1` and an `ExceptionCode` with `num = 1` produce the same hash, violating the `a.eql?(b) => a.hash == b.hash` contract and breaking `Set`/`Hash` correctness.

## Fix

Change both occurrences of `ExceptionCode` in `InterruptCode`'s methods to `InterruptCode`:

```ruby
# After (correct)
def <=>(other)
  return nil unless T.cast(other, Object).is_a?(InterruptCode)
  num <=> T.cast(other, InterruptCode).num
end

def hash = [InterruptCode, num].hash
```

## Tests

Add `test_exception_interrupt_code.rb` covering:
- Sorting `ExceptionCode` and `InterruptCode` collections by `num`
- `eql?` and `hash` consistency (objects with same `num` are equal and share a hash)
- Cross-type isolation: an `ExceptionCode` and `InterruptCode` with the same `num` are not equal and do not share a hash

All 3 tests pass. Sorbet type-checking (`./do test:sorbet`) passes with no errors.

Fixes #1979

### DIFF
--- a/tools/ruby-gems/udb/lib/udb/obj/exception_code.rb
+++ b/tools/ruby-gems/udb/lib/udb/obj/exception_code.rb
@@ -82,9 +82,9 @@ module Udb
 
     sig { override.params(other: BasicObject).returns(T.nilable(Integer)) }
     def <=>(other)
-      return nil unless T.cast(other, Object).is_a?(ExceptionCode)
+      return nil unless T.cast(other, Object).is_a?(InterruptCode)
 
-      num <=> T.cast(other, ExceptionCode).num
+      num <=> T.cast(other, InterruptCode).num
     end
 
     sig { override.params(other: BasicObject).returns(T::Boolean) }
@@ -93,6 +93,6 @@ module Udb
     end
 
     sig { override.returns(Integer) }
-    def hash = [ExceptionCode, num].hash
+    def hash = [InterruptCode, num].hash
   end
 end
--- a/tools/ruby-gems/udb/test/test_exception_interrupt_code.rb
+++ b/tools/ruby-gems/udb/test/test_exception_interrupt_code.rb
@@ -1,0 +1,91 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# typed: false
+# frozen_string_literal: true
+
+require_relative "test_helper"
+require "tmpdir"
+require "fileutils"
+
+require "udb/resolver"
+
+class TestExceptionInterruptCode < Minitest::Test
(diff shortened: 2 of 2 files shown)


krrishverma1805-web:
Closing this PR as a duplicate of #1980, which addresses the same bug and was opened earlier. I was not aware #1980 existed when I opened this.

One thing worth noting: #1980 does not include regression tests for the fixed behavior. If it would be useful, the test file from this PR (`test_exception_interrupt_code.rb`) covering sort correctness, `eql?`/`hash` consistency, and cross-type isolation could be incorporated into #1980 — happy to help if the author or a maintainer thinks it adds value. Otherwise, apologies for the noise.

PR #2037, opened 2026-07-20
fix: VersionSpec#<=> compared operands in reverse order

### LINKED ISSUE #2036: VersionSpec#<=> compares operands in reverse order when comparing against a String
## File
`tools/ruby-gems/udb/lib/udb/version_spec.rb`, line ~115, class `VersionSpec`

## Problem

`VersionSpec#<=>` reverses the comparison operands when the argument is a `String`:

```ruby
def <=>(other)
  if other.is_a?(String)
    VersionSpec.new(other) <=> self     # reversed
  elsif other.is_a?(VersionSpec)
    ...
```

Since `VersionSpec` includes `Comparable`, this flips the result of any comparison (`<`, `>`, `.between?`, `.sort`, etc.) made between a `VersionSpec` and a `String`. Comparisons between two `VersionSpec` objects are unaffected. `==` still returns the correct result since equality is symmetric, which is likely why this went unnoticed.

## How it was found

While auditing `Comparable`-including classes after finding a similar reversed-operand bug in `Extension#<=>`, checked sibling classes for the same pattern.

## Repro

```ruby
vs = Udb::VersionSpec.new("2.0")

vs <=> "1.0"   # => -1 (expected 1)
vs > "1.0"     # => false (expected true)
vs < "1.0"     # => true (expected false)

# control — VersionSpec vs VersionSpec, unaffected:
Udb::VersionSpec.new("1.0") <=> vs   # => -1 (correct)
```

Maanvi212006:
`VersionSpec#<=>` in `tools/ruby-gems/udb/lib/udb/version_spec.rb`
reversed the comparison operands when the argument was a `String`:

Since `VersionSpec` includes `Comparable`, this silently flipped `<`, `>`,
`.between?`, `.sort`, etc. when comparing a `VersionSpec` to a `String`.

Reverse the comparision.

Fixes #2036

### DIFF
--- a/tools/ruby-gems/udb/lib/udb/version_spec.rb
+++ b/tools/ruby-gems/udb/lib/udb/version_spec.rb
@@ -114,7 +114,7 @@ module Udb
     sig { params(other: T.untyped).returns(T.nilable(Integer)).checked(:never) }
     def <=>(other)
       if other.is_a?(String)
-        VersionSpec.new(other) <=> self
+        self <=> VersionSpec.new(other)
       elsif other.is_a?(VersionSpec)
         if @major != other.major
           @major <=> other.major


Maanvi212006:
Hii @ThinkOpenly . Could you pls review this fix when you have a chance?
Thanks!

ThinkOpenly (approved):

ThinkOpenly:
## Problem

`VersionSpec#<=>` in `tools/ruby-gems/udb/lib/udb/version_spec.rb`
reversed the comparison operands when the argument was a `String`:

```ruby
def <=>(other)
  if other.is_a?(String)
    VersionSpec.new(other) <=> self     # reversed
  elsif other.is_a?(VersionSpec)
```

Since `VersionSpec` includes `Comparable`, this silently flipped `<`, `>`,
`.between?`, `.sort`, etc. when comparing a `VersionSpec` to a `String`.
Comparisons between two `VersionSpec` objects were unaffected. `==` still
returned the correct result since equality is symmetric.

## Fix

```ruby
if other.is_a?(String)
  self <=> VersionSpec.new(other)
```

## Before / after

```ruby
vs = Udb::VersionSpec.new("2.0")

# before
vs <=> "1.0"   # => -1 (wrong)
vs > "1.0"     # => false (wrong)
vs < "1.0"     # => true (wrong)

# after
vs <=> "1.0"   # => 1
vs > "1.0"     # => true
vs < "1.0"     # => false

# control (VersionSpec vs VersionSpec, unaffected either way)
Udb::VersionSpec.new("1.0") <=> vs   # => -1
```

## Testing

Full suite run: `bin/bundle exec rake test:udb:unit`

## Impact

No live call site currently breaks on this — the only usage found was
`== "0"`, which is order-insensitive — so this was dormant rather than
actively wrong, same as the related `Extension#<=>` bug fixed in #<extension-pr-number>.

Fixes #2036

PR #2049, opened 2026-07-21
fix(isa): fp.idl F32 softfloat NV flag missing on signaling NaN inputs

### LINKED ISSUE #2048: fix(isa): fp.idl F32 softfloat NV flag missing on signaling NaN inputs
## Problem

The F32 softfloat helpers in `spec/std/isa/isa/fp.idl` (f32_muladd, f32_mul,
f32_div, f32_sqrt, f32_to_f64, softfloat_addMagsF32, softfloat_subMagsF32)
return `SP_CANONICAL_NAN` on NaN inputs without checking whether the NaN is
signaling. IEEE 754-2019 §7.2(a) requires NV whenever a signaling NaN is an
operand.

The `expDiff==0` branch in `softfloat_subMagsF32` has the opposite problem:
NV fires unconditionally there, including on quiet NaN inputs that should not
raise it.

## Scope

Pre-existing F extension softfloat code. Not part of the V extension
implementation. F64 equivalents will be addressed separately when the F64
helper functions are introduced.

## Fix

- Add `is_sp_signaling_nan?()` checks before each `return SP_CANONICAL_NAN`
  triggered by a NaN input (~15 call sites across 7 functions)
- Fix the `softfloat_subMagsF32` `expDiff==0` branch to distinguish NaN inputs
  (sNaN check) from inf-inf (always invalid)

The comparison functions (`f32_eq?`, `f32_lt?`, etc.) and integer conversion
functions were already correct and are not touched.

jpak-qc:
## Summary

IEEE 754-2019 section 7.2(a) requires the Invalid Operation (NV) flag whenever a signaling NaN (sNaN) is consumed as an operand, regardless of whether the result is a quiet NaN. The F32 softfloat helpers in `fp.idl` were silently returning `SP_CANONICAL_NAN` at NaN-input paths without checking for sNaN, so the NV flag was never raised on sNaN inputs to those functions.

This fix adds `is_sp_signaling_nan?` guards at 17 call sites across 6 functions:

- `f32_muladd` (5 sites: NaN in A, NaN in B, NaN in C paths)
- `f32_sqrt` (1 site)
- `f32_div` (3 sites: NaN A, NaN B, NaN A+B)
- `f32_mul` (2 sites)
- `softfloat_addMagsF32` (3 sites: expDiff==0, expDiff<0, expDiff>0 paths)
- `softfloat_subMagsF32` (3 sites: expDiff==0, expDiff<0, expDiff>0 paths)

At sites where either operand could be sNaN (the `expDiff != 0` paths in addMags/subMags, where either A or B triggered the NaN branch), both operands are checked with `is_sp_signaling_nan?(a) || is_sp_signaling_nan?(b)`. At sites where only one specific operand is known to be NaN by construction, only that operand is checked.

The F64 softfloat helpers have the same gap and will be addressed in a follow-on PR.

## Scope

This PR touches only `spec/std/isa/isa/fp.idl`. No instruction YAML files, no extension definitions, no CSR files.

## Testing

- `./do test:idl CFG=_` passes (IDL type-check against unconfigured architecture)
- The `is_sp_signaling_nan?` helper is defined in `fp.idl` and already used in the F32 comparison functions (`f32_le`, `f32_lt`, `f32_eq`) — this PR extends the same pattern to the arithmetic helpers

Closes #2048


henrikg-qc (approved):

jpak-qc:
Superseded by #1974 (merged 2026-07-21), which fixes the same 6 F32 softfloat functions for IEEE 754-2019 §7.2(a) NV-flag compliance using a cleaner single entry-guard approach per function. Closing this PR to avoid a conflicting fix on the same code.

PR #2064, opened 2026-07-21
fix(data): correct contradictory GVA field description in mstatus.yaml

### LINKED ISSUE #2061: mstatus.GVA field description is self-contradictory — same condition listed for both set and clear
**Describe the bug**
In `spec/std/isa/csr/mstatus.yaml`, the `GVA` field description at lines
113–114 contains two sentences that describe opposite effects (set vs. clear)
using the exact same condition, making the description logically contradictory:

    When a trap is taken and a guest virtual address is written into mtval, GVA is set.
    When a trap is taken and a guest virtual address is written into mtval, GVA is cleared.

The second sentence should describe when GVA is cleared, which per the
RISC-V Hypervisor specification is when a non-guest virtual address is
written into mtval — not a guest virtual address.

**To Reproduce**
1. On branch `main`,
2. Open `spec/std/isa/csr/mstatus.yaml` and search for `GVA is cleared`
3. Both lines 113 and 114 contain the word "guest virtual address" with
   opposite effects (set vs. cleared) under identical stated conditions.

No error — passes `./do test:schema` silently because this is a prose error,
not a structural one.

**Screenshots**
N/A

**Additional context**
The fix is to change line 114 from:
  "When a trap is taken and a guest virtual address is written into mtval, GVA is cleared."
to:
  "When a trap is taken and a non-guest virtual address is written into mtval, GVA is cleared."

Related file: `spec/std/isa/csr/mstatus.yaml`

tanukusaitejesh-prog:
The GVA field description in mstatus.yaml listed the same
condition for both setting and clearing the bit, making it self-contradictory.
The second sentence now correctly states that GVA is cleared for any trap
into HS mode that doesn't write a guest virtual address into `stval`.

Fixes #2061


spec/std/isa/csr/mstatus.yaml
```diff
@@ -111,7 +111,7 @@ fields:
     long_name: Guest Virtual Address
     description: |
       When a trap is taken and a guest virtual address is written into mtval, GVA is set.
-      When a trap is taken and a guest virtual address is written into mtval, GVA is cleared.
+      When a trap is taken and a non-guest virtual address is written into mtval, GVA is cleared.
```
ThinkOpenly: Let's just take the relevant text directly from the RISC-V ISA Manual:
```suggestion
      GVA is written by the implementation whenever a trap is taken into
      HS mode. For any trap (breakpoint, address misaligned, access fault,
      page fault, or guest-page fault) that writes a guest virtual address to
      `stval`, GVA is set to 1. For any other trap into HS-mode,
      GVA is set to 0.
```

ThinkOpenly (changes requested):
Thank you for this contribution! It's probably better to take text verbatim from the manual when possible.

ThinkOpenly:
Title:
fix(data): correct contradictory GVA field description in mstatus.yaml

Body:
## Summary
The GVA field description in mstatus.yaml (lines 113–114) listed the same
condition for both setting and clearing the bit, making it self-contradictory.
The second sentence now correctly states that GVA is cleared when a
non-guest virtual address is written into mtval.

## Motivation
Fixes #2061

## Changes
- `spec/std/isa/csr/mstatus.yaml` — corrected second sentence of GVA field description

## Verification
- [x] `./do test:schema` passes
- [x] `./bin/regress --tag smoke` passes

ThinkOpenly (approved):

PR #2069, opened 2026-07-21
Added params for sdtrig mcontrol6 tests

pclarkhmc:


ThinkOpenly:
It would help me to have references for where these parameters are described in The RISC-V Debug
Specification.

PR #2072, opened 2026-07-22
fix(udb): capture CSR and field name in CsrField#alias regex

### LINKED ISSUE #2018: CsrField#alias regex has no capture group for CSR/field name, mis-parsing every alias
**Describe the bug**

`CsrField#alias` in `tools/ruby-gems/udb/lib/udb/obj/csr_field.rb` (lines 300-306) parses an alias string like `csrname.FIELD[hi:lo]` with a regex whose capture groups do not wrap the CSR name or the field name:

```ruby
raise "Can't parse alias" unless data["alias"] =~ /^[a-z][a-z0-9]+\.[A-Z0-9]+(\[([0-9]+)(:[0-9]+)?\])?$/
csr_name    = T.must(Regexp.last_match(1))
csr_field   = Regexp.last_match(2)
range       = Regexp.last_match(3)
range_start = Regexp.last_match(4)   # nil - group 4 does not exist
range_end   = Regexp.last_match(5)   # nil - group 5 does not exist
```

The regex has only three groups: `(1)` = the whole optional `[..]` bracket, `(2)` = the high bit, `(3)` = the `:low` part. The CSR name `[a-z][a-z0-9]+` and field `[A-Z0-9]+` are not captured at all, and groups 4/5 do not exist. So `csr_name` is assigned the bracket text (or `nil`), and `T.must(cfg_arch.csr(csr_name))` looks up a nonexistent CSR. For an alias without a range, group 1 is `nil` and `T.must(...)` raises.

**To Reproduce**

```ruby
BUGGY = /^[a-z][a-z0-9]+\.[A-Z0-9]+(\[([0-9]+)(:[0-9]+)?\])?$/
FIXED = /^([a-z][a-z0-9]+)\.([A-Z0-9]+)(\[([0-9]+)(:[0-9]+)?\])?$/
["mstatus.MPP[12:11]", "sstatus.SPP"].each do |s|
  s =~ BUGGY; puts "current #{s}: csr_name=#{$1.inspect} field=#{$2.inspect}"
  s =~ FIXED; puts "fixed   #{s}: csr_name=#{$1.inspect} field=#{$2.inspect} start=#{$4.inspect} end=#{$5.inspect}"
end
```

```
current mstatus.MPP[12:11]: csr_name="[12:11]" field="12"
fixed   mstatus.MPP[12:11]: csr_name="mstatus" field="MPP" start="12" end=":11"
current sstatus.SPP: csr_name=nil field=nil
fixed   sstatus.SPP: csr_name="sstatus" field="SPP" start=nil end=nil
```

**Additional context**

Fix is to add capture groups around the name and field so the five `Regexp.last_match` reads line up:
`/^([a-z][a-z0-9]+)\.([A-Z0-9]+)(\[([0-9]+)(:[0-9]+)?\])?$/`
With this, the existing `range_end[1..]` on line 315 correctly strips the leading `:`. I can send a PR.

deepak0x:
## What this fixes

`CsrField#alias` in `tools/ruby-gems/udb/lib/udb/obj/csr_field.rb` parsed an alias string like `csrname.FIELD[hi:lo]` with a regex that only captured the optional bit-range. The CSR name and field name were never wrapped in capture groups, so:

- `csr_name` (`Regexp.last_match(1)`) received the `[..]` bracket text, or `nil` for a range-less alias
- `csr_field` (`Regexp.last_match(2)`) received the high bit
- groups 4 and 5 did not exist

The result was that `cfg_arch.csr(csr_name)` looked up a nonexistent CSR, and a range-less alias raised through `T.must`.

## The fix

Wrap the CSR name and field in capture groups so the five `Regexp.last_match` reads line up:

```
/^([a-z][a-z0-9]+)\.([A-Z0-9]+)(\[([0-9]+)(:[0-9]+)?\])?$/
```

The existing `range_end[1..]` on the following line still strips the leading `:` correctly.

## Tests

Added `test/test_csr_field_alias.rb` (registered in `test/run.rb`) covering a range-less alias (uses the target field location), a single-bit range, a multi-bit range, and rejection of an unparseable string. Verified before and after: against the old regex three of the four tests error with `Passed nil into T.must`; with the fix all pass (4 runs, 11 assertions, 0 failures).

Closes #2018


deepak0x:
@jordancarlin @dhower-qc when you have a moment, could you take a look? Small regex fix for CsrField#alias (Closes #2018) with unit tests, verified before and after. Thanks.

PR #2073, opened 2026-07-22
fix(udb): compare DatabaseObject by kind reader, not undefined @Kind

### LINKED ISSUE #2017: DatabaseObject#<=> references undefined @Kind, breaking comparison/sort
**Describe the bug**

`DatabaseObject#<=>` in `tools/ruby-gems/udb/lib/udb/obj/database_obj.rb` (line 118) guards on an instance variable `@Kind` (capital K) that is never assigned. The value set in `initialize` (line 114) is `@kind` (lowercase):

```ruby
def <=>(other)
  return nil unless other.is_a?(DatabaseObject)
  return nil unless @Kind == other.kind   # @Kind is undefined -> nil
  @name <=> other.name
end
```

`@Kind` is referenced only on this line (nowhere else in the repo), so it reads as `nil`. `nil == other.kind` is always false, so `<=>` returns `nil` for two same-kind `DatabaseObject`s instead of comparing by name. Any `sort`/`<`/`>`/`min`/`max` on objects using this base-class comparator raises `ArgumentError: comparison ... failed`.

**To Reproduce**

The comparison logic in isolation (Ruby 3.2):

```ruby
class DObj
  def initialize(name, kind) = (@name, @kind = name, kind)
  def name = @name
  def kind = @kind
  def cmp_buggy(o); return nil unless @Kind == o.kind; @name <=> o.name; end  # current
  def cmp_fixed(o); return nil unless @kind == o.kind; @name <=> o.name; end  # fix
end
a = DObj.new("aaa","csr"); b = DObj.new("bbb","csr")
p a.cmp_buggy(b)          # => nil   (should be -1)
p a.cmp_fixed(b)          # => -1
[b,a].sort { |x,y| x.cmp_buggy(y) or raise ArgumentError, "comparison failed" }
```

```
current (@Kind): a <=> b = nil   (expected -1)
fixed   (@kind): a <=> b = -1
current => sort RAISES: comparison of DObj failed
fixed   => sort works: ["aaa", "bbb"]
```

**Additional context**

Fix is to compare like-for-like using the assigned variable / reader, e.g. `return nil unless kind == other.kind`. Note `other.kind` uses the `kind` reader, so the fix should compare the same representation on both sides. I can send a PR.

deepak0x:
## What this fixes

`DatabaseObject#<=>` in `tools/ruby-gems/udb/lib/udb/obj/database_obj.rb` guarded on `@Kind` (capital K), an instance variable that is never assigned anywhere in the repo. It read as `nil`, so `@Kind == other.kind` was always false and the method returned `nil` for two same-kind objects instead of comparing by name. Any `sort`/`<`/`>`/`min`/`max` over these objects raised `ArgumentError: comparison ... failed`.

## The fix

Compare using the `kind` reader on both sides:

```
return nil unless kind == other.kind
```

`kind` returns `@kind.serialize`, so this compares the same representation on both operands (the right side already used `other.kind`).

## Tests

Added `test/test_database_obj_compare.rb` (registered in `test/run.rb`) covering same-kind ordering, `sort`, differing kinds returning `nil`, and a non-DatabaseObject operand returning `nil`. Verified before and after: against `@Kind` the ordering assertion returns `nil` (expected `-1`) and `sort` raises `ArgumentError`; with the fix all pass (4 runs, 6 assertions, 0 failures).

Closes #2017


deepak0x:
@jordancarlin @dhower-qc when you have a moment, could you take a look? One-line comparator fix for DatabaseObject#<=> (Closes #2017) with unit tests, verified before and after. Thanks.

ThinkOpenly (approved):

ThinkOpenly:
This apparently didn't make it through the merge queue. I believe it needs DCO ("Signed-off-by") for all commits. If you would add this to each commit (or feel free to squash them into a single commit), then force-push, we can then get this merged. Thank you!

ThinkOpenly (changes requested):
Please add DCO to your commits, force-push.

deepak0x:
Added DCO sign-off to both commits and force-pushed, as requested. The DCO check now passes. The branch was also rebased onto current main so it no longer carries a merge commit. Ready for another look whenever you get a chance.

ThinkOpenly (approved):

PR #2075, opened 2026-07-22
fix(data): correct Q extension ratified version to 2.2.0

### LINKED ISSUE #2068: Q extension ratified version is incorrect - listed as 1.0.0 instead of 2.2.0
Describe the bug
In spec/std/isa/ext/Q.yaml, the versions entry lists the Q (Quad-Precision Floating-Point) extension as ratified at version 1.0.0. This is inconsistent with the RISC-V ISA manual: Q shares the same floating-point ISA chapter as F and D, which are both correctly versioned 2.2.0 in this same repo (spec/std/isa/ext/F.yaml, spec/std/isa/ext/D.yaml). The official RISC-V unprivileged ISA manual colophon table lists all three together:

|F| |2.2| |Ratified|
|D| |2.2| |Ratified|
|Q| |2.2| |Ratified|

Q was ratified alongside F and D's 2.2 revision (NaN-boxing scheme / FMAX-FMIN redefinition, Dec 2019), not as a standalone 1.0 spec.

Confirmed against:
- ext/riscv-isa-manual colophon table (modules/unpriv/pages/colophon.adoc), which lists Q at version 2.2, Ratified.
- The chapter title in the official RISC-V User-Level ISA manual: "Q Standard Extension for Quad-Precision Floating-Point, Version 2.2".

Related file: spec/std/isa/ext/Q.yaml

sujalkurkute:
Q shares the same floating-point ISA chapter revision as F and D,
both of which are correctly versioned 2.2.0. The RISC-V unprivileged
ISA manual colophon table lists F, D, and Q together at version 2.2,
Ratified. Q was never ratified as a standalone 1.0 spec.

Fixes #2068"


jordancarlin:
Thanks for contributing, but this was already fixed in #2067. Closing.

PR #2084, opened 2026-07-23
fix(idlc): correct bits_needed for negative powers of two

### LINKED ISSUE #2083: `bits_needed` under-counts by one bit for negative powers of two (signed)
**Describe the bug**

`bits_needed(value, signed)` in `tools/ruby-gems/idlc/lib/idlc/ast.rb` returns one bit too few for **negative powers of two** when `signed` is true.

The signed / `value < 0` branch special-cases powers of two:

```ruby
# need bit_length plus a sign bit, unless value is a power of 2
if (value.abs & (value.abs - 1)) == 0
  value.bit_length          # <-- omits the sign bit
else
  value.bit_length + 1
end
```

The special case is incorrect. A negative power of two `-2^k` is exactly the *most-negative* value representable in a `(k+1)`-bit two's-complement type, so it needs `k+1` bits — i.e. `value.bit_length + 1`, the same as every other negative value. Omitting the sign bit under-counts by one:

| value | `bits_needed` (current) | correct (min signed width) |
|------:|:-----------------------:|:--------------------------:|
| `-1`  | **0** | 1 |
| `-2`  | **1** | 2 |
| `-4`  | **2** | 3 |
| `-8`  | **3** | 4 |
| `-16` | **4** | 5 |
| `-32` | **5** | 6 |

(Non-power-of-two negatives such as `-3`, `-5`, `-7` already return the correct `value.bit_length + 1`.)

**To Reproduce**

1. On branch `main`.
1. `bits_needed` is used by the signed integer constant-folding in `+`, `-`, and `*` (both narrowing and widening variants). The width check is `return sum if sum_bits_needed <= type.width` — so an under-count makes an over-wide value wrongly pass as "fits", and the truncation warning is skipped.
1. Concretely: a signed constant-folded result of `-16` targeting a 4-bit signed type. `bits_needed(-16, true)` returns `4`, so `4 <= 4` reports "fits" and no truncation warning is emitted — but `-16` does **not** fit in a 4-bit signed type (range `-8..7`); it silently truncates to a wrong value.

No crash — a silently wrong width. The observable effect is a *missing* truncation warning (and an accepted out-of-range value) for expressions whose folded signed result is a negative power of two.

**Additional context**

The `when 0`, `when 1`, positive, and unsigned branches are separate and correct; only the negative-power-of-two special case is wrong.

---

### Proposed change — what and why

**What:** remove the negative-power-of-two special case so the `value < 0` branch always returns `value.bit_length + 1`:

```diff
-            # need bit_length plus a sign bit, unless value is a power of 2
-            if (value.abs & (value.abs - 1)) == 0
-              value.bit_length
-            else
-              value.bit_length + 1
-            end
+            # need bit_length plus a sign bit
+            value.bit_length + 1
```

**Why:** every negative value — powers of two included — needs `bit_length + 1` signed bits, because `-2^k` is the most-negative value of a `(k+1)`-bit type. The special case was never correct; it under-counted by one for exactly the power-of-two inputs. This makes the truncation/width check accurate again.

Purav001:
## What

`bits_needed` under-counted signed negative powers of two: the special case
`(value.abs & (value.abs - 1)) == 0 -> value.bit_length` dropped the sign bit, so
e.g. `-16` returned 4 instead of 5 (it is the most-negative value of a 5-bit
signed type). In two's complement a signed value always needs `bit_length + 1`
bits, so the whole signed branch reduces to that; the unsigned branch keeps the
`value == 0 -> 1` guard since `0.bit_length` is 0.

### DIFF
--- a/tools/ruby-gems/idlc/lib/idlc/ast.rb
+++ b/tools/ruby-gems/idlc/lib/idlc/ast.rb
@@ -4799,27 +4799,12 @@ module Idl
     # @return [Integer] the number of bits needed to represent value in two's complement
     def bits_needed(value, signed)
       if signed
-        case value
-        when 0
-          1
-        when 1
-          2
-        else
-          if value > 0
-            # need bit_legnth plus a sign bit
-            bits = value.bit_length + 1
-          else
-            # need bit_length plus a sign bit, unless value is a power of 2
-            if (value.abs & (value.abs - 1)) == 0
-              value.bit_length
-            else
-              value.bit_length + 1
-            end
-          end
-        end
+        # add sign bit
+        value.bit_length + 1
       else
         internal_error "unsigned value is negative" if value < 0
 
+        # 0 needs a single bit; `0.bit_length` returns 0.
         value == 0 ? 1 : value.bit_length
       end
     end


tools/ruby-gems/idlc/lib/idlc/ast.rb
ThinkOpenly, line 4808: Good catch! Can this whole function be replaced with (lightly tested):
```
if signed
  value.bit_length + 1
else
  internal_error "unsigned value is negative" if value < 0

  value.bit_length
end
```

ThinkOpenly (changes requested):

Purav001:
Good idea @ThinkOpenly — that's much cleaner. I applied it, with one tweak: the unsigned branch still needs to special-case value == 0, since 0.bit_length is 0 but zero needs 1 bit. So:

```ruby
def bits_needed(value, signed)
  if signed
    value.bit_length + 1
  else
    internal_error "unsigned value is negative" if value < 0
    value == 0 ? 1 : value.bit_length
  end
end
```
Confirmed this matches the two's-complement minimum for the full signed and unsigned ranges (including the negative powers of two), and the existing test still passes.

tools/ruby-gems/idlc/lib/idlc/ast.rb
Purav001: updated

tools/ruby-gems/idlc/lib/idlc/ast.rb
```diff
@@ -4799,27 +4799,14 @@ def type_check(symtab, strict:)
...
-            if (value.abs & (value.abs - 1)) == 0
-              value.bit_length
-            else
-              value.bit_length + 1
-            end
-          end
-        end
+        # In two's complement, a value needs bit_length magnitude bits plus a
+        # sign bit (Integer#bit_length already returns the correct magnitude for
+        # negative values, so this also covers negative powers of two).
```
ThinkOpenly:
```suggestion
        # add sign bit
```

tools/ruby-gems/idlc/lib/idlc/ast.rb
```diff
@@ -4799,27 +4799,14 @@ def type_check(symtab, strict:)
...
+        value.bit_length + 1
       else
         internal_error "unsigned value is negative" if value < 0
 
+        # 0 needs a single bit; bit_length would return 0 for it.
```
ThinkOpenly:
```suggestion
        # 0 needs a single bit; `0.bit_length` returns 0.
```

tools/ruby-gems/idlc/test/test_values.rb
ThinkOpenly, line 1: I certainly appreciate the test case, but I'd prefer not to merge it. Could you drop it from this PR?

ThinkOpenly (changes requested):
Thanks for the additional testing!

I was going to edit the PR description, because it becomes the (squashed) commit message upon merging. Could you edit it? Please drop everything except the content of the "what" section, including the "what" title. (The reference to the GitHub issue contains all of that background.)

Purav001:
Thanks @ThinkOpenly ! Applied all four:

- Both comment suggestions (# add sign bit, and the 0.bit_length note)
- Dropped the test case from this PR
- Trimmed the description down to just the What section for the squash message

The PR now touches only ast.rb. 

Heads-up: the two codecov/patch checks now fail only because dropping the test leaves the changed lines uncovered — every actual test (incl. regress-idlc-unit) passes.

ThinkOpenly (approved):

PR #2090, opened 2026-07-23
Unify tvec mode and alignment parameters

jordancarlin:
- Add `STVEC_BASE_ALIGNMENT_VECTORED` to mirror the existing `mtvec` vectored alignment parameter.
- Replace `STVEC_MODE_DIRECT`/`STVEC_MODE_VECTORED` and `VSTVEC_MODE_DIRECT`/`VSTVEC_MODE_VECTORED` with array-based `STVEC_MODES` and `VSTVEC_MODES`, matching `MTVEC_MODES`.
- Convert the `*TVEC_BASE_ALIGNMENT*` parameter values to hexadecimal for readability.
- Express the related `*tvec` parameter requirements in YAML rather than IDL.

See https://github.com/riscv/riscv-arch-test/issues/1953


spec/std/isa/csr/stvec.yaml
```diff
@@ -42,13 +42,19 @@ fields:
...
-      } else {
-        return UNDEFINED_LEGAL_DETERMINISTIC;
+      if (csr_value.MODE == 0) {
+        if ($array_includes?(STVEC_MODES, 0)) {
+          return csr_value.MODE;
+        }
+      } else if (csr_value.MODE == 1) {
+        if ($array_includes?(STVEC_MODES, 1)) {
+          return csr_value.MODE;
+        }
```
ThinkOpenly: Shall we combine these?
```suggestion
      if ($array_includes?(STVEC_MODES, csr_value.MODE)) {
        return csr_value.MODE;
      }
```

spec/std/isa/csr/vstvec.yaml
```diff
@@ -43,13 +43,19 @@ fields:
...
-      } else {
-        return UNDEFINED_LEGAL_DETERMINISTIC;
+      if (csr_value.MODE == 0) {
+        if ($array_includes?(VSTVEC_MODES, 0)) {
+          return csr_value.MODE;
+        }
+      } else if (csr_value.MODE == 1) {
+        if ($array_includes?(VSTVEC_MODES, 1)) {
+          return csr_value.MODE;
+        }
```
ThinkOpenly: Same
```suggestion
      if ($array_includes?(VSTVEC_MODES, csr_value.MODE)) {
        return csr_value.MODE;
      }
```

ThinkOpenly (changes requested):
Tangential, but `mtvec` has some differences from `stvec` and `vstvec`, like "MTVEC_ILLEGAL_WRITE_BEHAVIOR". Do `stvec` and `vstvec` need analogous parameters?

One nit in `mtvec.yaml` where the code could (similarly) be simplified is in `sw_write()` method for the "MODE" field:
```
      if ($array_includes?(MTVEC_MODES, csr_value.MODE)) {
        return csr_value.MODE;
      if (MTVEC_ILLEGAL_WRITE_BEHAVIOR == "retain") {
        return CSR[mtvec].MODE;
      } else if (MTVEC_ILLEGAL_WRITE_BEHAVIOR == "custom") {
        return UNDEFINED_LEGAL_DETERMINISTIC;
      }
      unreachable();
```
Similar changes could also be made in `mtvec.yaml`, "BASE" field.

titoatwork:
Both MTVEC alignment enums contain `0xfff` (4095), although their schema requires a power of two. The sequence is `0x800`, `0xfff`, `0x2000`, and the analogous new STVEC parameter uses `0x1000`. Should both MTVEC entries be `0x1000`?

jordancarlin:
> Both MTVEC alignment enums contain `0xfff` (4095), although their schema requires a power of two. The sequence is `0x800`, `0xfff`, `0x2000`, and the analogous new STVEC parameter uses `0x1000`. Should both MTVEC entries be `0x1000`?

Looks like this was an issue in the old decimal version too (it had 4095) and I just did a raw conversion to hex. I agree that it should be 4096. Updated accordingly.

spec/std/isa/csr/stvec.yaml
jordancarlin: This was copied over from the existing `mtvec` approach. Agreed that this is simpler. I will update all of them accordingly.

jordancarlin:
> Tangential, but `mtvec` has some differences from `stvec` and `vstvec`, like "MTVEC_ILLEGAL_WRITE_BEHAVIOR". Do `stvec` and `vstvec` need analogous parameters?

They will likely eventually need similar parameters, but those *_ILLEGAL_WRITE_BEHAVIOR parameters are not all that useful in their current form and I expect them to change as the Parameters SIG progresses through its work, so I'm inclined not to add them yet unless you feel strongly about them.

ThinkOpenly:
> They will likely eventually need similar parameters, but those *_ILLEGAL_WRITE_BEHAVIOR parameters are not all that useful in their current form and I expect them to change as the Parameters SIG progresses through its work, so I'm inclined not to add them yet unless you feel strongly about them.

No worries.

ThinkOpenly (approved):
Thanks!

PR #1734, opened 2026-03-16
docs: add AGENTS.md and extract-instructions-from-subsection skill

adingank-qualcomm:
### AGENTS

- Adds `AGENTS.md` to provide Generative AI tools such as Claude Code with project-specific guidance
- Documents common commands for regression testing, generation, and development chores
- Describes repository structure, data model, Ruby gem architecture, IDL, backends, and CI/pre-commit workflow

### Skill to extract instructions from a subsection of a spec file
- Adds `SKILL.md` to `.claude/skills/extract-instructions-from-subsection/` directory

## Test plan

- [ ] No functional code changes; verify file renders correctly on GitHub
- [ ] Confirm pre-commit hooks pass (YAML/JSON linting, prettier)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2097, opened 2026-07-24
docs: add extract-parameters-from-subsection agent skill

### LINKED ISSUE #2096: Add an extract-parameters-from-subsection agent skill (parameter analogue of the instruction-extraction skill)
## Summary

Add an `extract-parameters-from-subsection` agent skill under `.agents/skills/`, as the parameter analogue of the existing [`extract-instructions-from-subsection`](https://github.com/riscv-software-src/riscv-unified-db/tree/main/.agents/skills/extract-instructions-from-subsection) skill (#1734).

## Motivation

The instruction-extraction skill gives Gen-AI tools a reproducible, repo-native recipe for pulling one class of data (instruction mnemonics) out of the ISA manual and into UDB shape. Architectural parameters are the natural next class, and are the subject of ongoing parameter-extraction work, but there is currently no public, reproducible agent workflow for them in the repo.

Instructions are a *lexical* class (uppercase mnemonics); parameters are *semantic* — a value the ISA leaves to the implementation. So a parameter skill can't just match a token class; it has to match the phrase families the spec uses to delegate a value to the implementation, and guard against hallucinated parameters.

## Proposal

A `SKILL.md` that, given a subsection title + AsciiDoc file:

- scans normative text (skipping NOTE/example blocks) for implementation-defined signals — `UNSPECIFIED`, `implementation-defined`/`-specific`, `WARL`, `platform-specific`, and contextual `the number/size of ...` — grounded in the phrases actually used in `ext/riscv-isa-manual` rather than invented keywords;
- captures the `[[norm:<id>]]` / `[#norm:<id>]#` anchors as provenance;
- requires a **verbatim excerpt** per candidate (anti-hallucination guard);
- classifies each candidate on the two axes the effort tracks — **named/unnamed** and **config-dependent/fixed**;
- emits a JSON-Schema fragment using only bounds the spec states explicitly;
- runs a **recall check** against the curated `spec/std/isa/param/*.yaml` list, tagging each candidate `existing` vs `new`.

Worked examples in the skill round-trip `ASID_WIDTH` (`satp` subsection, `norm:asidlen`) and `CACHE_BLOCK_SIZE` (Memory and Caches, `norm:cache_block_size`) to their existing param files, validating precision.

PR follows.

uditjainstjis:
Closes #2096

## Summary

Adds `.agents/skills/extract-parameters-from-subsection/SKILL.md` — a reproducible AI-coding-agent skill that extracts implementation-defined architectural parameters from a named subsection of a RISC-V AsciiDoc spec file and writes UnifiedDB parameter-candidate YAML.

It is the parameter analogue of the existing `extract-instructions-from-subsection` skill (#1734). Instruction extraction targets a *lexical* class (uppercase mnemonics); parameter extraction is *semantic*, so this skill matches the phrase families the spec actually uses to hand a value to the implementation, and is built around avoiding hallucinated parameters.

## What the skill does

- **Signal matching** — scans normative text (skipping NOTE / example blocks) for implementation-defined signals: `UNSPECIFIED`, `implementation-defined`/`-specific`, `WARL`, `platform-specific`, plus contextual `the number of …` / `the size of …`. The keyword families are grounded in phrases actually present in `ext/riscv-isa-manual`, not invented.
- **Provenance** — captures the `[[norm:<id>]]` / `[#norm:<id>]#` anchors the manual uses to tag normative sentences.
- **Anti-hallucination** — every candidate must carry a **verbatim excerpt** from the subsection; no excerpt, no candidate.
- **Classification** — tags each candidate on the two axes the parameter-extraction effort tracks: **named/unnamed** and **config-dependent/fixed**.
- **Schema** — emits a JSON-Schema-draft-07 fragment, adding `minimum`/`maximum`/`enum` bounds **only** when the spec states them.
- **Recall check** — compares each candidate against the curated `spec/std/isa/param/*.yaml` list and tags it `existing` vs `new`, so the skill doubles as a precision/recall probe on the current database.

## Validation

The worked examples in the skill were extracted with its own rules and round-tripped against committed param files:

- `ASID_WIDTH` — from *"Supervisor Address Translation and Protection (`satp`) Register"* (`norm:asidlen`): signals `is UNSPECIFIED` + `the number of … bits`; named (`ASIDLEN`), config-dependent, `integer [0, 16]`. Matches `spec/std/isa/param/ASID_WIDTH.yaml`.
- `CACHE_BLOCK_SIZE` — from *"Memory and Caches"* (`norm:cache_block_size`): signal `implementation-specific`; unnamed, fixed, `integer`. Matches `spec/std/isa/param/CACHE_BLOCK_SIZE.yaml`.

## Notes

- Single new markdown file under `.agents/skills/`; no code or data files changed.
- Formatted with the repo's pinned Prettier (3.9.6); REUSE lint clean (SPDX header present).

### DIFF
--- a/.agents/skills/extract-parameters-from-subsection/SKILL.md
+++ b/.agents/skills/extract-parameters-from-subsection/SKILL.md
@@ -56,7 +56,7 @@ Scan the normative text of the subsection for sentences that hand a value or beh
 
 - `UNSPECIFIED` / `is UNSPECIFIED` — the spec explicitly declines to fix a value (e.g., "The number of ASID bits is UNSPECIFIED").
 - `implementation-defined` / `IMPLEMENTATION-DEFINED` / `implementation-specific` — an implementation chooses (e.g., "the size of a cache block ... implementation-specific").
-- `WARL` (Write-Any-Read-Legal) — a field whose set of legal values is implementation-chosen; the _set_ is a parameter.
+- `WARL` (Write-Any-Read-Legal) — a parameter **only when the set of legal values is itself implementation-chosen**. The `WARL` label on its own is not a signal: a field can be WARL while the ISA fixes the legal set completely, in which case there is no implementation choice and no parameter. `mtvec.MODE` is a true positive (which modes are legal is up to the implementation); a WARL field whose only legal value is fixed by the ISA is not. Require the excerpt to show _who chooses the legal set_, not merely that the field is labelled WARL.
 - `platform-specific` / `vendor-defined` / `custom` — decision delegated below the ISA.
 
 **Contextual signals — a parameter only when paired with a quantity or a named field:**
@@ -65,6 +65,16 @@ Scan the normative text of the subsection for sentences that hand a value or beh
 - `the size of <X>` / `the width of <X>` — a size/width parameter.
 - `reserved` — a parameter only when the reservation is _for implementation use_; plain "reserved, must be zero" encodings are **not** parameters and must be excluded.
 - `may be` / `may support` / `optionally` guarding a concrete value or range.
+
+**Negative controls — text that looks like a signal but is not a parameter.** Each of these has produced a plausible-looking false positive, so check a candidate against this list before keeping it:
+
+- **WARL with an ISA-fixed legal set** — see above. The word `WARL` is present, the legal values are not implementation-chosen.
+- **Fixed encoding or address conventions** — a CSR number, an opcode, a field position. Fixed by the ISA for every implementation, so nothing is delegated.
+- **Advice aimed at software** — `should` / `may` / `is recommended` addressed to compilers, assemblers or programmers describes how to _use_ the ISA, not a hardware choice.
+- **A bare `shall` constraint** — a requirement every implementation must meet delegates nothing. A constraint becomes a parameter only when the spec also says the implementation picks the value.
+- **`reserved, must be zero`** — a fixed encoding, not a reservation for implementation use.
+
+A candidate that matches one of these must be dropped, not downgraded.
 
 **Named-value anchors:** the spec often tags the exact normative sentence with an inline anchor — a block anchor `[[norm:<id>]]` on its own line, or an inline anchor `[#norm:<id>]#...#` wrapping the sentence. When present, capture the `<id>`; it is the authoritative provenance link for the parameter and pairs the candidate to its normative text.
 
@@ -75,7 +85,9 @@ For every signal that survives step 3, build a candidate with:
 - **name** — an UPPER_SNAKE_CASE identifier in UnifiedDB style. Prefer the spec's own term when it names one (`ASIDLEN` → `ASID_WIDTH`, `PMLEN` → `PMLEN`); otherwise synthesize a descriptive name from the noun phrase (`CACHE_BLOCK_SIZE`).
 - **excerpt** — the verbatim sentence(s) from the subsection that justify the parameter. This is mandatory: no excerpt, no candidate. This is the anti-hallucination guard — a reviewer must be able to confirm the parameter against the copied text without re-reading the spec.
 - **source** — `file`, `subsection`, and the `norm:<id>` anchor if one exists.
-- **definedBy** — the extension(s) that introduce the parameter, as a `condition` per `param_schema.json` (a single `extension: {name: ...}`, or `anyOf`/`allOf` of extension names). Infer from the subsection's owning extension.
+- **definedBy** — the extension(s) that introduce the parameter, as a `condition` per `param_schema.json` (a single `extension: {name: ...}`, or `anyOf`/`allOf` of extension names). Infer from the subsection's owning extension **only when that extension is unambiguous**. The inference is unsafe in two common cases; in both, emit the best guess with `uncertain: true` and an `open_question` rather than asserting it:
+  - the parameter is gated by several extensions, none of which owns the subsection — `CACHE_BLOCK_SIZE` lives in a CMO subsection but is defined by `anyOf` Zicbom/Zicbop/Zicboz;
+  - applicability depends on another **parameter** rather than an extension — e.g. how many PMP entries are usable follows `NUM_PMP_ENTRIES`, which no `definedBy` condition expresses.
 - **schema** — a JSON-Schema-draft-07 fragment for the legal values. Emit `type: integer` / `type: boolean` / `enum: [...]`. **Only** add `minimum`/`maximum`/`enum` bounds that the spec states explicitly (e.g., "ASIDMAX ... is 9 for Sv32 or 16 for Sv39/48/57" → `maximum: 16`). If the spec gives no bound, leave the type unbounded — never invent limits.
 - **classification** (the two axes the parameter-extraction effort tracks):
   - `named`: `true` if the spec gives the value an explicit name (`ASIDLEN`, `PMLEN`), else `false`.
@@ -87,12 +99,22 @@ For every candidate, check whether UnifiedDB already models it: list `spec/std/i
 
 - `status: existing` — already in `spec/std/isa/param/` (report the matching filename). Useful as a precision/recall check on the extractor.
 - `status: new` — not yet modeled; a genuine recall candidate for a human to review and add.
+- `status: possible_duplicate` — a plausible match exists but the correspondence is not one-to-one: the existing file uses a different name for the same concept, the candidate covers only part of an existing parameter, or one candidate spans several existing files. Name the file(s) considered and state what is unresolved.
+
+Matching by name alone is not sufficient — renames and conceptual duplicates will not match textually. When in doubt use `possible_duplicate` rather than guessing: a wrong `new` inflates the apparent recall gap, and a wrong `existing` hides a real one.
 
 Do not overwrite or propose edits to existing files from this skill; it only surfaces candidates for review.
 
-### 7. Write the output
+### 7. Write the review envelope
 
-Derive the output filename from argument 1 (lowercased, spaces → hyphens). Write `/tmp/<derived-name>.yaml` in this review-friendly shape (a superset of `param_schema.json` — it carries the schema-relevant fields plus `excerpt`, `source`, `classification`, and `status` for the reviewer):
+This skill produces **two artifacts, in order**, and stops after the first:
+
+1. **The review envelope** — what is written here. It carries the evidence a reviewer needs: excerpt, anchor, classification, status, uncertainty and open questions.
+2. **Mergeable UnifiedDB YAML** — one `param_schema.json`-valid file per parameter, written **only after a human has accepted the candidate**, with every review-only key stripped.
+
+Keeping them separate matters because the envelope is deliberately **not** a valid UnifiedDB parameter file and must not be presented as one: it nests candidates under a `parameters:` wrapper that is outside the parameter object schema, and it omits the required `kind`, `long_name` and `description`. Emitting it is a review step, not a merge-ready result.
+
+Derive the output filename from argument 1 (lowercased, spaces → hyphens). Write `/tmp/<derived-name>.yaml` in this shape:
 
 ```yaml
 parameters:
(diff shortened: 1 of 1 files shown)


titoatwork:
Thanks for landing a parameter analogue of the instruction-extraction skill — the signal families, NOTE-skipping, and mandatory verbatim excerpt are the right anti-hallucination core for Part II-style work.

I reproduced the skill *rules* (not a paid re-run of the agent) against a small frozen adversarial pack built from existing open corpus/challenge fixtures in https://github.com/titoatwork/lfx-firstanalysis (`riscv-param-extraction/workflow_slice/eval_2097/`). Cases:

**Positives (5)** — DIRECT / CSR_RW / WARL
- `CACHE_BLOCK_SIZE` (implementation-specific cache block)
- `NUM_PMP_ENTRIES` (implementation-defined count)
- `MTVEC_ACCESS` (RO vs RW choice)
- `MTVEC_MODES` (true WARL: legal MODE set is implementation-chosen)
- `ASID_WIDTH` / ASIDLEN-style width

**Negatives (4)**
- Fixed CSR address-mapping convention (shared challenge negative)
- Software/compiler "should/may" advice (not a hardware parameter)
- Pure shall constraint without implementation delegation
- **WARL vocabulary with an ISA-fixed legal set** (legal value must be 0 for all implementations; not implementation-defined)

### Concrete observations

1. **WARL as almost-always-positive signal**
   The skill ranks `WARL` under "almost always name a parameter." That matches many real cases (`MTVEC_MODES`), but the fixed-legal-set negative shows a failure mode: the *word* WARL can appear where the legal *set* is fully fixed by the ISA. WARL-field misclassification is already a central Part II / Spring pain point; a skill aimed at parameters should require that the **set of legal values** is implementation-chosen, not merely that the field is labeled WARL.

2. **"Superset of param_schema" vs mergeable UDB files**
   The documented output is a top-level `parameters:` list with `excerpt` / `source` / `classification` / `status`. That is a valuable **review envelope**, but it is not a file that validates as `param_schema.json` (missing required fields such as `kind`, `long_name`, `description`, and the wrapper key is outside the parameter object schema). Calling it a schema superset can over-promise merge readiness.

3. **"Compare by name and meaning" for existing/new**
   Duplicate detection is underspecified for renames and conceptual duplicates. Spring Part I needed curated one-to-many alignments; a skill step should either point at an explicit alias/alignment table or mark uncertain duplicates as `possible_duplicate` rather than binary existing/new.

4. **`definedBy` from subsection ownership**
   Inferring `definedBy` only from the subsection's owning extension is unsafe for multi-extension gates (e.g. CMO `anyOf`) and param-gated parameters (e.g. usable PMP count depending on `NUM_PMP_ENTRIES`).

5. **No automated validation / negatives in the skill itself**
   The two worked examples help, but without negative controls and a machine check that (a) every candidate has an excerpt present in the subsection and (b) post-review emission validates as UDB YAML, regressions will be hard to catch.

### Suggested two-artifact design

Separate the products of the skill:

1. **Review envelope** (what the skill should emit first): evidence excerpt, anchor, classification, confidence/uncertainty, existing/new/possible-duplicate, proposed name, proposed `definedBy`, open questions.
2. **Clean UDB parameter YAML** (param_schema-valid only): emitted **after** human approval — no review-only keys, no `parameters:` wrapper.

Happy to contribute the frozen fixtures + a small validator (`workflow_slice/eval_2097/`) in whatever form is preferred (subtree under the skill, separate test path, or follow-up PR). No need to take any of this as blocking if you already planned a review-envelope split — flagging so Part II WARL and schema-validity expectations stay aligned.

Reproduction (no API):

```bash
cd riscv-param-extraction
python workflow_slice/eval_2097/scripts/validate_eval_pack.py
python workflow_slice/scripts/ci_slice_check.py
```

PATTASWAMY-VISHWAK-YASASHREE:
came here after exploring the phase 2 of the lfx program of the same topic

PATTASWAMY-VISHWAK-YASASHREE:
and since i am a ai junkie i am intrested in this and would love to know more about the AI-assisted extraction of architectural parameters from RISC-V specifications - Part II

titoatwork:
Following up on the review above with something concrete rather than more commentary.  The frozen fixtures are now public under `riscv-param-extraction/workflow_slice/eval_2097/` in https://github.com/titoatwork/lfx-firstanalysis. It holds 5 positives (DIRECT / CSR_RW / WARL) and 4 negatives, including the case I think is the sharpest for a parameter skill: WARL vocabulary where the legal value set is fixed by the ISA, so no implementation choice exists and it isn't an architectural parameter. The validator runs offline, no API calls:  ```bash cd riscv-param-extraction python workflow_slice/eval_2097/scripts/validate_eval_pack.py ```  Happy to contribute these in whatever shape suits you: a subtree under the skill, a separate test path, or a follow-up PR once this lands. Equally happy to leave them out if you'd rather keep the skill self-contained; nothing here is blocking.

uditjainstjis:
This is a genuinely useful review — thank you for putting real cases behind it rather than opinions. All five points are now in the skill (9d88fa4).

**1. WARL.** You're right, and this was the weakest rule in the file. `WARL` is now a signal only when the *set of legal values* is implementation-chosen; the label alone no longer qualifies. The rule asks the excerpt to show **who chooses the legal set**, with `mtvec.MODE` as the positive and an ISA-fixed legal set as the explicit negative.

**2. "Superset of param_schema".** Dropped — you're right that it over-promised merge readiness. Rewritten as a **two-artifact** flow, which matches what you suggested: the skill emits a *review envelope* and stops, and `param_schema.json`-valid YAML is a second artifact written only after a human accepts a candidate. The text now names why the envelope is not valid (the `parameters:` wrapper is outside the parameter object; `kind` / `long_name` / `description` are missing) so nobody mistakes it for mergeable output.

**3. Duplicates.** Added `status: possible_duplicate` for renames, partial overlaps and one-to-many alignments, with a note that name matching alone will not catch conceptual duplicates, and that guessing is worse than flagging — a wrong `new` inflates the recall gap, a wrong `existing` hides one.

**4. `definedBy`.** Now inferable from the owning subsection only when the extension is unambiguous. Both of your cases are called out by name: multi-extension gates (`CACHE_BLOCK_SIZE` sitting in a CMO subsection but defined by `anyOf` Zicbom/Zicbop/Zicboz) and parameter-gated applicability (usable PMP entries following `NUM_PMP_ENTRIES`). Neither may be asserted — they carry `uncertain: true` plus an open question.

**5. Negative controls and validation.** There is now an explicit negative-control list in step 4 covering all four of your negatives, plus a mechanical self-check step before reporting: every excerpt must be found verbatim in the subsection, must not originate in a skipped `[NOTE]`/example block, every candidate must cite the signal it matched, and drop counts are reported rather than hidden.

On the fixtures: the two skills in this repo are each a single `SKILL.md` with no test subtree, so a fixtures directory underneath one would be the first of its kind here — that's a maintainer call rather than mine. The substance is in the skill either way, as negative controls and self-checks. I'd rather your eval pack landed as **your own PR** than as a subtree of mine; it's your work, it deserves its own attribution, and if the maintainers want the skill to point at it, that's a one-line follow-up. Happy to review it if you open one.

titoatwork:
Thanks for working through all five, and for the reasoning on each rather than just the edits.

Two of the changes I think are stronger than what I suggested. Requiring the excerpt to show *who chooses the legal set* is a sharper test than my "the set must be implementation-chosen", because it is checkable against the text rather than requiring a judgement call. And on duplicates, stating that a wrong `new` inflates the recall gap while a wrong `existing` hides one is the asymmetry I did not articulate; it is a better argument for `possible_duplicate` than the one I gave.

Agreed on the fixtures, and thank you for saying so. I have raised the placement question as #2158 rather than opening a PR into your branch, since a test subtree under a skill would be the first of its kind and that is a maintainer call. If they want it, I will open it separately and it can point at the skill or not as they prefer. If they would rather skills stay documentation-only, that is a reasonable answer too and the pack costs nothing where it is.

Either way this PR no longer depends on it. The negative controls and the self-check step cover the failure modes the fixtures were testing for, which was the actual point.

uditjainstjis:
Revising the WARL rule in 72d18f7 — I had it on the wrong side of the line.

A field declared WARL or WLRL **is** a parameter; the parameter is its set of legal values. Requiring the excerpt to first prove the set is implementation-chosen rejects candidates that should instead carry a legal-value classification, with a degenerate classification covering the case where the ISA fixes the set. The self-check now asks for a classification rather than for proof of implementation choice.

It also cuts the other way, which I had missed entirely: the `WARL`/`WLRL` tokens **under-report**. A read/write field that is fixed to zero under some circumstances has a legal-value set that depends on the implementation or configuration, with no WARL token anywhere in the prose. Conditional-writability phrasing now counts as a parameter too.

titoatwork:
Checked the reframe against `param_schema.json` before replying, and it holds. Nine parameters already encode a legal-value set directly: `MSTATUS_FS_LEGAL_VALUES` and `MSTATUS_VS_LEGAL_VALUES` as `enum: [0,1,2,3]` with `minItems: 1, maxItems: 4`, plus `MTVEC_MODES`, `STVEC_MODES`, `VSTVEC_MODES`, `SXLEN`, `UXLEN`, `VSXLEN` and `VUXLEN`. So "the parameter is its set of legal values" is an existing UnifiedDB shape rather than a reframe that needs arguing for.

You were also right that requiring proof was the wrong instrument, and my own review contradicted itself on it. If a human accepts every candidate before any YAML is written, the filtering belongs at that gate, which is what the two-artifact point argued for two paragraphs earlier. Rejecting at extraction time only buys silent misses.

On the under-reporting I tried to size it. Grepping `ext/riscv-isa-manual` for conditional-writability phrasing, prose only with tables and captions dropped, gives 63 sentences carrying no `WARL` or `WLRL` token against 5 that do. The direction is clearly yours.

The 63 are not uniform though, and I think that matters more than the total. At least 19 are conditioned on an extension not being implemented, like "If Svpbmt is not implemented, PBMTE is read-only zero". UnifiedDB models that with `definedBy: {extension: {name: Svpbmt}}` and no `type()` at all, and there is no PBMTE parameter, nor one for STCE, CBZE, CBCFE or ADUE. That is extension gating rather than an implementation choice, so the rule as written would admit it as a legal-value parameter.

The remainder splits two ways rather than one, and your mechanism is exactly right for one of them. `vsstatus.UXL` is the case you described: `type()` is `$array_size(VUXLEN) > 1 ? CsrFieldType::RW : CsrFieldType::RO`, so the cardinality of a legal-value set decides writability, and `VUXLEN` is one of the nine. But the endianness fields resolve to `M_MODE_ENDIANNESS` and its four siblings as `['little','big','dynamic']`, and the counter fields to `HPM_COUNTER_EN` and `COUNTINHIBIT_EN` as boolean arrays. Those are mutability parameters, which `taxonomy.md` places under `NORM_CSR_RW`, "about the mutability of the field, not the set of legal values".

So the signal is worth adding and its output is three-way: extension gating, a legal-value set, or a read-write-behaviour parameter. A test that worked mechanically on the cases I looked at is whether the condition is a function of a set's cardinality. If it is, the parameter is the set. If it is a plain predicate, it is a mutability parameter. If it is an extension being absent, it is not a parameter at all.

Two structural things make either rule hard to execute as the file stands. The envelope has no key for the class, only `classification` with `named` and `config_dependent`, so neither `NORM_CSR_WARL` nor `NORM_CSR_RW` can be recorded. And step 5 tells the model to emit `type: integer` / `type: boolean` / `enum: [...]`, while all nine set parameters are `type: array` with `items.enum`, `minItems` and `maxItems`, so a correctly classified WARL candidate has no shape to emit.

The ambiguity underneath this is not yours. `taxonomy.md` defines `NORM_CSR_WARL` as "the spec describes a CSR field as WARL and the legal values are implementation-defined", while its decision tree at step 4 asks only whether the text describes a WARL field's legal values, and its `NORM_CSR_RW` section lists "MBE, SBE, and UBE bits in `mstatus` are WARL fields" as a read-write case. Your revision follows the tree, my review followed the class definition, and the same excerpt classifies differently depending on which half you read. Raised as #2200 rather than litigating it here.

My eval pack has the same problem in the other direction and I am fixing it. It only has positives and negatives, so a candidate that should be extracted and then classified out has nowhere to sit, and it tests over-firing while your finding is under-firing. Adding a candidate category and recall fixtures from those 63.

PR #2100, opened 2026-07-24
schema(inst): enforce access_detail when any access mode is 'sometimes'

### LINKED ISSUE #2098: schema(inst): inst_schema.json missing conditional constraint requiring access_detail when access mode is 'sometimes'
### Description

In `spec/schemas/inst_schema.json`, the `access_detail` property is defined with the description: *"Extra detail about access when at least one mode is 'sometimes'"*.

However, the JSON schema does not enforce that `access_detail` MUST be present when any access mode (`m`, `s`, `u`, `vs`, `vu`) is set to `sometimes`. As a result, instruction YAML files can declare `sometimes` access without providing `access_detail`, passing schema validation silently while leaving the generated documentation incomplete.

### Repro / Impact

A systematic audit revealed that 4 instructions in `spec/std/isa/inst/Svinval/` (`sinval.vma.yaml`, `hinval.gvma.yaml`, `sfence.w.inval.yaml`, and `sfence.inval.ir.yaml`) have `access: sometimes` but completely lack `access_detail`.

### Proposed Solution

1. Add a conditional `allOf` / `if-then` rule to `spec/schemas/inst_schema.json` requiring `access_detail` when any mode in `access` is set to `sometimes`.
2. Update affected instruction YAML files (`Svinval` extension instructions) with appropriate `access_detail` documentation.

krrishverma1805-web:
Fixes #2098

### Summary of Changes

1. **Schema ()**: Added a conditional `allOf` (`if-then`) rule requiring `access_detail` whenever any mode in `access` (`m`, `s`, `u`, `vs`, `vu`) is set to `sometimes`.
2. **Data ()**: Added missing `access_detail` documentation to `sinval.vma.yaml`, `hinval.gvma.yaml`, `sfence.w.inval.yaml`, and `sfence.inval.ir.yaml`.

### Verification

- `./do test:schema` passes.
- `./do test:sorbet` passes.
- `./do test:idl CFG=_` passes.

### DIFF
--- a/doc/docs/schemas/index.mdx
+++ b/doc/docs/schemas/index.mdx
@@ -54,6 +54,11 @@ Schemas are versioned independently of the UDB repository. Each schema file decl
 </div>
 
 <div style={{padding: '1rem', border: '1px solid var(--ifm-color-emphasis-300)', borderRadius: '8px'}}>
+  <strong><a href="./v0.3/inst_schema">Inst Schema</a></strong><br />
+  <span className="badge badge--secondary" style={{marginTop: '0.4rem', display: 'inline-block'}}>v0.3</span>
+</div>
+
+<div style={{padding: '1rem', border: '1px solid var(--ifm-color-emphasis-300)', borderRadius: '8px'}}>
   <strong><a href="./v0.1/inst_subtype_schema">Inst Subtype Schema</a></strong><br />
   <span className="badge badge--secondary" style={{marginTop: '0.4rem', display: 'inline-block'}}>v0.1</span>
 </div>
--- a/doc/docs/schemas/v0.1/_category_.json
+++ b/doc/docs/schemas/v0.1/_category_.json
@@ -1,6 +1,6 @@
 {
   "label": "v0.1",
-  "position": 2,
+  "position": 3,
   "collapsible": true,
   "collapsed": false
 }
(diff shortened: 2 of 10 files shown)


doc/docs/schemas/v0.2/inst_schema.mdx
```diff
@@ -91,52 +98,6 @@ Decode field
 
 
 
-## Properties
```
ThinkOpenly, line 94: Strange. Do you understand why the whole "Properties" section was deleted?

krrishverma1805-web:
The `## Properties` section was removed by `autofix.ci` automatically regenerating `inst_schema.mdx` via `bin/chore gen schema-docs` after the `allOf` was added to the top level of `inst_schema.json`.

Root cause: in `schema_doc_gen.rb` (line 78), the generator only renders the `## Properties` table when `allOf`/`anyOf`/`oneOf` are **absent** at the top level — when `allOf` is present, it switches to the `## Schema Structure` composite rendering mode instead. Since our PR adds `allOf` at the top level of `inst_schema.json` (to enforce the `access_detail` conditional), the generator intentionally drops the flat properties table and produces a 'Schema Structure: all of' representation.

The schema definition itself is unchanged — all properties are still defined. Only the doc rendering mode changed due to the generator's current design. This is the expected output from the generator given the schema structure.

PR #2103, opened 2026-07-24
Improve bounds for STVAL_WIDTH parameter

### LINKED ISSUE #2102: STVAL_WIDTH has no minimum and a nonsensical 2^64-1 maximum for a bit-width (should mirror MTVAL_WIDTH: 0..64)
## Describe the bug

`spec/std/isa/param/STVAL_WIDTH.yaml` bounds the parameter as:

```yaml
schema:
  type: integer
  maximum: 18446744073709551615 # 2^64 - 1
```

with **no `minimum`**. `STVAL_WIDTH` is "the number of implemented bits in `stval`", so its value is a *bit width*, not a byte count. The privileged spec states the register is at most `SXLEN` bits wide:

> "The `stval` CSR is an SXLEN-bit read-write register ..." — `riscv-isa-manual` `src/priv/supervisor.adoc` (norm anchor near line 709)

Since `SXLEN` is 32 or 64, a `stval` width of `2^64 - 1` bits is not representable and the upper bound is meaningless. The `2^64 - 1` idiom is appropriate for a byte-size parameter (e.g. `CACHE_BLOCK_SIZE`) but was misapplied to a bit-width here.

## Expected

The bound should match the M-mode twin `MTVAL_WIDTH.yaml`, which models the identical concept (implemented bits in `mtval`) correctly:

```yaml
schema:
  type: integer
  minimum: 0
  maximum: 64
```

`STVAL_WIDTH` should be brought to parity: `minimum: 0`, `maximum: 64`. Its `long_name` is also the placeholder `TODO`, where `MTVAL_WIDTH` has `Width of the mtval CSR`.

## Proposed fix

Bring `STVAL_WIDTH` to parity with `MTVAL_WIDTH`:

- `schema.minimum: 0`, `schema.maximum: 64`
- `long_name: "Width of the `stval` CSR"`

## Follow-up (needs maintainer confirmation, not in this fix)

`MTVAL_WIDTH` additionally carries an IDL requirement `MXLEN == 32 -> MTVAL_WIDTH <= 32`. The `stval` analogue would be `SXLEN == 32 -> STVAL_WIDTH <= 32`. I've left this out of the fix pending confirmation that `SXLEN` is available in the `requirements.idl()` evaluation context here (and that this constraint is wanted), to avoid an unverified IDL typecheck change — happy to add it in a follow-up.

uditjainstjis:
parameter "STVAL_WIDTH" should be constrained to at most the maximum width of a virtual address.

Also, add a "long_name".

Closes #2102


spec/std/isa/param/STVAL_WIDTH.yaml
```diff
@@ -10,10 +10,12 @@ description: |
   The number of implemented bits in `stval`.
 
   Must be greater than or equal to _max_(`PHYS_ADDR_WIDTH`, `VA_SIZE`)
-long_name: TODO
+long_name: |
+  Width of the `stval` CSR
```
ThinkOpenly: nit, but please make this a single line.

ThinkOpenly (changes requested):
> `MTVAL_WIDTH` also carries `MXLEN == 32 -> MTVAL_WIDTH <= 32`. The `stval` analogue would be `SXLEN == 32 -> STVAL_WIDTH <= 32`; I left it out pending confirmation that `SXLEN` is available in the `requirements.idl()` context here (noted in #2102).

Since this param is "definedBy: ... S", `SXLEN` is available here.

uditjainstjis:
Both done in 790be26.

`long_name` is a single line now.

On the requirement — thanks, that confirmation was the piece I was missing. One adjustment: `SXLEN` is a set rather than a scalar (`schema: type: array`, `[32]` / `[64]` / `[32, 64]`), so `SXLEN == 32` isn't the right shape here. I used the idiom already in `SXLEN.yaml` and `GSTAGE_MODE_BARE.yaml`:

```yaml
requirements:
  idl(): |
    $array_includes?(SXLEN, 32) -> STVAL_WIDTH <= 32;
  reason: |
    `stval` is an SXLEN-bit register, so it cannot hold more than 32 bits
    whenever S-mode can run with SXLEN == 32
```

which covers the `[32, 64]` case too — if S-mode can ever run with `SXLEN == 32`, `stval` has to fit in 32 bits. Happy to switch to a plain equality if you'd rather it only constrain the fixed-32 configuration.

I left out the S-mode analogues of `MTVAL_WIDTH`'s `REPORT_VA_IN_MTVAL_ON_*` clauses — say the word if you want those mirrored here as well and I'll add them in this PR.

spec/std/isa/param/STVAL_WIDTH.yaml
```diff
@@ -10,10 +10,17 @@ description: |
...
   type: integer
-  maximum: 18446744073709551615
+  minimum: 0
+  maximum: 64
+requirements:
+  idl(): |
+    $array_includes?(SXLEN, 32) -> STVAL_WIDTH <= 32;
+  reason: |
+    `stval` is an SXLEN-bit register, so it cannot hold more than 32 bits
+    whenever S-mode can run with SXLEN == 32
```
ThinkOpenly: I don't read the spec as having the requirement expressed here. If `SXLEN` _can be_ greater than 32, then `stval` can also be greater than 32, to my reading. I don't see anything that contradicts that. Per the spec:
> `stval` [...] must be able to hold all valid virtual addresses

If SXLEN can be 64, then UXLEN could be 64, and virtual addresses could be 64 bits wide.
```suggestion
```

ThinkOpenly (changes requested):
Nice catch on the minimum/maximum width, but let's remove the added requirement.

uditjainstjis:
Removed in aca925a — you're right, and my reasoning was backwards on exactly the case I claimed it handled.

I argued `$array_includes?` "covers the `[32, 64]` case too", but that's the case it gets wrong: if SXLEN can be 64 then UXLEN can be 64, virtual addresses can be 64 bits, and `stval` has to hold them. Requiring `<= 32` merely because 32 is *also* a legal SXLEN forbids a perfectly valid implementation. The MTVAL analogue only works because `MXLEN` is a fixed scalar — the moment the S-mode equivalent became a set, the analogy stopped holding and I carried it across anyway.

The PR is now just the `0..64` bounds and the single-line `long_name`.

spec/std/isa/param/STVAL_WIDTH.yaml
```diff
@@ -10,10 +10,11 @@ description: |
   The number of implemented bits in `stval`.
 
   Must be greater than or equal to _max_(`PHYS_ADDR_WIDTH`, `VA_SIZE`)
-long_name: TODO
+long_name: Width of the `stval` CSR
```
ThinkOpenly: Sorry, missed this before... to use more precise language:
```suggestion
long_name: `stval` implemented bits
```

ThinkOpenly (changes requested):

uditjainstjis:
Applied, thanks — no need to apologise, more precise is better.

ThinkOpenly:
[_original PR description message before editing_]
Closes #2102

## Summary

Brings `spec/std/isa/param/STVAL_WIDTH.yaml` to parity with its M-mode twin `MTVAL_WIDTH.yaml`, which models the identical concept correctly.

`STVAL_WIDTH` is "the number of implemented bits in `stval`" — a **bit width**. The privileged spec defines the register as at most `SXLEN` bits:

> "The `stval` CSR is an SXLEN-bit read-write register ..." — `riscv-isa-manual` `src/priv/supervisor.adoc`

Since `SXLEN` is 32 or 64, the file's `maximum: 18446744073709551615` (2^64 − 1) with **no `minimum`** admitted meaningless values — a byte-size idiom (fine for `CACHE_BLOCK_SIZE`) misapplied to a bit width.

## Change

| | before | after |
|---|---|---|
| `schema.minimum` | *(absent)* | `0` |
| `schema.maximum` | `2^64 − 1` | `64` |
| `long_name` | `TODO` | `Width of the `stval` CSR` |

These now match `MTVAL_WIDTH.yaml` (`minimum: 0`, `maximum: 64`, real `long_name`), the same-purpose parameter for `mtval`.

## Validation

- YAML parses; all `param_schema.json` required fields present; `schema` resolves to `{type: integer, minimum: 0, maximum: 64}`.
- Prettier (pinned 3.9.6) clean.
- Header (copyright + `yaml-language-server`) preserved; one-file, data-only change.

## Follow-up (not in this PR)

`MTVAL_WIDTH` also carries `MXLEN == 32 -> MTVAL_WIDTH <= 32`. The `stval` analogue would be `SXLEN == 32 -> STVAL_WIDTH <= 32`; I left it out pending confirmation that `SXLEN` is available in the `requirements.idl()` context here (noted in #2102). Happy to add it as a follow-up.

uditjainstjis:
The 19 red `regress-udb-unit-test` jobs weren't from this change: the `[autofix.ci]` commit on this branch deleted `tests/data/golden/Xqci@0.13.0.adoc` (~41k lines), which is still present on `main`. Restored it verbatim in 3d08600, so the diff against `main` is again just the five lines in `STVAL_WIDTH.yaml`.

Possibly the same interaction #2183 is addressing.

titoatwork:
Two pieces of corroboration for the bound you landed on, from a sweep I was running over the parameter files for something else.

`STVAL_WIDTH` is the only one of its kind. Of the nine `*_WIDTH` parameters, it is the sole one with no `minimum`, and the sole one whose `maximum` is not a plausible bit count:

```
ASID_WIDTH          0..16     MTVAL_WIDTH      0..64
DBG_HCONTEXT_WIDTH  0..14     PHYS_ADDR_WIDTH  1..64
DBG_SCONTEXT_WIDTH  0..32     RCID_WIDTH       1..12
MCID_WIDTH          1..12     VMID_WIDTH       0..14
STVAL_WIDTH         (none)..18446744073709551615
```

So `minimum: 0, maximum: 64` is not only the `MTVAL_WIDTH` analogue, it is the only value consistent with the other eight.

The missing `minimum` also has a concrete effect today. Validating candidate values against the current schema:

```
-1  accepted      0  accepted     64  accepted
65  accepted    2^63  accepted
```

A negative bit count validates, which the added `minimum: 0` fixes independently of the maximum.

For the ceiling itself, `spec/std/isa/csr/stval.yaml` declares `length: 64` with a single `VALUE` field at `63-0`, so 64 is the structural limit regardless of how the virtual-address wording is read.

No action needed from me here, just evidence in case it is useful for closing this out.

uditjainstjis:
That's a much better argument for the bound than mine was — thanks for digging it out.

The `-1 accepted` case is the part I hadn't checked; I'd been treating the missing `minimum` as tidiness rather than something that actually validates a negative bit count today. And `stval.yaml` declaring `length: 64` with `VALUE` at `63-0` settles the ceiling structurally, without needing the virtual-address reading at all.

ThinkOpenly (approved):

PR #2109, opened 2026-07-25
fix(csr): simplify reset_value and sw_write methods in mstatus.VS

### LINKED ISSUE #1057: Cleanup on aisle `mstatus.VS` methods
Unrelated to PR #999, but that PR seemed to expose that the `type()`, `reset()`, and `sw_write()` methods have extra code that can be similified to something like:
```
      if (CSR[misa].V == 1'b1 || CSR[misa].S == 1'b1) {
        return ary_includes?<$array_size(MSTATUS_VS_LEGAL_VALUES), 2>(MSTATUS_VS_LEGAL_VALUES, csr_value.VS) ? csr_value.VS : UNDEFINED_LEGAL_DETERMINISTIC;
      } else {
        # must be read-only-0
        return 0;
      }
```
(Whether the current code is correct or not is another question.)

_Originally posted by @ThinkOpenly in https://github.com/riscv-software-src/riscv-unified-db/pull/999#discussion_r2285707061_

krrishverma1805-web:
Simplify the `reset_value()` and `sw_write(csr_value)` method definitions for the `VS` field in `mstatus.yaml`:
- `sw_write(csr_value)` had two separate branches returning the exact same `$array_includes?(MSTATUS_VS_LEGAL_VALUES, csr_value.VS) ? csr_value.VS : UNDEFINED_LEGAL_DETERMINISTIC;` evaluation.
- `reset_value()` had redundant nested `$array_size` checks that mirrored the logic handled cleanly by sibling status fields like `FS`.
- The `$array_size(MSTATUS_VS_LEGAL_VALUES) == 0` case was unreachable: MSTATUS_VS_LEGAL_VALUES declares `minItems: 1`.

Closes #1057.

### DIFF
--- a/spec/std/isa/csr/mstatus.yaml
+++ b/spec/std/isa/csr/mstatus.yaml
@@ -499,31 +499,17 @@ fields:
         return $array_size(MSTATUS_VS_LEGAL_VALUES) == 1 ? CsrFieldType::RO : CsrFieldType::RW;
       }
     reset_value(): |
-      if (CSR[misa].V == 1'b1){
-        return UNDEFINED_LEGAL;
-      } else if ((CSR[misa].S == 1'b0) && (CSR[misa].V == 1'b0)) {
+      if ((!implemented?(ExtensionName::S) && !implemented?(ExtensionName::V)) || (MISA_CSR_IMPLEMENTED && (CSR[misa].S == 1'b0) && (CSR[misa].V == 1'b0))) {
         # must be read-only-0
         return 0;
-      } else {
-        # there will be no hardware update in this case because we know the V extension isn't implemented
-        if ($array_size(MSTATUS_VS_LEGAL_VALUES) == 0) {
-          return 0;
-        } else if ($array_size(MSTATUS_VS_LEGAL_VALUES) == 1) {
-          return MSTATUS_VS_LEGAL_VALUES[0];
-        } else {
-          return UNDEFINED_LEGAL;
-        }
       }
+      return $array_size(MSTATUS_VS_LEGAL_VALUES) == 1 ? MSTATUS_VS_LEGAL_VALUES[0] : UNDEFINED_LEGAL;
     sw_write(csr_value): |
-      if (implemented?(ExtensionName::V) && CSR[misa].V == 1'b1){
-        return $array_includes?(MSTATUS_VS_LEGAL_VALUES, csr_value.VS) ? csr_value.VS : UNDEFINED_LEGAL_DETERMINISTIC;
-      } else if (!implemented?(ExtensionName::S) && !implemented?(ExtensionName::V)) {
+      if ((!implemented?(ExtensionName::S) && !implemented?(ExtensionName::V)) || (MISA_CSR_IMPLEMENTED && (CSR[misa].S == 1'b0) && (CSR[misa].V == 1'b0))) {
         # must be read-only-0
         return 0;
-      } else {
-        # there will be no hardware update in this case because we know the V extension isn't implemented
-        return $array_includes?(MSTATUS_VS_LEGAL_VALUES, csr_value.VS) ? csr_value.VS : UNDEFINED_LEGAL_DETERMINISTIC;
       }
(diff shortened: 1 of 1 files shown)


titoatwork (commented):
I checked the two removed branches against the parameter schema and they look safe to drop.

The `$array_size(MSTATUS_VS_LEGAL_VALUES) == 0` case was unreachable: `MSTATUS_VS_LEGAL_VALUES` declares `minItems: 1`, so an empty list cannot validate. And the old first branch, `CSR[misa].V == 1'b1 -> UNDEFINED_LEGAL`, is preserved in effect, because that parameter's `requirements` say `implemented?(ExtensionName::V) -> $array_size(MSTATUS_VS_LEGAL_VALUES) > 1`, so whenever `misa.V` is 1 the size is greater than 1 and the new ternary lands on `UNDEFINED_LEGAL` anyway.

The `MISA_CSR_IMPLEMENTED` guard is a real improvement beyond the simplification. The old `reset_value()` read `CSR[misa].V` unconditionally, while the neighbouring `type()` already guarded that read. That inconsistency goes away here.

One thing that might be worth a second look, since the description says this matches the `FS` pattern. After this change `VS` is actually stricter than `FS` in two places:

- `FS.reset_value()` has no read-only-zero guard at all. It returns `$array_size(...) == 1 ? MSTATUS_FS_LEGAL_VALUES[0] : UNDEFINED_LEGAL` unconditionally, so when `FS` is read-only zero its reset value is not forced to 0. The new `VS.reset_value()` does force it.
- `FS.sw_write()` guards only on `MISA_CSR_IMPLEMENTED && S == 0 && F == 0`. The new `VS.sw_write()` also keeps `!implemented?(S) && !implemented?(V)` from the old code, which covers the case where `misa` is absent entirely.

So `VS` ends up more careful than the pattern it is modelled on, which reads like `FS` may have the same two gaps rather than anything wrong here. Might be worth a follow-up issue against `FS` if that is right.

krrishverma1805-web:
@titoatwork Thank you for the detailed review and thorough analysis! Appreciate you double-checking the parameter schema bounds and confirming the safety of dropping those unreachable branches.

You made a fantastic observation regarding `mstatus.FS`. You're completely spot on — `FS` was missing the explicit read-only-zero guard in `reset_value()` and the `!implemented?(ExtensionName::S) && !implemented?(ExtensionName::F)` check in `sw_write()`. I've submitted a dedicated follow-up PR (#2197) to align `mstatus.FS` with the complete read-only zero handling established here. Thanks again for catching that!

ThinkOpenly (approved):
Thank you!

ThinkOpenly:
## What
Simplify the `reset_value()` and `sw_write(csr_value)` IDL method definitions for the `VS` (Vector context Status) field in `spec/std/isa/csr/mstatus.yaml`.

## Why
Issue #1057 (filed by maintainer @ThinkOpenly) noted that the `reset_value()` and `sw_write()` methods in `mstatus.VS` contained redundant, duplicated conditional logic. Specifically:
- `sw_write(csr_value)` had two separate branches returning the exact same `$array_includes?(MSTATUS_VS_LEGAL_VALUES, csr_value.VS) ? csr_value.VS : UNDEFINED_LEGAL_DETERMINISTIC;` evaluation.
- `reset_value()` had redundant nested `$array_size` checks that mirrored the logic handled cleanly by sibling status fields like `FS`.

## How
1. Simplified `sw_write(csr_value)` to perform a single check for the read-only zero condition (when neither `S` nor `V` is active/implemented), returning `$array_includes?(MSTATUS_VS_LEGAL_VALUES, csr_value.VS)...` otherwise.
2. Simplified `reset_value()` to return `0` when read-only zero and `$array_size(MSTATUS_VS_LEGAL_VALUES) == 1 ? MSTATUS_VS_LEGAL_VALUES[0] : UNDEFINED_LEGAL` otherwise, matching the `mstatus.FS` implementation pattern.

## Testing
- `./do test:schema`: 100% Passed ("All files validate against their schema")
- `./do test:sorbet`: 100% Passed ("No errors! Great job.")
- `./do test:idl CFG=rv64`: 100% Passed
- `./do test:idl CFG=qc_iu`: 100% Passed
- `./bin/pre-commit`: 100% Passed

Closes #1057.

PR #2112, opened 2026-07-25
fix(generators): fail loudly when resolved exception codes cannot be loaded

### LINKED ISSUE #2111: load_exception_codes silently returns an empty list when the resolved codes JSON cannot be parsed
`load_exception_codes` in `backends/generators/generator.py` swallows errors and returns an empty list when the resolved exception codes JSON cannot be read.

The `try` block that parses the file ends in `except Exception as e: logging.error(...)` with no re-raise. Execution then falls through to code at the end of the function that was left behind when the raw YAML fallback was removed in #1233. That trailing code is unreachable in the success and missing-file cases, but the `except` path reaches it, and there it re-sorts the still-empty `exception_codes` list and returns it. So a corrupt or unreadable resolved-codes file produces a header or package with zero `CAUSE_*` definitions, exit code 0, and only an ERROR line in the log.

This is the same silent-bad-output problem as #1145, just on the parse-failure path instead of the missing-file path.

To reproduce:

```
./do gen:sverilog                       # populate gen/resolved_spec/_
echo 'not json' > /tmp/bad.json
uv run backends/generators/sverilog/sverilog_generator.py \
  --inst-dir=gen/resolved_spec/_/inst --csr-dir=gen/resolved_spec/_/csr \
  --ext-dir=gen/resolved_spec/_/ext --resolved-codes=/tmp/bad.json \
  --output=/tmp/out.svh --include-all
grep -c CAUSE_ /tmp/out.svh
```

The command exits 0 and `grep -c` reports 0. `generate_encoding.py` behaves the same way.

Two smaller things in the same function: the missing-file path returns `None`, which surfaces as `TypeError: object of type 'NoneType' has no len()` in the caller rather than a clear message, and `ext_dir`, `enabled_extensions` and `include_all` are now unused since only the dead code referenced them.

AlgoArtist06:
Closes #2111

`load_exception_codes` wrapped the resolved-codes JSON parsing in a bare `except Exception` that logged the error and did not re-raise. Execution then fell through into code left over from when the raw YAML fallback was removed in #1233. That trailing block is unreachable on the success and missing-file paths, but the parse-failure path reached it and returned the still-empty `exception_codes` list. The result was a generated header or SystemVerilog package with zero `CAUSE_*` definitions, exit code 0, and only an ERROR line in the log.

This raises `ValueError` for both the missing and the unparseable case, and deletes the unreachable block. The missing-file case previously returned `None`, which showed up as `TypeError: object of type 'NoneType' has no len()` in the caller rather than a clear message; it is the same root cause and the same few lines, so it is fixed here too.

`ext_dir`, `enabled_extensions` and `include_all` are gone from the signature since only the deleted code referenced them. The `--ext-dir` CLI flags are untouched.

This does not change the fallback behavior discussed in #1145, which was already removed in #1233.

## Testing

Added `backends/generators/test_generator.py` and a `regress-generators-unit` target that runs it. It covers the happy path plus four ways the load can fail: `None`, empty string, missing file, and corrupt JSON.

To confirm the test actually catches this regression, I reintroduced the old swallow-and-return-empty behavior and re-ran it. All four failure cases failed with `Failed: DID NOT RAISE ValueError`, and the happy-path test still passed. Restoring the fix returned all 5 to green.

End to end, with the fix applied, running `sverilog_generator.py` with `--resolved-codes` omitted and again with a deliberately corrupt JSON file both abort with a clear `ValueError` and write no output file. Previously the first crashed with a `TypeError` and the second silently emitted a package with no exception causes.

Ran locally, all passing:

- `./bin/regress -n regress-generators-unit`
- `./bin/regress -n regress-gen-sverilog`
- `./bin/regress -n regress-gen-c-header`
- `./bin/regress -n regress-gen-go`
- `./bin/regress -n regress-sorbet`
- `./bin/prek run --files` on all changed files

`.github/workflows/regress.yml` is regenerated by the `generate regress.yml` hook because of the new target; I did not hand-edit it.

I did not run `./bin/regress --all`, since some targets need Linux-only binaries that are unavailable on macOS.

### DIFF
--- a/.github/workflows/regress.yml
+++ b/.github/workflows/regress.yml
@@ -314,6 +314,16 @@ jobs:
       - name: Run tools/scripts unit tests
         run: ./bin/ruby -Itools/scripts tools/scripts/test/run.rb
 
+  regress-generators-unit:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone Github Repo Action
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - name: Set up mise environment
+        uses: ./.github/actions/mise-setup
+      - name: Run backends/generators unit tests
+        run: uv run pytest backends/generators/test_generator.py
+
   regress-idlc-unit:
     runs-on: ubuntu-latest
     steps:
@@ -951,6 +961,7 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - regress-scripts-unit
+      - regress-generators-unit
       - regress-idlc-unit
       - regress-udb-helpers-unit
       - regress-udb-gen-unit
--- a/backends/generators/c_header/generate_encoding.py
+++ b/backends/generators/c_header/generate_encoding.py
@@ -237,13 +237,7 @@ def main():
     )
 
     # Load exception codes
-    logging.info(f"Loading exception codes from {args.ext_dir}")
-    causes = load_exception_codes(
-        args.ext_dir,
-        args.extensions,
-        include_all=args.include_all,
-        resolved_codes_file=args.resolved_codes,
-    )
+    causes = load_exception_codes(args.resolved_codes)
 
     # Process instructions and calculate masks
     instr_dict = {}
--- a/backends/generators/generator.py
+++ b/backends/generators/generator.py
@@ -488,10 +488,13 @@ def load_csrs(csr_root, enabled_extensions, include_all=False, target_arch="RV64
 def load_exception_codes(resolved_codes_file):
     """Load exception codes from a pre-resolved JSON file.
 
-    Exception code names in the raw YAML contain unresolved ERB templates, so they
-    must be resolved by the Ruby side first (see with_resolved_exception_codes in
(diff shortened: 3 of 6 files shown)


ThinkOpenly:
It's been too long since I've looked here. Where does a pre-resolved exception codes file come from?

AlgoArtist06:
It is a temporary Ruby-to-Python handoff, not a checked-in artifact. `with_resolved_exception_codes` (`backends/generators/tasks.rake:11-44`) writes `[{num, name, var, ext}]` to a `Tempfile` and unlinks it when the generator exits. The only producers are `gen:c_header` (line 94) and `gen:sverilog` (line 123), and both always pass `--resolved-codes`, so no supported path reaches the new error.

It exists because only the Ruby side holds a `cfg_arch`, which selects the codes that apply to the configuration (`Extension#exception_codes`, `extension.rb:411-421`) and renders any ERB in a name. Reading the YAML directly would skip both.

One correction to my patch: the docstring I added said the raw YAML names contain ERB templates. None of the 22 files in `spec/std/isa/exception_code/` do, so `render_erb` is a pass-through today. Pushed a docstring fix, no behavior change.

PR #2117, opened 2026-07-26
fix(inst): correct Zbb prerequisite on c.zext.w and c.zext.b

RAJVEER42:
Two Zcb instructions are gated on the wrong bit-manipulation prerequisite.

`c.zext.w` requires Zba, not Zbb
`c.zext.b` requires nothing beyond Zcb

This one also carried an unnecessary `misa.B` guard in `operation()`.


ThinkOpenly:
I think you're gonna need to regenerate the golden instruction appendix:
```
```
Then add, commit, push.

ThinkOpenly:
## What this fixes

Two Zcb instructions are gated on the wrong bit-manipulation prerequisite.

### `c.zext.w` requires Zba, not Zbb

`c.zext.w` is the compressed form of `add.uw xd, xd, x0` — an instruction defined by **Zba**, not Zbb. The Zc specification's [`c_zext_w`](https://github.com/riscv/riscv-code-size-reduction/blob/main/Zc-specification/c_zext_w.adoc) page states verbatim:

> Zba is also required.

`spec/std/isa/inst/Zba/add.uw.yaml` also lists `zext.w xd, xs1` as its pseudo-op alias, confirming the mapping.

### `c.zext.b` requires nothing beyond Zcb

`c.zext.b` is `andi xd, xd, 0xff` — pure base-ISA behavior, no bit-manipulation extension involved. The Zc specification's [`c_zext_b`](https://github.com/riscv/riscv-code-size-reduction/blob/main/Zc-specification/c_zext_b.adoc) page lists:

> Prerequisites::
> None

This one also carried a `misa.B` guard in `operation()`. That guard is **not** boilerplate in this directory — it tracks the Zbb prerequisite:

| Instruction | Prerequisite | `misa.B` guard |
|---|---|---|
| `c.sext.b` | Zbb | yes |
| `c.not` | none | no |
| `c.lbu` | none | no |
| `c.zext.b` | none | **yes** ← inconsistent |

so it is removed alongside the `definedBy` change. Happy to split that part out if you'd rather land the gating fix alone.

## Why it matters

`definedBy` drives config resolution, so today a Zcb+Zba config (no Zbb) reports `c.zext.w` as unimplemented while a Zcb+Zbb config (no Zba) reports it as implemented — both backwards. Similarly a Zcb-only config wrongly rejects `c.zext.b`.

## Scope check

Checked the whole family against the same per-instruction spec pages. `c.sext.b`, `c.sext.h` and `c.zext.h` correctly require Zbb; `c.not` correctly requires nothing; `c.mul` correctly requires Zmmul. These two were the only outliers.

Note: the `Zc.adoc` overview page's summary of prerequisites disagrees with its own per-instruction pages (it claims `c.zext.w` needs no additional extension). I've gone with the per-instruction pages as authoritative — flagging in case anyone reads the overview and wonders.

## Verification

- Both files validate against `inst_schema.json`.
- `c.zext.w`'s `xlen: 64` restriction is unchanged and remains correct.

RAJVEER42:
Done — regenerated with `./do chore:update_golden_appendix` and pushed. The content delta is just the two expected rows:

- `c.zext.b` — dropped the `*Zbb* | any` row
- `c.zext.w` — `*Zbb*` → `*Zba*`

The `:wavedrom:` line also changed, since it embeds the generating machine's absolute path. I left it as generated rather than hand-editing the golden file — `test:instruction_appendix` filters `^:wavedrom:` out before diffing, and prior regenerations have carried whoever's path too.

RAJVEER42:
The 3 red checks here (`regress-schema-versions`, `regress-smoke-perf`, `regress-complete`) aren't caused by this diff — they're failing on every open PR, including Renovate's #2133 which touches no spec files.

Root cause filed as #2148: `config_schema.json` was changed in #1879 without bumping its `$id`, so the local `v0.1` no longer matches the published `v0.1`.

Everything else is green (67 passing), and the golden appendix is regenerated as requested.

ThinkOpenly (approved):

Issue #1026, opened 2025-08-25
Svadu ADUE field type logic is backwards in environment configuration CSRs


Sukuna0007Abhi:
## Bug Description
The type() function for ADUE fields in environment configuration CSRs has backwards logic.

## Current (Incorrect) Logic
yaml
```
return (implemented?(ExtensionName::Svadu)) ? CsrFieldType::RO : CsrFieldType::RW;
```

**Rationale**
When Svadu extension is implemented → ADUE field should be Read-Write (software can control it)
When Svadu extension is NOT implemented → ADUE field should be Read-Only-0 (hardwired to 0)

**Reference**
Identified by sir @ThinkOpenly in PR #1019 review feedback.

ThinkOpenly:
reference: https://github.com/riscv-software-src/riscv-unified-db/pull/1019#discussion_r2294795371

Sukuna0007Abhi:
Yes sir

PR #2118, opened 2026-07-26
fix(csr): correct inverted type() logic for scountovf OF fields

RAJVEER42:
`scountovf`'s `OF3`–`OF31` fields have their `type()` ternary the wrong way round:

```
return HPM_COUNTER_EN[n] ? CsrFieldType::RO : CsrFieldType::ROH;
```

Per `TYPE_DESC_MAP` in `tools/ruby-gems/udb/lib/udb/obj/csr_field.rb`:

- **RO** — "Field has a hardwired value that does not change."
- **RO-H** — "Reads reflect a value dynamically generated by hardware."

Each `OFn` aliases `mhpmevent<n>.OF`, a sticky bit set by hardware when the counter overflows. So when the counter *is* enabled the field is hardware-updated (`ROH`), and when it is *not* enabled it is hardwired zero (`RO`).

Reverse the ternary targets.

### DIFF
--- a/spec/std/isa/csr/Sscofpmf/scountovf.layout
+++ b/spec/std/isa/csr/Sscofpmf/scountovf.layout
@@ -41,7 +41,7 @@ fields:
       [when="HPM_COUNTER_EN[<%= of_num %>] == false"]
       This field is read-only zero because the counter is not enabled.
     type(): |
-      return HPM_COUNTER_EN[<%= of_num %>] ? CsrFieldType::RO : CsrFieldType::ROH;
+      return HPM_COUNTER_EN[<%= of_num %>] ? CsrFieldType::ROH : CsrFieldType::RO;
     reset_value(): |
       return HPM_COUNTER_EN[<%= of_num %>] ? UNDEFINED_LEGAL : 0;
 <%- end -%>
--- a/spec/std/isa/csr/Sscofpmf/scountovf.yaml
+++ b/spec/std/isa/csr/Sscofpmf/scountovf.yaml
@@ -43,7 +43,7 @@ fields:
       [when="HPM_COUNTER_EN[3] == false"]
       This field is read-only zero because the counter is not enabled.
     type(): |
-      return HPM_COUNTER_EN[3] ? CsrFieldType::RO : CsrFieldType::ROH;
+      return HPM_COUNTER_EN[3] ? CsrFieldType::ROH : CsrFieldType::RO;
     reset_value(): |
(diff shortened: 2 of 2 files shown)


ThinkOpenly (approved):

ThinkOpenly:
## What this fixes

`scountovf`'s `OF3`–`OF31` fields have their `type()` ternary the wrong way round:

```
return HPM_COUNTER_EN[n] ? CsrFieldType::RO : CsrFieldType::ROH;
```

Per `TYPE_DESC_MAP` in `tools/ruby-gems/udb/lib/udb/obj/csr_field.rb`:

- **RO** — "Field has a hardwired value that does not change."
- **RO-H** — "Reads reflect a value dynamically generated by hardware."

Each `OFn` aliases `mhpmevent<n>.OF`, a sticky bit set by hardware when the counter overflows. So when the counter *is* enabled the field is hardware-updated (`ROH`), and when it is *not* enabled it is hardwired zero (`RO`) — exactly as the field's own description already says:

```
[when="HPM_COUNTER_EN[n] == true"]
Shadow copy of mhpmevent<n> overflow (OF) bit.

[when="HPM_COUNTER_EN[n] == false"]
This field is read-only zero because the counter is not enabled.
```

## Two cross-checks

1. **`reset_value()` on the same field** returns `UNDEFINED_LEGAL` when enabled and `0` when disabled — which only pairs coherently with `ROH` and `RO` respectively. `type()` was contradicting its own sibling method in the same field.

2. **The aliased source field** `mhpmevent<n>.OF` (`spec/std/isa/csr/Zihpm/mhpmeventN.layout`) uses `RWH` when enabled and `RO` when disabled. `scountovf` is a read-only CSR (`writable: false`), so `ROH` is the correct read-only counterpart of `RWH` — confirming the ordering rather than the reverse.

Same bug class as #1026, fixed in #1027 for the Svadu `ADUE` fields.

## Why it matters

Backends consuming `type()` (e.g. `backends/cpp_hart_gen`) currently model an enabled `scountovf.OFn` as an immutable hardwired value that never reflects real overflow state, and a disabled one as dynamically hardware-updated when it must be pinned at 0 — wrong in both directions, across all 29 fields.

## Verification

- Regenerated `scountovf.yaml` from the corrected layout; the diff is the same single line in each of the 29 `OF` fields and nothing else.
- Regenerated file parses as valid YAML.
- Grepped the tree for the inverted pattern elsewhere — this CSR was the only occurrence.

PR #2128, opened 2026-07-27
fix(udb): default PartialConfig param_values to an empty Hash

### LINKED ISSUE #2116: `PartialConfig#param_values` defaults to an Array instead of a Hash, crashing configs with no `params` key
**Describe the bug**

In `tools/ruby-gems/udb/lib/udb/config.rb`, `PartialConfig#initialize` defaults `param_values` to an **Array** when the config has no `params` key:

```ruby
@param_values = @data.key?("params") ? @data["params"] : [].freeze
```

But `param_values` is expected to be a **Hash** everywhere it is consumed — it is declared to return `T::Hash` (`config.rb:270`) and callers index it by string key:

- `condition.rb`: `cfg_arch.config.param_values.key?("SXLEN")`, `param_values[term.name]`
- `portfolio_design.rb`: `param_values.key?(param.name)`, `param_values[param.name]`

So a `PartialConfig` whose YAML omits `params:` returns `[]`, and the first `.key?`/`[]` call on it raises. The two sibling config classes are correct — `FullConfig` (`config.rb:202`) defaults to `{}.freeze` and `UnConfig` uses the params hash — `PartialConfig` is the only one that defaults to an Array.

**To Reproduce**

1. On branch `main`.
1. Construct a partially-configured architecture from a config that has no `params` key, then evaluate anything that reads `param_values` (e.g. an XLEN/SXLEN condition check).
1. `param_values` is `[]`, and the consumer's `.key?`/`[]` call fails:

```ruby
[].key?("SXLEN")   # => NoMethodError: undefined method `key?' for []:Array
[]["SXLEN"]        # => TypeError: no implicit conversion of String into Integer
```

(With the correct Hash default, `{}.key?("SXLEN")` returns `false` and `{}["SXLEN"]` returns `nil` — the intended "no params configured" behavior.)

```
NoMethodError: undefined method `key?' for an instance of Array
```
(or a Sorbet runtime `TypeError: Expected type T::Hash[...], got T::Array[...]` at `PartialConfig#param_values`, since the declared return type is a Hash.)

**Additional context**

`FullConfig` already defaults `param_values` to `{}`; `PartialConfig` should do the same. The `@mxlen = @data.dig("params", "MXLEN")` line just below is unaffected (`dig` on an absent key returns `nil` regardless).

---

### Proposed change — what and why

**What:** default the empty case to an empty Hash instead of an empty Array:

```diff
-      @param_values = @data.key?("params") ? @data["params"] : [].freeze
+      @param_values = @data.key?("params") ? @data["params"] : {}.freeze
```

**Why:** `param_values` is declared and consumed as a `Hash` (callers use `.key?`/`[]`), so the "no params" default must be a Hash — matching `FullConfig`. `{}` makes those calls return `false`/`nil` cleanly instead of crashing, with no other behavior change.

Purav001:
`Udb::PartialConfig#param_values` returns an empty **Array** (`[].freeze`)
instead of an empty **Hash** when a partially configured config has no `params`
key. `param_values` is declared as `Hash<String, ParamValueType>` on the abstract
base (`AbstractConfig#param_values`) and is consumed as a Hash everywhere.

Make the empty default a Hash (`{}.freeze`).

Closes #2116

### DIFF
--- a/tools/ruby-gems/udb/lib/udb/config.rb
+++ b/tools/ruby-gems/udb/lib/udb/config.rb
@@ -238,7 +238,7 @@ module Udb
     def initialize(data, info)
       super(data, info)
 
-      @param_values = @data.key?("params") ? @data["params"] : [].freeze
+      @param_values = @data.key?("params") ? @data["params"] : {}.freeze
 
       @mxlen = @data.dig("params", "MXLEN")
 


ThinkOpenly:
Closes #2116

## What
`Udb::PartialConfig#param_values` returns an empty **Array** (`[].freeze`)
instead of an empty **Hash** when a partially configured config has no `params`
key. This one-line change makes the empty default a Hash (`{}.freeze`).

## Why
`param_values` is declared as `Hash<String, ParamValueType>` on the abstract
base (`AbstractConfig#param_values`) and is consumed as a Hash everywhere:

- `condition.rb`: `cfg_arch.config.param_values.key?("SXLEN")`, `param_values[term.name]`
- `portfolio_design.rb`: `param_values.key?(param.name)`, `param_values[param.name]`
- `cfg_arch.rb`, `logic.rb`, `obj/csr.rb`: `param_values.key?(...)`, `param_values[name]`

With the Array default, those Hash-only calls blow up the moment a
`PartialConfig` that omits `params` reaches the condition/parameter logic:

- `[].key?("SXLEN")` -> `NoMethodError: undefined method 'key?' for []:Array`
- `[]["SXLEN"]`      -> `TypeError: no implicit conversion of String into Integer`

The sibling config classes are already correct: `FullConfig` and `UnConfig`
both default parameters to a Hash. Only `PartialConfig` used an Array, so this
is a straightforward type-mismatch bug in the empty-default branch. The udb
gem's Sorbet runtime type check also flags it directly: `param_values` is
declared to return `T::Hash` but returns `T::Array` for a no-params partial
config.

## How
`tools/ruby-gems/udb/lib/udb/config.rb`, `PartialConfig#initialize`:

```diff
-      @param_values = @data.key?("params") ? @data["params"] : [].freeze
+      @param_values = @data.key?("params") ? @data["params"] : {}.freeze
```

`{}.key?("X")` returns `false` and `{}["X"]` returns `nil`, which is the
intended "no params configured" behavior — no consumer needs to change.

## Testing
Added `TestPartialConfigParamValues` in `tools/ruby-gems/udb/test/test_cfg.rb`,
built against the real `Udb::PartialConfig` via
`AbstractConfig.create_from_data`:

- `test_param_values_defaults_to_a_hash_when_params_absent` — a partial config
  with no `params` returns a Hash and `param_values.key?("SXLEN")` /
  `param_values["SXLEN"]` do not raise. This is red on the current code (the
  Array default fails the declared Hash return type) and green after the fix.
- `test_param_values_returns_the_params_hash_when_present` — guard test: a
  partial config with `params` returns that Hash intact (passes before and
  after, so the change isn't an over-correction).

Local results (macOS; the udb gem's native Boolean-minimization binaries
eqntott/espresso ship Linux-ELF only, so a subset of the suite that shells out
to them cannot run on this host — these run green on the Linux CI):

- New tests: red before the fix, green after (2 runs, 6 assertions, 0 failures).
- Full `test_cfg.rb` (all real-repo config-validity checks + the 2 new tests):
  25 runs, 29 assertions, 0 failures, 0 errors.
- `test_cfg_arch.rb` (the heaviest `param_values` consumer): identical results
  before and after the change (11 runs, 424 assertions, 1 failure, 3 errors);
  every one of those 4 is a pre-existing eqntott/espresso limitation on macOS,
  verified by re-running with the fix reverted — this change adds no new
  failures.
- RuboCop on both changed files: no offenses.

ThinkOpenly:
@Purav001, I took the liberty of removing the test case. It's a lot for a 2-character change. I hope you don't mind.

This repository added mandatory DCO a few days ago, so you'll now need to add a "sign-off" to your commits. Are you willing and able to do that?

Purav001:
@ThinkOpenly thanks, and no problem at all on removing the test — agreed it's a lot for a two-character change.

Yes, happy to sign off. I'll add the DCO sign-off.

ThinkOpenly (approved):

Issue #1781, opened 2026-04-09
idl_highlighter could not be loaded


Abdullah-10xe:
./bin/generate ext-doc -o Zba.pdf Zba -d debug 
Using docker environment
ℹ info    Constructing ConfiguredArchiture for _
Discovering which instructions need to be documented [                                                                                                                                              ] 1/1351• debug   Creating symbol table    
ℹ info    Compiling global IDL     
ℹ info    Found 8 instructions to include in documentation
ℹ info    ["add.uw", "sh1add.uw", "sh1add", "sh2add.uw", "sh2add", "sh3add.uw", "sh3add", "slli.uw"]
ℹ info    Creating instruction summary for Zba any
ℹ info    Determining relevant CSRs
ℹ info    Found 0 relevant CSRs    
Creating instruction definitions [========] 8/8
ℹ info    Running asciidoctor-pdf  
• debug   asciidoctor-pdf -w -v -a toc -a compress -a pdf-theme=/home/abdullah/udd_new/riscv-unified-db/tools/ruby-gems/udb-gen/riscv-docs-resources/themes/riscv-pdf.yml -a pdf-fontsdir=/home/abdullah/udd_new/riscv-unified-db/tools/ruby-gems/udb-gen/riscv-docs-resources/fonts -a imagesdir=/home/abdullah/udd_new/riscv-unified-db/tools/ruby-gems/udb-gen/riscv-docs-resources/images -r asciidoctor-diagram -r idl_highlighter -a wavedrom=/opt/node/node_modules/.bin/wavedrom-cli -o Zba.pdf/Zba.pdf Zba.pdf/Zba.adoc
/usr/local/bundle/ruby/3.4.0/gems/asciidoctor-diagram-3.1.0/lib/asciidoctor-diagram/barcode/converter.rb:98: warning: ambiguous `/`; wrap regexp in parentheses or add a space after `/` operator
asciidoctor: FAILED: 'idl_highlighter' could not be loaded
  Use --trace to show backtrace
SUCCESS! Wrote result to Zba.pdf/Zba.pd

it did not generate the PDF, though Zba.adoc is generated in the folder.

i did sucessfull generate the doc last month with the older version than. i reclone new repo and facing this issue .

iakshkhurana:
Can I work on this?

ThinkOpenly:
> Can I work on this?

@iakshkhurana, it's not the highest priority, but I have no objection. Shall I assign this issue to you?

[...]

PR #2136, opened 2026-07-27
fix: don't report success when asciidoctor-pdf fails in ext-doc

iakshkhurana:
`gen_pdf` ignored the return value of `system`, so it printed `SUCCESS! Wrote result to ...` even when asciidoctor-pdf exited non-zero and produced no PDF. The log in #1781 shows exactly this: `asciidoctor: FAILED: 'idl_highlighter' could not be loaded` immediately followed by `SUCCESS!`.

Now the generator exits with a software error and a clear message instead.

Verified locally:
- normal `ext-doc` generation still produces a PDF and prints `SUCCESS!`
- forcing asciidoctor-pdf to fail (by adding an unloadable `-r` library) exits 70 with `asciidoctor-pdf failed while generating <path>` and no `SUCCESS!` line

Relates to #1781

### DIFF
--- a/tools/ruby-gems/udb-gen/lib/udb-gen/generators/ext_doc/generator.rb
+++ b/tools/ruby-gems/udb-gen/lib/udb-gen/generators/ext_doc/generator.rb
@@ -187,7 +187,9 @@ module UdbGen
       ].join(" ")
 
       Udb.logger.debug cmd
-      system cmd
+      unless system(cmd)
+        exit_with(:software_error, "asciidoctor-pdf failed while generating #{pdf_filename}")
+      end
 
       puts "SUCCESS! Wrote result to #{pdf_filename}"
     end


ThinkOpenly:
All good here except for adding a "sign-off" to the commits.

ThinkOpenly (approved):

iakshkhurana:
Sir the pr is showing open right now after your approval to merge it.
Please review.

ThinkOpenly:
Please add a DCO ("sign-off") to your commits.

iakshkhurana:
Thanks @ThinkOpenly - sorry for the confusion earlier, I misread the status. I've now rebased the branch onto latest main and added the DCO sign-off:

Signed-off-by: Aksh Khurana <khuranaaksh08@gmail.com>

I dropped the earlier main merge commit in the process, since it was also unsigned and the DCO workflow checks every commit in the PR. The branch is now a single signed-off commit with no change to the actual diff. Force-pushed — please re-run CI when you get a chance.

PR #2138, opened 2026-07-27
fix(schema): replace non-power-of-two 4095 with 4096 in unsigned_pow2 enums

### LINKED ISSUE #2137: schema(defs): 32bit/64bit_unsigned_pow2 enums include 4095 which is not a power of two
## Describe the bug

In `spec/schemas/schema_defs.json`, both shared power-of-two integer defs list **4095** in their `enum` arrays:

- `$defs/32bit_unsigned_pow2`
- `$defs/64bit_unsigned_pow2`

Sequence is `..., 2048, **4095**, 8192, ...`.

4095 (`0xfff`) is **not** a power of two. The next true power of two after 2048 is **4096** (`0x1000`).

## Why this matters

These defs are the shared schema vocabulary for "unsigned power of 2" parameter values. The same non-power-of-two value previously appeared in MTVEC base-alignment parameter enums and was corrected in #2090 (4095/`0xfff` → 4096/`0x1000`), with maintainer agreement that the old decimal list already had 4095 and hex conversion preserved it.

Leaving 4095 in `schema_defs.json` reintroduces the same class of error for any parameter or tool that `$ref`s these defs, and undoes the intent of the #2090 correction at the schema layer.

## Expected

Both enums should list `4096` instead of `4095`, and ideally a regression test should reject non-powers-of-two in these enums so the failure cannot return unnoticed.

## Proposed fix

- `4095` → `4096` in both `32bit_unsigned_pow2` and `64bit_unsigned_pow2`
- Unit test: every enum entry is a positive power of two; assert 4096 present and 4095 absent

## Related

- #2090 / commit that fixed MTVEC alignment enums (same 4095 pattern)
- Not the same as CSR address-space `maximum: 4095` in `csr_schema.json` (12-bit CSR index range 0..4095 is correct and out of scope)

titoatwork:
`schema_defs.json` listed **4095** in both `32bit_unsigned_pow2` and `64bit_unsigned_pow2` enums. 4095 is not a power of two; the correct value is **4096**.

Closes #2137


ThinkOpenly:
## Summary

Closes #2137

`schema_defs.json` listed **4095** in both `32bit_unsigned_pow2` and `64bit_unsigned_pow2` enums. 4095 is not a power of two; the correct value is **4096**.

This is the same failure mode that previously appeared in MTVEC base-alignment parameter enums and was fixed in #2090 (maintainer agreement: old decimal list had 4095; hex conversion preserved it). Correcting the shared schema defs prevents the bad value from being reintroduced via `$ref`.

## Changes

- `spec/schemas/schema_defs.json`: `4095` → `4096` in both pow2 enums
- `tools/ruby-gems/udb/test/test_schema_defs_pow2.rb`: regression test that every enum entry is a positive power of two, and that 4096 is present / 4095 is absent
- Wire the test into `tools/ruby-gems/udb/test/run.rb` so the standard udb unit suite executes it

## Out of scope

- CSR address-space bounds of `4095` in `csr_schema.json` (valid 12-bit index range 0..4095)
- Bulk `long_name: TODO` cleanup

## Test plan

- [x] Enum math check: no non-powers-of-two remain; 4096 present
- [x] New test file path resolves to `spec/schemas/schema_defs.json`
- [ ] CI: `regress-udb-unit` / full PR checks

ThinkOpenly (approved):

titoatwork:
Approved and mergeable, but the required checks have not reported on the current head. Both workflow runs on `59e7afb` are sitting at `action_required`, which I think is the first-time-contributor approval gate rather than anything wrong with the change.

That is on me: checks were green on the earlier commit, and I merged `main` in afterwards without needing to, which reset them. Nothing in the diff changed. Whenever someone has a moment to approve the workflow run, it should go green again.

spec/schemas/schema_defs.json
ThinkOpenly: CI is failing due to the schema change, which needs an increment to the version.
```suggestion
  "$id": "v0.2",
```

ThinkOpenly (changes requested):

titoatwork:
Bumped to `v0.2`, thanks for the catch. All three failures were the same `check_schema_versions` run surfacing in three suites rather than three separate problems.

That is a new head commit, so the workflow will need approving again.

One question while I am here: this PR also updates the generated `doc/docs/schemas/v0.1/schema_defs.mdx`. With the schema now at v0.2, should that page regenerate under `v0.2/` and leave v0.1 documenting the published version? I can run `bin/chore gen schema-docs` and push if that is the convention.

doc/docs/schemas/v0.1/schema_defs.mdx
jordancarlin, line 1: Why is the v0.1 schema documentation being changed? Revert the changes to this file.

tools/ruby-gems/udb/test/test_schema_defs_pow2.rb
```diff
@@ -0,0 +1,51 @@
...
+    assert_equal "integer", schema["type"], "#{name} type"
+    enum = schema.fetch("enum")
+    assert !enum.empty?, "#{name} enum must not be empty"
+
+    bad = enum.reject { |v| power_of_two?(v) }
+    assert_empty bad, "#{name} enum contains non-powers-of-two: #{bad.inspect}"
+
+    # Specific historical failure mode: 4095 instead of 4096
+    refute_includes enum, 4095, "#{name} must not include 4095 (use 4096)"
+    assert_includes enum, 4096, "#{name} must include 4096"
```
jordancarlin: No need to explicitly check for 4095 if you are already checking that all values are powers of two above.

spec/schemas/schema_defs.json
```diff
@@ -873,7 +873,7 @@
       "description": "An unsigned power of 2 that fits in 64 bits",
       "type": "integer",
       "enum": [
-        1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4095, 8192, 16384,
+        1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384,
```
jordancarlin, line 876: Are we better off converting this list to hex? Especially for the larger numbers, it is easier to spot a non power-of-two number in hex.

jordancarlin (changes requested):

tools/ruby-gems/udb/test/test_schema_defs_pow2.rb
```diff
@@ -0,0 +1,52 @@
...
+    enum = schema.fetch("enum")
+    assert !enum.empty?, "#{name} enum must not be empty"
+
+    bad = enum.reject { |v| power_of_two?(v) }
+    assert_empty bad, "#{name} enum contains non-powers-of-two: #{bad.inspect}"
+
+    # 4095 is already excluded by the power-of-two assertion above. This checks
+    # the other half of the historical failure: that 4096 is actually present,
+    # rather than the bad entry simply having been deleted.
+    assert_includes enum, 4096, "#{name} must include 4096"
```
jordancarlin: Remove this comment and assert. There is nothing inherently special about 4096.

titoatwork:
First two done. The v0.1 page is restored to what is on `main`, and the test now asserts only that every value is a power of two.

On hex, I do not think it is available here. JSON numbers are decimal only, so `0x1000` is a parse error rather than a style choice, and there are no hex numeric literals in any `.json` in the repo. The hex enums are all YAML, `MTVEC_BASE_ALIGNMENT_DIRECT.yaml` and its siblings, where it is legal. If the readability point still stands I can add a `$comment` on each def recording that the sequences run 2^0 to 2^31 and 2^0 to 2^63, though the test now fails on any non-power-of-two rather than relying on spotting one.

The two remaining doc files, `v0.2/schema_defs.mdx` and the `index.mdx` entry, are autofix.ci output from `3240e68` after the `$id` bump, not hand edits. Say if you would rather this PR carried only the schema and the test and I will strip them.

jordancarlin:
> On hex, I do not think it is available here. JSON numbers are decimal only, so `0x1000` is a parse error rather than a style choice, and there are no hex numeric literals in any `.json` in the repo. The hex enums are all YAML, `MTVEC_BASE_ALIGNMENT_DIRECT.yaml` and its siblings, where it is legal. If the readability point still stands I can add a `$comment` on each def recording that the sequences run 2^0 to 2^31 and 2^0 to 2^63, though the test now fails on any non-power-of-two rather than relying on spotting one.

I forgot that JSON didn't support hex. This is fine then. No need for the $comment.

ThinkOpenly (approved):

jordancarlin (approved):

PR #2146, opened 2026-07-27
docs(param): correct SXLEN and UXLEN parameter descriptions

### LINKED ISSUE #2145: param: UXLEN description names SXLEN, and SXLEN option list contradicts its array schema
Two description defects in the XLEN parameter family. Both are documentation-only; no schema or value behaviour is affected.  ### 1. `UXLEN` description names the wrong parameter  `spec/std/isa/param/UXLEN.yaml`:  > Set of XLENs supported in U-mode. When both 32 and 64 are supported, **SXLEN** can be changed, via `mstatus.UXL`, between 32 and 64.  `mstatus.UXL` controls the effective XLEN in U-mode, i.e. UXLEN, not SXLEN. `mstatus.SXL` is the field that controls SXLEN, and `SXLEN.yaml` documents that correctly. The sentence appears to have been copied from `SXLEN.yaml` with the parameter name left unchanged.  As written, a reader consulting `UXLEN` to understand U-mode width control is told the wrong parameter is affected.  ### 2. `SXLEN` description contradicts its own schema  `spec/std/isa/param/SXLEN.yaml` lists its options as:  ```   * 32:   SXLEN is always 32   * 64:   SXLEN is always 64   * [32, 64]: SXLEN can be changed (via mstatus.SXL) between 32 and 64 ```  But the schema is `type: array` with `minItems: 1`, so the single-value cases are `[32]` and `[64]`, not the scalars `32` and `64`. The VS-mode twin `VSXLEN.yaml` already documents these correctly as `[32]` / `[64]`.  ### Suggested fix  Correct the parameter name in `UXLEN`, and bring `SXLEN`'s option list in line with its schema and with `VSXLEN`.  I have a branch ready and will open a PR against this issue.  ### Note on what is *not* a defect  While checking this I also compared `MXLEN` against `SXLEN`/`UXLEN`/`VSXLEN`: `MXLEN` is a scalar `integer` with `enum: [32, 64]` while the others are arrays. That difference is correct and intentional. M-mode XLEN is fixed for a hart, whereas SXLEN/UXLEN/VSXLEN describe the *set* of widths supported and are runtime-switchable via `mstatus.SXL`, `mstatus.UXL`, and `hstatus.VSXL`. Flagging it here so the asymmetry is not "fixed" by mistake later.

titoatwork:
`UXLEN` description uses "SXLEN" instead of "UXLEN".

`SXLEN` value is always an array, but the documentation example uses a scalar (e.g. "32") instead of an array ("[32]").

Fixes #2145.

### DIFF
--- a/spec/std/isa/param/SXLEN.yaml
+++ b/spec/std/isa/param/SXLEN.yaml
@@ -9,8 +9,8 @@ name: SXLEN
 description: |
   Set of XLENs supported in S-mode. Can be one of:
 
-    * 32:   SXLEN is always 32
-    * 64:   SXLEN is always 64
+    * [32]:     SXLEN is always 32
+    * [64]:     SXLEN is always 64
     * [32, 64]: SXLEN can be changed (via mstatus.SXL) between 32 and 64
 long_name: TODO
 schema:
--- a/spec/std/isa/param/UXLEN.yaml
+++ b/spec/std/isa/param/UXLEN.yaml
@@ -7,7 +7,7 @@ $schema: param_schema.json#
 kind: parameter
 name: UXLEN
 description: |
-  Set of XLENs supported in U-mode. When both 32 and 64 are supported, SXLEN can be changed,
+  Set of XLENs supported in U-mode. When both 32 and 64 are supported, UXLEN can be changed,
   via mstatus.UXL, between 32 and 64.
 long_name: TODO
 schema:


ThinkOpenly:
Fixes #2145.

Documentation-only corrections to two parameters in the XLEN family. No schema, value, or IDL changes.

### `UXLEN`: description named the wrong parameter

```diff
-  Set of XLENs supported in U-mode. When both 32 and 64 are supported, SXLEN can be changed,
+  Set of XLENs supported in U-mode. When both 32 and 64 are supported, UXLEN can be changed,
   via mstatus.UXL, between 32 and 64.
```

`mstatus.UXL` controls the effective XLEN in U-mode, i.e. UXLEN. `mstatus.SXL` is what controls SXLEN, and `SXLEN.yaml` states that correctly. The sentence looks copied from there with the name left unchanged.

### `SXLEN`: option list contradicted its own schema

```diff
-    * 32:   SXLEN is always 32
-    * 64:   SXLEN is always 64
+    * [32]:     SXLEN is always 32
+    * [64]:     SXLEN is always 64
     * [32, 64]: SXLEN can be changed (via mstatus.SXL) between 32 and 64
```

The schema is `type: array` with `minItems: 1`, so the single-value cases are `[32]` and `[64]`. `VSXLEN.yaml` already documents them that way; this makes SXLEN consistent with both its schema and its VS-mode twin.

### Checking

Both files still parse and satisfy `param_schema.json`'s required keys with no additional properties (`description` resolves to `spec_text`, which accepts a string). I re-parsed all 227 files under `spec/std/isa/param/` as a regression check, with no parse failures and no missing required keys.

No regression test is included: these are prose corrections, and a test asserting specific description wording would be brittle without adding value.

### Not a defect

`MXLEN` is a scalar `integer` with `enum: [32, 64]` while `SXLEN`/`UXLEN`/`VSXLEN` are arrays. That asymmetry is correct, because M-mode XLEN is fixed for a hart while the others describe the *set* of supported widths and are runtime-switchable via `mstatus.SXL`, `mstatus.UXL`, and `hstatus.VSXL`. Left unchanged deliberately, and noted in #2145 so it is not "corrected" later by mistake.

ThinkOpenly (approved):

titoatwork:
Same situation as #2138: approved and mergeable, but the runs on `ab185d5` are at `action_required` and no checks have reported. I re-synced with `main` after the approval, which re-armed the gate. The diff is unchanged and documentation-only. It just needs a workflow run approved.

PR #2150, opened 2026-07-27
fix(schemas): bump config_schema.json $id to v0.2

### LINKED ISSUE #2148: regress-schema-versions fails on every open PR: config_schema.json changed without an $id bump
## Summary

`regress-schema-versions` is currently failing on **every** open PR, including ones that touch no schema or spec files at all. It blocks the merge queue.

```
Schema mismatch for https://riscv.github.io/riscv-unified-db/schemas/config_schema.json/v0.1/config_schema.json:
  Local:  gen/schemas/config_schema.json/v0.1/config_schema.json
  Remote: https://riscv.github.io/riscv-unified-db/schemas/config_schema.json/v0.1/config_schema.json
  The published schema differs from the local version.
  To fix: bump the schema version ($id) to a new version number.
```

Every other schema reports `OK (matches published)` or `OK (not yet published)` — `config_schema.json` is the only mismatch.

## Cause

252dd19c ("fix: accept bare version strings in implemented_extensions", #1879) changed `spec/schemas/config_schema.json`:

```diff
               "version": {
-                "$ref": "schema_defs.json#/$defs/extension_version"
+                "$ref": "schema_defs.json#/$defs/implemented_extension_version"
               }
```

`$id` was left at `"v0.1"` (line 3). So local `v0.1` no longer matches the `v0.1` already published to `riscv.github.io`, which is exactly what this check exists to catch.

## Evidence it's not PR-specific

- #2133 (Renovate dependency bump — touches no spec/schema files) fails `regress-schema-versions`, `regress-smoke-perf`, and `regress-complete`
- #1991 and #2117 (unrelated data fixes) fail the same three

## Suggested fix

Bump `config_schema.json`'s `$id` to `v0.2`, since the `$ref` change narrows what `implemented_extensions.version` accepts and is therefore not backward compatible for existing `v0.1` consumers.

Happy to send that PR if the maintainers agree it's just the `$id` bump and not a revert.

RAJVEER42:
Fixes #2148

## Problem

`regress-schema-versions` fails on **every** open PR, including ones touching no schema or spec files, which blocks the merge queue.

252dd19c ("fix: accept bare version strings in implemented_extensions", #1879) changed `config_schema.json`:

```diff
               "version": {
-                "$ref": "schema_defs.json#/$defs/extension_version"
+                "$ref": "schema_defs.json#/$defs/implemented_extension_version"
               }
```

but left `$id` at `v0.1`. Local `v0.1` no longer matches the `v0.1` already published to `riscv.github.io`, which is precisely what `check_schema_versions.rb` exists to catch.

## Fix

Bump `config_schema.json`'s `$id` to `v0.2`. That is the remedy the check itself prescribes:

> To fix: bump the schema version ($id) to a new version number.
> Note: new versions are published automatically when merged to main.

and it matches how `csr_schema.json` and `inst_schema.json` are already at `v0.2`.

## Verification

Ran the real check before and after on a clean `gen/`:

**Before** — one failure:
```
Schema mismatch for .../config_schema.json/v0.1/config_schema.json
```

**After:**
```
OK (not yet published): .../config_schema.json/v0.2/config_schema.json
All schema version checks passed.
```

`./do test:schema` also passes — nothing resolves `config_schema` by version, since every reference is the versionless `"config_schema.json#"`.

## Notes

- **`schema_defs.json` deliberately not bumped.** The same commit added `implemented_extension_version` to it, but its `v0.1` reports `OK (not yet published)`, so the new content will publish as `v0.1` with no conflict. Only 3 schemas are currently published: `config_schema` v0.1 (the mismatch), `csr_schema` v0.2, and `exception_code_schema` v0.1.

- **On backward compatibility** — worth correcting something I wrote in #2148. I claimed the `$ref` change *narrows* what's accepted; it actually **widens** it. Old `extension_version` → `rvi_version` is `^[0-9]+(\.[0-9]+(\.[0-9]+(-pre)?)?)?$`; new `implemented_extension_version` is `^(=\s*)?[0-9]+(\.[0-9]+(\.[0-9]+(-pre)?)?)?$` — a strict superset adding an optional `=` prefix. Every document valid under published `v0.1` stays valid. So this is an additive change and `v0.2` (minor) is the right increment rather than a breaking bump.

- Docs regenerated with `./bin/chore gen schema-docs`, adding `doc/docs/schemas/v0.2/config_schema.mdx` and its index entry. The generator was idempotent for every other schema — no unrelated churn.

- `doc/docs/schemas/v0.1/config_schema.mdx` is left in place, since `v0.1` remains published and immutable so its docs stay accurate for consumers pinned to it. Flagging that the repo is inconsistent here: `csr_schema` was removed from `v0.1` when it bumped, but `inst_schema` was left behind, and `v0.1` also still holds `proc_cert_class_schema.mdx` / `proc_cert_model_schema.mdx` for schemas no longer in `spec/schemas`. Happy to follow whichever convention you prefer, or leave that cleanup to a separate PR.

- On @ThinkOpenly's question about automatic version bumping: I could not find any `$id`-bumping automation. `schema-release.yml` → `publish_schemas.rb` publishes changed versions after CI passes on main, and `check_schema_versions.rb` guards immutability, but nothing increments `$id`. So this class of breakage recurs whenever a published schema is edited without a manual bump. If that guard is wanted, it would probably belong as a pre-commit or PR check comparing changed `spec/schemas/*.json` against their published counterparts — happy to open a follow-up issue.


RAJVEER42:
Superseded by #2152 — please hold off on this one.

After digging into @jordancarlin's note on #2148 I found the actual root cause, and it isn't version drift: `check_schema_versions.rb` compares `Net::HTTP`'s `ASCII-8BIT` body against `Pathname#read`'s `UTF-8` string, which Ruby reports as unequal whenever the content contains a non-ASCII byte — even with identical bytes and matching MD5. `config_schema.json` is the only published schema with one (an em dash), which is why it was the only failure.

**The bump in this PR would not have held.** It passes only because `v0.2` currently 404s (`OK (not yet published)`). Once `v0.2` published automatically on the next merge to main, the same em dash would hit the same bug and break CI again.

#2152 fixes the comparison instead. With `$id` left at `v0.1`, the check reports `OK (matches published)` and still fails correctly on genuine drift.

Happy to close this, or rebase it down to just the `$id` bump if you decide on policy grounds that #1879 should have bumped the version regardless (there's a related wrinkle on #2148 — `publish_schemas.py` clobbers already-published versions, so `v0.1` was silently overwritten in place today).

jordancarlin:
I'm going to go ahead and close this since the versions do not actually mismatch.

PR #2155, opened 2026-07-27
Updated several ISA parameter spec files to replace placeholder long_name: TODO entries with concrete names. 

Bhupesh-081:
The change covers:

VLEN.yaml
ELEN.yaml
VMID_WIDTH.yaml
VS_MODE_ENDIANNESS.yaml
This is a metadata cleanup only. It does not change behavior, generated logic, or runtime code. Static validation of the edited YAML files reported no errors.

### DIFF
--- a/spec/std/isa/param/VS_MODE_ENDIANNESS.yaml
+++ b/spec/std/isa/param/VS_MODE_ENDIANNESS.yaml
@@ -13,7 +13,7 @@ description: |
    * big:     VS-mode data is always big endian
    * dynamic: VS-mode data can be either little or big endian,
               depending on the CSR field `hstatus.VSBE`
-long_name: Vs-mode data endianness
+long_name: VS-mode data endianness
 schema:
   type: string
   enum:
--- a/spec/std/isa/param/VS_MODE_ENDIANNESS.yaml
+++ b/spec/std/isa/param/VS_MODE_ENDIANNESS.yaml
@@ -13,7 +13,7 @@ description: |
    * big:     VS-mode data is always big endian
    * dynamic: VS-mode data can be either little or big endian,
               depending on the CSR field `hstatus.VSBE`
-long_name: TODO
+long_name: Vs-mode data endianness
 schema:
   type: string
   enum:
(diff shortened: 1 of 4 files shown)


spec/std/isa/param/VS_MODE_ENDIANNESS.yaml
```diff
@@ -13,7 +13,7 @@ description: |
    * big:     VS-mode data is always big endian
    * dynamic: VS-mode data can be either little or big endian,
               depending on the CSR field `hstatus.VSBE`
-long_name: TODO
+long_name: Vs-mode data endianness
```
titoatwork: The mode is spelled `VS-mode` everywhere else, including three times in this file's own description ("Endianness of data in VS-mode", "VS-mode data is always little endian"). Across `spec/std/isa/param` and `spec/std/isa/csr` it is `VS-mode` 469 times and `Vs-mode` zero, so this would be the only instance.

```suggestion
long_name: VS-mode data endianness
```

titoatwork (commented):
Three of the four read well and match their descriptions. `VLEN` as "Vector register length" and `ELEN` as "Maximum vector element size" line up with how the V extension defines them, and `VMID_WIDTH` is unambiguous. One correction on the fourth, inline.

Separately, and more of a coordination note than a review point: #2124 appears to carry these same four `long_name` changes already, with identical values. Its stated purpose is making the toolchain wrapper scripts independent of PATH-resolved core utilities, so the parameter edits look like they were picked up by accident there rather than intentionally. If that is right, whichever merges second will conflict, and it may be worth dropping the four parameter files from #2124 so this PR owns them cleanly.

Also worth knowing: #2171 touches `ELEN.yaml` too, adding `enum: [8, 16, 32, 64]` to the schema. That change sits a couple of lines below the `long_name` line, so the two may need a rebase depending on merge order.

spec/std/isa/param/VS_MODE_ENDIANNESS.yaml
Bhupesh-081: Thanks for the review, @titoatwork!

Fixed the `VS-mode` capitalization in `VS_MODE_ENDIANNESS.yaml` to match the rest of the codebase pushed to the branch.

Good catch on the overlap with #2124 too looks like those four param file changes ended up there unintentionally. Happy to have this PR own them once that's sorted out. Also noted the potential rebase with #2171 on `ELEN.yaml`.

PR #2156, opened 2026-07-27
fix(csr): correct copy-pasted table captions in hcounteren

RAJVEER42:
Some `hcounteren` field descriptions referred to the wrong counter name (copy-paste error).

Fix the layout file and include the generated YAML.

### DIFF
--- a/spec/std/isa/csr/H/hcounteren.layout
+++ b/spec/std/isa/csr/H/hcounteren.layout
@@ -76,7 +76,7 @@ fields:
 
       [separator="!",%autowidth]
       !===
-      .2+h! [.rotate]#`hcounteren.TM`# .2+h! [.rotate]#`mcounteren.TM`# .2+h! [.rotate]#`scounteren.TM`# 2+^.>! `cycle` access behavior
+      .2+h! [.rotate]#`hcounteren.TM`# .2+h! [.rotate]#`mcounteren.TM`# .2+h! [.rotate]#`scounteren.TM`# 2+^.>! `time` access behavior
       .>h! VS-mode .>h! VU-mode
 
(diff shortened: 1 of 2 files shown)


ThinkOpenly (approved):

ThinkOpenly:
## What this fixes

Each `hcounteren` field documents a table showing how the corresponding unprivileged counter behaves across the `hcounteren`/`mcounteren`/`scounteren` combinations. All four captions read `` `cycle` access behavior ``, so only the `CY` field's caption was correct:

| Field | Caption said | Should say |
|---|---|---|
| `CY` | `` `cycle` `` | `` `cycle` `` ✓ |
| `TM` | `` `cycle` `` | `` `time` `` |
| `IR` | `` `cycle` `` | `` `instret` `` |
| `HPM<n>` | `` `cycle` `` | `` `hpmcounter<n>` `` |

Each field's own row headers and surrounding prose already refer to the right CSR (`hcounteren.TM`/`mcounteren.TM`/…, "the `time` CSR … is accessible to VU-mode"), so the caption was the only thing out of step — a copy-paste leftover rather than an intentional simplification.

## Scope check

Confirmed the pattern is isolated to this file: `mcounteren.layout` and `scounteren.layout` contain no `access behavior` captions at all, so `hcounteren` was the only affected layout.

## Verification

- Regenerated `hcounteren.yaml` from the corrected layout.
- Captions in the generated file now read **1×** `cycle`, **1×** `time`, **1×** `instret`, **29×** `hpmcounter<n>` (32 total: CY + TM + IR + HPM3–31).
- Diff contains **zero** non-caption changed lines.
- Regenerated file parses as valid YAML.

Issue #2158, opened 2026-07-27
Convention question: evaluation fixtures for skills, and where they should live


titoatwork:
## Summary

Skills in this repo currently ship as a single `SKILL.md` with no fixtures, so there is no mechanical way to check that a skill's documented rules still behave as written after an edit. I would like to know whether maintainers want a small evaluation-fixture convention for skills, and if so where the fixtures should live.

Raising it as an issue rather than opening a PR directly because it would be the first test subtree under a skill, which is a convention decision rather than a data fix.

## Why now

While reviewing #2097 (`extract-parameters-from-subsection`) I built a frozen adversarial pack to check the skill's rules against concrete cases rather than against opinion. @uditjainstjis has since implemented all five review points in that PR, and suggested the fixtures would be better as their own contribution than as a subtree of his. That seems right, but where they belong is not mine to decide.

## What the fixtures are

Nine cases, each a `source.txt` excerpt plus an `expected.yaml` stating the expected outcome, with UDB `gold.yaml` attached for the positives:

**Positives (5)** covering DIRECT, CSR_RW and WARL: `CACHE_BLOCK_SIZE`, `NUM_PMP_ENTRIES`, `MTVEC_ACCESS`, `MTVEC_MODES`, `ASID_WIDTH`.

**Negatives (4)**: a fixed CSR address-mapping convention, software advice phrased with should/may, a bare `shall` constraint with no implementation delegation, and the one I think matters most:

> WARL vocabulary where the legal value set is fixed by the ISA. The field is WARL, software may write any value, but the only legal encoding is 0 for every implementation. There is no implementation choice, so it is not an architectural parameter.

That last case is the distinction between "the field is labelled WARL" and "the set of legal values is implementation-chosen". WARL misclassification is a known weak spot: measured against the pinned Spring gold, WARL is the worst class at 12/24, well below the 72.9% overall figure. A skill that treats the WARL keyword as sufficient will reproduce that error.

## Properties

- **No API calls and no model runs.** The pack tests the documented rules, so it is free to run and deterministic.
- **Frozen.** Cases are fixed text, not regenerated.
- **Fails closed.** The validator exits non-zero if a case is malformed, if a positive lacks its gold, or if an expected-invalid envelope would validate against `param_schema.json`.
- Currently 5 positives, 4 negatives, and 5 structural critique records, validated in one command.

## Questions

1. Are skill fixtures wanted at all, or is prose review the intended level for these?
[...]

Issue #2053, opened 2026-07-21
Question: scope of parameter extraction already completed (Spring 2026 mentorship term)


hjaat:
Hi — I'm preparing an application for the "AI-assisted extraction of
architectural parameters from RISC-V specifications (Part II)" mentorship,
and wanted to check something before applying, so I don't duplicate work
that's already done.

I searched the commit history on `main` between January and June 2026 and
found only one parameter-related commit (`MCOUNTINHIBIT_IMPLEMENTED`,
#1807). Since a full mentorship term should have produced more output than
that, I'm guessing the Spring 2026 work either landed in a separate
branch/PR that hasn't merged yet, or in one of the other sources mentioned
in the project description (the ISA Manual YAML repo, or the
`keyword_matches` spreadsheet).

Could someone point me to where that work actually lives, so I can build on
it rather than repeat it?

Separately — I had a very helpful exchange with Allen Baum on Slack about a
common misclassification: mistaking a CSR field (particularly WARL fields)
for a true architectural parameter, and the resulting duplication problem
when the same parameter is described in more than one place. I've written
up an extraction approach that tries to account for this, here:
https://gist.github.com/hjaat/7ab4f66dcda775fe089b28cb4d47e6bb

Thanks for your time.

Maanvi212006:
Hi @hjaat — good question, I dug into this too before applying.

The Spring 2026 work is still open/unmerged, which is why the commit search on main only surfaces #1807. Two PRs from Ishaan Arora (@ishaan-arora-1) cover the bulk of it:

#1831 — consolidated parameter spreadsheet merging LLM-found candidates against UDB's existing parameter list
#1832 — tagging the actual ISA manual spec text with the extracted parameters

Worth reading through both — helpful for scoping Part II.

Your gist looks like a solid start on the WARL/duplication problem, by the way.

hjaat:
Thank you — this is exactly what I needed. I'll go through both PRs before finalizing my approach. Appreciate you taking the time to dig into this and for the kind words on the gist.

ishaan-arora-1:
@hjaat @Maanvi212006 

The recent works that have been going on for parameter extraction are internal and are not uploaded on this repository yet. The PRs that you see are a few months old i think, and since then there have been a lot of changes to the work, so dont expect that these prs are where you'll be picking up from if you get into the project. those prs were the first version of the pipeline for parameter extraction, and the pipeline has evolved ever since.

[...]

PR #2164, opened 2026-07-28
test(param-eval): add parameter-extraction evaluation fixtures

titoatwork:
Eleven frozen cases pairing a specification excerpt with the outcome a parameter-extraction procedure should reach on it: six positives, four negatives, and one candidate. The tests check the fixtures rather than a model, so an upstream parameter rename fails here instead of silently invalidating a case.

The cases that carry weight are the ones where a plausible rule produces a parameter that should not exist. A fixed encoding convention, a bare `shall`, and software-facing `should`/`may` advice all read like implementation choices and are not. So does extension gating: "If Svpbmt is not implemented, PBMTE is read-only zero" describes writability that depends on configuration, but UnifiedDB expresses that with `definedBy` and no parameter exists. The candidate kind is separate from both. A WARL field whose legal set the ISA fixes yields no parameter either, but a procedure that drops it during extraction hides that decision instead of recording it.

The set also covers the opposite failure. A rule keyed on the `WARL` and `WLRL` tokens misses fields whose legal values depend on the implementation without either token appearing, so two cases carry none and the suite asserts their sources do not.

Relates to #2158 for placement, and to #2097, where these came from.


Ganesh-403:
@titoatwork I’ve been studying #2053 and the evaluation gaps around parameter extraction.

I found another useful negative case around implementation-specific cache properties that are discovered through the execution environment rather than being ISA architectural parameters.

I’ve prepared a small fixture for this case against the current #2164 test suite. I’d be happy to contribute it as a follow-up once #2164 lands, so it doesn’t duplicate the framework changes in this PR. Thanks for building the evaluation framework, it makes adding these targeted cases much easier.

titoatwork:
Thanks, and welcome to it. Before you build that fixture, two things that will save you the work I just spent.

The case you describe is already on #2053, and the interesting part of it is settled there. The distinction between a property that is discoverable through the execution environment and a property that is ISA-visible is the whole reason `CACHE_CAPACITY` is a trap rather than a parameter, and @RAJVEER42 and I worked through it in that thread across several rounds in late July. He also did the harder half, locating the passage, after I could not find the text I had been given and said so. Worth reading that exchange in full before writing anything, because a fixture that re-encodes a conclusion already reached adds a file rather than a case.

This PR also already covers that passage: `cases/positives/POS_DIRECT_CACHE_BLOCK` is built on it, as the positive. So the useful shape for what you are proposing is a `NEG_` sibling over the same excerpt, byte-identical to the positive, where only the expected classification differs. That is a stronger fixture than a fresh snippet would be, because it isolates the classification decision from the retrieval decision, and separating those two is what this set is for.

One correction of my own while I am here, since it bears on any fixture built from that passage. The header on `POS_DIRECT_CACHE_BLOCK/source.txt` cited "RISC-V Privileged ISA Manual, Section 19.3.1". That is wrong and I have just fixed it. The passage is in the Unprivileged volume, at `src/unpriv/cmo.adoc` lines 86-93, Chapter 19 "CMO Extensions", under Background > Memory and Caches. By the heading nesting in that file the section is 19.2.1, and 19.3.1 is "Memory Ordering". This is identical at `b2e69ab2`, which this branch pins, and at `310a111`, which main pins. `src/cmo.adoc`, which the header also named, exists at neither.

The label came from the coding challenge, which supplied the snippet under that name, so if you are working from the same document you have inherited the same wrong citation. The quoted text itself is verbatim apart from "(or NAPOT)", where the manual reads "(NAPOT)".

titoatwork:
Correcting my own correction, on the one part of it that depends on which rendering you read.

I gave the section number above as a flat fact. It is not. `modules/unpriv/nav.adoc` hardcodes the label "Chapter 19. \"CMO\" Extensions for Base Cache Management Operation ISA" for `cmo.adoc`, and under that page numbering the heading nesting in the file does put the passage at 19.2.1, with 19.3.1 being "Memory Ordering". That is the rendering in which a citation shaped like 19.3.1 refers to CMO at all, so it is the one the challenge's label has to be read against.

The monolithic build numbers it differently. There is no `leveloffset` anywhere on the chain from `src/riscv-spec.adoc` to `unpriv/unpriv.adoc` to `unpriv/zi.adoc` to `cmo.adoc`, so `cmo.adoc`'s `===` heading is a section inside the "Scalar Integer Extensions" chapter rather than a chapter of its own, and the passage sits a level deeper than I implied.

Nothing above was false and the substance does not move: the volume is Unprivileged under either reading, `src/cmo.adoc` exists under neither, and the passage is at `src/unpriv/cmo.adoc` lines 86-93 at both pinned revisions. The fixture header now carries the rendering alongside the number rather than stating it as though there were only one.

Ganesh-403:
Thanks, this makes sense. I’ll rework the fixture as a negative sibling of `POS_DIRECT_CACHE_BLOCK`, using the exact same `source.txt` so the case isolates classification from retrieval.

I’ll also use the corrected provenance from your latest commits and avoid introducing a new interpretation of the cache-capacity case.

RAJVEER42:
Both of your corrections check out against `b2e69ab`, and both land on my own snippet, so thanks for chasing the section number rather than stopping where I did.

On the heading nesting: `==== Background` at line 48 and `===== Memory and Caches` at line 53 put the passage at 19.2.1, while `==== Coherent Agents and Caches` at line 146 and `===== Memory Ordering` at line 221 put 19.3.1 elsewhere in the same chapter. So the challenge label is wrong twice over, not once. I had only caught the volume and had written "that label is incorrect: the text is in the UNprivileged manual", which is true and stops one step short. `src/cmo.adoc` returns 404 at that revision, as you say.

The `(or NAPOT)` divergence is the one that mattered most to me. Manual line 87 reads `_naturally aligned power-of-two_ (NAPOT) range of memory`, and the challenge supplied `(or NAPOT)`. My snippet reproduces what the challenge gave me, so it is verbatim to the challenge and not to the manual, and my header's "verified at src/unpriv/cmo.adoc" line could fairly be read as claiming the stronger thing. I have separated the two claims and recorded the divergence explicitly. The excerpt substring checks in my repo run against the supplied body, so no result moves, but the provenance line was overclaiming.

Agreed on the fixture shape. A `NEG_` sibling over a byte-identical `source.txt` isolates classification from retrieval, which is the distinction the cache-capacity case exists to test. Worth stating in the case file that the excerpt is the challenge's wording rather than the manual's, otherwise the next person to diff it against `cmo.adoc` will find the same two words and think the fixture is wrong.

PR #2171, opened 2026-07-28
fix(param): constrain ELEN to power-of-two element widths

### LINKED ISSUE #2170: ELEN parameter is an unconstrained integer; intended power-of-two bound sits commented out and references a nonexistent `popcount` IDL function
## Describe the bug

`spec/std/isa/param/ELEN.yaml` declares the parameter as an unconstrained integer:

```yaml
schema:
  type: integer
```

There is no `minimum`, no `maximum`, and no `enum`, so any integer validates — including `0`, `7`, and negative values. ELEN is the maximum size in bits of a vector element, so none of those are meaningful.

The file already states the intended constraint, but as a commented-out block:

```yaml
schema:
  type: integer
#    requirements:
#      idl(): |
#        -> (ELEN >='h8) && (popcount(ELEN) == 1);
#      reason: ELEN >= 8, which must be a power of 2.
```

## Why the constraint is currently unenforceable as written

Two independent reasons, both worth recording because they explain why this has sat commented out:

1. **`popcount` is not an IDL function.** There is no `popcount` in `spec/std/isa/isa/builtin_functions.idl`, in `globals.isa`, or anywhere else under `spec/`. The only occurrences in the repo are the `Zbb` `cpop` instruction prose and a C++ `__builtin_popcount` in `backends/`. Uncommenting the block as-is would reference an undefined function.

2. **The block is indented inside `schema:`.** `requirements` is a top-level property in `param_schema.json`, and the 46 parameters that use it correctly place it at top level (e.g. `ASID_WIDTH.yaml`). Uncommenting these lines verbatim produces a YAML scanner error, since `requirements:` would sit at a deeper indent than the `type: integer` mapping it follows.

## Proposed fix

Rather than reviving an IDL constraint that needs a function that does not exist, the constraint is directly expressible in JSON Schema, which is where the other bounded parameters express theirs:

```yaml
schema:
  type: integer
  enum: [8, 16, 32, 64]
```

This captures "power of two, at least 8" exactly over the architecturally meaningful range. ELEN is bounded above by 64 in practice: `Zve32x` documents "maximum element width (ELEN) of 32 bits" and `Zve64x` documents 64, and `SEW_MIN` — which the spec constrains to `SEW_MIN <= ELEN` — is itself already `enum: [8, 16, 32, 64]`. Using the same enum keeps the two consistent.

An `enum` is preferred over `minimum`/`maximum` here because it also excludes non-powers-of-two such as 24 or 48, which a range cannot express. This mirrors the existing `32bit_unsigned_pow2` / `64bit_unsigned_pow2` idiom in `schema_defs.json`.

### Verified

- All three configs that set `ELEN` (`rv32-vector`, `rv64-vector`, `example_rv64_with_overlay`) use `ELEN: 64` and continue to validate.
- `ELEN.yaml` validates against `param_schema.json` with the change applied.

## Related, not in scope here

- The same commented-out-`requirements` pattern appears in `VLEN.yaml`, `SEW_MIN.yaml`, and `VECTOR_LS_INDEX_MAX_EEW.yaml`. Those need cross-parameter constraints (`VLEN >= ELEN`, `SEW_MIN <= ELEN`), which JSON Schema cannot express per-file and which therefore genuinely need the `requirements`/IDL path — related to #1890. Keeping this issue to the single-parameter case that is cleanly fixable.
- `long_name: TODO` in this file is being handled separately (#2155, #2124).

Hiteshsai007:
ELEN was declared as a bare `type: integer`, so any value validated, including 0, 7 and negatives. VLEN used `minimum: 32` / `maximum: 65536`, which admitted every non-power-of-two in between. Both quantities are powers of two.

Both now use explicit hex power-of-two enums, matching the convention already used by MTVEC_BASE_ALIGNMENT_DIRECT. ELEN runs 0x8 to 0x200000, the theoretical maximum of (max VLEN) * (#VRs) = 64Ki * 32, since the ISA states no maximum for ELEN itself. VLEN runs 0x20 to 0x10000.

This also retires the commented-out `requirements` block in each file. Neither could be enabled as written: they call `popcount`, which does not exist in the IDL, and they are indented inside `schema:`, so uncommenting them is a YAML parse error. The `VLEN >= ELEN` constraint in VLEN's block was additionally incorrect, since the spec explicitly admits ELEN > VLEN with an element held across multiple vector registers.

Closes #2170

### DIFF
--- a/spec/std/isa/param/ELEN.yaml
+++ b/spec/std/isa/param/ELEN.yaml
@@ -8,13 +8,35 @@ kind: parameter
 name: ELEN
 description: |
   The maximum size in bits of a vector element that any operation can produce or consume.
-long_name: TODO
+long_name: maximum vector element bit width
 schema:
+  description: |
+    An unsigned power of 2 that is at least 8. The ISA states no maximum; the
+    absolute upper bound is (max VLEN) * (#VRs) = 64Ki * 32 = 2Mi, since an
+    element cannot exceed the total vector register state.
   type: integer
-#    requirements:
-#      idl(): |
-#        -> (ELEN >='h8) && (popcount(ELEN) == 1);
-#      reason: ELEN >= 8, which must be a power of 2.
+  enum:
+    [
+      0x8,
+      0x10,
+      0x20,
+      0x40,
+      0x80,
+      0x100,
+      0x200,
+      0x400,
+      0x800,
+      0x1000,
+      0x2000,
+      0x4000,
+      0x8000,
+      0x10000,
+      0x20000,
+      0x40000,
+      0x80000,
+      0x100000,
+      0x200000,
+    ]
 definedBy:
   extension:
     name: Zvl32b
--- a/spec/std/isa/param/VLEN.yaml
+++ b/spec/std/isa/param/VLEN.yaml
@@ -8,15 +8,24 @@ kind: parameter
 name: VLEN
 description: |
   The number of bits in a single vector register.
-long_name: TODO
+long_name: vector register bit width
 schema:
   type: integer
-  minimum: 32
-  maximum: 65536
-#    requirements:
-#      idl(): |
-#        -> (VLEN >= ELEN) && (VLEN <='h10000) && (popcount(VLEN) == 1);
-#      reason: VLEN >= ELEN, which must be a power of 2, and must be no greater than 2^16.
+  enum:
+    [
+      0x20,
+      0x40,
+      0x80,
+      0x100,
+      0x200,
+      0x400,
+      0x800,
+      0x1000,
+      0x2000,
+      0x4000,
+      0x8000,
+      0x10000,
+    ]
 definedBy:
   extension:
     name: Zvl32b


spec/std/isa/param/ELEN.yaml
```diff
@@ -11,10 +11,7 @@ description: |
 long_name: TODO
 schema:
   type: integer
-#    requirements:
-#      idl(): |
-#        -> (ELEN >='h8) && (popcount(ELEN) == 1);
-#      reason: ELEN >= 8, which must be a power of 2.
+  enum: [8, 16, 32, 64]
```
ThinkOpenly: This seems like a good approach. An open question: the spec doesn't constrain ELEN to be <= 64. Indeed, it explicitly states the possibility of ELEN > VLEN, although there are no ratified use-cases to my knowledge. Should it be constrained to ratified use-cases, or "future proof" it to _possible_ use-cases?
```suggestion
  enum: [8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8196, 16384, 32768, 65536]
```

ThinkOpenly: Technically, ELEN can be as large as max VLEN (64k) * #VRs (32) = 2M.

ThinkOpenly:
Would you entertain similarly fixing VLEN in this PR as well? VLEN is a power of 2 up to 64k.

Hiteshsai007:
Applied, with one correction: the suggested list had `8196` where `8192` (2^13) was intended. It was the only non-power-of-two in the sequence — same class of slip as the `4095` in `schema_defs.json` (#2137), which is a decent argument that these lists want a test rather than careful reading. I've also extended to `32768` and `65536` to reach the ceiling the ISA states for VLEN.

On the open question — I've gone with future-proofing to possible use cases. Restricting to ratified ones means editing the enum every time a new `Zvl*b` or `Zve*` lands, and a schema that must be revised before a legal implementation can be described is the more annoying failure mode. Your 2M figure (VLEN 64k × 32 VRs) is a fair reading of the theoretical maximum; I stopped at 65536 to match the bound the ISA states directly rather than encoding a derived one. Happy to widen if you'd prefer.

Added VLEN as suggested:

```yaml
schema:
  type: integer
  enum: [32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536]
```

That replaces `minimum: 32` / `maximum: 65536`, which admitted every non-power-of-two in between. It also retires VLEN's commented-out `requirements` block, which couldn't be enabled as written: it calls `popcount`, which doesn't exist in the IDL, and it's indented inside `schema:` so uncommenting it verbatim is a YAML parse error.

What it does **not** carry over is the cross-parameter half, `VLEN >= ELEN`. JSON Schema can't express a relationship between two parameter files. That matters more now than before: with ELEN reaching 65536 and VLEN capped at 65536, `ELEN > VLEN` is representable and nothing rejects it. You noted the spec permits that anyway, so it may be fine — but if it should be constrained, it needs the `requirements`/IDL path, related to #1890. Happy to follow up separately.

Verified: both files validate against `param_schema.json`, every enum entry is a power of two, and the three configs setting these (`ELEN: 64`, `VLEN: 128`) remain valid.

jordancarlin:
> Applied, with one correction: the suggested list had `8196` where `8192` (2^13) was intended. It was the only non-power-of-two in the sequence — same class of slip as the `4095` in `schema_defs.json` (#2137), which is a decent argument that these lists want a test rather than careful reading. I've also extended to `32768` and `65536` to reach the ceiling the ISA states for VLEN.

Should we use hex instead of decimal to avoid this? Power of 2 are easier to recognize in hex.

Hiteshsai007:
Yes, and there's precedent — `MTVEC_BASE_ALIGNMENT_DIRECT`, `MTVEC_BASE_ALIGNMENT_VECTORED`, `STVEC_BASE_ALIGNMENT_VECTORED` and `JVT_BASE_MASK` all already use hex. Consistency alone is a decent argument, and hex does make the shape self-evident: a power of two is `1`, `2`, `4` or `8` followed by zeros, so `0x2000` reads as correct in a way `8192` doesn't.

I've switched both files:

```yaml
# ELEN
enum: [0x8, 0x10, 0x20, 0x40, 0x80, 0x100, 0x200, 0x400, 0x800, 0x1000, 0x2000, 0x4000, 0x8000, 0x10000]

# VLEN
enum: [0x20, 0x40, 0x80, 0x100, 0x200, 0x400, 0x800, 0x1000, 0x2000, 0x4000, 0x8000, 0x10000]
```

One caveat worth putting on the record, though, because I don't think hex fully solves the problem it's being asked to solve.

The `4095` in `schema_defs.json` (#2137) traces back to exactly this situation. Per #2137, MTVEC alignment enums were converted to hex in #2090, and the note there is that *"the old decimal list already had 4095 and hex conversion preserved it"* — the bad value survived the conversion as `0xfff` and was only caught later by someone checking the arithmetic. `0xfff` and `0x1000` are one character apart just as `8196` and `8192` are.

So hex makes a correct list easier to *read*, but an incorrect entry still passes a careful reading. What actually catches this class is a test asserting every entry satisfies `v > 0 && (v & (v-1)) == 0` — which is what @titoatwork added for `schema_defs.json` in #2138, and what I've added for the Z3 side in #2179.

Happy with hex here for consistency and readability. Just flagging that the guarantee comes from the test rather than the notation, in case that affects whether these enums are worth deriving rather than hand-maintaining.

Hiteshsai007:
Whenever you have a moment, would you mind approving the workflow runs? They're sitting at action_required since this is my first PR here. Happy to fix whatever CI surfaces.

ThinkOpenly:
> Applied, with one correction: the suggested list had `8196` where `8192` (2^13) was intended.

Yep, typo. Thanks for catching.

> Your 2M figure (VLEN 64k × 32 VRs) is a fair reading of the theoretical maximum; I stopped at 65536 to match the bound the ISA states directly rather than encoding a derived one.

I don't think the ISA states a maximum for ELEN. It only states:
> The maximum size in bits of a vector element that any operation can produce or consume, ELEN ≥ 8, which must be a power of 2.

So, complete "future proof" would include them all:
```
enum: [0x8, 0x10, 0x20, 0x40, 0x80, 0x100, 0x200, 0x400, 0x800, 0x1000, 0x2000, 0x4000, 0x8000, 0x10000, 0x20000, 0x40000, 0x80000, 0x100000, 0x200000 ]
```

Add a comment that says something like "absolute maximum ELEN is (max VLEN) * (#VRs)".

> What it does **not** carry over is the cross-parameter half, `VLEN >= ELEN`.

The spec doesn't state that, either. Indeed, it states exactly:
> this specification has been revised to admit the possibility of future extensions that allow ELEN > VLEN, wherein one element is held using bits from multiple vector registers.

Hiteshsai007:
You're right on both, and the first one is my error — I claimed the ISA states a maximum for ELEN and it doesn't. It says only "ELEN ≥ 8, which must be a power of 2". I'd taken `Zve64x`'s documented maximum element width as a universal bound, which it isn't; it's that extension's maximum.

Extended to `0x200000`, with the rationale in `schema.description`:

```yaml
schema:
  description: |
    An unsigned power of 2 that is at least 8. The ISA states no maximum; the
    absolute upper bound is (max VLEN) * (#VRs) = 64Ki * 32 = 2Mi, since an
    element cannot exceed the total vector register state.
  type: integer
  enum: [0x8, ..., 0x100000, 0x200000]
```

That matches the `schema.description` convention already used by `MTVEC_BASE_ALIGNMENT_DIRECT`. 19 entries, 2^3 through 2^21, no gaps — and `0x200000` is exactly 65536 × 32, so the bound is derived rather than asserted.

On `VLEN >= ELEN`: thanks, that's a better reason to drop it than the one I gave. I removed the commented block because it couldn't be enabled — `popcount` doesn't exist in the IDL and the indentation makes it a parse error — but you're pointing out the constraint was wrong on its own terms, since the spec explicitly admits `ELEN > VLEN` with an element held across multiple vector registers. So it shouldn't be reinstated in any form, and my note about it being a gap for #1890 to close was misdirected. Withdrawn.

VLEN keeps its own `0x20`–`0x10000` range, which the ISA does bound directly.

spec/std/isa/param/VLEN.yaml
```diff
@@ -11,12 +11,7 @@ description: |
 long_name: TODO
```
ThinkOpenly:
```suggestion
long_name: vector register bit width
```

spec/std/isa/param/ELEN.yaml
```diff
@@ -10,11 +10,16 @@ description: |
   The maximum size in bits of a vector element that any operation can produce or consume.
 long_name: TODO
```
ThinkOpenly:
```suggestion
long_name: maximum vector element bit width
```

ThinkOpenly (changes requested):
If you'll tolerate a couple more nits, since we're already editing these files.

ThinkOpenly (approved):

ThinkOpenly:
Oh, yeah: we enabled mandatory "sign-offs" for this repository yesterday. Consider squashing your stack of commits into a single commit, appending a "Signed-off-by" line for yourself, and force-pushing.

Secondly, please revise your PR description to match the current PR content, which now includes VLEN. Also, the description is FAR too long, so reduce it to: a brief description of the issue, and, if necessary, a description of the fix. No Markdown headings, please. Thanks!

Hiteshsai007:
Squashed to a single signed-off commit and force-pushed, and trimmed the description to match the current content including VLEN. Also fixed the prettier failure — the enums needed one value per line.

Workflows are pending approval again after the force-push.

ThinkOpenly (approved):

PR #2175, opened 2026-07-28
fix(schema): enforce draft-07 keyword allowlist on param schema blocks

### LINKED ISSUE #2173: param_schema.json: additionalProperties is ignored due to draft-07 $ref semantics
## Describe the bug

`param_schema.json` sets `additionalProperties: false` on the `schema` property, but it has no effect. Every parameter's `schema` block currently accepts arbitrary unknown keys, including typos, which are silently ignored rather than rejected.

```json
"schema": {
  "$ref": "json-schema-draft-07.json#",
  "additionalProperties": false
}
```

In JSON Schema **draft-07**, any keyword placed as a *sibling* of `$ref` is ignored — the `$ref` replaces the entire schema object. `param_schema.json` declares `"$schema": "http://json-schema.org/draft-07/schema#"`, so this rule applies. The `additionalProperties: false` is dead.

(This was changed in draft 2019-09, where `$ref` became an ordinary applicator and siblings *are* honoured. So the construct reads as correct to anyone used to the newer drafts, which is likely how it slipped in.)

## Reproduction

Validating a parameter against `param_schema.json` with the repo's own schemas:

| `schema` block under test | Current result |
|---|---|
| `{type: integer, minimum: 8}` | accepted (correct) |
| `{type: integer, minimun: 8}` — typo | **accepted** |
| `{type: integer, requirements: {...}}` — misplaced key | **accepted** |
| `{type: integer, bogus_xyz: 1}` — nonsense | **accepted** |

A misspelled `minimum` silently drops the bound: the parameter ends up unconstrained while appearing to be constrained in review.

## Why this is worth fixing

This is the enforcement gap behind a class of parameter defects that keep being found one at a time. Recent examples where a bound was missing or wrong and nothing caught it: #2102 (`STVAL_WIDTH`), #2137 (non-power-of-two in the shared pow2 enums), #2170 (`ELEN` unconstrained). A working constraint here turns "reviewer notices" into "CI notices".

It also matters for tooling that *emits* parameter YAML — a generator producing a slightly-wrong key name gets no feedback from schema validation.

Related: #817 tracks enforcing `additionalProperties: false` across all schemas, and #2108 is landing that for several of them. This issue is a distinct failure mode: `param_schema.json` **already has** the setting and it *looks* enforced, so an audit that greps for the key would mark it done. I checked the other schema files — `param_schema.json` is the only place in `spec/schemas/` where `additionalProperties` sits next to a `$ref` and is therefore inert.

## Proposed fix

The `$ref` needs to be moved inside a subschema so `propertyNames` is no longer a sibling:

```json
"schema": {
  "allOf": [{ "$ref": "json-schema-draft-07.json#" }],
  "propertyNames": {
    "enum": ["$id", "$schema", "$comment", "$ref", "title", "description",
             "default", "readOnly", "writeOnly", "examples", "multipleOf",
             "maximum", "exclusiveMaximum", "minimum", "exclusiveMinimum",
             "maxLength", "minLength", "pattern", "additionalItems", "items",
             "maxItems", "minItems", "uniqueItems", "contains", "maxProperties",
             "minProperties", "required", "additionalProperties", "definitions",
             "properties", "patternProperties", "dependencies", "propertyNames",
             "const", "enum", "type", "format", "contentMediaType",
             "contentEncoding", "if", "then", "else", "allOf", "anyOf",
             "oneOf", "not"]
  }
}
```

### Why `propertyNames` rather than just relocating `additionalProperties`

`additionalProperties` only knows about keys named in a *sibling* `properties`. Since the allowed keys live in the referenced draft-07 metaschema rather than locally, moving `additionalProperties: false` alongside `allOf` rejects **everything**, including `type`. I tested that variant: all 227 parameter files fail.

`propertyNames` constrains key names directly and composes correctly with the `$ref`. The enum is the draft-07 keyword list, taken from `properties` in the vendored `spec/schemas/json-schema-draft-07.json` so it stays in step with the metaschema the repo actually uses.

### Verified

| Case | Current | With fix |
|---|---|---|
| `{type, minimum}` | accept | accept |
| typo `minimun` | accept | **reject** |
| nested `requirements` | accept | **reject** |
| garbage key | accept | **reject** |

- **All 227 files in `spec/std/isa/param/` still validate** under the proposed fix — zero regressions.
- I enumerated every key currently used inside `schema:` blocks across those files: `type`, `enum`, `minimum`, `maximum`, `items`, `minItems`, `maxItems`, `additionalItems`, `uniqueItems`, `default`, `const`, `description`, `allOf`, `not`, `$ref`. All are draft-07 keywords and all are in the enum above.

## Notes for whoever picks this up

- This edits a published schema, so it likely needs a `$id` bump (`v0.1` → `v0.2`) per the immutability rule enforced by `tools/scripts/check_schema_versions.py`, and a `bin/chore gen schema-docs` run to refresh `doc/docs/schemas/.../param_schema.mdx`. Worth confirming with a maintainer whether a validation-tightening change warrants a version bump or is treated as a fix.
- The repo validates with **json_schemer** (Ruby) as well as Python `jsonschema`. My testing was Python-only — I could not run the Ruby path locally (no Ruby toolchain in my environment; `bin/mise` cannot reach `mise.jdx.dev` to bootstrap). json_schemer is draft-07 compliant and should behave identically, but that is worth confirming in CI rather than taking on my word.
- If tightening this surfaces violations in **out-of-tree overlays**, that would be a deliberate behaviour change for downstream users, not just an internal fix — flagging in case that affects whether this is wanted.

I'm happy to open a PR for this if the approach looks right.

Hiteshsai007:
param_schema.json sets `additionalProperties: false` on the `schema` property, but the keyword sits as a sibling of `$ref`. JSON Schema draft-07 ignores keywords beside `$ref`, so the constraint is inert and every parameter's `schema` block currently accepts arbitrary unknown keys. A misspelled `minimun` is accepted silently, dropping the bound it was meant to express.

Move the `$ref` inside `allOf` so it no longer suppresses siblings, and constrain key names with `propertyNames` against the draft-07 keyword list, read from the vendored metaschema so it cannot drift. `additionalProperties: false` cannot be used here, because the permitted keys come from the referenced metaschema rather than a local `properties` map, so it would reject everything including `type`.

Bumps `$id` to v0.2, since v0.1 is already published as a release asset.

All 462 `kind: parameter` files in the repository still validate, including the udb and idlc test fixtures.

Closes #2173

### DIFF
--- a/spec/schemas/param_schema.json
+++ b/spec/schemas/param_schema.json
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "v0.1",
+  "$id": "v0.2",
   "type": "object",
   "required": [
     "$schema",
@@ -33,8 +33,62 @@
       "$ref": "schema_defs.json#/$defs/condition"
     },
     "schema": {
-      "$ref": "json-schema-draft-07.json#",
-      "additionalProperties": false
+      "allOf": [
+        {
+          "$ref": "json-schema-draft-07.json#"
+        }
+      ],
+      "propertyNames": {
+        "description": "Only JSON Schema draft-07 keywords are permitted. Prevents typos and misplaced keys from being silently ignored.",
+        "enum": [
+          "$comment",
+          "$id",
+          "$ref",
+          "$schema",
+          "additionalItems",
+          "additionalProperties",
+          "allOf",
+          "anyOf",
+          "const",
+          "contains",
+          "contentEncoding",
+          "contentMediaType",
+          "default",
+          "definitions",
+          "dependencies",
+          "description",
+          "else",
+          "enum",
+          "examples",
+          "exclusiveMaximum",
+          "exclusiveMinimum",
+          "format",
+          "if",
+          "items",
+          "maxItems",
+          "maxLength",
+          "maxProperties",
+          "maximum",
+          "minItems",
+          "minLength",
+          "minProperties",
+          "minimum",
+          "multipleOf",
+          "not",
+          "oneOf",
+          "pattern",
+          "patternProperties",
+          "properties",
+          "propertyNames",
+          "readOnly",
+          "required",
+          "then",
+          "title",
+          "type",
+          "uniqueItems",
+          "writeOnly"
+        ]
+      }
     },
     "requirements": {
       "$ref": "schema_defs.json#/$defs/condition"
--- a/doc/docs/schemas/index.mdx
+++ b/doc/docs/schemas/index.mdx
@@ -109,6 +109,11 @@ Schemas are versioned independently of the UDB repository. Each schema file decl
 </div>
 
 <div style={{padding: '1rem', border: '1px solid var(--ifm-color-emphasis-300)', borderRadius: '8px'}}>
+  <strong><a href="./v0.2/param_schema">Param Schema</a></strong><br />
+  <span className="badge badge--secondary" style={{marginTop: '0.4rem', display: 'inline-block'}}>v0.2</span>
+</div>
+
+<div style={{padding: '1rem', border: '1px solid var(--ifm-color-emphasis-300)', borderRadius: '8px'}}>
   <strong><a href="./v0.1/prm_schema">Prm Schema</a></strong><br />
   <span className="badge badge--secondary" style={{marginTop: '0.4rem', display: 'inline-block'}}>v0.1</span>
 </div>
--- a/doc/docs/schemas/v0.2/param_schema.mdx
+++ b/doc/docs/schemas/v0.2/param_schema.mdx
@@ -1,0 +1,44 @@
+---
+title: Param Schema (v0.2)
+sidebar_label: Param Schema
+custom_edit_url: null
+# This file is auto-generated from param_schema.json
+# Do not edit manually - run `bin/chore gen schema-docs` to regenerate
+---
+
+
+# Param Schema
+
+<span class="badge badge--secondary">v0.2</span>
+
+<br />
+
+:::note Auto-generated
+This page is generated from [`param_schema.json`](https://github.com/riscv/riscv-unified-db/blob/main/spec/schemas/param_schema.json) by the [schema doc generator](https://github.com/riscv/riscv-unified-db/blob/main/tools/internal-gems/schema_doc_gen/lib/schema_doc_gen.rb). To update this page, edit the schema file and run `bin/chore gen schema-docs`.
+:::
+
+
+## Properties
+
+| Property | Type | Required | Description |
+|----------|------|----------|-------------|
+| `$schema` | `any` | ✓ |  |
+| `kind` | `any` | ✓ |  |
+| `long_name` | `string` | ✓ | Short description of the parameter |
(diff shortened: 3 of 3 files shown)


spec/schemas/param_schema.json
```diff
@@ -33,8 +33,62 @@
       "$ref": "schema_defs.json#/$defs/condition"
     },
     "schema": {
-      "$ref": "json-schema-draft-07.json#",
-      "additionalProperties": false
+      "allOf": [
+        {
+          "$ref": "json-schema-draft-07.json#"
+        }
+      ],
```
ThinkOpenly, line 40: Ignoring the "propertyNames" added below, would the following be sufficient to address the issue where "additionalProperties" is ignored?
```suggestion
      "allOf": [
        {
          "$ref": "json-schema-draft-07.json#"
        },
        "additionalProperties": false
      ],
```

ThinkOpenly: And if the above is true, is the "propertyNames" restriction added below necessary?

Hiteshsai007: Thanks for the review, Paul — good questions; I went down both of these paths before settling on this shape.
**Would `additionalProperties: false` (inside `allOf`) be sufficient?** No — it can't work here, in any placement. `additionalProperties` only considers keys named by `properties`/`patternProperties` in the *same schema object*, and the permitted keys come from the referenced draft-07 metaschema rather than a local `properties` map. So `additionalProperties: false` — whether as a second `allOf` element or as a sibling of `allOf` — rejects *every* key, including `type`. I tested that variant against the vendored `spec/schemas/json-schema-draft-07.json` with a draft-07 validator:
| Schema block | `{type: integer, minimum: 0}` | `{type: integer, minimun: 0}` (typo) |
|---|---|---|
| `allOf: [{$ref}]` + `propertyNames` enum (this PR) | ✅ accepted | ✅ rejected |
| `allOf: [{$ref}, {"additionalProperties": false}]` | ❌ rejected (`type`, `minimum` unexpected) | ❌ rejected |
| `allOf: [{$ref}]` + sibling `additionalProperties: false` | ❌ rejected (`type`, `minimum` unexpected) | ❌ rejected |
Even a perfectly valid schema fails with "Additional properties are not allowed ('minimum', 'type' were unexpected)", so all 462 parameter files would become invalid. (As written, the suggested diff also isn't valid JSON — the second `allOf` element would need to be `{"additionalProperties": false}`.)
**Is the `propertyNames` restriction necessary?** Yes, for exactly that reason: it's the only keyword that constrains key names without a local `properties` map, so it composes correctly with the `$ref`-through-`allOf` pattern. The alternative that does work — a local `properties: {<each draft-07 keyword>: true}` map plus `additionalProperties: false` — duplicates the keyword list and can drift; the `propertyNames` enum here is taken from the `properties` of the vendored `json-schema-draft-07.json`, so it stays in step with the metaschema the repo actually uses.
If you'd prefer the local-`properties`-map shape over `propertyNames`, the behavior is identical and I'm happy to switch.
---

Hiteshsai007:

Issue #2176, opened 2026-07-28
z3: pow2 $ref branches contain an unsatisfiable zero disjunct, and nothing pins the solver's accepted set


Hiteshsai007:
## Describe the bug

`Udb::Z3ParameterTerm.constrain_int` handles the `32bit_unsigned_pow2` and `64bit_unsigned_pow2` `$ref`s by re-deriving the power-of-two constraint algebraically (`tools/ruby-gems/udb/lib/udb/z3.rb:414-420`):

```ruby
elsif schema_hsh.fetch("$ref").split("/").last == "32bit_unsigned_pow2"
  assertions << ((term == 0) | (0 == (term & (term - 1))))
  assertions << ((term.unsigned_gt(0)) & (term.unsigned_le(2**32 - 1)))
elsif schema_hsh.fetch("$ref").split("/").last == "64bit_unsigned_pow2"
  assertions << ((term == 0) | (0 == (term & (term - 1))))
  assertions << ((term.unsigned_gt(0)) & (term.unsigned_le(2**64 - 1)))
```

The two assertions are conjoined and contradict each other on zero:

- assertion 1 — `(term == 0) | (0 == (term & (term - 1)))` — permits `term == 0`
- assertion 2 — `(term.unsigned_gt(0)) & (...)` — requires `term > 0`

So the `(term == 0)` disjunct can never be satisfied. It is dead code.

## Why it is worth removing

Validation is correct today: zero is rejected, just by the lower bound rather than by the bit identity. The problem is what the code communicates. It reads as "zero is an accepted power of two" to anyone maintaining the function, and the `uint32`/`uint64` arms immediately above genuinely do use `unsigned_ge(0)`, which makes the inconsistency easy to misread as intentional rather than vestigial — and easy to "fix" in the wrong direction by relaxing the bound to match.

## Second, related gap: the solver side is untested

Nothing pins which values these `$ref` branches accept. The two existing tests in `test_conditions.rb` (`test_z3_constrain_int_allof_with_32bit_pow2_ref` and its 64-bit twin) assert only that such a schema is *satisfiable* — they never check *which* values are admitted. Any change to the assertions, including relaxing that lower bound, would pass silently.

## Expected

Remove the `(term == 0)` disjunct from both arms, and add tests characterising the accepted set: in-range powers of two accepted, zero and non-powers-of-two rejected, and the 32-bit upper bound enforced while the 64-bit branch accepts `2**32`.

## Verification notes

[...]

PR #2179, opened 2026-07-28
fix(z3): drop unsatisfiable zero disjunct from pow2 constraints

Hiteshsai007:
The 32bit_unsigned_pow2 and 64bit_unsigned_pow2 `$ref` branches of `constrain_int` each build two assertions that are conjoined: the first permits `term == 0`, the second requires `term > 0`. The `(term == 0)` disjunct can therefore never be satisfied, so it is dead code. Behaviour does not change by removing it — zero is still excluded, by the lower bound rather than by the bit identity.

It is worth removing because of what it communicates. It reads as "zero is an accepted power of two" to anyone maintaining the function, and the uint32/uint64 branches immediately above genuinely do use `unsigned_ge(0)`, which makes the inconsistency easy to misread as intentional and to "fix" in the wrong direction.

Also adds tests characterising which values these branches accept, which was previously unpinned: the two existing tests in test_conditions.rb assert only that such a schema is satisfiable, never which values are admitted, so any change to the assertions would pass silently. The new tests cover in-range powers of two, zero, non-powers-of-two including 4095, and the 32-bit upper bound.

I could not run these locally — ./bin/ruby reports `unsupported OS: MINGW64_NT` — so CI is their first execution.

Refs #2176


Hiteshsai007 (commented):
This PR correctly removes a dead (term == 0) disjunct from the Z3 constraints for 32- and 64-bit “unsigned power of two” branches and adds a focused test suite that characterises the solver-side behaviour. I recommend merging after CI passes; the change is low-risk and improves readability and correctness of the assertions.

Hiteshsai007:

tools/ruby-gems/udb/test/run.rb
ThinkOpenly, line 1: Sincerely appreciate the effort, but I prefer removing the tests from this PR.

ThinkOpenly (changes requested):

Hiteshsai007:
Thanks for the review @ThinkOpenly! Removed the tests as requested — the PR now only contains the z3.rb fix. Happy to submit the tests in a separate follow-up PR if desired.

PR #2187, opened 2026-07-28
fix(deps): verify downloaded binaries before making them executable

Hiteshsai007:
The 32bit_unsigned_pow2 and 64bit_unsigned_pow2 `$ref` branches of `constrain_int` each build two assertions that are conjoined: the first permits `term == 0`, the second requires `term > 0`. The `(term == 0)` disjunct can therefore never be satisfied, so it is dead code. Behaviour does not change by removing it — zero is still excluded, by the lower bound rather than by the bit identity.

It is worth removing because of what it communicates. It reads as "zero is an accepted power of two" to anyone maintaining the function, and the uint32/uint64 branches immediately above genuinely do use `unsigned_ge(0)`, which makes the inconsistency easy to misread as intentional and to "fix" in the wrong direction.

Also adds tests characterising which values these branches accept, which was previously unpinned: the two existing tests in test_conditions.rb assert only that such a schema is satisfiable, never which values are admitted, so any change to the assertions would pass silently. The new tests cover in-range powers of two, zero, non-powers-of-two including 4095, and the 32-bit upper bound.

I could not run these locally — ./bin/ruby reports `unsupported OS: MINGW64_NT` — so CI is their first execution.

Refs #2176

### DIFF
--- a/tools/ruby-gems/udb/ext/udb_download/extconf.rb
+++ b/tools/ruby-gems/udb/ext/udb_download/extconf.rb
@@ -82,7 +82,6 @@ unless File.exist?(espresso_file)
   $stderr.puts "Downloading espresso (#{espresso_version}, #{cpu}) from GitHub releases..."
   $stderr.puts "  URL: #{url_str}"
   File.binwrite(espresso_file, download_with_redirects(url_str))
-  File.chmod(0o755, espresso_file)
   $stderr.puts "  Saved to #{espresso_file}"
 
   # Download and verify checksum
@@ -102,10 +101,15 @@ unless File.exist?(espresso_file)
     $stderr.puts "  Expected: #{expected}"
     $stderr.puts "  Got:      #{actual}"
     $stderr.puts "  The downloaded file may be corrupted or tampered with."
+    # Remove the unverified artifact. Leaving it in place makes the next
+    # run skip both the download and this check via File.exist?.
+    FileUtils.rm_f(espresso_file)
     abort "Checksum verification failed"
   end
 
   $stderr.puts "  Checksum verified successfully."
+
+  File.chmod(0o755, espresso_file)
 end
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Hiteshsai007:
The two failing checks are codecov/patch at 10% against a 50% target. Worth flagging what the uncovered lines are, since I don't think they're straightforwardly testable.

Most of the diff is in ext/udb_download/extconf.rb, which runs at gem-install time and is never loaded by the test suite — there is no existing coverage for it, or for dep_paths.rb. Testing the changed paths means either performing real downloads or stubbing Net::HTTP inside a file that only executes during native extension builds.

The dep_paths.rb half is more testable: download_binary could take an injectable fetcher, which would let the checksum-mismatch and non-HTTPS paths be exercised without network access. That's a slightly larger refactor of a file this PR otherwise only adds to, so I'd rather ask than assume — happy to add it if you'd like the coverage, or to leave the check red if the policy is understood to not apply to install-time code.

Hiteshsai007:
Added dep_paths to the regress-udb-unit-test matrix in tools/test/regress-tests.yaml, and mirrored it into the generated regress.yml by hand since I can't run ./bin/chore gen regress locally. Happy for autofix to normalise it if the generator formats it differently.

Hiteshsai007:
The new test exposed a latent issue: dep_paths.rb calls T.cast but never requires sorbet-runtime, so it only loads when something else has already required it. The test requires it explicitly rather than adding the require to dep_paths.rb, since that's arguably a separate fix — happy to move it into the lib file if preferred.

Hiteshsai007:
Update: I've removed the test file I'd added and reverted this to the original two-file change.

I did try covering the dep_paths.rb half — the tests worked in principle, but ran into a chain of environment-specific issues I couldn't verify locally (dep_paths.rb calls T.cast without requiring sorbet-runtime, and download_with_redirects is private, which breaks the obvious stubbing approach). Rather than keep iterating on CI, I'd rather land the security fix and offer tests as a follow-up if wanted.

On the coverage number itself: SimpleCov roots at the udb gem, so extconf.rb — 18 of the 22 changed lines — is never measured. It only executes during native gem installation. That leaves dep_paths.rb, which has no existing coverage either.

Happy to add tests in a separate PR if a maintainer would like the coverage, or to have the patch threshold waived here if install-time code is understood to be out of scope.

PR #2189, opened 2026-07-28
fix(param): constrain CACHE_BLOCK_SIZE to a power of two

### LINKED ISSUE #2188: param: CACHE_BLOCK_SIZE schema admits non-power-of-two sizes and a 2**64-1 maximum
## Describe the bug

`spec/std/isa/param/CACHE_BLOCK_SIZE.yaml` constrains the parameter with a plain integer range:

```yaml
schema:
  type: integer
  minimum: 1
  maximum: 18446744073709551615
```

That admits `3`, `7`, `100` and `18446744073709551615` as cache block sizes.

## Why this matters

The CMO specification defines the shape of a cache block normatively (`riscv-isa-manual`, `src/cmo.adoc`):

> Caches organize copies of data into _cache blocks_, each of which represents a contiguous, naturally aligned power-of-two (or _NAPOT_) range of memory locations.

So a cache block size is a power of two by definition, and the schema does not say so. The upper bound is separately implausible: 2**64-1 bytes is not a cache block, and it is not a power of two either.

The repository already relies on the power-of-two property elsewhere. `Zic64b` pins the value with `requirements: param: CACHE_BLOCK_SIZE equal: 64`, and the reference pseudocode in `cbo.clean`, `cbo.flush` and `cbo.inval` computes the block address as `X[xs1] & ~(CACHE_BLOCK_SIZE-1)`, an alignment mask that is only correct for a power-of-two size. Those pseudocode lines are commented out, so the normative text above is the argument rather than them.

## Expected

The schema should permit only powers of two. `spec/schemas/schema_defs.json` already defines exactly this vocabulary as `$defs/64bit_unsigned_pow2`, and three parameters already reference `schema_defs` this way (`ARCH_ID_VALUE`, `CONFIG_PTR_ADDRESS`, `IMP_ID_VALUE`).

## Proposed fix

```yaml
schema:
  $ref: schema_defs.json#/$defs/64bit_unsigned_pow2
```

No configuration changes. The only value set anywhere in the repository is `CACHE_BLOCK_SIZE: 64` in `cfgs/example_rv64_with_overlay.yaml`, which remains valid, and `Zic64b`'s requirement is still satisfied.

## One interaction worth flagging

`64bit_unsigned_pow2` currently lists **4095**, which is itself not a power of two. That is the subject of #2137 and PR #2138. Until that lands, referencing the def would still admit 4095 for this parameter. The change here is still an improvement on its own, since it removes 2**64-1 and every odd size, and it gives that def a second consumer. I mention it so the residual gap is on the record rather than discovered later.

titoatwork:
CACHE_BLOCK_SIZE allows any integer from 1 through 2^64-1, even though the CMO specification defines a cache block as a naturally aligned power-of-two range of memory locations.

Constrain it to an enum of 2^0 through 2^63, written in hex to match `MTVEC_BASE_ALIGNMENT_DIRECT` and its siblings, where a non-power-of-two stands out visually.

The shared `schema_defs.json#/$defs/64bit_unsigned_pow2` would be the natural thing to reference instead, but `Idl::Type.from_json_schema_scalar_type` resolves only `uint32` and `uint64` and raises `unhandled ref` on anything else, so no parameter can use it. Filed as #2199.

Closes #2188.


spec/std/isa/param/CACHE_BLOCK_SIZE.yaml
```diff
@@ -11,9 +11,10 @@ description: "The observable size of a cache block, in bytes
   "
 long_name: TODO
 schema:
-  type: integer
-  minimum: 1
-  maximum: 18446744073709551615
+  # A cache block is a "naturally aligned power-of-two (or NAPOT) range of memory
+  # locations" (RISC-V CMO spec, Zicbom/Zicbop/Zicboz). The previous bounds
+  # (1 .. 2**64-1) admitted non-power-of-two sizes.
```
ThinkOpenly: Move the first sentence into the "description" field above.
Remove the second sentence. "previous" will have no context in isolation.

ThinkOpenly (changes requested):
We use a merge queue, which squashes all commits and uses the PR description as the commit message. The only things required there are:
- Describe the issue (in present tense), like "The parameter allows too many values" (but better than that).
- Describe the fix if it is anything more complicated than an obvious fix to the issue.
No test results, although they would be helpful in a follow-on comment. No "nothing breaks". No fancy formatting. Thanks!

spec/std/isa/param/CACHE_BLOCK_SIZE.yaml
titoatwork: Got it, thanks for clarifying. I’ll trim the PR description to just the issue and fix, move the first sentence into the parameter description, and remove the second one.

spec/std/isa/param/CACHE_BLOCK_SIZE.yaml
```diff
@@ -6,14 +6,79 @@
 $schema: param_schema.json#
 kind: parameter
 name: CACHE_BLOCK_SIZE
-description: "The observable size of a cache block, in bytes
-
-  "
+description: The observable size of a cache block, in bytes. A cache block is a
+  naturally aligned power-of-two (NAPOT) range of memory locations, per the CMO
+  specification.
 long_name: TODO
```
ThinkOpenly: Since we're here, if you don't mind:
```suggestion
long_name: cache block size
```

spec/std/isa/param/CACHE_BLOCK_SIZE.yaml
```diff
@@ -6,14 +6,79 @@
...
 schema:
   type: integer
-  minimum: 1
-  maximum: 18446744073709551615
+  enum:
```
ThinkOpenly, line 14: I see you converted from a `$ref` to a full expansion. Was there a reason?

spec/std/isa/param/CACHE_BLOCK_SIZE.yaml
```diff
@@ -6,14 +6,79 @@
...
+    [
+      1,
```
ThinkOpenly: We're now leaning toward hexadecimal strings, as they catch power-of-2 discrepancies (like "4095") easier, visually. Would you mind either converting back to the `$ref` or using hex here?

ThinkOpenly (changes requested):

titoatwork:
Both done in `145dde6b`.

On the `$ref`: `Idl::Type.from_json_schema_scalar_type` resolves only `schema_defs.json#/$defs/uint32` and `.../uint64`, and raises `unhandled ref` on anything else, so a parameter cannot point at `64bit_unsigned_pow2`. Wrapping it in `allOf` does not help either, since that branch recurses through the same method. The Z3 constraint path does support both, in `Z3ParameterTerm.constrain_int`, which is where the asymmetry comes from. That is #2199. Happy to open the `type.rb` patch, since the two cases would mirror the existing ones, and this file can go back to the `$ref` once it lands.

Hex in the meantime, matching `MTVEC_BASE_ALIGNMENT_DIRECT` and its siblings. Values unchanged, 2^0 through 2^63.

ThinkOpenly:
> On the `$ref`: `Idl::Type.from_json_schema_scalar_type` resolves only `schema_defs.json#/$defs/uint32` and `.../uint64`, and raises `unhandled ref` on anything else

Well, I learned something today. _sigh_

Hex it is.

spec/std/isa/param/CACHE_BLOCK_SIZE.yaml
```diff
@@ -6,14 +6,79 @@
...
+      0x40000000000000,
+      0x80000000000000,
+      0x100000000000000,
+      0x200000000000000,
+      0x400000000000000,
+      0x800000000000000,
+      0x1000000000000000,
+      0x2000000000000000,
+      0x4000000000000000,
+      0x8000000000000000,
```
ThinkOpenly, line 79: I didn't think YAML supported a trailing comma in lists?

titoatwork:
It does, and it is already the convention here. On `main`, `MTVEC_BASE_ALIGNMENT_DIRECT` closes the same way, `0x8000000000000000,` on line 92 and `]` on line 93, as do `MTVEC_BASE_ALIGNMENT_VECTORED` at the same lines and `STVEC_BASE_ALIGNMENT_VECTORED` at 89 and 90. I took the shape from those.

It also survives the tooling on this head: inside `regress-pre-commit`, both `check-yaml` and `prettier` pass and prettier left the list unchanged, and `regress-schema-docs`, `regress-schema-versions` and the udb unit tests pass alongside them.

Happy to drop the comma if you would rather, though I would want to confirm prettier does not simply put it back, since it is what settles the layout of this block.

ThinkOpenly:
If it works, it works. A web search said it didn't, but you can't trust the web! ;-) Moving on...

spec/std/isa/param/CACHE_BLOCK_SIZE.yaml
```diff
@@ -6,14 +6,79 @@
 $schema: param_schema.json#
 kind: parameter
 name: CACHE_BLOCK_SIZE
-description: "The observable size of a cache block, in bytes
-
-  "
-long_name: TODO
+description: The observable size of a cache block, in bytes. A cache block is a
+  naturally aligned power-of-two (NAPOT) range of memory locations, per the CMO
+  specification.
```
ThinkOpenly: last nit
```suggestion
  naturally aligned power-of-two (NAPOT) range of memory.
```

titoatwork:
Applied in `cc02f45a`.

One red check that is not from this diff. `regress-riscv-tests-vector` failed while setting up the environment, before any test ran: `docker/setup-buildx-action` could not pull `moby/buildkit:buildx-stable-1`.

```
Head "https://registry-1.docker.io/v2/moby/buildkit/manifests/buildx-stable-1":
read tcp 10.1.0.252:49376->104.18.43.178:443: read: connection reset by peer
```

Steps 4 and 5, `check out riscv-tests` and `Run vector riscv-tests`, were both skipped, and the job ended after 16 seconds. I do not have rights to re-run it.

ThinkOpenly:
Closes #2188.

CACHE_BLOCK_SIZE allows any integer from 1 through 2^64-1, even though the CMO specification defines a cache block as a naturally aligned power-of-two range of memory locations.

Constrain it to an enum of 2^0 through 2^63, written in hex to match `MTVEC_BASE_ALIGNMENT_DIRECT` and its siblings, where a non-power-of-two stands out visually.

The shared `schema_defs.json#/$defs/64bit_unsigned_pow2` would be the natural thing to reference instead, but `Idl::Type.from_json_schema_scalar_type` resolves only `uint32` and `uint64` and raises `unhandled ref` on anything else, so no parameter can use it. Filed as #2199.

ThinkOpenly (approved):
Thank you!

Issue #2141, opened 2026-07-27
PRM PDF generator builds a shell string instead of an argv array (breaks on paths with spaces; latent command injection)


AlgoArtist06:
### Describe the bug

The PRM PDF generator builds its command as an argv array and then flattens it back into a single string, which hands the whole thing to a shell.

`tools/ruby-gems/udb/lib/udb/prm_generator.rb:673`:

```ruby
cmd = build_asciidoctor_command(adoc_path, pdf_path)   # proper argv array
...
Open3.popen3(cmd.join(" ")) do |stdin, stdout, stderr, wait_thr|
```

Two consequences:

**Word splitting - reproducible today.** Any path component containing a space is split into separate arguments. `-a pdf-themesdir=#{@template_dir / 'pdf-theme'}` is derived from the checkout location, so `./do prm:pdf[...]` is simply broken for anyone whose clone sits under a directory with a space in the name.

**Shell metacharacter interpretation.** `pdf_path` is `output_dir / prm_name / "pdf" / "#{prm_name}-specification.pdf"` and `prm_name` is the rake task argument. Anything in it that a shell treats specially is interpreted rather than passed through. Reaching it requires a matching `spec/custom/non_isa/prm_example/<name>.yaml` to exist (`load_prm` checks), so this is latent rather than directly exploitable from the CLI today - but the string is shell-evaluated, and the guard is an accident of an unrelated existence check rather than any escaping.

Same pattern in `backends/prm_pdf/tasks.rake:66-70`, where an unquoted path is interpolated into `system` for all three platforms:

```ruby
system("xdg-open #{pdf_path} &")
system("open #{pdf_path}")
system("start \"\" #{pdf_path}")
```

### To Reproduce

Word splitting, using the real `build_asciidoctor_command` output shape:

```ruby
require "open3"; require "pathname"
template_dir = Pathname.new("/home/me/my repos/riscv-unified-db/backends/prm_pdf")
cmd = ["echo", "-a", "pdf-themesdir=#{template_dir / 'pdf-theme'}", "-o", "/tmp/out.pdf", "main.adoc"]
Open3.popen3(cmd.join(" ")) { |i,o,e,t| i.close; puts "ARGS SEEN: #{o.read}" }
```

Metacharacter interpretation:

```ruby
require "open3"; require "pathname"
prm_name = "demo; touch /tmp/PWNED_BY_PRM;"
[...]

PR #2190, opened 2026-07-28
fix(prm): exec asciidoctor-pdf directly instead of through a shell

Hiteshsai007:
build_asciidoctor_command returns a correct argv array, but the call site flattened it with cmd.join(" ") and handed the result to Open3, which then ran it through a shell.

Any path component containing a space was word-split. The theme directory is derived from the checkout location, so `-a pdf-themesdir=#{@template_dir / 'pdf-theme'}` becomes two arguments for anyone whose clone sits under a directory with a space in its name, and asciidoctor-pdf receives a truncated theme dir plus a stray positional. Shell metacharacters in the output path were likewise interpreted rather than passed through.

Splat the array so Open3 execs the command without a shell. The two cmd.join(' ') uses in the log and error messages are unchanged, since those are for display.

Refs #2141


Hiteshsai007:

PR #2192, opened 2026-07-28
feat(tools): add parameter emitter and validator

### LINKED ISSUE #2191: Nothing in the repository emits parameter YAML; add a parameter emitter and validator

The 227 files in `spec/std/isa/param` are written by hand, and nothing in the repository emits them. `grep "kind: parameter" --include=*.py --include=*.rb tools/ backends/` returns nothing.

That is fine while parameters are added one at a time by people who already know the conventions. It becomes a problem for anything that proposes parameters in bulk — an extraction flow reading the ISA manual, a script porting a vendor list, or the parameter-extraction work tracked in #1750 and #2096. Each such tool has to rediscover the file shape by hand: the license header, the `yaml-language-server` hint, which of the nine keys are permitted, which six are required, and the `definedBy` condition grammar. Nothing checks the result until a reviewer reads it.

There is a second, quieter problem. `param_schema.json` cannot currently reject a malformed `schema` block, because its `additionalProperties: false` sits as a sibling of `$ref` and draft-07 ignores keywords beside `$ref` (#2173). A misspelled `minimun` is accepted and silently drops the bound it was meant to express. So even a generator that validates against the schema can emit a parameter that looks constrained and is not.

A small Python module under `tools/python/` that closes the gap from both ends:

- a `ParamCandidate` describing a proposed parameter, rendered to YAML matching the conventions of the existing hand-written set
- validation that runs *before* emission, so a candidate that would produce an invalid file raises rather than writing it
- an `audit` subcommand applying the same rules to files already in the database, so the checks are usable on current data rather than only on new candidates

Two checks would go beyond `param_schema.json` as it stands. Both describe intent rather than present enforcement, so both should be reported rather than assumed, and skippable with a flag:

1. **Unknown draft-07 keywords in a `schema` block**, taken from the vendored metaschema so the list cannot drift. This catches the `minimun` class of typo that #2173 makes invisible.
2. **Integer schemas with no `minimum`, `maximum`, `enum`, `const` or `$ref`.** Running this against the current database finds exactly two: `ELEN` and `PMLEN` — #2170 and #2008 respectively. Both are real, and neither was found by schema validation.

Emission would also need to reproduce the house style rather than generic YAML: literal block scalars for multi-line prose, consistent key order, LF endings, and a trailing newline.

*Leave it to each consumer.* This is the status quo. It works, but every tool reimplements the same file shape, and nothing catches the two defect classes above.

*Extend the schema instead.* Repairing the inert `additionalProperties` (#2173) would fix check 1 at the schema layer, which is better, and I have a PR open for it. It does not help with check 2, with emission, or with the audit over existing files — a schema can reject a bad file but cannot write a good one.

*Put it in the Ruby gem.* `Udb::Schema` already validates. But the parameter-extraction work is Python, `tools/python/` already exists with `pytest` wired into the Rakefile, and `ruamel.yaml` is already a declared dependency. Python keeps this close to its likely callers. Happy to be told otherwise if maintainers would rather it lived in Ruby.

**Additional context**

This is proposal point 4 of the *AI-assisted extraction of architectural parameters* mentorship — exporting extracted parameters in UDB YAML format. The extraction side is being explored in #2096, #2097 and #1750; this is the piece that turns a candidate into a file the database will accept.

I have an implementation ready and will open a PR against this issue. It includes a round-trip test over all 227 existing parameter files: each is loaded, rebuilt as a candidate, re-emitted, and compared field by field, which exercises the full variety of real `schema` and `definedBy` shapes rather than only simple ones.

Since this is new tooling rather than a fix, I would rather agree the shape before it is reviewed in detail — where it should live, whether the two advisory checks belong there at all, and whether the CLI is the right entry point. Happy to reshape or drop parts of it.

Hiteshsai007:
Nothing in the repository emits parameter YAML. The 227 files in spec/std/isa/param are written by hand, so anything proposing parameters in bulk — an extraction flow reading the ISA manual, a script porting a vendor list — has to reproduce the file shape by hand, with nothing checking the result until review.

Adds tools/python/param_emitter: a ParamCandidate that renders to house-style YAML, validation that runs before emission so an invalid candidate raises rather than writing a file, and an audit subcommand that applies the same rules to files already in the database.

Two checks go beyond param_schema.json and are reported rather than assumed, since they describe intent rather than current enforcement. First, unknown draft-07 keywords in a `schema` block: param_schema sets `additionalProperties: false` there, but as a sibling of `$ref`, which draft-07 ignores, so a misspelling such as `minimun` is silently accepted (#2173). Second, integer schemas with no bounds, enum, const or `$ref`; auditing the current database finds exactly two, ELEN and PMLEN (#2170, #2008). Pass --schema-only to apply just the schema.

Tests round-trip every file in spec/std/isa/param: each is loaded, rebuilt as a candidate, re-emitted and compared field by field, which exercises the real variety of schema and definedBy shapes. Run via ./do test:param_emitter.

Since this is new tooling rather than a fix, I would rather agree the shape than present it as settled — in particular whether tools/python is the right home, and whether the two advisory checks belong here at all. Happy to relocate or drop them.

Closes #2191

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -136,6 +136,11 @@ namespace :test do
       raise unless e.message.include?("status (5)") # don't fail on skipped tests
     end
   end
+  # "Run the parameter emitter/validator test suite"
+  task :param_emitter do
+    sh "uv run pytest tools/python/param_emitter/test_param_emitter.py -v"
+  end
+
   # "Run the IDL compiler test suite"
   task :idl_compiler do
     test_files = Dir["#{$root}/lib/idl/tests/test_*.rb"]
--- a/tools/python/param_emitter/param_emitter.py
+++ b/tools/python/param_emitter/param_emitter.py
@@ -1,0 +1,375 @@
+#!/usr/bin/env python3
+# Copyright (c) 2026 Hitesh Sai
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+"""Emit and validate UDB architectural parameter YAML.
+
+Parameters in ``spec/std/isa/param`` are currently written by hand. Any tool that
+proposes new parameters -- for example an extraction flow reading the ISA manual --
+has to reproduce the file shape by hand as well, with nothing checking the result
+beyond a reviewer's eye.
+
+This module provides that missing piece: a ``ParamCandidate`` describing a
+parameter, a validator that rejects anything ``param_schema.json`` would reject,
+and an emitter that writes a file matching the conventions of the existing
+hand-written set.
+
+Emission always validates first, so a candidate that would produce an invalid file
+raises rather than writing it.
+
+Two checks here are deliberately stricter than the schema as it stands on main:
+
+* ``schema`` keys are restricted to JSON Schema draft-07 keywords. ``param_schema``
+  sets ``additionalProperties: false`` on that subschema, but the keyword sits as a
+  sibling of ``$ref``, and draft-07 ignores siblings of ``$ref``, so the constraint
+  is currently inert and typos such as ``minimun`` pass silently (issue #2173).
+* An integer ``schema`` with neither bounds nor an enum is reported. The schema
+  permits it, but an unbounded integer parameter is almost always an oversight
+  (issues #2170, #2102).
+
+Both are advisory in the sense that they describe intent rather than current
+enforcement, and both are reported through the same mechanism as hard schema
+errors so a caller cannot silently ignore them.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import sys
+from collections.abc import Iterable, Sequence
+from dataclasses import dataclass, field
+from io import StringIO
+from pathlib import Path
+from typing import Any
+
+from ruamel.yaml import YAML
+from ruamel.yaml.scalarstring import LiteralScalarString
+
+# --------------------------------------------------------------------------- #
+# Repository layout
+# --------------------------------------------------------------------------- #
+
+# tools/python/param_emitter/param_emitter.py -> repository root
+REPO_ROOT = Path(__file__).resolve().parents[3]
+SCHEMAS_DIR = REPO_ROOT / "spec" / "schemas"
+PARAM_DIR = REPO_ROOT / "spec" / "std" / "isa" / "param"
+
+LICENSE_HEADER = (
+    "# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.\n"
+    "# SPDX-License-Identifier: BSD-3-Clause-Clear\n"
+    "\n"
+    "# yaml-language-server: $schema=../../../schemas/param_schema.json\n"
+    "\n"
+)
+
+# Key order used when emitting. Existing files are not consistent about this --
+# most lead with `name` then `description`, a few put `long_name` or `definedBy`
+# earlier -- so a single order is chosen for newly emitted files rather than
+# trying to reproduce the variation.
+KEY_ORDER = (
+    "$schema",
+    "kind",
+    "name",
+    "long_name",
+    "description",
+    "schema",
+    "definedBy",
+    "requirements",
+)
+
+# JSON Schema draft-07 keywords, read from the metaschema the repository vendors
+# so this list cannot drift from the one actually in use.
+_METASCHEMA = SCHEMAS_DIR / "json-schema-draft-07.json"
+
+
+def draft7_keywords() -> frozenset[str]:
+    """Return the set of draft-07 keywords, from the vendored metaschema."""
+    with _METASCHEMA.open(encoding="utf-8") as f:
+        return frozenset(json.load(f)["properties"].keys())
+
+
+def _block(text: str) -> Any:
+    """Render multi-line prose as a literal block scalar, matching existing files.
+
+    Most hand-written parameter descriptions use ``|``. Single-line values are left
+    as plain scalars, which is also what the existing set does.
+    """
+    if isinstance(text, str) and "\n" in text.strip():
+        return LiteralScalarString(text if text.endswith("\n") else text + "\n")
+    return text
+
+
+def _yaml() -> YAML:
+    y = YAML()
+    y.default_flow_style = False
+    y.width = 4096  # do not wrap; prettier owns line breaking for YAML
+    y.indent(mapping=2, sequence=4, offset=2)
+    y.preserve_quotes = True
+    return y
+
+
+# --------------------------------------------------------------------------- #
+# Candidate
+# --------------------------------------------------------------------------- #
+
+
+@dataclass
+class ParamCandidate:
+    """A proposed architectural parameter, prior to validation or emission.
+
+    ``name``, ``long_name``, ``description``, ``schema`` and ``defined_by`` map
+    directly onto the fields of ``param_schema.json``. ``requirements`` is
+    optional and carries either an IDL condition or a YAML condition structure.
+
+    ``source_excerpt`` is not part of the schema and is never emitted. It exists so
+    that a caller extracting parameters from specification prose can carry the
+    supporting quotation alongside the candidate for review.
+    """
+
+    name: str
+    long_name: str
+    description: str
+    schema: dict[str, Any]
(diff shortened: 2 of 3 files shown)


titoatwork:
Read this closely because it fills a gap I hit from the other direction. I built frozen fixtures for parameter extraction in #2164 and had no way to check that a candidate would survive the database. An emitter that validates before writing is the piece that was missing, and deriving `draft7_keywords()` from the vendored metaschema rather than hardcoding a list is the right call.

Both claims in your description hold. I checked them rather than take them:

- Integer schemas with no bounds, enum, const or `$ref`: exactly `ELEN` and `PMLEN` across all 227 files.
- `additionalProperties` as a sibling of `$ref` under draft-07: reproduced, a validator accepts an unknown key in that position. So `minimun` really is swallowed today, and #2173 is a live hole rather than a theoretical one.

One substantive gap, and it sits in the check you added this for. `validate_doc` inspects only the top-level `schema` dict:

```python
subschema = doc.get("schema")
if isinstance(subschema, dict):
    unknown = sorted(set(subschema) - draft7_keywords())
```

Fifteen of the 227 parameters carry a nested schema dict under `items` or `additionalItems`:

```
HPM_EVENTS                schema.items -> {type, minimum, maximum}
MSTATUS_FS_LEGAL_VALUES   schema.items -> {type, enum}
HPM_COUNTER_EN            schema.additionalItems -> {type}
COUNTINHIBIT_EN           schema.additionalItems -> {type}
MCOUNTENABLE_EN           schema.items -> {type}
HCOUNTENABLE_EN           schema.items -> {type}
```

`HPM_EVENTS.schema.items` is a bounded integer, so it is exactly where a `minimun` typo would land, and the check would not see it. Walking the subschema recursively would close it. The round-trip tests pass either way, so nothing currently in the database would catch that regression for you.

Two smaller notes, both latent rather than active.

The unconstrained-integer check tests `{"enum", "const", "minimum", "maximum", "$ref"}`, which omits `exclusiveMinimum` and `exclusiveMaximum`. Both are in the metaschema's `properties`, so they pass the unknown-keyword check while still being reported as unconstrained. No parameter uses them today, so it is only a question of which way you want the two checks to disagree.

That same check keys on a top-level `type == "integer"`, so an array of unbounded integers under `items` would not be reported. Currently zero of those exist, so it is a gap in reach rather than a missed finding.

On placement, from someone with the same problem rather than an opinion about yours. I put fixtures under `tools/python/param-extraction-eval` in #2164 and raised the convention question separately as #2158, which is still unanswered. Your `tools/python/param_emitter` sits beside it consistently, so if a maintainer decides that tree is wrong, both move together and one answer covers both. I would not hold this for it.

The two advisory checks read to me as worth keeping. Reporting intent the schema cannot yet enforce, with the issue number attached, is more useful than silence, and you were explicit that they are advisory.

Incidentally, the two red checks are one ruff error, `EXE001` on the shebang in `param_emitter.py`, with `regress-complete` failing after it as a dependent job. Mentioning it only because the rollup makes it look like two separate problems.

Hiteshsai007:
> Read this closely because it fills a gap I hit from the other direction. I built frozen fixtures for parameter extraction in #2164 and had no way to check that a candidate would survive the database. An emitter that validates before writing is the piece that was missing, and deriving `draft7_keywords()` from the vendored metaschema rather than hardcoding a list is the right call.
> 
> Both claims in your description hold. I checked them rather than take them:
> 
> * Integer schemas with no bounds, enum, const or `$ref`: exactly `ELEN` and `PMLEN` across all 227 files.
> * `additionalProperties` as a sibling of `$ref` under draft-07: reproduced, a validator accepts an unknown key in that position. So `minimun` really is swallowed today, and [param_schema.json: additionalProperties is ignored due to draft-07 $ref semantics #2173](https://github.com/riscv/riscv-unified-db/issues/2173) is a live hole rather than a theoretical one.
> 
> One substantive gap, and it sits in the check you added this for. `validate_doc` inspects only the top-level `schema` dict:
> 
> ```python
> subschema = doc.get("schema")
> if isinstance(subschema, dict):
>     unknown = sorted(set(subschema) - draft7_keywords())
> ```
> 
> Fifteen of the 227 parameters carry a nested schema dict under `items` or `additionalItems`:
> 
> ```
> HPM_EVENTS                schema.items -> {type, minimum, maximum}
> MSTATUS_FS_LEGAL_VALUES   schema.items -> {type, enum}
> HPM_COUNTER_EN            schema.additionalItems -> {type}
> COUNTINHIBIT_EN           schema.additionalItems -> {type}
> MCOUNTENABLE_EN           schema.items -> {type}
> HCOUNTENABLE_EN           schema.items -> {type}
> ```
> 
> `HPM_EVENTS.schema.items` is a bounded integer, so it is exactly where a `minimun` typo would land, and the check would not see it. Walking the subschema recursively would close it. The round-trip tests pass either way, so nothing currently in the database would catch that regression for you.
> 
> Two smaller notes, both latent rather than active.
> 
> The unconstrained-integer check tests `{"enum", "const", "minimum", "maximum", "$ref"}`, which omits `exclusiveMinimum` and `exclusiveMaximum`. Both are in the metaschema's `properties`, so they pass the unknown-keyword check while still being reported as unconstrained. No parameter uses them today, so it is only a question of which way you want the two checks to disagree.
> 
> That same check keys on a top-level `type == "integer"`, so an array of unbounded integers under `items` would not be reported. Currently zero of those exist, so it is a gap in reach rather than a missed finding.
> 
> On placement, from someone with the same problem rather than an opinion about yours. I put fixtures under `tools/python/param-extraction-eval` in #2164 and raised the convention question separately as #2158, which is still unanswered. Your `tools/python/param_emitter` sits beside it consistently, so if a maintainer decides that tree is wrong, both move together and one answer covers both. I would not hold this for it.
> 
> The two advisory checks read to me as worth keeping. Reporting intent the schema cannot yet enforce, with the issue number attached, is more useful than silence, and you were explicit that they are advisory.
> 
> Incidentally, the two red checks are one ruff error, `EXE001` on the shebang in `param_emitter.py`, with `regress-complete` failing after it as a dependent job. Mentioning it only because the rollup makes it look like two separate problems.


You checked my claims rather than taking them, so I did the same with yours. All three hold.

The nested gap is real, and it is worse than "would not see it" — I reproduced it directly. Injecting `minimun` into `HPM_EVENTS.schema.items` and running `validate_doc` returns `[]`. Exactly 15 of the 227 files carry a nested schema dict, matching your count:

```
COUNTINHIBIT_EN, HCOUNTENABLE_EN, HPM_COUNTER_EN, HPM_EVENTS, MCOUNTENABLE_EN,
MSTATUS_FS_LEGAL_VALUES, MSTATUS_VS_LEGAL_VALUES, MTVEC_MODES, SCOUNTENABLE_EN,
STVEC_MODES, SXLEN, UXLEN, VSTVEC_MODES, VSXLEN, VUXLEN
```

Fixed by extracting the checks into a `_check_subschema` helper that recurses through `items`, `additionalItems`, `contains`, `propertyNames`, `not`, and the `allOf`/`anyOf`/`oneOf` lists. Problems now carry the path, so the `HPM_EVENTS` case reports:

```
schema.items: unknown draft-07 keyword(s) ['minimun']
```

The `exclusiveMinimum`/`exclusiveMaximum` omission is fixed too — both are now in the bounding set, so the two checks agree. You were right that nothing uses them today; it was only a question of which way they should disagree, and agreeing seemed better than choosing.

Your third point falls out of the same change: an unbounded integer under `items` is now reported, since the recursion re-applies the `type == "integer"` test at each level rather than only at the root.

Added four tests: nested unknown keyword under `items`, under `additionalItems`, unbounded integer under `items`, and a parametrised check that the exclusive bounds count as constraining. 31 pass. The audit still reports exactly `ELEN` and `PMLEN`, so the recursion adds reach without adding false positives on current data.

On placement — that reasoning is more useful than an opinion would have been, and I agree the two should move together if #2158 lands somewhere else.

And thanks for the `EXE001` diagnosis. That was the whole failure: the shebang was there but the file was not mode 755 in the index, which is fixed now.

Hiteshsai007:

PR #2196, opened 2026-07-28
fix: refactored Trusim and unit tests to not be depenedent on fmtlib

henrikg-qc:
Removed fmtlib dependency


jordancarlin (changes requested):
Looks pretty good. Can you remove the custom regex matcher for fmtlib in `renovate.json`?

jordancarlin:
Here is an AI generated review. Not sure how valid all of these points are, but some of them seem worthwhile:

> Reviewed PR #2196 (fix: refactored Trusim and unit tests to not be depenedent on fmtlib).
> 
>  Blocking issues
> 
>  - backends/cpp_hart_gen/cpp/include/udb/htif.hpp:64: what() returns std::format(...).c_str() from a temporary std::string, so the returned pointer is dangling immediately. Store the formatted string in the exception object (preferred) or otherwise return storage with a stable lifetime.
> 
>  - backends/cpp_hart_gen/templates/params.hxx.erb:61-64: replacing fmt/ranges.h with std::format likely breaks configs with array/vector parameters under the project’s GCC 14 toolchain. rv64-vector has array params like MTVEC_MODES, SXLEN, STVEC_MODES, UXLEN, etc., and regress-riscv-tests-vector is currently failing. Either keep fmt for range formatting here, add a project-local stringify helper for param values, or gate use of std::format ranges on toolchain support. Reference: libstdc++ range formatter support appears to have landed in GCC 15-era work: https://gcc.gnu.org/pipermail/gcc-cvs/2025-April/421435.html
> 
>  - backends/cpp_hart_gen/templates/hart_impl.hxx.erb:33-35: rf_width >= 64 sends normal RV64 registers through the 128-bit split path, so a 64-bit value like 0x123 prints as 0x00x123. Wider values are also not padded, so the concatenation is ambiguous/incorrect. Handle exactly-64-bit registers with the normal single-value path, and if wider register files are needed, print one properly formatted value or zero-pad high/low chunks.
> 
>  Nits / non-blocking
> 
>  - backends/cpp_hart_gen/cpp/include/udb/defines.hpp:37: assertion failures now print to stdout via std::cout; previous behavior used stderr. Prefer std::cerr or std::print(stderr, ...).
> 
>  - backends/cpp_hart_gen/lib/template_helpers.rb:19: comment still says fmt::format; update to std::format.

henrikg-qc:
The port to format from fmtlib in its current implementation requires gcc 15.  Do we, or can, we require a certain version or as new version as gcc 15?

PR #2197, opened 2026-07-28
fix(csr): add MISA guard to mstatus.FS reset_value method

krrishverma1805-web:
Add MISA_CSR_IMPLEMENTED read-only zero guard to mstatus.FS reset_value() method.

Issue #1057 cleanup on mstatus.VS revealed that mstatus.FS reset_value() lacked an explicit MISA_CSR_IMPLEMENTED read-only zero check (when CSR[misa].S == 0 && CSR[misa].F == 0). While sw_write() already included this guard, reset_value() returned UNDEFINED_LEGAL when MSTATUS_FS_LEGAL_VALUES size was greater than 1 instead of returning 0.

This change aligns reset_value() with sw_write() in mstatus.FS by returning 0 when MISA_CSR_IMPLEMENTED is true and both misa.S and misa.F are 0.

### DIFF
--- a/spec/std/isa/csr/mstatus.yaml
+++ b/spec/std/isa/csr/mstatus.yaml
@@ -436,6 +436,10 @@ fields:
           - name: F
           - name: S
     reset_value(): |
+      if (MISA_CSR_IMPLEMENTED && (CSR[misa].S == 1'b0) && (CSR[misa].F == 1'b0)) {
+        # must be read-only-0
+        return 0;
+      }
       return $array_size(MSTATUS_FS_LEGAL_VALUES) == 1 ? MSTATUS_FS_LEGAL_VALUES[0] : UNDEFINED_LEGAL;
     sw_write(csr_value): |
       if (MISA_CSR_IMPLEMENTED && (CSR[misa].S == 1'b0) && (CSR[misa].F == 1'b0)) {


spec/std/isa/csr/mstatus.yaml
```diff
@@ -436,9 +436,13 @@ fields:
           - name: F
           - name: S
     reset_value(): |
+      if ((!implemented?(ExtensionName::S) && !implemented?(ExtensionName::F)) || (MISA_CSR_IMPLEMENTED && (CSR[misa].S == 1'b0) && (CSR[misa].F == 1'b0))) {
```
ThinkOpenly: nit: make this LONG line shorter
```suggestion
      if ((!implemented?(ExtensionName::S) && !implemented?(ExtensionName::F))
          || (MISA_CSR_IMPLEMENTED
              && (CSR[misa].S == 1'b0) && (CSR[misa].F == 1'b0))) {
```

spec/std/isa/csr/mstatus.yaml
```diff
@@ -436,9 +436,13 @@ fields:
...
+        # must be read-only-0
+        return 0;
+      }
       return $array_size(MSTATUS_FS_LEGAL_VALUES) == 1 ? MSTATUS_FS_LEGAL_VALUES[0] : UNDEFINED_LEGAL;
     sw_write(csr_value): |
-      if (MISA_CSR_IMPLEMENTED && (CSR[misa].S == 1'b0) && (CSR[misa].F == 1'b0)) {
+      if ((!implemented?(ExtensionName::S) && !implemented?(ExtensionName::F)) || (MISA_CSR_IMPLEMENTED && (CSR[misa].S == 1'b0) && (CSR[misa].F == 1'b0))) {
```
ThinkOpenly: here, too
```suggestion
      if ((!implemented?(ExtensionName::S) && !implemented?(ExtensionName::F))
          || (MISA_CSR_IMPLEMENTED
              && (CSR[misa].S == 1'b0) && (CSR[misa].F == 1'b0))) {
```

ThinkOpenly (changes requested):
a couple of nits, but good otherwise

titoatwork:
This one is on me. I raised the FS gap on #2109 and I now think the guard cannot fire.

`mstatus.FS` carries `definedBy: {extension: {anyOf: [{name: F}, {name: S}]}}`, so the field exists only when F or S is implemented. The condition being added here, `!implemented?(ExtensionName::S) && !implemented?(ExtensionName::F)`, is exactly the complement of that gate.

The two meet in `exists_in_cfg?` in `tools/ruby-gems/udb/lib/udb/obj/csr_field.rb`, which for a fully configured arch requires `defined_by_condition.satisfied_by_cfg_arch?(cfg_arch) == SatisfiedResult::Yes`. The partially configured branch is looser, but `implemented?` cannot be reached there: `tools/ruby-gems/idlc/lib/idlc/ast.rb` raises `implemented? is only known when evaluating in the context of a fully-configured arch def` rather than returning false. So wherever `FS.reset_value()` or `FS.sw_write()` evaluates this condition, the field exists, which means F or S is implemented, which means the new disjunct is false.

The reason the same change was right for VS is that those two conditions are not complementary. `mstatus.VS` is gated on `anyOf[Zvl32b, S]` while its guard tests `!implemented?(S) && !implemented?(V)`, and neither `V` nor `Zvl32b` implies the other. A hart with `Zvl32b` but neither V nor S has the field present and the guard true, so #2109 does real work. FS has no equivalent gap because its gate names the same two extensions its guard tests.

What survives is a spec asymmetry rather than a code one. `norm:mstatus_fs_acc2` requires FS to be read-only zero when neither F nor S is implemented, and UnifiedDB satisfies that by the field not existing. `norm:mstatus_vs_acc2` says the same for VS and the `v` registers, but the gate is written in terms of `Zvl32b`, which is why the explicit guard is needed there and not here.

I should have checked the `definedBy` before raising it on your PR rather than after, and I am sorry for the wasted work. If you or @ThinkOpenly can point at a configuration where the field exists and the new disjunct is true, I am glad to be wrong twice.

ThinkOpenly:
You're correct. Sadly, I looked at exactly that and concluded differently earlier. But, we indeed will never get into the function if both S and F are compile-time unimplemented extensions. The `misa` checks need to stay, but the `(!S and !F)` checks can be removed. Good catch!

krrishverma1805-web:
@ThinkOpenly @titoatwork Thank you both for the sharp analysis!

@titoatwork is spot on: because `mstatus.FS` is gated by `definedBy: {extension: {anyOf: [{name: F}, {name: S}]}}`, whenever the `FS` field exists in a hart configuration, at least one of `F` or `S` is compile-time implemented. Thus, `!implemented?(ExtensionName::S) && !implemented?(ExtensionName::F)` is indeed redundant and unreachable.

I have updated the PR to remove the redundant `(!S and !F)` compile-time checks while retaining the `MISA_CSR_IMPLEMENTED && (CSR[misa].S == 1'b0) && (CSR[misa].F == 1'b0)` read-only-zero check in `reset_value()`. All schema, Sorbet, and IDL tests pass cleanly!

ThinkOpenly:
Great! Now, please update the PR description to match the PR content. Remove the "Testing" section. Also, remove the Markdown headings. Commit messages should be concise, state the issue (including "why" if necessary) and fix (if non-obvious).

Yesterday, this repository enabled mandatory sign-offs for commits, so you are encouraged to (and may need to) append a "Signed-off-by" line in your commit (and force-push it here).

krrishverma1805-web:
Updated the PR description as requested: removed all Markdown headings and the Testing section, updating the text to concisely describe the problem and fix. The commit is signed off with DCO as well. PTAL

ThinkOpenly (approved):

PR #2204, opened 2026-07-29
Update Ruff to 0.16.0

jordancarlin:
Update Ruff and the Ruff pre-commit hooks to 0.16.0. Ruff 0.16 dramatically expands the default lint rule set, so remove the explicit `select` block and rely on the new defaults.

Keep a small `extend-select` list for higher-value rules that are not part of the default set, remove the remaining lint ignores, and fix the resulting lint issues.


ThinkOpenly:
I presume all of the `generators` content are fixes due to the upgrade?
This is one of those cases where 2 commits would make the review a bit easier. ;-)

ThinkOpenly:
...`generators` and `tools`

jordancarlin:
> I presume all of the `generators` content are fixes due to the upgrade? This is one of those cases where 2 commits would make the review a bit easier. ;-)

It was two commits, but then I had to rebase them and accidentally squashed them. The non-lint fix changes are fairly minimal (pyproject.toml, uv.lock, .pre-commit-config.yaml), so hopefully it isn't too bad to review. Let me know if you want me to split it back apart though.

ThinkOpenly:
> Let me know if you want me to split it back apart though.

No. Desquashing is non-trivial and for no value here, given the directory separation.

ThinkOpenly (approved):

PR #2207, opened 2026-07-29
Simplify package manager setup

jordancarlin:
Remove the custom shared install handling for uv and npm and let the tools use their default workspace locations again. `bin/setup` now runs `uv sync` and `npm ci` directly. A temporary migration path for existing checkouts is added to replace the old shared `.venv` or `node_modules` symlinks.

The shared-directory logic added complexity for little benefit, made setup behavior harder to understand, and makes it harder to update tooling in the future. The original motivation for this setup was so that multiple worktrees could share the same `venv` and `node_modules`, but `uv` already symlinks everything from a shared cache, so there is almost no speed or disk advantage to shared the `venv` itself. Using the default uv/npm behavior should make setup simpler and easier to maintain.


bin/doctor
```diff
@@ -236,33 +229,22 @@ _check_python_ecosystem() {
...
     _fail "node_modules not found"
-    _info "Fix: run \`bin/setup\`"
+    _info "Fix: run \`bin/setup\` or \`bin/npm ci\`"
     _record_fail
   fi
 }
 
 # ---------------------------------------------------------------------------
-# Section 5: Developer Setup
+# Section 6: Developer Setup
```
ThinkOpenly: nit: should we even keep the section _numbers_?

bin/setup
```diff
@@ -31,41 +30,6 @@ _info()    { printf "    ${_DIM}%s${_RESET}\n" "$1"; }
...
-    printf "    ${_CYAN}export MISE_DATA_DIR=\"...\"${_RESET}   # mise tools only\n"
-    printf "    ${_CYAN}export BUNDLE_PATH=\"...\"${_RESET}     # Ruby gems only\n\n"
-    printf "    After updating your profile, open a new shell (or source it), then re-run:\n\n"
-    printf "        ${_BOLD}bin/setup${_RESET}\n\n"
-    exit 0
-  fi
-fi
-
 # ---------------------------------------------------------------------------
 # 1. Bootstrap mise via bin/mise
```
ThinkOpenly: More section numbers. Heh.

ThinkOpenly (approved):

bin/doctor
jordancarlin: Removed

bin/setup
jordancarlin: Removed

PR #2213, opened 2026-07-29
feat: add Zibi extension

wangpc-pp:
Add the Zibi (Branch with Immediate) extension, which provides conditional branch instructions that compare an integer register with a small immediate constant instead of a second register.

The extension adds two B-type instructions, beqi and bnei (funct3 010/011), which reuse the rs2 encoding slot for a 5-bit cimm field. The cimm field is not sign-extended: cimm=0 decodes to the value -1, and cimm values 1 through 31 decode to the corresponding positive integers.

Spec: https://github.com/riscv/zibi/blob/main/src/unpriv/zibi.adoc

Signed-off-by: Pengcheng Wang <wangpengcheng.pp@bytedance.com>


spec/std/isa/ext/Zibi.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+  The `beqi` and `bnei` instructions use the B-type instruction format. The
+  instruction bits that would otherwise carry _rs2_ in a B-type branch carry a
+  5-bit _cimm_ field. All _cimm_ encodings are valid: the _cimm_ field is not
+  sign-extended; it decodes to an XLEN-bit comparison constant where `cimm=0`
+  encodes the value -1, and `cimm` values 1 through 31 encode the corresponding
+  positive integer values 1 through 31.
+type: unprivileged
+versions:
+  - version: "0.6.0"
```
ThinkOpenly: I believe the latest version is 0.7 (2026-03-07), and this implementation is compatible with that version. Please confirm.
```suggestion
  - version: "0.7.0"
```

ThinkOpenly (changes requested):
Nice work, @wangpc-pp !

We enabled mandatory "sign-offs" for this repository yesterday, so please amend your commit with a "Signed-off-by" line. Thanks!

One additional comment inline below...

spec/std/isa/ext/Zibi.yaml
wangpc-pp: Thanks! It it frozen so the version should be 0.7. :-)

spec/std/isa/inst/Zibi/beqi.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - name: xs1
+      location: 19-15
+access:
+  s: always
+  u: always
+  vs: always
+  vu: always
+operation(): |
+  XReg lhs = X[xs1];
+  XReg rhs = (cimm == 0) ? {MXLEN{1'b1}} : {{MXLEN-5{1'b0}}, cimm};
```
jordancarlin: Should this use `xlen()` instead of `MXLEN` so that it gets the dynamic XLEN length if it is different in different priv modes?

jordancarlin (changes requested):

spec/std/isa/inst/Zibi/beqi.yaml
wangpc-pp: Done in https://github.com/riscv/riscv-unified-db/pull/2213/commits/5bcff572ab0ea226f618f5bbd30606f1084fe579! I know little about the IDL, please double check and many thanks for that!

ThinkOpenly: I had the same question. I see both uses in other places. I originally figured `MXLEN` bits are more than enough to cover either case, plus the assignment to `rhs` would be truncated in the questionable case, and it all works out.

It is probably preferred (and more straightforward) to limit everything to `xlen()`, as you suggest. In addition, the latter ternary expression can be greatly simplified.
```suggestion
  XReg rhs = (cimm == 0) ? {xlen(){1'b1}} : cimm;
```
and I think it can be further simplified:
```suggestion
  XReg rhs = (cimm == 0) ? -1'sd : cimm;
```
...certainly more readable, but I'd like to see that tested. (Maybe @henrikg-qc knows off the top of his head?)

ThinkOpenly: AI and I chatted. We agreed that the best solution is:
```
(cimm == 0) ? {MXLEN{1'b1}} : cimm
```
This does rely on truncation, but it told me that the `{xlen(){value}}` method is not valid because IDL replication requires a compile-time constant.

wangpc-pp: Yes, I asked AI and it said yes. Done in https://github.com/riscv/riscv-unified-db/pull/2213/commits/529b3d91202416c7eb6593b60acbbdbbe0a31314.

ThinkOpenly (approved):

wangpc-pp:
It seems this was not merged because of some blocking reviews? @jordancarlin Hi Jordan! Can you approve this PR?

jordancarlin (approved):

PR #2215, opened 2026-07-29
fix(param): typo: SV32_VSMODE_TRANSLATION should require SV32, not SV39

### LINKED ISSUE #2214: SV32_VSMODE_TRANSLATION requirement concludes SV39_VSMODE_TRANSLATION, rejecting a legal configuration
## Describe the bug

`spec/std/isa/param/SV32_VSMODE_TRANSLATION.yaml` states a requirement whose conclusion names a different parameter than the one the file, its comment and its `reason` all describe:

```yaml
requirements:
  idl(): |
    # at least one of Bare, Sv32 must be supported
    !VSSTAGE_MODE_BARE
      -> SV39_VSMODE_TRANSLATION;
  reason: at least one of Bare, Sv32 must be supported
```

The comment says Sv32, the `reason` says Sv32, and the parameter is `SV32_VSMODE_TRANSLATION`, but the conclusion is `SV39_VSMODE_TRANSLATION`. The two are gated on different XLENs: this file is `VSXLEN includes 32`, while `SV39_VSMODE_TRANSLATION` is `VSXLEN includes 64`.

Every sibling concludes its own parameter:

```
SV39:  !BARE && !SV48 && !SV57  ->  SV39_VSMODE_TRANSLATION
SV48:  !BARE && !SV39 && !SV57  ->  SV48_VSMODE_TRANSLATION
SV57:  !BARE && !SV39 && !SV48  ->  SV57_VSMODE_TRANSLATION
SV32:  !BARE                    ->  SV39_VSMODE_TRANSLATION
```

So does the equivalent in the G-stage family, which is the same construct with the same comment shape and the same 32-bit gate:

```yaml
# SV32X4_TRANSLATION.yaml
# at least one of Bare, Sv32x4 must be supported
!GSTAGE_MODE_BARE
  -> SV32X4_TRANSLATION;
```

## Effect

This is not a missing constraint. `VSSTAGE_MODE_BARE` already enforces "at least one mode per XLEN" completely, from both directions, so nothing is unguarded:

```yaml
$array_includes?(VSXLEN, 32) && !SV32_VSMODE_TRANSLATION
  -> VSSTAGE_MODE_BARE;
$array_includes?(VSXLEN, 64) && !SV39_VSMODE_TRANSLATION && !SV48_VSMODE_TRANSLATION && !SV57_VSMODE_TRANSLATION
  -> VSSTAGE_MODE_BARE;
```

The effect is the opposite: the clause over-constrains and rejects a legal configuration. Take a hypervisor hart where 32-bit guests use Sv32 and 64-bit guests use Sv48:

```
VSXLEN [32, 64], VSSTAGE_MODE_BARE false, SV32 true, SV48 true, SV39 false, SV57 false
```

Evaluating every relevant rule against it:

| rule | antecedent | result |
|---|---|---|
| `VSSTAGE_MODE_BARE` 32-bit clause | false | vacuous |
| `VSSTAGE_MODE_BARE` 64-bit clause | false | vacuous |
| `SV39` | false | vacuous |
| `SV48` | true | satisfied |
| `SV57` | false | vacuous |
| `SV32` as written, `!BARE -> SV39` | true | **violated** |
| `SV32` with `-> SV32`, for comparison | true | satisfied |

Only the mistyped clause rejects it, by demanding Sv39 support from a hart that has no need of it.

Separately, in a configuration where `VSXLEN` is `[32]` alone, which the schema permits and `cfgs/example_rv64_with_overlay.yaml` documents as "32: VSXLEN is always 32", `SV39_VSMODE_TRANSLATION` does not exist at all, so the clause references an undefined parameter.

## Expected

The conclusion should name the parameter the file describes:

```yaml
!VSSTAGE_MODE_BARE
  -> SV32_VSMODE_TRANSLATION;
```

Introduced in #891. Happy to open the PR.

titoatwork:
`SV32_VSMODE_TRANSLATION`'s requirement reads "at least one of Bare, Sv32 must be supported" in both its comment and its `reason`, but concludes `SV39_VSMODE_TRANSLATION`. The two are gated on different XLENs: this parameter on `VSXLEN` including 32, that one on 64. Every sibling concludes its own parameter, as does the same construct in the G-stage family, `SV32X4_TRANSLATION`.

The constraint itself is not missing. `VSSTAGE_MODE_BARE` already enforces "at least one mode per XLEN" from both directions, so the effect is over-constraint rather than a hole: a hart where 32-bit guests use Sv32 and 64-bit guests use Sv48 is currently rejected, because the clause demands Sv39 support it has no need of. Every other rule is vacuous or satisfied for that configuration.

Conclude the parameter the file describes.

Closes #2214.

### DIFF
--- a/spec/std/isa/param/SV32_VSMODE_TRANSLATION.yaml
+++ b/spec/std/isa/param/SV32_VSMODE_TRANSLATION.yaml
@@ -16,7 +16,7 @@ requirements:
   idl(): |
     # at least one of Bare, Sv32 must be supported
     !VSSTAGE_MODE_BARE
-      -> SV39_VSMODE_TRANSLATION;
+      -> SV32_VSMODE_TRANSLATION;
   reason: at least one of Bare, Sv32 must be supported
 definedBy:
   allOf:


titoatwork:
Checked this for regressions by enumerating the configuration space rather than reasoning about it, since no configuration in `cfgs/` exercises the clause and CI therefore cannot show the property either way.

In the regime where all four parameters exist, `VSXLEN` including both 32 and 64, there are 32 combinations of `VSSTAGE_MODE_BARE`, `SV32`, `SV39`, `SV48` and `SV57`. Evaluating every rule in the family against each, before and after the change:

| | configurations |
|---|---:|
| newly accepted | 3 |
| newly rejected | 0 |

The three newly accepted are `BARE=false, SV32=true, SV39=false` with Sv48 and/or Sv57 supported, that is a hart whose 32-bit guests use Sv32 and whose 64-bit guests use Sv48 or Sv57. The change is a strict relaxation: nothing that validated before stops validating.

Separately, where `VSXLEN` is `[32]` alone, `SV39_VSMODE_TRANSLATION` does not exist, so the old clause referenced an undefined parameter and the new one does not.

ThinkOpenly (approved):

PR #2225, opened 2026-07-29
Migrate Node tooling to aube

jordancarlin:
Switch the Node package flow from `npm` to `aube`. `aube` is a `node` equivalent of `uv`, so it automatically updates local dependencies when they are outdated, manages Node itself, uses a single on-disk cache for parallel installs, is significantly faster, and has much better security defaults.

Prepared with AI assistance.


jordancarlin:
Follow up to #2207. Once this is in place, a further PR can drop the manual logic that installs Python and Node packages whenever you pull/switch branches because `uv` and `aube` both automatically update dependencies when invoked if the `package.json`/`pyproject.toml` have changed.

tests/golden/all_instructions.golden.adoc
```diff
@@ -1,6 +1,6 @@
 = Instruction Appendix
 :doctype: book
-:wavedrom: /Users/rajveerbishnoi/riscv-unified-db/node_modules/.bin/wavedrom-cli
+:wavedrom: /home/jcarlin/riscv-projects/riscv-unified-db/bin/wavedrom
```
ThinkOpenly, line 3: You don't need to include this, but it also is innocuous.
(The many updates to this line are annoying, frankly. I wonder if there's a way to turn them off.)

ThinkOpenly (approved):

tests/golden/all_instructions.golden.adoc
jordancarlin: Yeah. It's been on my wishlist for a while to add some kind of sanitizer that normalizes this line when we generate the golden appendix.

PR #2227, opened 2026-07-29
fix(csr): set priv_mode VS on vstval and vstvec

### LINKED ISSUE #2226: vstval and vstvec carry priv_mode: S, gating field widths on mstatus.SXL instead of hstatus.VSXL
## Describe the bug

`vstval` and `vstvec` carry `priv_mode: S`. Every other CSR in the VS block carries `priv_mode: VS`:

```
0x200 vsstatus   VS      0x250 vsiselect  VS
0x241 vsepc      VS      0x251 vsireg     VS
0x242 vscause    VS      0x252 vsireg2    VS   (and vsireg3-6)
0x24e vsctrctl   VS      0x280 vsatp      VS

0x205 vstvec     S    <--
0x243 vstval     S    <--
```

Twelve of the fourteen use `VS`. The specification is explicit for both of the exceptions:

> The `vstvec` register is a VSXLEN-bit read/write register […], `hypervisor.adoc` line 1304, `norm:vstvec_sz_acc_op`

> The `vstval` register is a VSXLEN-bit read/write register […], `hypervisor.adoc` line 1368, `norm:vstval_sz_acc_op`

`vstval` also contradicts itself, declaring `length: VSXLEN` alongside `priv_mode: S`.

## Effect

`priv_mode` selects which XLEN control gates a field's 32-bit and 64-bit locations, in `CsrField#location_cond32` and `#location_cond64`:

```ruby
when "S"   then "CSR[mstatus].SXL == 0"
when "VS"  then "CSR[hstatus].VSXL == 0"
```

`vstval`'s `VALUE` field has both `location_rv32` and `location_rv64`, so the branch is reached. Its width is therefore gated on `mstatus.SXL` rather than `hstatus.VSXL`. Those differ on any hart where SXLEN is 64 and VSXLEN is 32, which is the case `hstatus.VSXL` exists to express, and which `VSXLEN` is shaped to allow with `minItems: 1, maxItems: 2`.

`vscause` at 0x242 is the same kind of register with the same split locations and correctly resolves to `hstatus.VSXL`.

`modes_with_access` returns `["M", "S", "VS"]` for both values, so access checking is unaffected. The difference is confined to the width conditions.

## Origin

Both values date from `923badf6` "Change CSR schema", which set them inconsistently in a single commit:

```
vscause -> +priv_mode: VS
vsepc   -> +priv_mode: VS
vstval  -> +priv_mode: S
vstvec  -> +priv_mode: S
```

Neither file's `priv_mode` has been revisited since.

The distinction is applied deliberately elsewhere. #617 adds nine H-extension CSRs and splits them by kind: `hgeie`, `hgeip`, `hideleg`, `hie`, `hip` and `hvip` get `S`, while `vsie` and `vsip` get `VS`.

## Expected

`priv_mode: VS` on both, consistent with the other twelve and with the specification's VSXLEN wording.

Separately and not proposed here: `vstvec` declares `length: 64` where the specification says VSXLEN. `vsepc` does the same, so either both are wrong or `64` is an intentional shorthand in this family. I have not established which and have left it out.

Happy to open the PR.

titoatwork:
`vstval` and `vstvec` are marked `priv_mode: S`, but should be `VS` like all of the other `vs` CSRs.

Closes #2226.

### DIFF
--- a/spec/std/isa/csr/vstval.yaml
+++ b/spec/std/isa/csr/vstval.yaml
@@ -11,7 +11,7 @@ address: 0x243
 writable: true
 virtual_address: 0x143
 description: Holds trap-specific information
-priv_mode: S
+priv_mode: VS
 length: VSXLEN
 definedBy:
   extension:
--- a/spec/std/isa/csr/vstvec.yaml
+++ b/spec/std/isa/csr/vstvec.yaml
@@ -10,7 +10,7 @@ long_name: Supervisor Trap Vector
 address: 0x205
 writable: true
 virtual_address: 0x105
-priv_mode: S
+priv_mode: VS
 length: 64
 description: Controls where traps jump.
 definedBy:


ThinkOpenly:
Closes #2226.

`vstval` and `vstvec` carried `priv_mode: S` while the other twelve CSRs in the VS block carry `VS`, and the specification calls each of them a VSXLEN-bit register. `vstval` also declared `length: VSXLEN` alongside `priv_mode: S`.

`priv_mode` selects which XLEN control gates a field's 32-bit and 64-bit locations, in `CsrField#location_cond32` and `#location_cond64`. `vstval`'s `VALUE` field has both `location_rv32` and `location_rv64`, so its width was gated on `mstatus.SXL` rather than `hstatus.VSXL`. Those differ whenever SXLEN is 64 and VSXLEN is 32. `vscause` is the same kind of register with the same split locations and already resolves correctly.

Both values date from `923badf6`, which set `vscause` and `vsepc` to `VS` and these two to `S` in the same commit.

ThinkOpenly (approved):

PR #2230, opened 2026-07-30
Add Shlcofideleg extension data

jordancarlin:
Add the ratified Shlcofideleg extension record.

AI-generated.

### DIFF
--- a/spec/std/isa/ext/Shlcofideleg.yaml
+++ b/spec/std/isa/ext/Shlcofideleg.yaml
@@ -1,0 +1,27 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../schemas/ext_schema.json
+
+$schema: "ext_schema.json#"
+kind: extension
+name: Shlcofideleg
+long_name: Hypervisor local counter overflow interrupt delegation
+description: |
+  Shlcofideleg supports delegating Local Counter Overflow Interrupts
+  (LCOFIs) to VS-mode.
+
+  When Shlcofideleg is implemented, `hideleg` bit 13 is writable. When
+  `hideleg`.LCOFI is set, `vsip`.LCOFIP and `vsie`.LCOFIE are aliases of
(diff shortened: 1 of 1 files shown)


spec/std/isa/ext/Shlcofideleg.yaml
```diff
@@ -0,0 +1,29 @@
...
+  When Shlcofideleg is implemented, `hideleg` bit 13 is writable. When
+  `hideleg`.LCOFI is set, `vsip`.LCOFIP and `vsie`.LCOFIE are aliases of
+  `sip`.LCOFIP and `sie`.LCOFIE, respectively. Otherwise, `vsip`.LCOFIP and
+  `vsie`.LCOFIE are read-only zeros.
+type: privileged
+versions:
+  - version: "1.0.0"
+    state: ratified
+    ratification_date: 2024-10
+requirements:
```
jordancarlin, line 20: Every other Sh* extension file that requires H (Shvstvala.yaml, Shtvala.yaml, Shcounterenw.yaml, Shgatpa.yaml) expresses that dependency as an unconstrained `extension: { name: H }` (or via `idl()`/`implemented?(ExtensionName::H)`), not a pinned `version: "= 1.0.0"`. Since H currently has only one ratified version (1.0.0), pinning to an exact version here is stricter than the convention used by its siblings and than the manual's own wording ("Extension Shlcofideleg supports delegating LCOFI interrupts to VS-mode", hypervisor.adoc:1278 — no version constraint stated). Recommend dropping the version pin on both `H` and `Sscofpmf` to match sibling style, unless there's a specific reason to require exactly 1.0.0.

```suggestion
requirements:
  extension:
    allOf:
      - name: H
      - name: Sscofpmf
```

---
AI-generated.

jordancarlin (commented):
The H and Sscofpmf dependencies, description, `type: privileged`, and ratification_date (2024-10, matching the "Preface to Version 20241017" ratified table in src/priv/preface.adoc:196-223) all check out against hypervisor.adoc:1278-1284 and supervisor.adoc:448-455. One convention deviation from sibling Sh* files on the version pinning.

---
AI-generated.

ThinkOpenly (approved):

PR #2231, opened 2026-07-30
Fix CFI dependency and shadow stack data

jordancarlin:
Encodes CFI extension requirements and adds the compressed Zicfiss shadow-stack instructions provided through Zcmop.

AI-generated.


spec/std/isa/ext/Zicfiss.yaml
```diff
@@ -8,9 +8,21 @@ kind: extension
...
 versions:
   - version: "1.0.0"
     state: ratified
     ratification_date: 2024-07
+requirements:
+  extension:
+    allOf:
+      - name: Zicsr
+      - name: Zimop
+      - name: Zaamo
```
jordancarlin: Missing requirement: the manual states use of Zicfiss in U-mode requires S-mode to be implemented (`ext/riscv-isa-manual/src/unpriv/cfi.adoc:385-386`): "Moreover, use of ext:zicfiss[] in U-mode requires S-mode to be implemented." This is analogous to the conditional pattern already used in `spec/std/isa/ext/Supm.yaml` (`if: extension: {name: S} / then: ...` else branch). Zicsr/Zimop/Zaamo are correctly captured, but the U-implies-S condition is absent.

```suggestion
requirements:
  extension:
    allOf:
      - name: Zicsr
      - name: Zimop
      - name: Zaamo
      - if:
          extension:
            name: U
        then:
          extension:
            name: S
```

---
AI-generated.

spec/std/isa/inst/Zicfiss/c.sspush.yaml
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../../schemas/inst_schema.json
+
+$schema: "inst_schema.json#"
```
jordancarlin: `$schema` is quoted here (`"inst_schema.json#"`) but every existing sibling file in this directory (`sspush.x1.yaml`, `sspush.x5.yaml`, `sspopchk.x1.yaml`, `sspopchk.x5.yaml`, `ssrdp.yaml`, `ssamoswap.*.yaml`) uses the unquoted form `inst_schema.json#`. Nit for directory consistency; both are schema-valid.

---
AI-generated.

spec/std/isa/inst/Zicfiss/c.sspopchk.yaml
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../../schemas/inst_schema.json
+
+$schema: "inst_schema.json#"
```
jordancarlin: Same nit as `c.sspush.yaml`: quoted `$schema` differs from the unquoted convention used by all other files in `spec/std/isa/inst/Zicfiss/`.

---
AI-generated.

jordancarlin (commented):
Encoding/dependency data for c.sspush/c.sspopchk checks out against the manual (C.MOP.1/C.MOP.5 match strings, x1/x5 operands, Zicfiss+Zcmop definedBy). One missing requirement: Zicfiss's U-mode dependency on S-mode is not encoded. Two minor style nits below.

---
AI-generated.

ThinkOpenly (approved):

PR #2232, opened 2026-07-30
Encode Zcb dependency on Zca

jordancarlin:
Encodes the Zcb dependency on Zca.

AI-generated.

### DIFF
--- a/spec/std/isa/ext/Zcb.yaml
+++ b/spec/std/isa/ext/Zcb.yaml
@@ -43,3 +43,6 @@ versions:
       - name: Graeme Smecher
       - name: Nicolas Brunie
       - name: Jiawei
+requirements:
+  extension:
+    name: Zca


jordancarlin (commented):
Verified against the manual: `ext/riscv-isa-manual/src/unpriv/zcb.adoc:6` states "The ext:zcb[] extension depends on the ext:zca[] extension." The added `requirements:` block matches this exactly, and its shape/placement (top-level, after `versions:`, single `extension: {name: ...}`) is consistent with `Zcf.yaml` and `Zcd.yaml`. The manual's other Zcb interactions (`c.mul` needing M/Zmmul, `c.sext.b`/`c.zext.h`/etc. needing Zbb, `c.zext.w` needing Zba) are per-instruction conditional availability, not extension-level requirements, and are already correctly encoded as `definedBy: allOf` conditions on the individual instructions (e.g. `spec/std/isa/inst/Zcb/c.mul.yaml`, `c.sext.b.yaml`) rather than as a Zcb extension requirement. No changes needed.

---
AI-generated.

ThinkOpenly (approved):

PR #2233, opened 2026-07-30
Encode Zihpm dependency on Zicsr

jordancarlin:
Encodes the Zihpm dependency on Zicsr.

AI-generated.

### DIFF
--- a/spec/std/isa/ext/Zihpm.yaml
+++ b/spec/std/isa/ext/Zihpm.yaml
@@ -13,3 +13,6 @@ versions:
   - version: "2.0.0"
     state: ratified
     ratification_date: 2023-03
+requirements:
+  extension:
+    name: Zicsr


jordancarlin (commented):
Verified against the manual: `ext/riscv-isa-manual/src/unpriv/zihpm.adoc:8` states "The ext:zihpm[] extension depends on the ext:zicsr[] extension." This is word-for-word the same dependency statement already used to justify `Zicntr`'s existing `requirements: {extension: {name: Zicsr}}` block (see `ext/riscv-isa-manual/src/unpriv/zicntr.adoc:13`, and `spec/std/isa/ext/Zicntr.yaml`), so this PR makes Zihpm consistent with its sibling counter extension rather than introducing any new inconsistency. Placement (top-level `requirements:` after `versions:`) and shape match `Zicntr.yaml`. Sscofpmf/Smcntrpmf interactions are optional extensions that add fields to the mhpmevent/hpmcounter CSRs when present, not prerequisites for Zihpm itself, so no additional requirement is warranted. No changes needed.

---
AI-generated.

ThinkOpenly (approved):

PR #2234, opened 2026-07-30
Fix vector crypto base-vector requirements

jordancarlin:
Adds missing Zve32x/Zve64x requirements for vector crypto extensions.

AI-generated.

### DIFF
--- a/spec/std/isa/ext/Zvkt.yaml
+++ b/spec/std/isa/ext/Zvkt.yaml
@@ -26,3 +26,6 @@ versions:
   - version: "1.0.0"
     state: ratified
     ratification_date: 2023-09
+requirements:
+  extension:
+    name: Zve32x
--- a/spec/std/isa/ext/Zvkt.yaml
+++ b/spec/std/isa/ext/Zvkt.yaml
@@ -28,4 +28,4 @@ versions:
     ratification_date: 2023-09
 requirements:
   extension:
-    name: Zve32x
+    name: Zvl32b
--- a/spec/std/isa/ext/Zvbc.yaml
+++ b/spec/std/isa/ext/Zvbc.yaml
@@ -17,3 +17,6 @@ versions:
   - version: "1.0.0"
     state: ratified
     ratification_date: 2023-09
+requirements:
+  extension:
+    name: Zve64x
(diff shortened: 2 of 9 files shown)


jordancarlin (commented):
Verified against `ext/riscv-isa-manual/src/unpriv/zvk.adoc` and `zvbc.adoc`. All 9 base-vector requirements are correct: the manual states explicitly (zvk.adoc:477) "The Zvknhb, Zvkn, Zvknc, Zvkng, and Zvksc depend on ext:zve64x[]" and (zvk.adoc:479) "All of the other Vector Crypto Extensions depend on ext:zve32x[]" (covers Zvkb, Zvkg, Zvkned, Zvknha, Zvksed, Zvksh, Zvkt), and zvbc.adoc:6 states "ext:zvbc[] depends upon the ext:zve64x[] extension." The Zvknhb merge correctly preserves the pre-existing Zvknha `= 1.0.0` requirement inside the new `allOf` rather than clobbering it. Checked `V.yaml` on main: `V` requires `Zve64d` which transitively implies `Zve64f -> Zve64x -> Zve32x`, so the bare `Zve32x`/`Zve64x` requirements do not wrongly reject V-only implementations. No findings.

---
AI-generated.

spec/std/isa/ext/Zvkt.yaml
```diff
@@ -26,3 +26,6 @@ versions:
   - version: "1.0.0"
     state: ratified
     ratification_date: 2023-09
+requirements:
+  extension:
+    name: Zve32x
```
ThinkOpenly: It's probably correct, but how was this determined?

jordancarlin: For Zvkt in particular, I guess we could do either Zvl32b or Zve32x. Seems like we use Zvl32b to represent that "vector" in general must be supported, so I can switch it to that.

ThinkOpenly: OK, let's. There's an interesting note in the manual:
> Explicit use of the `Zvl32b` extension string is not required for any standard vector extension as they all effectively mandate at least this minimum, but the string can be useful when stating hardware capabilities.

jordancarlin: Updated.

ThinkOpenly (approved):

PR #2235, opened 2026-07-30
Fix Zfbfmin dependency and instruction data

jordancarlin:
Corrects Zfbfmin to depend on F, not full Zfh, and marks the included half-width transfer/load/store instructions as also defined by Zfbfmin.

AI-generated.

### DIFF
--- a/spec/std/isa/ext/Zfbfmin.yaml
+++ b/spec/std/isa/ext/Zfbfmin.yaml
@@ -12,8 +12,8 @@ description: |
   It enables BF16 as an interchange format as it provides conversion between BF16 values and
   FP32 values.
 
-  This extension depends upon the single-precision floating-point extension `F`,
-  and the `flh`, `fsh`, `fmv.x.h`, and `fmv.h.x` instructions as defined in the `Zfh` extension.
+  This extension includes the `flh`, `fsh`, `fmv.x.h`, and `fmv.h.x`
+  instructions defined in the `Zfh` extension.
 
 type: unprivileged
 versions:
@@ -22,6 +22,4 @@ versions:
     ratification_date: 2024-06
 requirements:
   extension:
-    allOf:
-      - name: F
-      - name: Zfh
+    name: F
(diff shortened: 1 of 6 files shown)


jordancarlin (commented):
Reviewed against the vendored `riscv-isa-manual` (pinned at `310a111`, 2026-07-29). Both claims in this PR check out; no findings.

**Dependency change (F, not Zfh) — correct.** `src/unpriv/zfbfmin.adoc:8`: "This extension depends upon the ext:f[] extension." The manual states no Zfh or Zfhmin dependency, so narrowing `allOf: [F, Zfh]` to `F` is right.

**Instruction `definedBy` changes — correct.** `src/unpriv/zfbfmin.adoc:9-12`: "This extension includes six instructions: the insn:fcvt.bf16.s[] and insn:fcvt.s.bf16[] instructions, defined below, and the insn:flh[], insn:fsh[], insn:fmv.x.h[], and insn:fmv.h.x[] instructions, defined in extlink:zfh[]." Widening those four to `anyOf: [Zfhmin, Zfbfmin]` rather than replacing `Zfhmin` is the correct modeling, since Zfbfmin now reaches them without requiring Zfhmin. The `anyOf` shape matches `schema_defs.json` `$defs/extension_condition` and the existing precedent in `inst/B/ror.yaml`. `fcvt.bf16.s`/`fcvt.s.bf16` were already correct and rightly left untouched.

**Golden-file regeneration — correct.** The 41 removed lines are the old single-extension table blocks for `flh`/`fsh`/`fmv.h.x`/`fmv.x.h`; the replacement `xref:exts:` form matches `backends/instructions_appendix/templates/instructions.adoc.erb` and what `ror` already renders on main. Nothing correct was lost.

**Downstream — no inconsistency.** `Zvfbfwma.yaml` already requires `allOf: [Zvfbfmin, Zfbfmin]`, consistent with `src/unpriv/zvfbfwma.adoc:7`, and is unaffected by this change.

One non-blocking observation: the updated description says the reused instructions are "defined in the `Zfh` extension" while the condition uses `Zfhmin`. The manual's own prose says the same thing (`extlink:zfh[]`), so this mirrors the source rather than introducing an error.

---
AI-generated.

ThinkOpenly (approved):

PR #2236, opened 2026-07-30
Add seed CSR data for Zkr

jordancarlin:
Adds the missing seed CSR record for Zkr, including the mseccfg seed-access bits and seed access semantics.

AI-generated.

### DIFF
--- a/spec/std/isa/csr/Zkr/seed.yaml
+++ b/spec/std/isa/csr/Zkr/seed.yaml
@@ -1,0 +1,94 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: seed
+long_name: Seed for cryptographic random bit generators
+address: 0x015
+writable: true
+priv_mode: M
+length: 32
+description: |
+  The `seed` CSR provides an interface to a physical Entropy Source.
+
+  The CSR provides up to 16 physical entropy bits that can be used to seed
+  cryptographic random bit generators. The write value is ignored by
+  implementations; the write is used to signal polling and flushing.
+definedBy:
+  extension:
+    name: Zkr
+fields:
+  OPST:
+    location: 31-30
+    description: |
+      Status bits returned by the entropy source.
+
+      Encodings are:
+
+      * 0 - BIST
+      * 1 - WAIT
+      * 2 - ES16
+      * 3 - DEAD
+    type: RW-H
+    reset_value: UNDEFINED_LEGAL
+    sw_write(csr_value): |
+      if (mode() == PrivilegeMode::U && CSR[mseccfg].USEED == 1'b0) {
+        raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+      } else if (mode() == PrivilegeMode::S && CSR[mseccfg].SSEED == 1'b0) {
+        raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+      } else if (mode() == PrivilegeMode::VS || mode() == PrivilegeMode::VU) {
+        if (CSR[mseccfg].SSEED == 1'b1) {
+          raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
+        } else {
+          raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+        }
+      }
+      return CSR[seed].OPST;
+  ENTROPY:
+    location: 15-0
+    description: |
+      Sixteen bits of randomness.
+
+      This field is valid only when `OPST` is ES16. When `OPST` is not ES16,
+      `ENTROPY` must be zero.
+    type: RW-H
+    reset_value: UNDEFINED_LEGAL
+    sw_write(csr_value): |
+      if (mode() == PrivilegeMode::U && CSR[mseccfg].USEED == 1'b0) {
+        raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+      } else if (mode() == PrivilegeMode::S && CSR[mseccfg].SSEED == 1'b0) {
+        raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+      } else if (mode() == PrivilegeMode::VS || mode() == PrivilegeMode::VU) {
+        if (CSR[mseccfg].SSEED == 1'b1) {
+          raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
+        } else {
+          raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+        }
+      }
+      return CSR[seed].ENTROPY;
+sw_read(): |
+  Bits<3> csr_funct3 = $encoding[14:12];
+  Bits<5> csr_rs1_uimm = $encoding[19:15];
+
+  if (((csr_funct3 == 3'b010) || (csr_funct3 == 3'b011) ||
+       (csr_funct3 == 3'b110) || (csr_funct3 == 3'b111)) &&
+      (csr_rs1_uimm == 5'b00000)) {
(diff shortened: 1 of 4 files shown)


jordancarlin (commented):
Two blocking issues: `priv_mode` and the missing access-control gating, plus the missing read-only-instruction-illegal semantics that the manual specifically calls out as normative for this CSR.

---
AI-generated.

spec/std/isa/csr/seed.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+# yaml-language-server: $schema=../../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: seed
+long_name: Seed for cryptographic random bit generators
+address: 0x015
+writable: true
+priv_mode: U
```
jordancarlin: `priv_mode: U` is wrong. The manual states `seed` is an M-mode CSR whose accessibility from less-privileged modes is controlled by `mseccfg.SSEED`/`USEED`: "The Zkr extension adds the `SSEED` and `USEED` fields to the `mseccfg` CSR to control access to the `seed` CSR from modes less privileged than M" and the access-control table states "M ... The `seed` CSR is always available in machine mode as normal ... U ... `USEED`=`0` ... Any `seed` CSR access raises an illegal-instruction exception" (`ext/riscv-isa-manual/src/priv/machine.adoc` lines 2331-2371, table at 2352-2371). Compare `mseccfg`/`mseccfgh` in this repo which are `priv_mode: M`. This should be `priv_mode: M` with `sw_read()`/`sw_write(csr_value)` implementing the mseccfg-gated access (illegal-instruction below M without the seed bit set; virtual-instruction in VS/VU per the same table), the way `ssp.yaml` and `jvt.yaml` gate access via `menvcfg`/`senvcfg`/`henvcfg`.

```suggestion
priv_mode: M
```

---
AI-generated.

spec/std/isa/csr/seed.yaml
```diff
@@ -0,0 +1,45 @@
...
+    description: |
+      Status bits returned by the entropy source.
+
+      Encodings are:
+
+      * 0 - BIST
+      * 1 - WAIT
+      * 2 - ES16
+      * 3 - DEAD
+    type: RO-H
```
jordancarlin: `seed` reads are destructive and the manual makes read-only access illegal: "Attempts to access the csr:seed[] CSR using a read-only CSR-access instruction (csrrs/csrrc with rs1=x0 or csrrsi/csrrci with uimm=0) raise an illegal-instruction exception; any other CSR-access instruction may be used to access csr:seed[]" (`ext/riscv-isa-manual/src/unpriv/zk.adoc` line 3574-3577, `norm:seed_ro_illegal`). As written with plain `RO-H` fields and no `sw_read()`/`sw_write(csr_value)`, this models `seed` as an ordinary readable CSR — a CSRRS/CSRRC read would succeed and return stale/garbage data instead of raising `IllegalInstruction`. This needs a top-level `sw_read()` (and/or field-level `sw_write(csr_value)`) that raises `ExceptionCode::IllegalInstruction` when the access is not a genuine read-write, following the pattern used elsewhere in this repo (e.g. `cycle.yaml`, `ssp.yaml`) of raising via `$encoding` in `sw_read()`. This is a blocking correctness gap, not just a style nit.

---
AI-generated.

spec/std/isa/csr/Zkr/seed.yaml
```diff
@@ -0,0 +1,45 @@
...
+writable: true
+priv_mode: U
+length: 32
+description: |
+  The `seed` CSR provides an interface to a physical Entropy Source.
+
+  The CSR provides up to 16 physical entropy bits that can be used to seed
+  cryptographic random bit generators. The write value is ignored by
+  implementations; the write is used to signal polling and flushing.
+definedBy:
```
jordancarlin, line 21: No mseccfg gating is modeled at all (no reference to `CSR[mseccfg].SSEED`/`USEED`). Per `ext/riscv-isa-manual/src/priv/machine.adoc` lines 2334-2371, access from S/HS and U mode must raise `IllegalInstruction` when the corresponding bit is 0, and from VS/VU an HS-qualified access raises `VirtualInstruction`. This should be implemented in `sw_read()`/`sw_write(csr_value)` alongside the read-only-instruction check noted below.

---
AI-generated.

jordancarlin:
This is now cleaned up and complete, but we need to implement builtins of some kind for the actual random number generation. We also need to decide how to handle OPST. For context, sail takes the approach of making OPST always ES16. @henrikg-qc @ThinkOpenly

henrikg-qc:
This has overlap with PR #913 
I implemented the entropy functionality in the ISS but @blazethunderstorm never responded nor completed PR #913 to merge

jordancarlin:
> This has overlap with PR #913 I implemented the entropy functionality in the ISS but ANIRUDDH (@blazethunderstorm) never responded nor completed PR #913 to merge

There were several other issues with #913 (for example, it doesn't check for read-only accesses, which need to trap). I will pull your entropy implementation into this PR.

spec/std/isa/csr/Zkr/seed.yaml
```diff
@@ -0,0 +1,93 @@
...
+    raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+  } else if (mode() == PrivilegeMode::VS || mode() == PrivilegeMode::VU) {
+    if (CSR[mseccfg].SSEED == 1'b1) {
+      raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
+    } else {
+      raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+    }
+  }
+
+  return $bits(CSR[seed]);
```
ThinkOpenly, line 93: This needs to be a builtin that returns the values set by hardware upon reading... the actual entropy value and associated status.

spec/std/isa/csr/Zkr/seed.yaml
```diff
@@ -0,0 +1,93 @@
...
+        if (CSR[mseccfg].SSEED == 1'b1) {
+          raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
+        } else {
+          raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+        }
+      }
+      return CSR[seed].ENTROPY; # TODO: Implement ENTROPY as a builtin?
+sw_read(): |
+  # CSRRS/CSRRC/CSRRSI/CSRRCI are the CSR instructions with funct3[1] set.
+  Boolean read_only_access = ($encoding[13] == 1'b1) && ($encoding[19:15] == 5'b00000);
```
ThinkOpenly, line 75: I guess we're presuming here that only CSR instructions will use the `sw_read()` method. OK. I wonder if it would be helpful to also say that the `19:15` check is for the source register that must be `x0` (read-only-zero). These semantics are not otherwise defined in this file and may also be useful in the CSR description well above.

spec/std/isa/csr/Zkr/seed.yaml
```diff
@@ -0,0 +1,93 @@
...
+      } else if (mode() == PrivilegeMode::S && CSR[mseccfg].SSEED == 1'b0) {
+        raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+      } else if (mode() == PrivilegeMode::VS || mode() == PrivilegeMode::VU) {
+        if (CSR[mseccfg].SSEED == 1'b1) {
+          raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
+        } else {
+          raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+        }
+      }
+      return CSR[seed].ENTROPY; # TODO: Implement ENTROPY as a builtin?
```
ThinkOpenly, line 72: The `ENTROPY` field is technically read-only. Writes are ignored. It's probably fine to just return the existing value as you already have here, but the comment should be removed. The builtin for entropy should be in the `sw_read()` method below.
```suggestion
      return CSR[seed].ENTROPY;
```

ThinkOpenly (changes requested):
This is a weird one. `seed` is read-write, except it's not writable. Hmm. I guess it's OK that an illegal write will be detected twice?

PR #2237, opened 2026-07-30
Add Sstc timer compare CSR data

jordancarlin:
Adds the missing Sstc timer-compare CSR records for supervisor and virtual-supervisor modes, including STCE and counter-enable access gating.

AI-generated.

### DIFF
--- a/spec/std/isa/csr/Sstc/stimecmp.yaml
+++ b/spec/std/isa/csr/Sstc/stimecmp.yaml
@@ -1,0 +1,42 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: stimecmp
+long_name: Supervisor timer compare
+address: 0x14D
+writable: true
+priv_mode: S
+length: 64
+description: |
+  Supervisor timer compare register.
+
+  When `menvcfg.STCE` is set, `stimecmp` is operational and provides
+  supervisor mode with a CSR-based timer interrupt facility.
+definedBy:
+  extension:
+    name: Sstc
+fields:
+  TIME:
+    location: 63-0
+    description: |
+      Time compare value for supervisor timer interrupts.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+    sw_write(csr_value): |
+      if (mode() != PrivilegeMode::M) {
+        if (CSR[menvcfg].STCE == 1'b0 || CSR[mcounteren].TM == 1'b0) {
+          raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+        }
+      }
+      return csr_value.TIME;
+sw_read(): |
+  if (mode() != PrivilegeMode::M) {
+    if (CSR[menvcfg].STCE == 1'b0 || CSR[mcounteren].TM == 1'b0) {
+      raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+    }
+  }
+  return CSR[stimecmp].TIME;
--- a/spec/std/isa/csr/Sstc/stimecmph.yaml
+++ b/spec/std/isa/csr/Sstc/stimecmph.yaml
@@ -1,0 +1,43 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: stimecmph
+long_name: Supervisor timer compare high
+address: 0x15D
+writable: true
+priv_mode: S
+length: 32
+description: |
+  Upper 32 bits of the `stimecmp` CSR.
+definedBy:
+  allOf:
+    - xlen: 32
+    - extension:
+        name: Sstc
+fields:
+  TIME:
+    location: 31-0
+    alias: stimecmp.TIME[63:32]
+    description: |
+      Upper 32 bits of the supervisor timer compare value.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+    sw_write(csr_value): |
+      if (mode() != PrivilegeMode::M) {
+        if (CSR[menvcfg].STCE == 1'b0 || CSR[mcounteren].TM == 1'b0) {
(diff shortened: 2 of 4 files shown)


spec/std/isa/csr/Sstc/stimecmp.yaml
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
jordancarlin, line 1: Consider placing this under `spec/std/isa/csr/Sstc/` rather than the top level. The repo's convention for single/small extensions that own CSRs is to give them their own subdirectory (e.g. `Sscofpmf/scountovf.yaml`, `Ssqosid/srmcfg.yaml`, `Smrnmi/mncause.yaml`, `Smctr/{m,s,vs}ctrctl.yaml`). Sstc owns exactly these 4 CSRs, so a subdirectory fits the pattern (note: the layout is inconsistent overall, e.g. many `H`-required CSRs like `hstatus.yaml`/`hedeleg.yaml` stay at top level, so this isn't a hard rule).

---
AI-generated.

spec/std/isa/csr/Sstc/stimecmp.yaml
```diff
@@ -0,0 +1,28 @@
...
+name: stimecmp
+long_name: Supervisor timer compare
+address: 0x14D
+writable: true
+priv_mode: S
+length: 64
+description: |
+  Supervisor timer compare register.
+
+  When `menvcfg.STCE` is set, `stimecmp` is operational and provides
```
jordancarlin, line 17: Access to `stimecmp` is not unconditional. The manual states two independent gates that this record doesn't implement:

- `machine.adoc:2222-2224` (`norm:menvcfg_stce_op2`): "The `STCE` bit enables `stimecmp` for S-mode when set to one. When this extension is implemented and `STCE` in `menvcfg` is zero, an attempt to access `stimecmp` in a mode other than M-mode raises an illegal-instruction exception."
- `machine.adoc:1586-1588` (`norm:mcounteren_tm_clr`): "when the [tm] bit in the `mcounteren` register is clear, attempts to access the `stimecmp` ... register while executing in a mode less privileged than M will cause an illegal-instruction exception."

Sibling CSRs with equivalent conditional access (`time.yaml`, `timeh.yaml`, `Ssqosid/srmcfg.yaml`) implement this via `sw_read()`/`sw_write(csr_value)` with explicit `raise(ExceptionCode::IllegalInstruction, mode(), $encoding)`. This CSR has no `sw_read()`/`sw_write()` at all, so it's modeled as always accessible from S-mode regardless of `menvcfg.STCE` or `mcounteren.TM`.

---
AI-generated.

jordancarlin (commented):
Addresses, `definedBy` shapes (including the H+Sstc `allOf` for VS variants), `priv_mode`, `length`, and `virtual_address` all check out against the manual and sibling CSR conventions. The main gap is that none of the four CSRs implement the normative access-control gating (`menvcfg.STCE`/`henvcfg.STCE` and `mcounteren.TM`/`hcounteren.TM`) that the manual requires and that comparable CSRs in this repo (`time.yaml`, `timeh.yaml`, `srmcfg.yaml`) implement via `sw_read()`/`sw_write()` with explicit `raise()` calls — as written these CSRs are unconditionally accessible.

---
AI-generated.

spec/std/isa/csr/Sstc/stimecmph.yaml
```diff
@@ -0,0 +1,33 @@
...
+    location: 31-0
+    alias: stimecmp.TIME[63:32]
+    description: |
+      Upper 32 bits of the supervisor timer compare value.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+    sw_write(csr_value): |
+      CSR[stimecmp].TIME = {csr_value.TIME, CSR[stimecmp].TIME[31:0]};
+      return csr_value.TIME;
+sw_read(): |
```
jordancarlin, line 37: Same gap as `stimecmp.yaml`: the `sw_write`/`sw_read` here only handle the high/low split, not the `menvcfg.STCE` / `mcounteren.TM` access gating described in `machine.adoc:2222-2224` and `machine.adoc:1586-1588`. As written, `stimecmph` is accessible from S-mode unconditionally.

---
AI-generated.

spec/std/isa/csr/Sstc/vstimecmp.yaml
```diff
@@ -0,0 +1,31 @@
...
+long_name: Virtual supervisor timer compare
+address: 0x24D
+writable: true
+virtual_address: 0x14D
+priv_mode: VS
+length: 64
+description: |
+  Virtual supervisor timer compare register.
+
+  When `menvcfg.STCE` and `henvcfg.STCE` are set, `vstimecmp` is operational
```
jordancarlin, line 18: No gating is implemented for VS-mode access either. Per the manual:

- `hypervisor.adoc:748-751` (`norm:henvcfg_stce`): "When [the] `STCE` bit [in] `henvcfg` is zero, an attempt to access `stimecmp` (really `vstimecmp`) when V=1 raises a virtual-instruction exception."
- `hypervisor.adoc:866-871` (`norm:hcounteren_acc`): "when the TM bit in the `hcounteren` register is clear, attempts to access the `vstimecmp` register (via `stimecmp`) while executing in VS-mode will cause a virtual-instruction exception if the same bit in `mcounteren` is set."
- plus the S-mode-side `menvcfg.STCE`/`mcounteren.TM` gates from `machine.adoc` apply too (access when V=1 is gated by both levels).

None of this is implemented via `sw_read()`/`sw_write()`, so `vstimecmp` is modeled as unconditionally accessible in VS-mode.

---
AI-generated.

spec/std/isa/csr/Sstc/vstimecmph.yaml
```diff
@@ -0,0 +1,36 @@
...
+    location: 31-0
+    alias: vstimecmp.TIME[63:32]
+    description: |
+      Upper 32 bits of the virtual supervisor timer compare value.
+    type: RW
+    reset_value: UNDEFINED_LEGAL
+    sw_write(csr_value): |
+      CSR[vstimecmp].TIME = {csr_value.TIME, CSR[vstimecmp].TIME[31:0]};
+      return csr_value.TIME;
+sw_read(): |
```
jordancarlin, line 40: Same gap as `vstimecmp.yaml`: no `henvcfg.STCE`/`hcounteren.TM` (or `menvcfg.STCE`/`mcounteren.TM`) gating is implemented, so this is modeled as unconditionally accessible in VS-mode when XLEN=32.

---
AI-generated.

ThinkOpenly (approved):

PR #2256, opened 2026-07-30
docs(idl): state what read_memory returns, and that the width is MXLEN

### LINKED ISSUE #2253: Improve documentation for "read_memory" in IDL standard library
What type is returned from `read_memory`?

Add to `doc/docs/idl/standard-library.mdx`

titoatwork:
The standard library page documented `read_memory`'s arguments and gave three examples but never said what it returns. The answer is not obvious from the examples: it returns `XReg`, which is `Bits<MXLEN>`, so the returned width is the register width and not the access width, and the `len` bits that were read are zero-extended into it.

That is the reason signed loads look the way they do. `lb` and `lw` wrap the call in `sext` because the value arrives zero-extended, while `lbu` and `lwu` assign it straight to `X[xd]`. Without the return type written down, the `sext` in those examples reads like a style choice rather than a requirement.

Also notes that the doubleword example needs MXLEN to be 64, since `Bits<64>` cannot hold a `Bits<MXLEN>` result on a 32-bit config, and that `write_memory` returns nothing.

Closes #2253

### DIFF
--- a/doc/docs/idl/standard-library.mdx
+++ b/doc/docs/idl/standard-library.mdx
@@ -120,20 +120,22 @@ Memory access functions handle virtual-to-physical address translation, PMP/PMA 
 
 Reads `len` bits from virtual memory. `len` must be 8, 16, 32, or 64.
 
+Returns `XReg`, which is `Bits<MXLEN>` and not `Bits<len>`. The width of the returned value is the register width, whatever the access size, and the `len` bits that were read are zero-extended into it. A signed load therefore has to widen the result itself with `sext(value, len)`, which is why `lb` and `lw` call `sext` while `lbu` and `lwu` assign the result directly.
+
 ```idl
-# Load byte (zero-extended)
+# Load byte, zero-extended into the full XReg
 X[xd] = read_memory(8, virtual_address, $encoding);
 
-# Load word (sign-extended via sext helper)
+# Load word, sign-extended from bit 31 by the sext helper
 X[xd] = sext(read_memory(32, virtual_address, $encoding), 32);
 
-# Load doubleword
+# Load doubleword. MXLEN must be 64 here, since the return type is Bits<MXLEN>
 Bits<64> data = read_memory(64, virtual_address, $encoding);
 ```
 
 ### `write_memory(len, virtual_address, value, encoding)`
 
-Writes `len` bits to virtual memory.
+Writes `len` bits to virtual memory. Returns nothing.
 
 ```idl
 # Store byte
--- a/doc/docs/idl/standard-library.mdx
+++ b/doc/docs/idl/standard-library.mdx
@@ -120,10 +120,10 @@ Memory access functions handle virtual-to-physical address translation, PMP/PMA 
 
 Reads `len` bits from virtual memory. `len` must be 8, 16, 32, or 64.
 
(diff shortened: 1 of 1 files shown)


doc/docs/idl/standard-library.mdx
```diff
@@ -120,20 +120,22 @@ Memory access functions handle virtual-to-physical address translation, PMP/PMA
 
 Reads `len` bits from virtual memory. `len` must be 8, 16, 32, or 64.
 
+Returns `XReg`, which is `Bits<MXLEN>` and not `Bits<len>`. The width of the returned value is the register width, whatever the access size, and the `len` bits that were read are zero-extended into it. A signed load therefore has to widen the result itself with `sext(value, len)`, which is why `lb` and `lw` call `sext` while `lbu` and `lwu` assign the result directly.
```
ThinkOpenly: wordsmithing...
```suggestion
Returns `XReg` (`Bits<MXLEN>`). The `len` bits that were read are zero-extended to MXLEN. To read a signed quantity, the caller must sign-extend the result itself with `sext(result, len)`.
```

doc/docs/idl/standard-library.mdx
```diff
@@ -120,20 +120,22 @@ Memory access functions handle virtual-to-physical address translation, PMP/PMA
...
+
 ```idl
-# Load byte (zero-extended)
+# Load byte, zero-extended into the full XReg
```
ThinkOpenly:
```suggestion
# Load byte, zero-extended to full XReg width
```

doc/docs/idl/standard-library.mdx
```diff
@@ -120,20 +120,22 @@ Memory access functions handle virtual-to-physical address translation, PMP/PMA
...
 X[xd] = read_memory(8, virtual_address, $encoding);
 
-# Load word (sign-extended via sext helper)
+# Load word, sign-extended from bit 31 by the sext helper
 X[xd] = sext(read_memory(32, virtual_address, $encoding), 32);
 
-# Load doubleword
+# Load doubleword. MXLEN must be 64 here, since the return type is Bits<MXLEN>
```
ThinkOpenly, line 132: That's a good point, making me wonder if `ld` (RV32) is broken:
```
  if (xlen() == 32) {
    if (implemented?(ExtensionName::Zilsd)) {
      Bits<64> data = read_memory(64, virtual_address, $encoding);
```
That's obviously testing `xlen()` instead of MXLEN, but if MXLEN == 32, `xlen()` == 32, and I think the above is broken in that case. Maybe `read_memory` should always return `Bits<64>`? Hmm. (@henrikg-qc?)

doc/docs/idl/standard-library.mdx
```diff
@@ -120,20 +120,22 @@ Memory access functions handle virtual-to-physical address translation, PMP/PMA
...
 Bits<64> data = read_memory(64, virtual_address, $encoding);
 ```
 
 ### `write_memory(len, virtual_address, value, encoding)`
 
-Writes `len` bits to virtual memory.
+Writes `len` bits to virtual memory. Returns nothing.
```
ThinkOpenly:
```suggestion
Writes `len` bits to virtual memory. Returns nothing (`void`).
```

ThinkOpenly (commented):
Thanks, this is very helpful! Some wordsmithing, and a thought that we have a bug unrelated to this PR.

ThinkOpenly (changes requested):
Forgot to select the "request changes" button for the above comments.

ThinkOpenly (approved):

PR #2266, opened 2026-07-30
fix(param): enforce the counter-enable rules the descriptions already state

### LINKED ISSUE #2265: MCOUNTENABLE_EN and HCOUNTENABLE_EN do not enforce the counter rules their own descriptions state

`spec/std/isa/param/MCOUNTENABLE_EN.yaml`, `SCOUNTENABLE_EN.yaml` and `HCOUNTENABLE_EN.yaml` carry the same three sentences in their `description`, differing only in the parameter name:

- an unimplemented counter cannot be specified, so `HPM_COUNTER_EN[3]` false makes enabling index 3 illegal
- `[0:2]` must all be false if `Zicntr` is not implemented
- `[3:31]` must all be false if `Zihpm` is not implemented

Only `SCOUNTENABLE_EN` encodes all three in `requirements`. `MCOUNTENABLE_EN` encodes one, the counter-existence rule written as `MCOUNTENABLE_EN[i] -> HPM_COUNTER_EN[i]`. `HCOUNTENABLE_EN` has no `requirements` block at all.

`Parameter` requirements are checked when a configuration is resolved, in `cfg_arch.rb`, which reports "Parameter requirements not met". So a config that enables `mcounteren` bits 0 to 2 without `Zicntr`, or any `hcounteren` bit for a counter that does not exist, is accepted today while the equivalent `scounteren` config is rejected.

**Expected value**

The two incomplete parameters carrying the clauses `SCOUNTENABLE_EN` already uses:

```
for (U32 i = 0; i < 3; i++) {
  !implemented?(ExtensionName::Zicntr) -> !<PARAM>[i];
}
for (U32 i = 3; i < 32; i++) {
  !implemented?(ExtensionName::Zihpm) -> !<PARAM>[i];
  (!HPM_COUNTER_EN[i]) -> !<PARAM>[i];
}
```

This is not a change of interpretation. All three files already state the rule; two of them do not enforce it, and the third shows the encoding.

The asymmetry is not explained by a privilege-level difference. The privileged spec is explicit that there is no cross-level writability constraint to model here: "The setting of a bit in `mcounteren` does not affect whether the corresponding bit in `scounteren` is writable" (`supervisor.adoc`, the note under `norm:scounteren_acc`). `Shcounterenw` and `Sscounterenw` tie counter-enable writability to whether the `hpmcounter` is read-only zero, which is the same direction the existing clauses encode.

Nothing in the repository is affected. Seven configurations under `cfgs/` set at least one of these three parameters, and none would be newly rejected.

titoatwork:
`MCOUNTENABLE_EN`, `SCOUNTENABLE_EN` and `HCOUNTENABLE_EN` carry the same three sentences in their descriptions: an unimplemented counter cannot be enabled, bits 0 to 2 must be false without `Zicntr`, and bits 3 to 31 must be false without `Zihpm`. Only `SCOUNTENABLE_EN` encodes all three in `requirements`. `MCOUNTENABLE_EN` encodes one and `HCOUNTENABLE_EN` has no `requirements` block, so both accept configurations their own descriptions forbid.

This adds the missing clauses, copying the encoding `SCOUNTENABLE_EN` already uses. `MCOUNTENABLE_EN` keeps its existing counter-existence clause, which is the same rule written as a contrapositive. No cross-level clause is added, since the privileged spec states that a bit in `mcounteren` does not affect whether the corresponding bit in `scounteren` is writable.

No configuration changes status. Of the seven under `cfgs/` that set any of these parameters, none is newly rejected.

Closes #2265

### DIFF
--- a/spec/std/isa/param/HCOUNTENABLE_EN.yaml
+++ b/spec/std/isa/param/HCOUNTENABLE_EN.yaml
@@ -25,3 +25,16 @@ schema:
 definedBy:
   extension:
     name: H
+requirements:
+  idl(): |
+    for (U32 i = 0; i < 3; i++) {
+      !implemented?(ExtensionName::Zicntr) -> !HCOUNTENABLE_EN[i];
+    }
+    for (U32 i = 3; i < 32; i++) {
+      !implemented?(ExtensionName::Zihpm) -> !HCOUNTENABLE_EN[i];
+      (!HPM_COUNTER_EN[i]) -> !HCOUNTENABLE_EN[i];
+    }
+  reason: |
+    Counters 0-2 are defined by Zicntr
+    Counters 3..31 are defined by Zihpm
+    When mhpmcounter[i] does not exist, it cannot be enabled.
--- a/spec/std/isa/param/HCOUNTENABLE_EN.yaml
+++ b/spec/std/isa/param/HCOUNTENABLE_EN.yaml
@@ -37,4 +37,4 @@ requirements:
   reason: |
     Counters 0-2 are defined by Zicntr
     Counters 3..31 are defined by Zihpm
-    When mhpmcounter[i] does not exist, it cannot be enabled.
+    When mhpmcounter[i] is not implemented, it cannot be enabled.
--- a/spec/std/isa/param/MCOUNTENABLE_EN.yaml
+++ b/spec/std/isa/param/MCOUNTENABLE_EN.yaml
@@ -27,6 +27,14 @@ definedBy:
     name: Sm
 requirements:
   idl(): |
+    for (U32 i = 0; i < 3; i++) {
+      !implemented?(ExtensionName::Zicntr) -> !MCOUNTENABLE_EN[i];
+    }
     for (U32 i=3; i < 32; i++) {
+      !implemented?(ExtensionName::Zihpm) -> !MCOUNTENABLE_EN[i];
       MCOUNTENABLE_EN[i] -> HPM_COUNTER_EN[i];
     }
+  reason: |
+    Counters 0-2 are defined by Zicntr
+    Counters 3..31 are defined by Zihpm
+    When mhpmcounter[i] does not exist, it cannot be enabled.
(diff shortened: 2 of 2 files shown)


spec/std/isa/param/HCOUNTENABLE_EN.yaml
```diff
@@ -25,3 +25,16 @@ schema:
...
+      !implemented?(ExtensionName::Zicntr) -> !HCOUNTENABLE_EN[i];
+    }
+    for (U32 i = 3; i < 32; i++) {
+      !implemented?(ExtensionName::Zihpm) -> !HCOUNTENABLE_EN[i];
+      (!HPM_COUNTER_EN[i]) -> !HCOUNTENABLE_EN[i];
+    }
+  reason: |
+    Counters 0-2 are defined by Zicntr
+    Counters 3..31 are defined by Zihpm
+    When mhpmcounter[i] does not exist, it cannot be enabled.
```
ThinkOpenly:
```suggestion
    When mhpmcounter[i] is not implemented, it cannot be enabled.
```

spec/std/isa/param/MCOUNTENABLE_EN.yaml
```diff
@@ -27,6 +27,14 @@ definedBy:
...
+      !implemented?(ExtensionName::Zicntr) -> !MCOUNTENABLE_EN[i];
+    }
     for (U32 i=3; i < 32; i++) {
+      !implemented?(ExtensionName::Zihpm) -> !MCOUNTENABLE_EN[i];
       MCOUNTENABLE_EN[i] -> HPM_COUNTER_EN[i];
     }
+  reason: |
+    Counters 0-2 are defined by Zicntr
+    Counters 3..31 are defined by Zihpm
+    When mhpmcounter[i] does not exist, it cannot be enabled.
```
ThinkOpenly:
```suggestion
    When mhpmcounter[i] is not implemented, it cannot be enabled.
```

ThinkOpenly (changes requested):
This is great! Thank you! Just some tiny nits.

ThinkOpenly (approved):

PR #2273, opened 2026-07-31
fix(data): write minstreth updates to minstret

jordancarlin:
`minstreth.COUNT` aliases the upper half of `minstret`, but its write operation updated `mcycle` instead. This corrects the write target to update `minstret`.

Generated with AI assistance.


ThinkOpenly (approved):
Big oops.

PR #2274, opened 2026-07-31
fix(data): preserve correct upper mcycle bits on RV32 writes

jordancarlin:
RV32 writes to `mcycle.COUNT` should update the low 32 bits while preserving bits 63:32. The existing write path preserved bits 63:31, so this corrects the preserved slice.

Generated with AI assistance.


ThinkOpenly (approved):
More subtle, but oops.

PR #2289, opened 2026-07-31
fix(data): validate parameter enum string literals in IDL compares

### LINKED ISSUE #2285: need validation that parameter enum conditionals use valid enum values
Do all enum checks use a valid enum?

spec/std/isa/param/TINST_VALUE_ON_STORE_AMO_ACCESS_FAULT.yaml:
```
  enum:
    - always zero
    - always transformed standard instruction
    - custom
```

```
TINST_VALUE_ON_STORE_AMO_ACCESS_FAULT != "always_zero"
```
(note the `_`)

_Originally posted by @ThinkOpenly in https://github.com/riscv/riscv-unified-db/issues/2271#issuecomment-5146023912_

titoatwork:
String-enum parameters list legal values under `schema.enum`, but IDL comparisons in CSR `type()` and similar sites are free text. A typo such as `"always_zero"` where the enum only has `"always zero"` is never rejected by schema validation and changes the meaning of the condition.

The instance in `htinst` and `mtinst` was fixed in #2271. This adds `tools/scripts/check_param_enum_literals.py` and a smoke CI job so any later invalid literal fails the PR. Unit tests cover a valid compare, an invalid typo, and ignoring non-parameter identifiers.

Closes #2285


tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,153 @@
...
+from __future__ import annotations
+
+import argparse
+import re
+import sys
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parents[2]
+PARAM_DIR = ROOT / "spec" / "std" / "isa" / "param"
+SCAN_ROOTS = [
```
ThinkOpenly: We should look at everything under `spec`.

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,153 @@
...
+    ROOT / "spec" / "std" / "isa" / "csr",
+    ROOT / "spec" / "std" / "isa" / "isa",
+    ROOT / "spec" / "std" / "isa" / "inst",
+    ROOT / "spec" / "std" / "isa" / "param",
+]
+SCAN_SUFFIXES = {".yaml", ".yml", ".isa", ".idl"}
```
ThinkOpenly: There are no `.yml` files.

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,153 @@
...
+
+# PARAM == "literal" or PARAM != "literal" (double or single quotes)
+COMPARE_RE = re.compile(r"""\b([A-Z][A-Z0-9_]*)\s*(==|!=)\s*(?:"([^"]*)"|'([^']*)')""")
+
+
+def load_param_string_enums(param_dir: Path) -> dict[str, set[str]]:
+    """Map parameter name -> set of schema.enum string values."""
+    enums: dict[str, set[str]] = {}
+    for path in sorted(param_dir.glob("*.yaml")):
+        text = path.read_text(encoding="utf-8")
```
ThinkOpenly: we should use `ruamel` to load the YAML data into a Python `dict`. Doing that will change all of the logic here, so I won't review it now.

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,153 @@
...
+SCAN_ROOTS = [
+    ROOT / "spec" / "std" / "isa" / "csr",
+    ROOT / "spec" / "std" / "isa" / "isa",
+    ROOT / "spec" / "std" / "isa" / "inst",
+    ROOT / "spec" / "std" / "isa" / "param",
+]
+SCAN_SUFFIXES = {".yaml", ".yml", ".isa", ".idl"}
+
+# PARAM == "literal" or PARAM != "literal" (double or single quotes)
+COMPARE_RE = re.compile(r"""\b([A-Z][A-Z0-9_]*)\s*(==|!=)\s*(?:"([^"]*)"|'([^']*)')""")
```
ThinkOpenly, line 36: This is a little fragile. It assumes that all comparisons will look like `PARAM op "string"`, but it's legal to have `"string" op PARAM`. Unlikely, certainly. I doubt there are any cases (I didn't look). Maybe just add a note about that lack of robustness and the expectation of rarity.

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,153 @@
...
+            enums[name] = values
+    return enums
+
+
+def iter_scan_files(roots: list[Path]) -> list[Path]:
+    files: list[Path] = []
+    for root in roots:
+        if not root.is_dir():
+            continue
+        for path in root.rglob("*"):
```
ThinkOpenly: I'd consider limiting the pattern some (so you don't get _everything_) to files with a 'dot', something like `?*.?*`

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,153 @@
...
+        except OSError as exc:
+            print(f"ERROR: could not read {path}: {exc}", file=sys.stderr)
+            continue
+        for line_no, line in enumerate(text.splitlines(), 1):
+            for match in COMPARE_RE.finditer(line):
+                param = match.group(1)
+                if param not in enums:
+                    continue
+                op = match.group(2)
+                literal = match.group(3) if match.group(3) is not None else match.group(4)
```
ThinkOpenly, line 100: You might want to explain somewhere why there is a `group(3)` and a `group(4)`. These are for single quote and double quote cases?

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,153 @@
...
+    parser.add_argument(
+        "--root",
+        type=Path,
+        default=ROOT,
+        help="Repository root (default: inferred from script location)",
+    )
+    args = parser.parse_args(argv)
+    root: Path = args.root.resolve()
+
+    param_dir = root / "spec" / "std" / "isa" / "param"
```
ThinkOpenly: Use the global `PARAM_DIR` here?

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,153 @@
...
+        print(f"ERROR: param directory not found: {param_dir}", file=sys.stderr)
+        return 2
+
+    enums = load_param_string_enums(param_dir)
+    scan_roots = [
+        root / "spec" / "std" / "isa" / "csr",
+        root / "spec" / "std" / "isa" / "isa",
+        root / "spec" / "std" / "isa" / "inst",
+        root / "spec" / "std" / "isa" / "param",
+    ]
```
ThinkOpenly: Use the global `SCAN_ROOTS` here?

Although, as I commented above, we probably just want to look at everything under `spec`.

tools/scripts/test/test_check_param_enum_literals.py
```diff
@@ -0,0 +1,144 @@
...
+    def setUpClass(cls):
+        cls.mod = load_module()
+
+    def _write_tree(self, root: Path, param_yaml: str, csr_yaml: str) -> None:
+        param_dir = root / "spec" / "std" / "isa" / "param"
+        csr_dir = root / "spec" / "std" / "isa" / "csr" / "H"
+        param_dir.mkdir(parents=True)
+        csr_dir.mkdir(parents=True)
+        (root / "spec" / "std" / "isa" / "isa").mkdir(parents=True)
+        (root / "spec" / "std" / "isa" / "inst").mkdir(parents=True)
```
ThinkOpenly: Lots of repetition here. Also, since we'll be looking at everything under "spec" (per an earlier comment), you don't need this level of complexity. You could just drop everything under "root".

.github/workflows/regress.yml
```diff
@@ -863,6 +863,18 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone Github Repo Action
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - name: Set up mise environment
+        uses: ./.github/actions/mise-setup
+      - name: Check parameter enum string literals in IDL comparisons
+        run: ./bin/python tools/scripts/check_param_enum_literals.py
+      - name: Unit test parameter enum literal checker
+        run: ./bin/python -m unittest tools.scripts.test.test_check_param_enum_literals
```
ThinkOpenly, line 874: Not exactly sure whether to lump the unit test with the repo-wide test, but this is probably fine. It would make more sense to me to perform the unit test first. Then, if that fails, you don't waste time on the repo-wide test.

ThinkOpenly (changes requested):
Awesome, and thanks for taking this on! Great first attempt. I do have some comments inline...

tools/scripts/test/test_check_param_enum_literals.py
```diff
@@ -0,0 +1,193 @@
...
+        cls.mod = load_module()
+
+    def _write_fixture(self, root: Path, compare_body: str, param_yaml: str = PARAM_YAML) -> None:
+        """Minimal tree: one param + one file under spec/ with the compare."""
+        param_dir = root / "spec" / "std" / "isa" / "param"
+        other = root / "spec" / "std" / "isa" / "csr"
+        param_dir.mkdir(parents=True)
+        other.mkdir(parents=True)
+        (param_dir / "DEMO_ENUM.yaml").write_text(param_yaml, encoding="utf-8")
+        (other / "demo.yaml").write_text(
```
ThinkOpenly: Why not just use `root` here?
```suggestion
        (root / "DEMO_ENUM.yaml").write_text(param_yaml, encoding="utf-8")
        (root / "demo.yaml").write_text(
```

tools/scripts/test/test_check_param_enum_literals.py
```diff
@@ -0,0 +1,193 @@
...
+    def test_ruamel_loads_multiword_enum_members(self):
+        """Enums with spaces (always zero) must load via ruamel, not only simple tokens."""
+        with tempfile.TemporaryDirectory() as tmp:
+            root = Path(tmp)
+            param_dir = root / "spec" / "std" / "isa" / "param"
+            param_dir.mkdir(parents=True)
+            (param_dir / "DEMO_ENUM.yaml").write_text(PARAM_YAML, encoding="utf-8")
+            enums = self.mod.load_param_string_enums(param_dir)
+            self.assertIn("DEMO_ENUM", enums)
+            self.assertEqual(enums["DEMO_ENUM"], {"always zero", "custom"})
```
ThinkOpenly: This seems like overkill
```suggestion
```

tools/scripts/test/test_check_param_enum_literals.py
```diff
@@ -0,0 +1,193 @@
...
+                name: Sm
+            """
+        )
+        with tempfile.TemporaryDirectory() as tmp:
+            root = Path(tmp)
+            param_dir = root / "spec" / "std" / "isa" / "param"
+            param_dir.mkdir(parents=True)
+            (param_dir / "DEMO_NO_TYPE.yaml").write_text(yaml, encoding="utf-8")
+            enums = self.mod.load_param_string_enums(param_dir)
+            self.assertEqual(enums.get("DEMO_NO_TYPE"), {"retain", "custom"})
```
ThinkOpenly: same
```suggestion
```

tools/scripts/test/test_check_param_enum_literals.py
```diff
@@ -0,0 +1,193 @@
...
+    def test_missing_spec_dir_exits_2(self):
+        with tempfile.TemporaryDirectory() as tmp:
+            root = Path(tmp)
+            # param path missing too → still 2 (param checked first)
+            self.assertEqual(self.mod.main(["--root", str(root)]), 2)
+
+    def test_repo_paths_uses_globals_for_default_root(self):
+        param_dir, spec_root = self.mod.repo_paths(self.mod.ROOT)
+        self.assertEqual(param_dir, self.mod.PARAM_DIR)
+        self.assertEqual(spec_root, self.mod.SPEC_ROOT)
```
ThinkOpenly: same
```suggestion
```

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,199 @@
+#!/usr/bin/env python3
+# Copyright (c) 2026 titoatwork
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+"""Fail if IDL/YAML compares a string-enum parameter to a literal not in its enum.
+
+Parameter files under spec/std/isa/param declare legal string values with
```
ThinkOpenly:
```suggestion
Parameter files declare legal string values with
```

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,199 @@
...
+those parameters to string literals. A typo such as "always_zero" where the
+enum only lists "always zero" is silently wrong: the comparison is always true
+or always false and no schema check catches it.
+
+See https://github.com/riscv/riscv-unified-db/issues/2285
+
+Scope and known limits
+----------------------
+* Parameter enums are loaded with ruamel YAML (typ=safe) from
+  ``spec/std/isa/param/*.yaml``.
```
ThinkOpenly: Can be gleaned from the code, and I prefer not to put fragile path names in multiple places.
```suggestion
```

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,199 @@
...
+* Call sites are scanned under the entire ``spec/`` tree, restricted to ordinary
+  files whose names contain a dot (e.g. ``foo.yaml``), so extension-less junk is
+  skipped.
```
ThinkOpenly:
```suggestion
* The entire ``spec/`` tree is scanned, restricted to ordinary
  files whose names contain a dot (e.g. ``foo.yaml``).
```

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,199 @@
...
+from pathlib import Path
+
+from ruamel.yaml import YAML
+from ruamel.yaml.error import YAMLError
+
+ROOT = Path(__file__).resolve().parents[2]
+PARAM_DIR = ROOT / "spec" / "std" / "isa" / "param"
+SPEC_ROOT = ROOT / "spec"
+
+# Safe loader shared with other UDB Python tooling (see tools/python/udb.py).
```
ThinkOpenly:
```suggestion
```

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,199 @@
...
+import sys
+from pathlib import Path
+
+from ruamel.yaml import YAML
+from ruamel.yaml.error import YAMLError
+
+ROOT = Path(__file__).resolve().parents[2]
+PARAM_DIR = ROOT / "spec" / "std" / "isa" / "param"
+SPEC_ROOT = ROOT / "spec"
+
```
ThinkOpenly: Let's simplify this by specifying a default "root" parameter of "."... more below.
```suggestion
```

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,199 @@
...
+    Uses module globals PARAM_DIR / SPEC_ROOT when ``root`` is this checkout,
+    so the default path is not re-derived by string concatenation in main().
+    """
+    root = root.resolve()
+    if root == ROOT:
+        return PARAM_DIR, SPEC_ROOT
+    return root / "spec" / "std" / "isa" / "param", root / "spec"
+
+
```
ThinkOpenly: Simplify...
```suggestion
```

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,199 @@
...
+                bad.append((path, line_no, param, op, literal, line.strip()))
+    return bad
+
+
+def main(argv: list[str] | None = None) -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--root",
+        type=Path,
+        default=ROOT,
```
ThinkOpenly:
```suggestion
        default=".",
```

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,199 @@
...
+        help="Repository root (default: inferred from script location)",
+    )
+    args = parser.parse_args(argv)
+    root: Path = args.root.resolve()
+
+    param_dir, spec_root = repo_paths(root)
```
ThinkOpenly:
```suggestion
    spec_root = root / "spec"
    param_dir = spec_root / "std" / "isa" / "param"
```

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,199 @@
...
+    Uses module globals PARAM_DIR / SPEC_ROOT when ``root`` is this checkout,
+    so the default path is not re-derived by string concatenation in main().
+    """
+    root = root.resolve()
+    if root == ROOT:
+        return PARAM_DIR, SPEC_ROOT
+    return root / "spec" / "std" / "isa" / "param", root / "spec"
+
+
+# PARAM == "literal" or PARAM != 'literal'
```
ThinkOpenly:
```suggestion
# Regular expression to match either:
# - PARAM [op] "literal"
# - PARAM [op] 'literal'
# ... where "op" can be either "==" or "!=".
```

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,199 @@
...
+# group(3) = literal when double-quoted
+# group(4) = literal when single-quoted
+# (Only one of group(3)/group(4) is set per match.)
+#
+# Robustness: does not match "literal" == PARAM (parameter on the right).
+COMPARE_RE = re.compile(r"""\b([A-Z][A-Z0-9_]*)\s*(==|!=)\s*(?:"([^"]*)"|'([^']*)')""")
+
+
+def load_param_string_enums(param_dir: Path) -> dict[str, set[str]]:
+    """Map parameter name -> set of schema.enum string values (ruamel YAML)."""
```
ThinkOpenly:
```suggestion
    """Map parameter name -> set of schema.enum string values."""
```

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,199 @@
...
+        if not isinstance(doc, dict):
+            continue
+
+        # Only true parameter objects (defensive; param/ should be all kind: parameter).
+        if doc.get("kind") != "parameter":
+            continue
+
+        name = doc.get("name")
+        schema = doc.get("schema")
+        if not isinstance(name, str) or not isinstance(schema, dict):
```
ThinkOpenly: If name is not a string, we have bigger problems.
```suggestion
        if not isinstance(schema, dict):
```

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,199 @@
...
+            continue
+
+        enum_vals = schema.get("enum")
+        if not isinstance(enum_vals, list):
+            continue
```
ThinkOpenly: If `enum` is not a list, we have bigger problems.
```suggestion
```

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,199 @@
...
+    files: list[Path] = []
+    if not spec_root.is_dir():
+        return files
+
+    for path in spec_root.rglob("*"):
+        if not path.is_file():
+            continue
+        # ?*.?* style: require a dot in the filename so we skip extension-less paths.
+        if "." not in path.name:
+            continue
```
ThinkOpenly: I expected the fancy glob pattern to be used here
```suggestion
    for path in spec_root.rglob("?*.?*"):
        if not path.is_file():
            continue
```
Does that not work for you?

tools/scripts/check_param_enum_literals.py
```diff
@@ -0,0 +1,199 @@
...
+----------------------
+* Parameter enums are loaded with ruamel YAML (typ=safe) from
+  ``spec/std/isa/param/*.yaml``.
+* Call sites are scanned under the entire ``spec/`` tree, restricted to ordinary
+  files whose names contain a dot (e.g. ``foo.yaml``), so extension-less junk is
+  skipped.
+* Comparisons are matched only in the form ``PARAM op "literal"`` or
+  ``PARAM op 'literal'`` (parameter name on the left). The reverse form
+  ``"literal" op PARAM`` is legal IDL but is not checked; it is expected to be
+  rare or absent in this tree. Extend COMPARE_RE if that form appears.
```
ThinkOpenly: Add that only comparisons on a single line are tested. (We could probably eliminate that restriction by joining all of the file lines.)

ThinkOpenly (changes requested):

tools/scripts/check_param_enum_literals.py
ThinkOpenly: Oh, previously, you limited the files to "*.yaml", "*.idl", "*.isa". Any reason to remove that?

tools/scripts/test/test_check_param_enum_literals.py
titoatwork: Mostly yes, but not the two lines exactly as written. The loader reads parameters from root/spec/std/isa/param, so with both files directly under root it finds 0 parameters and 0 files to scan, main() returns 0, and the invalid-literal test would go green for the wrong reason. I checked that rather than guessing.

The rest of the point stands and I took it: the second file now goes straight into spec/, which drops one mkdir, one local variable and the dedent block around it.

tools/scripts/test/test_check_param_enum_literals.py
titoatwork: Removed.

tools/scripts/test/test_check_param_enum_literals.py
titoatwork: Removed. I left one line of comment where the enum is read, noting that parameters which set enum without schema.type are included deliberately, since a few real ones are, and that is easy to break by accident later.

tools/scripts/test/test_check_param_enum_literals.py
titoatwork: All four removed, including the one asserting the globals, which had nothing left to point at once they were deleted. Five tests remain.

tools/scripts/check_param_enum_literals.py
titoatwork: Taken.

tools/scripts/check_param_enum_literals.py
titoatwork: Removed. I took the same view a few lines further down and dropped the pointer to tools/python/udb.py for the same reason.

tools/scripts/check_param_enum_literals.py
titoatwork: The bullet list this belonged to is gone, so the sentence has nowhere to sit. The docstring is short prose now and keeps only the two limits a reader can actually trip over, single line comparisons and the parameter on the left. Say the word if you would rather keep the list and I will put it back in your wording.

tools/scripts/check_param_enum_literals.py
titoatwork: Gone.

tools/scripts/check_param_enum_literals.py
titoatwork: Done. The three module globals are deleted and --root defaults to ".".

tools/scripts/check_param_enum_literals.py
titoatwork: repo_paths deleted.

tools/scripts/check_param_enum_literals.py
titoatwork: Applied.

tools/scripts/check_param_enum_literals.py
titoatwork: Applied as written. I also dropped the separate spec_root.is_dir() check below it, since param_dir lives inside spec_root and is already checked, so the second one could never fire on its own.

tools/scripts/check_param_enum_literals.py
titoatwork: Applied.

tools/scripts/check_param_enum_literals.py
titoatwork: Applied.

tools/scripts/check_param_enum_literals.py
titoatwork: Applied. I checked the premise too: all 227 parameter files have a string name, so the guard was never doing anything.

tools/scripts/check_param_enum_literals.py
titoatwork: Agreed on the premise, no parameter file in the tree has a non-list enum. Deleting just those two lines does break it though: it leaves `enum_vals = schema.get("enum")` returning None for every parameter with no enum key, and the set comprehension underneath then raises TypeError. That is 157 of the 227 files, so effectively every run.

So the guard is gone and the read is `schema.get("enum", [])` inline at the point of use, which is what I think you were after anyway.

tools/scripts/check_param_enum_literals.py
titoatwork: It works fine, that one was just me being inconsistent. It is spec_root.rglob("?*.?*") with the is_file() check, as you wrote it.

tools/scripts/check_param_enum_literals.py
titoatwork: Added.

On eliminating the restriction by joining the lines: I tried it before deciding. Collapsing newlines and rescanning finds the same 217 comparisons as the line by line pass, so nothing in the tree spans lines today. Joining also costs the line number in the error, which is most of what makes the output useful. Documenting the limit seemed better than carrying the machinery for a case that has not appeared yet.

tools/scripts/check_param_enum_literals.py
titoatwork: There is, and I should have led with it instead of quietly dropping the filter.

The restriction loses four live comparisons, all on LRSC_MISALIGNED_BEHAVIOR, in spec/std/isa/inst/Zalrsc/lr.SIZE.AQRL.layout and sc.SIZE.AQRL.layout. Those two templates are what the generated lr.*.yaml and sc.*.yaml are built from, and the generated files open with "WARNING: This file is auto-generated from ... .layout".

So with the filter, a typo introduced in a template is invisible until somebody regenerates, and when it does surface it is reported against a file the header tells you not to edit. Without the filter it is reported against the template you actually have to fix.

I put a typo into lr.SIZE.AQRL.layout without regenerating, to be sure rather than to argue it: unfiltered the check exits 1 and names the layout file and line 141, filtered it exits 0 and says everything is fine.

The cost is 60 extra files, 2577 against 2517, which does not show up in the runtime. There is a test covering the .layout path so it does not regress quietly. Happy to put the filter back with .layout added if you would rather be explicit about the set.

Issue #2263, opened 2026-07-30
Add Python type checker


jordancarlin:
We should enable a Python type checker to complement the existing standard linting we use Ruff for. Seems like `pyrefly` is the most popular choice currently.

PR #2293, opened 2026-08-01
Add pyrefly static type checker alongside Ruff

sujalkurkute:
#2263 
#2278 
Changes:

Added pyrefly as a project dependency in pyproject.toml
Configured [tool.pyrefly] using the basic preset — a low-noise starting point suitable for a codebase adopting type checking for the first time
Excluded ext/** (matches the existing Ruff exclusion) and tools/mcp_gen_server/** (this tool depends on the mcp package, which isn't currently listed as a project dependency)
Added search-path entries for a few tool scripts (tools/python/auto-inst, backends/generators, tools/python) that import sibling modules at runtime rather than as installed packages, so pyrefly can resolve them correctly

Findings:
Running pyrefly check against this config surfaces one pre-existing issue:

tools/ruby-gems/udb/python/yaml_resolver.py:94 — passes an unexpected retrieve keyword argument to Registry(...)

I've left this as-is rather than fixing it in this PR, since it's out of scope for adding the tool itself. Happy to fix it here if preferred, or leave it for a follow-up — let me know which you'd rather have.

Testing:
Verified locally with Python 3.13 in a clean virtual environment; pyrefly check runs cleanly against the configured scope with only the one finding above.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,17 +6,16 @@ name = "riscv-unified-db"
 version = "0.1.0"
 requires-python = ">=3.12"
 dependencies = [
+    "check-jsonschema>=0.36.1",
     "deepmerge>=2.1.0",
     "jsonschema>=4.26.0",
+    "pycobertura>=4.1.0",
+    "pyrefly>=1.1.1",
+    "pytest>=9.0.2",
+    "reuse>=6.2.0",
     "ruamel-yaml>=0.19.1",
+    "ruff>=0.15.0",
     "tqdm>=4.67.3",
-]
-
-[dependency-groups]
-dev = [
-    "pytest==9.1.1",
-    "reuse==6.2.0",
-    "ruff==0.16.0",
 ]
 
 [tool.ruff]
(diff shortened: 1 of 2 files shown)


pyproject.toml
jordancarlin, line 1: Why is this changing so many other things in `pyproject.toml`? Revert the changes to the Ruff rules and restore the dev dependency group. `pyrefly` should be a dev dependency.

pyproject.toml
```diff
@@ -6,47 +6,59 @@ name = "riscv-unified-db"
...
+    "ext/**",
+    "tools/mcp_gen_server/**",  # optional MCP tool, 'mcp' package not in project deps
+]
+search-path = [
+    "tools/python/auto-inst",
+    "backends/generators",
+    "tools/python",
 ]
+python-version = "3.12"
+preset = "basic"
```
jordancarlin: Turning on basic mode for `pyrefly` is the almost the same as not enabling it at all. This disables all of the type checking, which is the whole reason we are adding it. We should be using `default`.

jordancarlin (changes requested):
It looks like something went wrong here. If you add something to `pyproject.toml`, `uv.lock` should automatically update.

jordancarlin (changes requested):
There are still a lot of changes to `pyproject.toml` that have nothing to do with adding `pyrefly`. Please remove all of those other changes.

PR #2294, opened 2026-08-01
Add pyright as Python type checker

### LINKED ISSUE #2263: Add Python type checker

lucifer4330k:
The repository has a Ruby type checker (Sorbet) wired into Rake and CI but no equivalent for the Python tooling under `tools/` and `backends/`. Adding pyright gives early signal on type errors in those packages and keeps the Python code base held to the same standard as Ruby.

This mirrors the existing Sorbet setup: a dev dependency plus `[tool.pyright]` config in `pyproject.toml`, a `test:python_typecheck` Rake task, and a `regress-python-typecheck` job registered in `tools/test/regress-tests.yaml`. The workflow file `.github/workflows/regress.yml` is regenerated from `regress-tests.yaml` by the existing `regress` pre-commit hook (via `tools/test/gen_regress.py`), so the new job appears there automatically and is included in the `regress-complete` dependency closure.

Pyright runs in `basic` mode and is scoped to `tools/` and `backends/`. `ext/` (git submodules), `gen/`, `.github/`, and `tools/mcp_gen_server` (which depends on the external `mcp` package not declared in `pyproject.toml`) are excluded so the check stays focused and unblocked.

The rollout surfaced 17 pre-existing type errors across three files, all fixed in the same change:

- `tools/python/auto-inst/parsing.py`: annotations on module-level `yaml_instructions` and `REPO_DIRECTORY`, a too-narrow `default` argument type on `safe_get`, and a `None` guard before `os.path.join(REPO_DIRECTORY, ...)`.
- `tools/python/auto-inst/test_parsing.py`: explicit `Any` / `str | None` annotations on the test-fixture globals.
- `tools/ruby-gems/udb/python/yaml_resolver.py`: switched the `deepmerge.Merger` import to the public `deepmerge.merger.Merger`, wrapped `os.getenv("UDB_ROOT")` in `str(...)` so the inferred type is strictly `str`, and added `assert overlay_dir is not None` for narrowing in `merge_file`.

Verified locally with `uv run pyright` (0 errors, 0 warnings) and `uv run ruff check` (clean).

Fixes: #2263

### DIFF
--- a/.github/workflows/regress.yml
+++ b/.github/workflows/regress.yml
@@ -373,6 +373,16 @@ jobs:
       - name: Run sorbet checks
         run: ./do test:sorbet
 
+  regress-python-typecheck:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone Github Repo Action
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - name: Set up mise environment
+        uses: ./.github/actions/mise-setup
+      - name: Run pyrefly checks
+        run: ./do test:python_typecheck
+
   regress-ext-explorer:
     runs-on: ubuntu-latest
     steps:
@@ -971,6 +981,7 @@ jobs:
       - regress-udb-gen-integration
       - regress-udb-gen-chore
       - regress-sorbet
+      - regress-python-typecheck
       - regress-ext-explorer
       - regress-idl-typecheck-smoke
       - regress-idl-typecheck-other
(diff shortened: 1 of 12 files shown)


jordancarlin (changes requested):
Thanks for opening this PR. I think we've decided to go with `pyrefly` instead of Pyright. See discussion in #2263 and #2293.

PR #2295, opened 2026-08-01
fix(inst): remove redundant encoding notes in Zfa quad-precision instructions

aashu2006:
Four Zfa quad-precision files each carried a sentence describing the instruction's encoding relative to a counterpart, and all four contradicted the `match:` string a few lines below them in the same file: `fltq.q` and `fleq.q` named each other's base instruction, and `fminm.q`/`fmaxm.q` cited bit 14 when the bit that actually differs is 13.


ThinkOpenly (changes requested):
I certainly appreciate the attention to detail here!

I feel that there is no need to describe these particulars of the encodings when the actual encoding is right below in the same file.

I ask that you instead delete all of the lines in question.

ThinkOpenly (approved):

PR #2299, opened 2026-08-01
Simplify M/Zmmul definedBy and add explicit M requires Zmmul

### LINKED ISSUE #2224: Make Zmmul a proper subset of M
M and Zmmul are defined as a standalone extensions.
Instructions that are part of Zmmul are "definedBy: anyOf: [M, Zmmul]"

**Expected value**
M should require Zmmul.
Zmmul instructions should be "definedBy: Zmmul".

amartyatatspandey:
M and Zmmul were defined as independent extensions, with M-only
instructions using `definedBy: anyOf: [M, Zmmul]` even though every M
Implementation includes Zmmul's multiply behavior.

This PR makes M depend on Zmmul, and makes the Zmmul-only instructions
(mul, mulh, mulhu, mulhsu, mulw) be defined by just Zmmul instead of the
anyOf.

Fixes #2224


spec/std/isa/ext/M.yaml
```diff
@@ -12,6 +12,10 @@ versions:
   - version: "2.0.0"
     state: ratified
     ratification_date: 2019-12
+    requirements:
+      extension:
+        name: Zmmul
+        version: "= 1.0.0"
```
ThinkOpenly: Let's relax this to "compatible with version 1.0.0"
```suggestion
        version: "~> 1.0.0"
```

ThinkOpenly (changes requested):
This looks great! Thanks for the contribution! I only have one nit in the code (see below). Also, I ask that you modify the PR description:
- remove the "Testing" section (leave "Fixes")
- remove the headings for "Problem" and "Fix"
- reduce the "Fix" section to just a few sentences (in your own words):
  - make M depend on Zmmul
  - make Zmmul instructions be defined by just Zmmul

ThinkOpenly (approved):

PR #2316, opened 2026-08-02
fix(param): generalize NUM_USABLE_PMP_ENTRIES constraint

### LINKED ISSUE #2313: NUM_USABLE_PMP_ENTRIES: requirement only enforces the NUM_PMP_ENTRIES==16 case
## Problem I came across

The description of `spec/std/isa/param/NUM_USABLE_PMP_ENTRIES.yaml` clearly states:

> "Must be between 0 and NUM_PMP_ENTRIES, inclusive."

However, the `requirements` block only enforces this for one specific value:

```yaml
requirements:
  idl(): |
    NUM_PMP_ENTRIES == 16 -> NUM_USABLE_PMP_ENTRIES <= 16;
  reason: Usable PMP entries cannot exceed the number of implemented PMP entries
```

When `NUM_PMP_ENTRIES == 64`, no constraint prevents `NUM_USABLE_PMP_ENTRIES` from being set to a value greater than 64, which the description prohibits. Its contradicting.

## Proposed Fix

Replace the specific case constraint with a general one:

```yaml
requirements:
  idl(): |
    NUM_USABLE_PMP_ENTRIES <= NUM_PMP_ENTRIES;
  reason: Usable PMP entries cannot exceed the number of implemented PMP entries
```

This covers all valid values of `NUM_PMP_ENTRIES` (0, 16, 64) without  needing separate cases.

## Reference

Privileged ISA Manual 3.7.1: implemented PMP entries that are hardwired to zero are still accessible but cannot guard a region.....the number of usable entries is always bounded by the number of implemented entries.

I can create  a PR for the same.

Princess0407:
The constraint associated with `NUM_PMP_ENTRIES` only enforces
an upper bound for a single value:
```
    NUM_PMP_ENTRIES == 16 -> NUM_USABLE_PMP_ENTRIES <= 16;
```

This leaves the `NUM_PMP_ENTRIES == 64` case unconstrained, contradicting
the description which states the value must be between 0 and
`NUM_PMP_ENTRIES` inclusive.

Replace with a general constraint that covers all valid values:

    NUM_USABLE_PMP_ENTRIES <= NUM_PMP_ENTRIES;

Also fix `long_name: TODO` -> "number of usable PMP entries".

Fixes #2313


Maanvi212006 (commented):
I walked through the diff, checked the linked issue matched the fix, verified the new constraint syntax matched patterns used elsewhere in the codebase, and confirmed the repo's existing IDL/schema tests would catch this on CI. Found no issues.

spec/std/isa/param/NUM_USABLE_PMP_ENTRIES.yaml
```diff
@@ -31,5 +31,5 @@ definedBy:
     greaterThan: 0
 requirements:
   idl(): |
-    NUM_PMP_ENTRIES == 16 -> NUM_USABLE_PMP_ENTRIES <= 16;
+    NUM_USABLE_PMP_ENTRIES <= NUM_PMP_ENTRIES;
```
ThinkOpenly, line 34: There is some interest in reducing instances where UDB specifies IDL when YAML would suffice. I thought this was an easy one, but a "param <= param" condition isn't directly expressible. Could you add this minimally sufficient YAML as a comment, in case we do want to convert later?

```
    # YAML version (minimally sufficient)
    #allOf:
    #  - if:
    #      param:
    #        name: NUM_PMP_ENTRIES
    #        equal: 16
    #    then:
    #      param:
    #        name: NUM_USABLE_PMP_ENTRIES
    #        lessThanOrEqual: 16
    #  - if:
    #      param:
    #        name: NUM_PMP_ENTRIES
    #        equal: 64
    #    then:
    #      param:
    #        name: NUM_USABLE_PMP_ENTRIES
    #        lessThanOrEqual: 64
```

PR #2330, opened 2026-08-03
fix(data): generalize Zicfiss sspush/sspopchk into register-constrained encodings

### LINKED ISSUE #779: Need to complete adoption of classified register names
Many instruction definition still have generic register names like `rd`, `rs1`, `rs2`.

In assembly syntax:
```
arch/inst/Zicfiss/ssrdp.yaml:assembly: rd
arch/inst/Zilsd/sd.yaml:assembly: rs2, offset(rs1)
arch/inst/Zilsd/ld.yaml:assembly: rd, offset(rs1)
arch/inst/Zfh/fsh.yaml:assembly: xs2, imm12($rs1)
arch/inst/Zfh/flh.yaml:assembly: xd, imm12($rs1)
arch/inst/Zfh/fli.h.yaml:assembly: rd, imm
arch/inst/C/c.addw.yaml:assembly: xd, rs2
arch/inst/C/c.subw.yaml:assembly: xd, rs2
arch/inst/C/c.and.yaml:assembly: xd, rs2
arch/inst/C/c.xor.yaml:assembly: xd, rs2
arch/inst/C/c.add.yaml:assembly: xd, rs2
arch/inst/C/c.or.yaml:assembly: xd, rs2
arch/inst/C/c.sub.yaml:assembly: xd, rs2
arch/inst/F/froundnx.s.yaml:assembly: fd, rs1, rm
```

And in field names (too numerous to list here):
```
$ grep -e '- name:' $(find arch/inst -name '*.yaml') | grep -E 'rd|rs1|rs2' | wc -l
1306
```

**Expected value**
UDB has adopted the convention of a leading `x` for general purpose register indices, `f` for floating-point, `v` for vector.

### LINKED ISSUE #1061: Can't describe instructions with overlapping match strings discriminated by "not" constraints
**Describe the bug**
There is a need to describe `sspush` and `sspopchk`, but their respective match strings ostensibly overlap:
```
sspush  : 1100111-----00000100000001110011
sspopchk: 110011011100-----100000001110011
```
But the `not` constraint for each distinguishes the two unambiguously, as the first string's variable must be 1 or 5, not 28 (`11100`) and the second string's variable must be 1 or 5, not 0 (`00000`).

**To Reproduce**
With `sspush` defined with:
```
encoding:
  match: "1100111-----00000100000001110011"
  variables:
    - name: xs2
      location: 24-20
      # prettier-ignore
      not: [ 0, 2, 3, 4, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31 ]
```
and `sspopchk` defined with:
```
encoding:
  match: "110011011100-----100000001110011"
  variables:
    - name: xs1
      location: 19-15
      # prettier-ignore
      not: [ 0, 2, 3, 4, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31 ]
```

```
/workspace/riscv-unified-db/backends/cpp_hart_gen/lib/decode_tree.rb:122:in `block in construct_decode_tree': undefined method `match?' for nil:NilClass (NoMethodError)

      if inst_format.reverse[cur_range].match?(/^[01]+$/)
                                       ^^^^^^^
```

**Additional context**
These two instructions are currently represented with incorrect mnemonics as in `sspush.x1` and `sspush.x5`.

iakshkhurana:
Replaces the four fixed-encoding Zicfiss shadow-stack stubs with a single
`sspush` and a single `sspopchk`. The register field is a proper encoding
variable (`xs2` at 24-20 for sspush, `xs1` at 19-15 for sspopchk) restricted to
`x1`/`x5` via `not:`, the assembly uses the generic register name, and the
long_name/description are register-neutral. The `hints` in `mop.rr.7`/`mop.r.28`
(layout templates and generated files) reference the merged definitions, and the
instruction appendix golden file and inst-table fixture are updated to match.

On the encoding-overlap concern: `Encoding#indistinguishable?` already takes
`not:` exclusions into account, so `sspopchk` vs `ssrdp` is distinguishable —
their only pattern intersection (xs1=0, xd=0) is excluded on both sides — and the
remaining containments (`sspopchk` within `mop.r.28`, `sspush` within `mop.rr.7`)
are hint-excused, same as before the rework.

Fixes #779

Signed-off-by: Aksh Khurana <khuranaaksh08@gmail.com>


ThinkOpenly (changes requested):
Thank you for working on this!

The implementation needs to be a little more complex.
- There should be a single of each "sspush.yaml" and "sspopchk.yaml".
- The respective names should be `sspush` and `sspopchk`.
- The long names should not include a specific register.
- Same for the descriptions.
- The assembly will have to use a generic register name (`xd` or `xs1`).
- The encoding strings will need "unknown/variable" bits for the register encoding.
- The variables will need to be defined, but restricted to the allowed values (`x1` and `x5`, so `not: [ 0, 2, 3, 4, 6, 7, ...`).
- The hints in "mop.r.28" and "mop.rr.7" will need to be updated.

Then, you need `./do test:inst_encodings` (at least) to succeed. I think this is where I got stuck. You end up with overlapping encodings _if_ the `not` attribute, above, is not taken into consideration. So, that part may need additional effort.

backends/cpp_hart_gen/lib/decode_tree.rb
```diff
@@ -236,7 +236,30 @@ def has_hints?(node, inst_list, xlen)
 
     return false unless node.insts.size == 1
 
-    !node.insts[0].hints.select { |hint_inst| hint_inst.defined_in_base?(xlen) && inst_list.include?(hint_inst) }.empty?
+    !node.insts[0].hints.select { |hint_inst| hint_inst.defined_in_base?(xlen) && inst_list.include?(hint_inst) && !dispatchable_hint?(hint_inst, xlen) }.empty?
```
ThinkOpenly: Due to this somewhat subtle change, should the name of the function be changed? `has_dispatchable_hints`? Maybe the whole concept of "dispatchable hint" should be renamed to "exclusion constrained hint"? So, `exclusion_constrained_hints`? All (defined) hints are ultimately dispatchable, right?

backends/cpp_hart_gen/lib/decode_tree.rb
```diff
@@ -298,7 +324,72 @@ def decode_c(encoding_var_name, xlen, inst_list, node = nil, indent = 0)
...
+      }
+      NEW_INST
+      # code += "#{' ' * (indent + 2)}return new #{tenv.name_of(:inst, @cfg_arch, node.insts[0].name)}<#{xlen}, SocType>(this, pc, #{encoding_var_name});\n"
+    end
+    code
+  end
+  private :decode_c
+
+  # @return [String] C++ condition that inst's defining extension is implemented
+  def implemented_cond_cxx(inst)
```
ThinkOpenly, line 393: This refactoring, while certainly helpful overall, makes the PR difficult to review. Could you restack your commits so that the refactoring is done in the first commit such that there are no functional changes, then a 2nd commit has the rest?

ThinkOpenly (changes requested):

backends/cpp_hart_gen/lib/decode_tree.rb
ThinkOpenly: should this be `@cfg_arch`, as Copilot suggested?

iakshkhurana: Yes — fixed in 4160f702 (the functional commit): the extracted helper now uses
`@cfg_arch`. It was latent before the refactor too, since that branch only runs
for parameter-conditioned instructions, which this path never hit.

backends/cpp_hart_gen/lib/decode_tree.rb
```diff
@@ -231,12 +231,37 @@ def extract_dv(dv, encoding_var_name)
...
+  def dv_allowed_cond(dv, encoding_var_name)
+    width = dv.encoding_fields.sum { |ef| ef.range.size }
+    allowed = (0...(1 << width)).to_a - dv.excludes
+    dv_val = extract_dv(dv, encoding_var_name)
+    if allowed.size <= dv.excludes.size
+      "(#{allowed.map { |val| "(#{dv_val} == #{val}_b)" }.join(' || ')})"
+    else
+      "(#{dv.excludes.map { |val| "(#{dv_val} != #{val}_b)" }.join(' && ')})"
+    end
   end
```
ThinkOpenly, line 265: @henrikg-qc, are you able to take a peek at this? I don't think anybody is a Ruby expert here, but this looks like where a decision is made how to represent (or not) in C++ a "hint" instruction that has exclusions (like "not valid with register `x0`").

backends/cpp_hart_gen/lib/decode_tree.rb
```diff
@@ -322,6 +351,29 @@ def decode_c(encoding_var_name, xlen, inst_list, node = nil, indent = 0)
...
+        conds << implemented_cond_cxx(hint_inst) if needs_to_check_implemented?(hint_inst)
+        code += <<~HINT_INST
+        if (#{conds.join(' && ')}) {
+            std::construct_at(
+              reinterpret_cast<#{tenv.name_of(:inst, @cfg_arch, hint_inst.name)}<#{xlen}, SocType>*>(inst),
+              this, pc, #{encoding_var_name}
+            );
+            return true;
+        }
+        HINT_INST
```
ThinkOpenly, line 374: @henrikg-qc more C++ stuff. Might be useful to see what's generated for a config that supports SSPUSH/SSPOPCHK.

spec/std/isa/inst/Zicfiss/sspopchk.yaml
```diff
@@ -0,0 +1,63 @@
...
+definedBy:
+  extension:
+    name: Zicfiss
+assembly: xs1
+encoding:
+  match: "110011011100-----100000001110011"
+  variables:
+    - name: xs1
+      location: 19-15
+      not:
```
jordancarlin, line 24: Might be worth eventually adding a key that lists only allowed values for cases like this. Not in this PR though.

ThinkOpenly (approved):
Full speed ahead

Issue #2327, opened 2026-08-03
Failures during autofix CI are silently causing trouble


ThinkOpenly:
**Describe the bug**
The "autofix" CI job [.github/workflows/autofix.yaml](blob/main/.github/workflows/autofix.yaml) runs `chore gen all`.

`chore gen all` removes all artifacts and regenerates them, to see if there are any differences that should be automatically added to the PR.

If there is a failure during this run, it is not detected, and there are bad side-effects, including `autofix` pushing commits into PRs that delete artifacts.

**To Reproduce**
`chore gen all` seems to fail in a number of ways:
- including a non-ASCII character in a non-quoted "long_name" field in a YAML file
- having an invalid requirement in a parameter YAML file
  PR #2316 had:
  ```
    idl(): |
      NUM_USABLE_PMP_ENTRIES <= NUM_PMP_ENTRIES;
  ```

**Text output/Error Messages**
There are error messages in the CI logs:
```
ℹ info    Compiling global IDL     
Running with 1 job(s)
⨯ error   Could not compile IDL constraint: 
NUM_USABLE_PMP_ENTRIES <= NUM_PMP_ENTRIES;
rake aborted!
SyntaxError: While parsing NUM_USABLE_PMP_ENTRIES <= NUM_PMP_ENTRIES; (SyntaxError)
:1:42
[... traceback ...]
```
...but `autofix` continues unimpeded. Once it thinks it is done generating all of the artifacts, it notices that some are missing, presumes that is expected, and creates a commit that deletes the missing files and adds it to the PR.

Princess0407:
I just read through the autofix workflow and the chore script. The root cause comes out to be because of these two things I guess:

1. `do_gen_all()` in `bin/chore` does not propagate failures between steps....if any sub function fails, execution continues to the next step.

2. `.github/workflows/autofix.yaml` runs `autofix-ci/action` unconditionally 
   after `chore gen all`, even if it exited with an error.

A potential fix can be to add `set -e` to `do_gen_all()` so it aborts on first failure, 
and add an `if: success()` to the autofix action step so it doesn't run after a 
failed generation.

Would this approach be appropriate?

ShivangSingh68:
I'd like to work on this issue if it's still available.

[...]

PR #2345, opened 2026-08-03
fix(ci): audit and fix error propagation gaps in bin/chore

Princess0407:
## Summary

Follow up to #2342  which fixed:
`gen all` and `profile-cfgs`/`cfg-headers`  failure propagation. 
This PR audits and fixes all the other remaining cases in `bin/chore`.

## Root Cause

`bin/chore` had two categories of silent failure:

1. **Helper functions** called external commands (`bundle exec`, `python`, 
   `node`, `ruby`, `cp`, `sed`) without `|| return 1`, swallowing failures.
2. **Subcommand dispatchers** used `exit 0` unconditionally after calling 
   helper functions, discarding any non-zero return code.

## Changes

| Function | Commands fixed |
|----------|---------------|
| `do_gen_regress()` | `python tools/test/gen_regress.py` |
| `do_gen_vscode_idl()` | `node idl-grammar-gen/index.js` |
| `do_gen_schema_docs()` | `ruby schema-docs-all` |
| `do_gen_gem_versions()` | `ruby gen_gem_versions.rb` (both branches) |
| `do_gen_udb_gen_fixtures()` | `do chore:udb_gen:update_fixtures` |
| `do_gen_smoke()` | All helper calls |
| `do_gen_xqci_golden()` | `ruby gen_xqci.rb`, `cp`, `sed` |
| `do_gen_instruction_appendix_golden()` | `do chore:update_golden_appendix` |
| `do_install_gems()` | `bundle install`, `do_ruby_type_def` |
| `do_ruby_type_def()` | `tapioca dsl`, `tapioca annotations` |

### Dispatchers — `exit 0` → `exit $?`:

All `gen`, `install`, `update`, and `build` subcommand dispatchers.

## Verification

- `bash -n bin/chore` passes
- `./bin/chore -h` runs without error

Fixes #2342 and addresses issue #2327

Signed-off-by: Priyanka Tiwari <princessss1404@gmail.com>


bin/chore
ThinkOpenly: This is very selective. Do the unchanged lines not have useful return codes?

Princess0407: Yeah youre right..I was selective. Every helper call in an orchestrator should propagate failure regardless of internal error handling. With the new commit I have updated `do_gen_all() ` and `do_gen_smoke()` to add `|| return 1` on all helper calls. And, pushed the same!!

PR #2395, opened 2026-08-05
fix(csr): sstateen0-3 take the stateen family's fixed length of 64

### LINKED ISSUE #2394: sstateen0-3 carry length: MXLEN instead of SXLEN
`sstateen0` through `sstateen3` (`0x10C`-`0x10F`) declare `length: MXLEN`. They are supervisor CSRs, so their width follows S-mode rather than M-mode, and the rest of the `stateen` family is already modelled that way.

## What the specification says

From `ext/riscv-isa-manual/src/priv/smstateen.adoc`, at the submodule commit this repository pins (`310a1114`). `norm:stateen_rv32_upper_bits_csrs` at line 65 lists the high-half CSRs that exist for RV32, and the supervisor registers are not among them:

> For RV32, there are CSR addresses for accessing the upper 32 bits of corresponding machine-level and hypervisor CSRs: `mstateen0h`, `mstateen1h`, `mstateen2h`, `mstateen3h`, `hstateen0h`, `hstateen1h`, `hstateen2h`, and `hstateen3h`.

Line 71 gives the reason they are absent:

> For the supervisor-level `sstateen` registers, high-half CSRs are not added at this time because it is expected the upper 32 bits of these registers will always be zeros, as explained later below.

A register that needs a high-half CSR on RV32 is 64 bits wide whatever XLEN is. One that does not is the width of its own mode. That is the distinction between `mstateen`/`hstateen` and `sstateen`, and it is the distinction `length:` exists to record.

## What the data already says

The other eight registers in the family are modelled correctly:

| Register | `length:` | high half in `spec/std/isa/csr/` |
|---|---|---|
| `mstateen0`-`mstateen3` | `64` | `mstateen0h`-`mstateen3h` |
| `hstateen0`-`hstateen3` | `64` | `hstateen0h`-`hstateen3h` |
| `sstateen0`-`sstateen3` | `MXLEN` | none, correctly |

So the four supervisor registers already agree with the specification about having no high half, and disagree with it about where their width comes from.

Twenty-one supervisor CSRs in the repository already use `length: SXLEN`, including `sstatus`, `satp`, `stvec`, `stval`, `scause`, `sepc`, `sscratch`, `sie` and `sip`. Five of those moved from `length: 64` to `SXLEN` in #2362 for the same reason these four should move from `MXLEN`.

## Where the two values differ

`MXLEN` and `SXLEN` diverge when `mstatus.SXL` makes S-mode narrower than M-mode. `cfgs/example_rv64_with_overlay.yaml:501` documents that as a supported configuration:

```yaml
  # XLENs supported in S-mode. Can be one of:
  #
  #  * 32:   SXLEN is always 32
  #  * 64:   SXLEN is always 64
  #  * [32, 64]: SXLEN can be changed (via mstatus.SXL) between 32 and 64
  SXLEN: [64]
```

On a configuration where they differ, the current declaration is wrong in two places.

`Csr#dynamic_length?` (`tools/ruby-gems/udb/lib/udb/obj/csr.rb:185`) answers `MXLEN` with `cfg_arch.mxlen.nil?`, because MXLEN cannot change at runtime. That is true of MXLEN and false of these registers: with `SXLEN: [32, 64]` the width does change at runtime, and the `SXLEN` branch four lines below already handles exactly that case. So the register is modelled as fixed while S-mode is switching width underneath it.

`sstateen1`, `sstateen2` and `sstateen3` declare their `DATA` field as `location_rv64: 63-0` with `location_rv32: 31-0`. `CsrField#width` selects `location_rv#{effective_xlen}`, so with the width pinned to MXLEN the 64-bit form is chosen for a register S-mode accesses as 32 bits. `sstateen0` needs no field change, since `C`, `FCSR` and `JVT` sit at bits 0, 1 and 2.

## Scope, and what I am not claiming

No configuration in `cfgs/` currently sets `SXLEN` to anything other than `MXLEN`, so nothing in the repository observes the difference today. This is a correctness fix to the model rather than a repair of an output that is presently wrong, and I would rather say so than imply a live failure.

I am also not claiming anything about the VS-mode branch in these registers' `sw_read()`, which masks against `hstateen` when `mode() == PrivilegeMode::VS`. That is a separate question from the declared width and I have not analysed it.

One note on the `Bits<64>` locals in those `sw_read()` bodies, since it is the objection that deferred `senvcfg` out of #2362. It does not constrain the declared width, and these four registers are the precedent I cited in #2364 for pairing a variable `length:` with a fixed 64-bit read body. Moving them to `SXLEN` keeps that precedent and makes it an exact parallel to the `senvcfg` case in #2384 rather than an approximate one.

## Why my earlier sweep missed these

The sweep behind #2361 matched supervisor and VS-mode CSRs carrying a literal `length: 64` with no high-half companion. These four match neither half of that: the width is symbolic, and the absence of a high half is correct. Widening the check to "supervisor CSR whose width does not come from `SXLEN`" is what surfaced them.

titoatwork:
`sstateen0` through `sstateen3` carry `length: MXLEN`, which sizes a supervisor CSR by a mode that does not define its width. They now take the `*stateen*` family's fixed `length: 64`. `hstateen0`-`hstateen3` are already `priv_mode: S` at a fixed 64, so this is an established shape rather than a new one; eight supervisor-visible CSRs use it today.

The `DATA` field on `sstateen1`, `sstateen2` and `sstateen3` collapses to `location: 31-0`, leaving bits 63-32 undefined and read-only zero. That is what the description already in these files says: "4 * 64 = 256 bits for machine and hypervisor levels, and 4 * 32 = 128 bits for supervisor level", and the upper 32 bits of an `mstateen` CSR control state "inherently inaccessible to user level, so no corresponding enable bits in the supervisor-level `sstateen` CSR are applicable". The absence of `sstateen0h`-`sstateen3h` in the tree agrees. A range narrower than its fixed-width register is the ordinary shape here: 291 fields already do it.

Keeping the `location_rv32`/`location_rv64` pair inside a fixed-width register would instead have made these the only CSRs of their kind. Of the 461 CSRs under `spec/`, the 137 with an XLEN-dependent field location all have a symbolic `length`, and none of the 288 with a fixed integer `length` do.

Closes #2394

### DIFF
--- a/spec/std/isa/csr/sstateen0.yaml
+++ b/spec/std/isa/csr/sstateen0.yaml
@@ -9,7 +9,7 @@ name: sstateen0
 long_name: Supervisor State Enable 0 Register
 address: 0x10C
 priv_mode: S
-length: SXLEN
+length: 64
 writable: true
 description: |
   Each bit of a `stateen` CSR controls less-privileged access to an extension’s state,
--- a/spec/std/isa/csr/sstateen0.yaml
+++ b/spec/std/isa/csr/sstateen0.yaml
@@ -9,7 +9,7 @@ name: sstateen0
 long_name: Supervisor State Enable 0 Register
 address: 0x10C
 priv_mode: S
-length: MXLEN
+length: SXLEN
 writable: true
 description: |
   Each bit of a `stateen` CSR controls less-privileged access to an extension’s state,
--- a/spec/std/isa/csr/sstateen1.yaml
+++ b/spec/std/isa/csr/sstateen1.yaml
@@ -9,7 +9,7 @@ name: sstateen1
 long_name: Supervisor State Enable 1 Register
 address: 0x10D
 priv_mode: S
-length: MXLEN
+length: SXLEN
 writable: true
 description: |
   Each bit of a `stateen` CSR controls less-privileged access to an extension’s state,
--- a/spec/std/isa/csr/sstateen1.yaml
+++ b/spec/std/isa/csr/sstateen1.yaml
@@ -9,7 +9,7 @@ name: sstateen1
 long_name: Supervisor State Enable 1 Register
 address: 0x10D
 priv_mode: S
-length: SXLEN
+length: 64
 writable: true
 description: |
   Each bit of a `stateen` CSR controls less-privileged access to an extension’s state,
@@ -61,8 +61,7 @@ definedBy:
 
 fields:
   DATA:
-    location_rv64: 63-0
-    location_rv32: 31-0
+    location: 31-0
     description: Data value
     type: RW
     sw_write(csr_value): |
--- a/spec/std/isa/csr/sstateen2.yaml
+++ b/spec/std/isa/csr/sstateen2.yaml
@@ -9,7 +9,7 @@ name: sstateen2
 long_name: Supervisor State Enable 2 Register
 address: 0x10E
 priv_mode: S
-length: SXLEN
+length: 64
 writable: true
 description: |
   Each bit of a `stateen` CSR controls less-privileged access to an extension’s state,
@@ -61,8 +61,7 @@ definedBy:
 
 fields:
   DATA:
-    location_rv64: 63-0
-    location_rv32: 31-0
+    location: 31-0
     description: Data value
     type: RW
     sw_write(csr_value): |
--- a/spec/std/isa/csr/sstateen2.yaml
+++ b/spec/std/isa/csr/sstateen2.yaml
@@ -9,7 +9,7 @@ name: sstateen2
 long_name: Supervisor State Enable 2 Register
 address: 0x10E
 priv_mode: S
-length: MXLEN
+length: SXLEN
 writable: true
 description: |
   Each bit of a `stateen` CSR controls less-privileged access to an extension’s state,
--- a/spec/std/isa/csr/sstateen3.yaml
+++ b/spec/std/isa/csr/sstateen3.yaml
@@ -9,7 +9,7 @@ name: sstateen3
 long_name: Supervisor State Enable 3 Register
 address: 0x10F
 priv_mode: S
-length: MXLEN
+length: SXLEN
 writable: true
 description: |
   Each bit of a `stateen` CSR controls less-privileged access to an extension’s state,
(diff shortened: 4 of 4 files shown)


ThinkOpenly (approved):

spec/std/isa/csr/sstateen0.yaml
```diff
@@ -9,7 +9,7 @@ name: sstateen0
 long_name: Supervisor State Enable 0 Register
 address: 0x10C
 priv_mode: S
-length: MXLEN
+length: SXLEN
```
ThinkOpenly:
```suggestion
length: 64
```

spec/std/isa/csr/sstateen1.yaml
```diff
@@ -9,7 +9,7 @@ name: sstateen1
 long_name: Supervisor State Enable 1 Register
 address: 0x10D
 priv_mode: S
-length: MXLEN
+length: SXLEN
```
ThinkOpenly:
```suggestion
length: 64
```

spec/std/isa/csr/sstateen2.yaml
```diff
@@ -9,7 +9,7 @@ name: sstateen2
 long_name: Supervisor State Enable 2 Register
 address: 0x10E
 priv_mode: S
-length: MXLEN
+length: SXLEN
```
ThinkOpenly:
```suggestion
length: 64
```

spec/std/isa/csr/sstateen3.yaml
```diff
@@ -9,7 +9,7 @@ name: sstateen3
 long_name: Supervisor State Enable 3 Register
 address: 0x10F
 priv_mode: S
-length: MXLEN
+length: SXLEN
```
ThinkOpenly:
```suggestion
length: 64
```

ThinkOpenly (changes requested):
After thinking about this a bit, I think we should follow the lead of the `mstateen*` and `hstateen*` CSRs. The length of the `sstateen*` CSRs is not terribly important at this time or in the foreseeable future, but the most consistent implementation would be to match the other `*stateen*` CSRs at a fixed 64 bits.

ThinkOpenly (approved):

PR #2419, opened 2026-08-06
fix(isa): c.sdsp must use its full 5-bit register index

### LINKED ISSUE #2418: `c.sdsp` maps its 5-bit `rs2` through `creg2reg`, so it stores from the wrong register
`c.sdsp` uses the CSS format, in which `rs2` is a full 5-bit register specifier. `spec/std/isa/inst/C/c.sdsp.yaml` declares it that way and then passes it to `creg2reg`, which takes a 3-bit compressed index and can only produce `x8`-`x15`. Both the RV64 (`Zca`) and RV32 (`Zclsd`) paths do this, so on RV64 `c.sdsp xs2, offset(sp)` stores from `x8 + (xs2 & 7)` instead of from `xs2`.

## What the specification says

`ext/riscv-isa-manual/src/unpriv/images/wavedrom/c-sp-load-store-css.edn` draws C.SWSP, C.SDSP, C.FSWSP and C.FSDSP as one CSS group sharing a single 5-bit `rs2`:

```
{bits: 5, name: 'rs2',    attr: ['5', 'src', 'src', 'src', 'src']},
{bits: 6, name: 'imm',    attr: [...]},
{bits: 3, name: 'funct3', attr: ['3','C.SWSP', 'C.SDSP', 'C.FSWSP', 'C.FSDSP']},
```

`images/bytefield/rvc-instr-quad2.edn:103` draws the same field for C.SDSP (RV64) as `(draw-box "rs2" {:span 5})`.

For the Zclsd form, `src/unpriv/zclsd.adoc` gives `{bits: 5, name: 'rs2', attr: ['src, src[0]=0']}`, and line 15 restricts only the CS-format pair: "insn:c.ld[] and insn:c.sd[] instructions can only use `x8`-`x15`". No such restriction is stated for `c.sdsp`.

## What the model does

`c.sdsp.yaml` declares `xs2` at `location: 6-2` in both encodings, five bits wide:

```yaml
  RV32:
    variables:
      - name: xs2
        location: 6-2
        not: [1, 3, 5, 7, 9, 11, 13, 15, 17, 19, 21, 23, 25, 27, 29, 31]
  RV64:
    variables:
      - name: xs2
        location: 6-2
```

and then routes it through `creg2reg` in both branches:

```idl
  if (xlen() == 32) {
    if (implemented?(ExtensionName::Zclsd)) {
      Bits<64> data = {X[creg2reg(xs2) + 1], X[creg2reg(xs2)]};
      write_memory(64, virtual_address, data, $encoding);
    ...
  } else {
    write_memory(64, virtual_address, X[creg2reg(xs2)], $encoding);
  }
```

`creg2reg` is declared for a three-bit index (`spec/std/isa/isa/globals.isa:348-350`) and its body is

```idl
    return {2'b01, creg_idx};
```

so its range is exactly `x8`-`x15`. A `c.sdsp x20, 0(sp)` therefore stores from `x12`.

## The sibling shows the intended idiom

`c.swsp.yaml` has the identical CSS field, `xs2` at `location: 6-2`, and writes:

```idl
  write_memory(32, virtual_address, X[xs2][31:0], $encoding);
```

`c.sw.yaml`, which is CS format with a genuine 3-bit field, correctly uses `X[creg2reg(xs2)]`. `spec/std/isa/inst/I/sd.yaml` builds its RV32 pair as `{X[xs2 + 1], X[xs2]}`.

## Scope

I enumerated every `creg2reg` call site under `spec/std/isa/inst/` and compared each argument against the declared bit width of the variable it names: 75 call sites across 33 files. 72 pass a 3-bit field. The three that do not are all in `c.sdsp.yaml`, and it is the only file in the tree with the mismatch.

## Why the type checker does not reject it

`Type#convertable_to?` in `tools/ruby-gems/idlc/lib/idlc/type.rb:291-296` returns

```ruby
      when :bits
        if type.kind == :enum_ref
          warn "You seem to be missing an $enum cast"
          return false
        end
        return type.kind == :bits || type.kind == :bitfield
```

for a `:bits` source, without comparing widths, and `FunctionCallExpressionAst` checks arguments with `convertable_to?` (`ast.rb:7868`). So a `Bits<5>` argument to a `Bits<3>` parameter passes type checking silently.

## Reachability

This one is generated today rather than latent. `c.sdsp` is `definedBy: anyOf: [allOf: [xlen: 64, Zca], Zclsd]`, and three configurations in `cfgs/` implement `Zca` at `MXLEN: 64`: `rv64-riscv-tests.yaml`, `rv64-vector.yaml` and `example_rv64_with_overlay.yaml`. The RV32 branch is unreachable, since no configuration implements `Zclsd` or `Zilsd`.

## Suggested fix

Drop `creg2reg` from both branches, matching `c.swsp` and `I/sd.yaml`:

```idl
      Bits<64> data = {X[xs2 + 1], X[xs2]};
...
    write_memory(64, virtual_address, X[xs2], $encoding);
```

The encodings, `definedBy` and `not:` lists are already correct and need no change.

## What I am not claiming

- That the two replacement lines compile. No `idlc` here, so this is worth an actual build.
- Anything about the `x0` rule at `zclsd.adoc:13`, that `c.sdsp` from `x0` stores a zero operand without accessing `x1`. That is unimplemented here as it is in `sd`, and it belongs with the register-pair work under #2282 rather than with a register-index fix.
- Anything about `c.ld`, `c.sd` or `c.ldsp`, whose `creg2reg` use is correct for their formats.

titoatwork:
`c.sdsp` is a CSS-format instruction, so its `rs2` is a full 5-bit register specifier, and `c.sdsp.yaml` declares it that way (`location: 6-2`). Both branches then pass that field to `creg2reg`, which takes a 3-bit compressed index and returns `{2'b01, creg_idx}`. Its range is `x8`-`x15`, so on RV64 `c.sdsp x20, 0(sp)` stores from `x12`.

Drop `creg2reg` from both branches.

Closes #2418

### DIFF
--- a/spec/std/isa/inst/C/c.sdsp.yaml
+++ b/spec/std/isa/inst/C/c.sdsp.yaml
@@ -8,7 +8,7 @@ kind: instruction
 name: c.sdsp
 long_name: Store doubleword to stack
 description: |
-  Stores a 64-bit value in register rs2 to memory.
+  Stores a 64-bit value in register xs2 to memory.
   It computes an effective address by adding the zero-extended offset, scaled by 8,
   to the stack pointer, x2.
   It expands to `sd` `xs2, offset(x2)`.
--- a/spec/std/isa/inst/C/c.sdsp.yaml
+++ b/spec/std/isa/inst/C/c.sdsp.yaml
@@ -53,11 +53,11 @@ operation(): |
 
   if (xlen() == 32) {
     if (implemented?(ExtensionName::Zclsd)) {
-      Bits<64> data = {X[creg2reg(xs2) + 1], X[creg2reg(xs2)]};
+      Bits<64> data = {X[xs2 + 1], X[xs2]};
       write_memory(64, virtual_address, data, $encoding);
     } else {
       raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
     }
   } else {
-    write_memory(64, virtual_address, X[creg2reg(xs2)], $encoding);
+    write_memory(64, virtual_address, X[xs2], $encoding);
   }
--- a/tests/golden/all_instructions.golden.adoc
+++ b/tests/golden/all_instructions.golden.adoc
@@ -14358,7 +14358,7 @@ RV64::
 ....
 
 Description::
-Stores a 64-bit value in register rs2 to memory.
+Stores a 64-bit value in register xs2 to memory.
 It computes an effective address by adding the zero-extended offset, scaled by 8,
 to the stack pointer, x2.
 It expands to xref:insts:sd.adoc#udb:doc:inst:sd[sd] `xs2, offset(x2)`.


spec/std/isa/inst/C/c.sdsp.yaml
ThinkOpenly: While you are here, could you fix this?
```suggestion
  Stores a 64-bit value in register xs2 to memory.
```

ThinkOpenly (approved):

spec/std/isa/inst/C/c.sdsp.yaml
titoatwork: Done in bbcc681b. Thanks for catching it, that was the only remaining rs2 in the file, the other seven references were already xs2.

ThinkOpenly (approved):

PR #2435, opened 2026-08-06
fix(isa): remove misa guards that gate nothing, correct the rest

### LINKED ISSUE #1175: MISA extension bits mean "supported" and "maybe?"
From the spec:
> Although one of the bits 25—0 in `misa` being set to 1 implies that the corresponding feature is implemented, the inverse is not necessarily true: one of these bits being clear does not necessarily imply that the corresponding feature is not implemented.

Using `misa.B` as an example:
> When the "B" bit is 1, the implementation supports the instructions provided by the Zba, Zbb, and Zbs extensions. When the "B" bit is 0, it indicates that the implementation might not support one or more of the Zba, Zbb, or Zbs extensions.

So, I don't think we can rely on `misa` to unconditionally raise an illegal instruction error, as in `Zba/add.uw.yaml`:
```
operation(): |
  if (implemented?(ExtensionName::B) && (CSR[misa].B == 1'b0)) {
  }
```

### LINKED ISSUE #2434: misa-gated illegal-instruction checks do not honor `MISA_CSR_IMPLEMENTED`
## The rule

A `misa` that reads zero does not mean "no extensions are implemented". It means `misa` itself is not implemented, and the ISA has to be discovered some other way. From the privileged specification:

> This register must be readable in any implementation, but a value of zero can be returned to indicate the `misa` register has not been implemented, requiring that CPU capabilities be determined through a separate non-standard mechanism.

UDB models this with `MISA_CSR_IMPLEMENTED`, and the parameter's own description spells out the consequence (`spec/std/isa/param/MISA_CSR_IMPLEMENTED.yaml`):

```yaml
  false::
    The `misa` CSR is read-only-0.
```

So a check of the form "extension bit is 0, therefore raise `IllegalInstruction`" is only meaningful when `misa` is implemented. Otherwise it reads a zero that carries no architectural claim, and traps an instruction the hart supports.

## The idiom is already in the tree, in two independent places

`spec/std/isa/csr/mstatus.yaml` consults `misa` only when `misa` exists, in both polarities:

```
426:  if (implemented?(ExtensionName::F) && (!MISA_CSR_IMPLEMENTED || CSR[misa].F == 1'b1)) {
429:  } else if (MISA_CSR_IMPLEMENTED && (CSR[misa].S == 1'b0) && (CSR[misa].F == 1'b0)) {
```

`spec/std/isa/inst/Zaamo/` and `spec/std/isa/inst/Zabha/` do the same across 153 sites, all spelled:

```
if (implemented?(ExtensionName::A) && MISA_CSR_IMPLEMENTED && CSR[misa].A == 1'b0) {
```

## The defect

139 of the 292 misa-gated illegal-instruction checks omit the guard:

```
if (implemented?(ExtensionName::B) && (CSR[misa].B == 1'b0)) {
```

| directory | sites | gates on | source |
|---|---:|---|---|
| `C` | 23 | `misa.C` | hand-written |
| `Zalrsc` | 18 | `misa.A` | 2 `.layout` + 16 generated |
| `Zcb` | 17 | `misa.B/C/M` | hand-written |
| `Zbb` | 14 | `misa.B` | hand-written |
| `M` | 13 | `misa.M` | hand-written |
| `B` | 12 | `misa.B` | hand-written |
| `Zba` | 8 | `misa.B` | hand-written |
| `Zbs` | 8 | `misa.B` | hand-written |
| `Xqccmp` (custom, `qc_iu`) | 7 | `misa.C` | hand-written |
| `Zcd` | 6 | `misa.C/D` | hand-written |
| `Zcf` | 6 | `misa.C/F` | hand-written |
| `Zcmp` | 6 | `misa.C` | hand-written |
| `Zbc` | 1 | `misa.B` | hand-written |
| total | 139 | | |

The clearest case is `misa.A`, which is gated both ways for the same extension. `Zaamo` and `Zabha` guard it across 153 sites, and `Zalrsc` does not, across 18. `Zaamo` and `Zalrsc` are the two halves of `A`, reading the same bit for the same purpose, so one of the two spellings is wrong. It is not the one used 153 times and independently in `mstatus`.

## Is it live

No, this one is latent. All seven `fully configured` configs set `MISA_CSR_IMPLEMENTED: true`, so every gate currently evaluates as if the guard were present. The remaining six configs are `unconfigured` or `partially configured` and pin no value.

A config that sets it `false` gets a hart where every `B`, `C`, `M`, `Zb*`, `Zc*` and `Zalrsc` instruction raises `IllegalInstruction`, while the `Zaamo` and `Zabha` instructions beside them execute normally.

## A second, separate observation

`spec/std/isa/csr/misa.yaml` never references `MISA_CSR_IMPLEMENTED`. There are zero occurrences and no CSR-level `sw_read()`, so every field still resets to `implemented?(ExtensionName::X) ? 1 : 0` regardless of the parameter. Repo-wide the only consumers are `mstatus.yaml`, the 153 guarded instruction sites, the configs, and two C++ test fixtures. No backend code reads it.

So today, setting `MISA_CSR_IMPLEMENTED: false` does not actually make `misa` read-only-0, contrary to the parameter's description.

That looks worth fixing separately, and the ordering matters: this guard fix would need to land first. If `misa` were made to read zero while 139 sites still consult it unguarded, the latent defect above would become live in exactly the configuration the change is meant to support. Happy to file it as its own issue if you want it tracked.

## Reproduction

```bash
# total misa-gated illegal-instruction checks
git grep -h "CSR\[misa\]" origin/main -- 'spec/*/isa/**/inst/**' 'spec/*/isa/inst/**' \
  | grep "== 1'b0" | grep -c "implemented?(ExtensionName::"
# 292

# of those, the ones missing the guard
git grep -h "CSR\[misa\]" origin/main -- 'spec/*/isa/**/inst/**' 'spec/*/isa/inst/**' \
  | grep "== 1'b0" | grep "implemented?(ExtensionName::" | grep -vc "MISA_CSR_IMPLEMENTED"
# 139
```

Measured on `origin/main` at `e909b6c9`.

## Proposed fix

Add `MISA_CSR_IMPLEMENTED &&` to the 139 sites, matching the existing spelling. On every config in the tree this is a no-op, since `X && true` is `X`, and it introduces no new parameter dependency, because instruction IDL already references this parameter in 153 places.

PR to follow.

titoatwork:
Of the 292 misa-gated illegal-instruction checks, 224 test a bit whose clearing does not make the instruction unimplemented, and are removed: all 171 `misa.A` sites, all 47 `misa.B`, and the 6 `misa.M` sites on instructions that are `definedBy: Zmmul`. Zalrsc and Zaamo are independently implementable, as are Zba, Zbb and Zbs, so clearing the bit leaves them implemented.

The remaining 68 keep the guard and are brought into line: they check `MISA_CSR_IMPLEMENTED` before reading the bit, and call `reserved_instruction()` instead of raising directly. Those are 56 for `misa.C`, which `ialign()` and the four `*epc` registers also read, the 8 division instructions that are `definedBy: M`, and 4 for `misa.F` and `misa.D`. `reserved_instruction()` honours `TRAP_ON_UNIMPLEMENTED_INSTRUCTION`, which `Ssstrict` requires to be true, so routing through it is what lets `Ssstrict` govern these traps.

`misa.yaml` loses the `misa.A` and `misa.B` sentences claiming that writing 0 raises `IllegalInstruction`, since neither bit gates anything now, and the `misa.M` sentence is narrowed to divide and remainder. No behaviour change in the tree: all 38 `MUTABLE_MISA_*` assignments in `cfgs/` are `false`. The 23 `misa.S` and `misa.H` sites test `== 1` to select behaviour rather than to trap, so they remain a separate question.

Closes #2434
Closes #1175

### DIFF
--- a/spec/std/isa/inst/B/andn.yaml
+++ b/spec/std/isa/inst/B/andn.yaml
@@ -32,8 +32,4 @@ access:
   vu: always
 data_independent_timing: true
 operation(): |
-  if (implemented?(ExtensionName::B) && MISA_CSR_IMPLEMENTED && (CSR[misa].B == 1'b0)) {
-    reserved_instruction($encoding);
-  }
-
   X[xd] = X[xs1] & ~X[xs2];
--- a/spec/std/isa/inst/B/andn.yaml
+++ b/spec/std/isa/inst/B/andn.yaml
@@ -32,7 +32,7 @@ access:
   vu: always
 data_independent_timing: true
 operation(): |
-  if (implemented?(ExtensionName::B) && (CSR[misa].B == 1'b0)) {
+  if (implemented?(ExtensionName::B) && MISA_CSR_IMPLEMENTED && (CSR[misa].B == 1'b0)) {
     raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
   }
 
--- a/spec/std/isa/inst/B/andn.yaml
+++ b/spec/std/isa/inst/B/andn.yaml
@@ -33,7 +33,7 @@ access:
 data_independent_timing: true
 operation(): |
   if (implemented?(ExtensionName::B) && MISA_CSR_IMPLEMENTED && (CSR[misa].B == 1'b0)) {
-    raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
+    reserved_instruction($encoding);
   }
 
   X[xd] = X[xs1] & ~X[xs2];
--- a/spec/std/isa/inst/Zalrsc/lr.d.aq.yaml
+++ b/spec/std/isa/inst/Zalrsc/lr.d.aq.yaml
@@ -73,7 +73,7 @@ access:
   vs: always
   vu: always
 operation(): |
-  if (implemented?(ExtensionName::A) && (CSR[misa].A == 1'b0)) {
+  if (implemented?(ExtensionName::A) && MISA_CSR_IMPLEMENTED && (CSR[misa].A == 1'b0)) {
     # even though this is a memory operation, the exception occurs before that would be known,
     # so mode() is the correct reporting mode rathat than effective_ldst_mode()
     raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
--- a/spec/std/isa/inst/Zalrsc/lr.d.aq.yaml
+++ b/spec/std/isa/inst/Zalrsc/lr.d.aq.yaml
@@ -74,9 +74,7 @@ access:
   vu: always
 operation(): |
   if (implemented?(ExtensionName::A) && MISA_CSR_IMPLEMENTED && (CSR[misa].A == 1'b0)) {
-    # even though this is a memory operation, the exception occurs before that would be known,
-    # so mode() is the correct reporting mode rathat than effective_ldst_mode()
-    raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
+    reserved_instruction($encoding);
   }
 
   XReg virtual_address = X[xs1];
--- a/spec/std/isa/inst/Zalrsc/lr.d.aq.yaml
+++ b/spec/std/isa/inst/Zalrsc/lr.d.aq.yaml
@@ -73,10 +73,6 @@ access:
   vs: always
   vu: always
 operation(): |
-  if (implemented?(ExtensionName::A) && MISA_CSR_IMPLEMENTED && (CSR[misa].A == 1'b0)) {
-    reserved_instruction($encoding);
-  }
-
   XReg virtual_address = X[xs1];
 
   if (!is_naturally_aligned(64, virtual_address)) {
--- a/spec/custom/isa/qc_iu/inst/Xqccmp/qc.cm.mva01s.yaml
+++ b/spec/custom/isa/qc_iu/inst/Xqccmp/qc.cm.mva01s.yaml
@@ -28,7 +28,7 @@ access:
   vu: always
 operation(): |
   if (implemented?(ExtensionName::Zcmp) && MISA_CSR_IMPLEMENTED && (CSR[misa].C == 1'b0)) {
-    raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+    reserved_instruction($encoding);
   }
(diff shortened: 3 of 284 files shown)


spec/std/isa/inst/B/andn.yaml
```diff
@@ -32,8 +32,8 @@ access:
   vu: always
 data_independent_timing: true
 operation(): |
-  if (implemented?(ExtensionName::B) && (CSR[misa].B == 1'b0)) {
-    raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
+  if (implemented?(ExtensionName::B) && MISA_CSR_IMPLEMENTED && (CSR[misa].B == 1'b0)) {
+    reserved_instruction($encoding);
```
ThinkOpenly: Yeah, Jordan's right: `misa.B` isn't directly associated with any instructions. Setting `misa.B` to 0 has no architectural impact.

Remove all tests where any behavior is associated with testing `misa.B`.

spec/std/isa/inst/Zalrsc/lr.d.aq.yaml
```diff
@@ -73,10 +73,8 @@ access:
   vs: always
   vu: always
 operation(): |
-  if (implemented?(ExtensionName::A) && (CSR[misa].A == 1'b0)) {
-    # even though this is a memory operation, the exception occurs before that would be known,
-    # so mode() is the correct reporting mode rathat than effective_ldst_mode()
-    raise (ExceptionCode::IllegalInstruction, mode(), $encoding);
+  if (implemented?(ExtensionName::A) && MISA_CSR_IMPLEMENTED && (CSR[misa].A == 1'b0)) {
+    reserved_instruction($encoding);
   }
```
ThinkOpenly: `misa.A` is like `misa.B` in that clearing it has no architectural impact. Extension Zalrsc may still be implemented even if "A" isn't.
```suggestion
```
So, like `misa.B`, remove any tests where any behavior is associated with testing `misa.A`.

ThinkOpenly (changes requested):
`misa.B` and `misa.A` are weird because the original extensions were broken up into individually implementable extensions, so clearing them does not necessarily mean that the new extensions appear unimplemented.

ThinkOpenly (approved):
Thanks for your persistence and attention!

PR #2447, opened 2026-08-10
Fix Zvkb instruction ownership

abkshaqster:
Move the nine instructions defined by Zvkb from the Zvbb instruction directory to the Zvkb instruction directory.

Update their definedBy fields from Zvbb to Zvkb. Zvbb already requires Zvkb, so no extension dependency changes are needed.

Signed-off-by: Abhik Gude <abhigude@qti.qualcomm.com>


ThinkOpenly (approved):
Thanks!

PR #2455, opened 2026-08-10
Fix EBADDEVENGINES node error

jordancarlin:
Sync node version with mise to avoid prek errors in CI.


ThinkOpenly (approved):
magic incantations. 🤞

Issue #2432, opened 2026-08-06
Atomics narrower than XLEN are modeled incorrectly


aashu2006:
While reading the atomics IDL I hit three problems. They all type-check, and in each case the instruction's own `description:` field says the opposite of what the `operation()` does.

Spec quotes below are from the ISA manual commit this repo pins (`ext/riscv-isa-manual` @ `310a111`).

### 1. lr.w never sign-extends

**Current value and location**

`spec/std/isa/inst/Zalrsc/lr.SIZE.AQRL.layout:149-155`, generated into `lr.w.yaml`, `lr.w.aq.yaml`, `lr.w.rl.yaml`, `lr.w.aqrl.yaml` at lines 105-110:

```idl
XReg load_value = load_reserved(32, virtual_address, aq, rl, $encoding);
if (xlen() == 64) {
  X[xd] = load_value;
} else {
  X[xd] = sext(load_value[31:0], 32);
}
```

The two branches are swapped. `sext()` returns early when `first_extended_bit == MXLEN`, so on RV32 that call does nothing, and the RV64 path has no `sext` at all. `load_reserved(32, ...)` ends up returning `read_physical_memory_32()`, a `Bits<32>` widened into `XReg`, which is a zero-extension. So `rd` gets a zero-extended word on RV64.

`lr.w.yaml:14` and `:19` say the opposite: "places the sign-extended value in xd" and "The 32-bit load result is sign-extended to 64-bits".

`zalrsc.adoc`: "For RV64, `lr.w` and `sc.w` sign-extend the value placed in _rd_."

**Expected value**

```idl
X[xd] = sext(load_value[31:0], 32);
```

in both cases, or just swap the branches.

Affects `lr.w`, `lr.w.aq`, `lr.w.rl`, `lr.w.aqrl` on RV64.

### 2. AMO results are not sign-extended into rd

**Current value and location**

`spec/std/isa/inst/Zaamo/amo*.SIZE.AQRL.layout:80-81` (9 files), generated into e.g. `Zaamo/amoadd.w.yaml:45` and `Zabha/amoadd.b.yaml:45`:

```idl
X[xd] = amo(32, virtual_address, X[xs2][31:0], AmoOperation::Add, 1'b0, 1'b0, $encoding);
X[xd] = amo(8,  virtual_address, X[xs2][7:0],  AmoOperation::Add, 1'b0, 1'b0, $encoding);
```

[...]

PR #2460, opened 2026-08-11
fix(zalrsc): sign-extend the lr.w result on RV64

aashu2006:
The `size == "w"` path in `lr.SIZE.AQRL.layout` had its two branches swapped: `sext()` was called on the RV32 path, where it is a no-op because it returns early when `first_extended_bit == MXLEN`, and the RV64 path assigned `load_value` directly. Since `load_reserved(32, ...)` returns a zero-extended `Bits<32>`, `lr.w` put a zero-extended word in `xd` on RV64.

That contradicts the instruction's own description ("places the sign-extended value in xd") and the Zalrsc spec: "For RV64, LR.W and SC.W sign-extend the value placed in rd." Removing the `xlen()` check and always calling `sext(load_value[31:0], 32)` is correct on both bases and matches how `lw` is written.

Part 1 of #2432.


ThinkOpenly (approved):
Thanks!

Issue #2377, opened 2026-08-04
menvcfgh does not forward accesses to the parent menvcfg CSR


cavemansatyn-design:
### Description

`menvcfgh` is documented as providing access to bits `[63:32]` of the 64-bit `menvcfg` CSR on RV32.

Unlike other unified upper-half CSRs such as `mstateen0h`, `menvcfgh` does not provide a CSR-level `sw_read()` or field-level `sw_write()` forwarding to the parent CSR.

As a result, accesses through `menvcfgh` operate on the local CSR object instead of the logical `menvcfg` storage.

### Expected behavior

Reads and writes through `menvcfgh` should access the upper 32 bits of `menvcfg`, matching the documented alias relationship.

### Proposed fix

- Add a CSR-level `sw_read()` returning `$bits(CSR[menvcfg])[63:32]`.
- Forward writable fields (`STCE`, `PBMTE`, `ADUE`) to `CSR[menvcfg]`, following the existing `mstateen0h` pattern.

cavemansatyn-design:
@ThinkOpenly
please review this one

PR #2462, opened 2026-08-11
fix(csr): forward mcyclecfgh and minstretcfgh accesses to their parent CSRs

RAJVEER42:
`mcyclecfgh` and `minstretcfgh` declare `alias` on every field but have no forwarding, so accesses hit the local CSR object rather than the upper half of the 64-bit parent. This adds a CSR-level `sw_read()` and per-field `sw_write()` in the shape `mstateen0h` already uses.

Same defect #2377 reports for `menvcfgh`. Following the count I posted on #781: of the 104 high-half CSRs carrying aliases, 71 already have a CSR-level `sw_read()` and 33 do not, 29 of them generated from `mhpmeventNh.layout`. This takes two of the four hand-written ones and deliberately leaves `menvcfgh` to #2377 and `mstatush` until #2427 merges.

### DIFF
--- a/spec/std/isa/csr/Smcntrpmf/mcyclecfgh.yaml
+++ b/spec/std/isa/csr/Smcntrpmf/mcyclecfgh.yaml
@@ -27,6 +27,9 @@ fields:
         name: Sm
     description: If set, then counting of events in M-mode is inhibited.
     reset_value: UNDEFINED_LEGAL
+    sw_write(csr_value): |
+      CSR[mcyclecfg].MINH = csr_value.MINH;
+      return csr_value.MINH;
 
   SINH:
     alias: mcyclecfg.SINH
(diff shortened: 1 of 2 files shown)


spec/std/isa/csr/Smcntrpmf/mcyclecfgh.yaml
```diff
@@ -27,6 +29,9 @@ fields:
         name: Sm
     description: If set, then counting of events in M-mode is inhibited.
     reset_value: UNDEFINED_LEGAL
+    sw_write(csr_value): |
+      CSR[mcyclecfg].MINH = csr_value.MINH;
```
ThinkOpenly: You've uncovered an interesting problem in UDB. The CI tests are failing because the new values are being forwarded to the upper half of the 64-bit register (exactly as desired), but using the field names that are only defined for XLEN=64, so these two things will never be defined at the same time.

For now, can you change the forwarding to the (much uglier) respective field bit positions?
```suggestion
      CSR[mcyclecfg][62] = csr_value.MINH;
```
Here, and elsewhere, of course.

ThinkOpenly (changes requested):
I'm really starting to think we should approach this "alias" problem differently, but let's see if we can get the CI tests to pass with this (modified) approach.

PR #2467, opened 2026-08-13
fix(tools): clean up Docker resources on failure in eqntott build script

nXtCyberNet:
Add an EXIT trap handler to `tools/scripts/build_eqntott_with_docker.sh` to ensure temporary directories, Docker containers, and intermediate build images are automatically cleaned up when the script exits. 

Previously, if container extraction failed or the script exited early, temporary Docker containers and images were leaked on the host system.

### DIFF
--- a/tools/scripts/build_eqntott_with_docker.sh
+++ b/tools/scripts/build_eqntott_with_docker.sh
@@ -63,9 +63,20 @@ build_eqntott_with_docker() {
 
     local temp_dir
     temp_dir=$(mktemp -d -t eqntott-build-XXXXXX)
-    trap "rm -rf '$temp_dir'" EXIT
-
     info "Using temporary directory: $temp_dir"
+
+    local container_name=""
+    local image_name=""
+    cleanup() {
+        if [[ -n "$container_name" ]]; then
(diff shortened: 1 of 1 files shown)


tools/scripts/build_eqntott_with_docker.sh
```diff
@@ -63,10 +63,21 @@ build_eqntott_with_docker() {
...
+    cleanup() {
+        if [[ -n "$container_name" ]]; then
+            docker rm "$container_name" >/dev/null 2>&1 || true
+        fi
+        if [[ -n "$image_name" ]]; then
+            docker rmi "$image_name" >/dev/null 2>&1 || true
+        fi
+        rm -rf "$temp_dir"
+    }
+    trap cleanup EXIT
```
ThinkOpenly, line 79: There's a tiny window between when the temporary directory is created and when the cleanup handler is defined. This existed before your changes, but let's fix it now. Please move this new code above the `mktemp` call, initialize `temp_dir` variable to the null string, and check that variable before calling `rm -rf`.

ThinkOpenly (changes requested):
Thanks for this submission!

PR #2468, opened 2026-08-13
feat(csr): add privilege mode check into indirect CSR lookup

inochisa:
Many indirect CSRs have a different definition depending on privilege mode (for example `eidelivery` for M/S and VS has different available values, and `iprio*` is only defined in M/S mode).

Add privilege mode as a parameter when storing and looking up indirect CSRs.

Note:
1. This PR changes the indirect container type, as `std::hash<std::tuple<>>` is not defined.


titoatwork (commented):
The design question I had about the key is answered over on #2478, so this is code-level only. Checked against `bfff85bf`.

## The container change is required, not a preference

Worth stating explicitly, because it reads like an incidental change and a reviewer could reasonably ask you to revert it:

```
- std::unordered_map<std::pair<Bits<64>, Bits<4>>, CsrBase*> m_csr_indirect_addr_map;
+ std::map<std::tuple<PrivilegeMode, Bits<64>, Bits<4>>, CsrBase*> m_csr_indirect_addr_map;
```

The standard specialises `std::hash` for neither `std::pair` nor `std::tuple`, and this tree specialises it only for `udb::_Bits` and `udb::_PossiblyUnknownBits` (`backends/cpp_hart_gen/cpp/include/udb/bits.hpp:1448` and `:3810`). So the old `unordered_map` line could not have compiled had it ever been instantiated, and it never was, because the surrounding `if cfg_arch.not_prohibited_csrs.any?(&:indirect?)` is false for every config today. `std::map` needs only `operator<`, which the tuple has.

## Three questions on behaviour

**`csr_handle.valid` changes meaning.**

```
- csr_handle.valid = true;
+ csr_handle.valid = csr->second->defined();
```

Every `*ireg*` `sw_read`/`sw_write` calls `unimplemented_csr($encoding)` when `!handle.valid`, so this moves when those trap. It is not mentioned in the description. Intended as part of this change, or separable from it?

**Unknown select values now hard-error rather than trapping.** The lookup calls `csr_indirect_addr.to_defined()`, and `to_defined()` raises `UNDEFINED_VALUE_ERROR` when any bit is unknown (`bits.hpp:2438`). If `miselect`/`siselect` holds unknowns, that becomes a model error instead of `valid = false` into `unimplemented_csr`. Which of the two do you want there?

**The duplicate-key `raise` cannot fire yet.** No CSR under `spec/std/isa/csr/` sets `indirect_address` or `indirect_slot`; `schema.adoc` is the only file mentioning them. So `select(&:indirect?)` is empty in every config and the guard is unreachable. Good to land it ahead of the data rather than after, but nothing exercises it today, and the first thing that will is whatever gets authored for AIA or CTR.

## One typo left behind

`spec/std/isa/csr/schema.adoc`, on a line this PR already edits:

```
| *description* | Indirect sddress of the CSR, ...
```

`sddress` should be `address`. Thanks for catching `Smcdrind`/`Sscdrind`/`Sscrind` in the same pass, those had been wrong in three places.

backends/cpp_hart_gen/templates/hart.hxx.erb
```diff
@@ -118,10 +118,15 @@ namespace udb {
             <%- end -%>
           },
           <%- if cfg_arch.not_prohibited_csrs.any?(&:indirect?) -%>
+          <%- duplicate_indirect_keys = cfg_arch.not_prohibited_csrs.select(&:indirect?).group_by { |csr| [csr.priv_mode, csr.indirect_address, csr.indirect_slot] }.select { |_key, csrs| csrs.size > 1 } -%>
+          <%- unless duplicate_indirect_keys.empty? -%>
+          <%- duplicate_indirect_key_details = duplicate_indirect_keys.map { |(mode, addr, slot), csrs| "#{mode}:#{addr}:#{slot} => #{csrs.map(&:name).join(', ')}" }.join("; ") -%>
+          <%- raise "Duplicate indirect CSR keys: #{duplicate_indirect_key_details}" -%>
```
ThinkOpenly: I'm not an "ERB" expert, but can we merge these into a single multi-line statement with better indentation? Like (untested):
```suggestion
          <%-
            duplicate_indirect_keys = cfg_arch.not_prohibited_csrs.select(&:indirect?).group_by { |csr| [csr.priv_mode, csr.indirect_address, csr.indirect_slot] }.select { |_key, csrs| csrs.size > 1 }
            unless duplicate_indirect_keys.empty?
              duplicate_indirect_key_details = duplicate_indirect_keys.map { |(mode, addr, slot), csrs| "#{mode}:#{addr}:#{slot} => #{csrs.map(&:name).join(', ')}" }.join("; ")
              raise "Duplicate indirect CSR keys: #{duplicate_indirect_key_details}"
            end
          -%>
```
(and doesn't the `unless` need an `end` (added) ?)

spec/std/isa/csr/Smcsrind/mireg.yaml
```diff
@@ -46,7 +46,7 @@ fields:
       Register state of the CSR selected by the current `miselect` value
     reset_value: UNDEFINED_LEGAL
     sw_write(csr_value): |
-      Csr handle = indirect_csr_lookup(CSR[miselect].VALUE, 1);
+      Csr handle = indirect_csr_lookup(PrivilegeMode::M, CSR[miselect].VALUE, 1);
```
ThinkOpenly, line 49: Let's be more generic here.
```suggestion
      Csr handle = indirect_csr_lookup(mode(), CSR[miselect].VALUE, 1);
```

spec/std/isa/csr/Smcsrind/mireg.yaml
```diff
@@ -56,7 +56,7 @@ fields:
       csr_sw_write(handle, csr_value.VALUE);
       return csr_hw_read(handle);
 sw_read(): |
-  Csr handle = indirect_csr_lookup(CSR[miselect].VALUE, 1);
+  Csr handle = indirect_csr_lookup(PrivilegeMode::M, CSR[miselect].VALUE, 1);
```
ThinkOpenly, line 59:
```suggestion
  Csr handle = indirect_csr_lookup(mode(), CSR[miselect].VALUE, 1);
```

spec/std/isa/csr/Smcsrind/mireg2.yaml
```diff
@@ -46,7 +46,7 @@ fields:
       Register state of the CSR selected by the current `miselect` value
     reset_value: UNDEFINED_LEGAL
     sw_write(csr_value): |
-      Csr handle = indirect_csr_lookup(CSR[miselect].VALUE, 2);
+      Csr handle = indirect_csr_lookup(PrivilegeMode::M, CSR[miselect].VALUE, 2);
```
ThinkOpenly, line 49:
```suggestion
      Csr handle = indirect_csr_lookup(mode(), CSR[miselect].VALUE, 2);
```

...and for all of the rest, please!

backends/cpp_hart_gen/templates/csrs.hxx.erb
```diff
@@ -207,7 +207,9 @@ namespace udb {
     unsigned address() const override { <%= !csr.indirect? ? "return #{csr.address};" : "throw CsrAddressTypeError(\"#{csr.name} is not direct addressible.\")" %>; }
     uint64_t indirect_address() const override { <%= csr.indirect? ? "return #{csr.indirect_address};" : "throw CsrAddressTypeError(\"#{csr.name} is not indirect.\")" %>; }
     uint8_t indirect_slot() const override { <%= csr.indirect? ? "return #{csr.indirect_slot};" : "throw CsrAddressTypeError(\"#{csr.name} is not indirect.\")" %>; }
+    <%- unless csr.address.nil? -%>
     static constexpr Bits<12> _address() { return <%= csr.address %>_b; }
+    <%- end -%>
```
ThinkOpenly, line 212: Should this instead be modeled like the methods immediately preceding, where the method is always defined, but will `throw` if the CSR is indirect-only?

ThinkOpenly (changes requested):

backends/cpp_hart_gen/templates/csrs.hxx.erb
inochisa: I think this is the thing that is related to how to use this. We can change it if a binding wants to access the "address" field of all CSR (even for indirect CSR). But I prefer to keep this as we can throw a early compiler time error, so the bad code can be find in an early stage.

spec/std/isa/csr/Smcsrind/mireg.yaml
inochisa: After some lookup I found the `mode()` is used as a runtime hart mode. This is what I found in CSR scountovf binding
https://github.com/riscv/riscv-unified-db/blob/ea764633cd73795f8858e29b9dd0e9758fa277af/spec/std/isa/csr/Sscofpmf/scountovf.layout#L49-L57

But here we always needs a argument to represent the CSR privilege mode. So `mode()` is not a good choice....

backends/cpp_hart_gen/templates/hart.hxx.erb
inochisa: > I'm not an "ERB" expert, but can we merge these into a single multi-line statement with better indentation? Like (untested):
> 

Yes, it is possible, I am OK to merge this.

> (and doesn't the `unless` need an `end` (added) ?)

The code already has, I guess you miss the last line (if no `end`, it can not compile.)

spec/std/isa/csr/Smcsrind/mireg.yaml
ThinkOpenly: If we're trying to look up an indirect CSR, we should be looking it up with respect to the current mode, no?

inochisa: In fact we should, but I think we need a new argument for the current mode. The current `PrivilegeMode::M` means we want to lookup an indirect CSR in M mode indirect CSR space, if we remove this argument, the space information is lost so the `indirect_csr_lookup` will not know which indirect CSR space should be lookup. So I think it could be better for `Csr handle = indirect_csr_lookup(mode(), PrivilegeMode::M, CSR[miselect].VALUE, 1);` ?

ThinkOpenly: I don't understand why a lookup in one mode would need to perform the lookup from the context of another mode. A mode-specific lookup should be performed in that mode's CSR address space.

inochisa: This is what the thing introduced by this PR. Without this PR, all the indirect CSRs are in the same global indirect CSR space. However, some indirect CSR has different definition for different privilege mode, this means using one global indirect CSR space makes it hard to describe this different definition clearly, especially for the CSR that has separate storage/valid candidates for different privilege mode.

As a result, to distinguish different indirect CSR space. a new argument is needed to tell the `indirect_csr_lookup()` which indirect CSR space to looked up. This is why you see a `PrivilegeMode::M` is added. For adding `mode()`, I think is could be used for privilege check. Or maybe we should add a check before the lookup, as it is always illegal for a low privilege mode to access a high privilege mode CSR.

I hope this could answer your question.

ThinkOpenly: I think I understand what you are saying. What I don't understand is why we need both an "absolute" mode (by name) AND a "relative" (current) mode? What's the use-case for an "absolute" lookup? In this case for example, `mireg` is defined as `priv_mode: M`, so if the current mode is M, the lookup succeeds, otherwise, it fails. Or, the lookup succeeds, but the later access test fails because the current mode is not sufficiently privileged.

(I haven't explored these paths, so maybe some of that infrastructure is not present? If so, then we should fix _that_ instead of performing a lookup as if in M mode, and then doing an access check as if in current mode. There's only one `mireg`. Maybe I'm still missing something?)

inochisa: > I think I understand what you are saying. What I don't understand is why we need both an "absolute" mode (by name) AND a "relative" (current) mode? What's the use-case for an "absolute" lookup? In this case for example, `mireg` is defined as `priv_mode: M`, so if the current mode is M, the lookup succeeds, otherwise, it fails. Or, the lookup succeeds, but the later access test fails because the current mode is not sufficiently privileged.
> 

tl;dr: Only  "absolute" mode is needed as `indirect_csr_lookup` wants to know which indirect CSR space of current `*ireg*` to lookup, not the indirect CSR space for current hart mode.

Maybe I understand you suggestion in a wrong way? Does the all means all `*ireg*` or `mireg*`? If it is `mireg*`, you suggestion is right as `mode() == PrivilegeMode::M` is always right for M mode CSR.

I do not think we need current mode. After rechecking the indirect CSR access. The thing will be:
1. Apply the common CSR check to `*ireg*` (privilege mode check, access check and other checks)
2. Invoke `sw_read`/`sw_write` of `*ireg*`
3. Execute logic of the selected indirect CSR.

For privilege mode check, it is already applied before entering `*ireg*`. So no current mode is needed (maybe vsireg* need this for virtualization, but it is another topic).

For your question why the current mode can not be used. For a successful CSR access, the current mode could be:

1. M: M
2. S: M,S
3. VS: M,S, VS

This means for `sireg*` and `vsireg*`, may select wrong CSR space if we use `mode()`

The HS mode access `vsireg` as an example 

```c
indirect_csr_lookup(mode(), CSR[vsiselect].VALUE, 1);
```
will be expanded to 
```
indirect_csr_lookup(PrivilegeMode::M, CSR[vsiselect].VALUE, 1);
```
This is totally wrong as `vsireg` is a VS mode CSR. it needs to use `PrivilegeMode::VS` all the time.

For `mode()`, I misunderstand this as I thought `indirect_csr_lookup` may need this for privilege check. But after checking the whole process, I think it is not needed, and it will be better to `indirect_csr_lookup` only an indirect CSR lookup function without other function.

As a result, only an "absolute" mode.

> (I haven't explored these paths, so maybe some of that infrastructure is not present? If so, then we should fix _that_ instead of performing a lookup as if in M mode, and then doing an access check as if in current mode. There's only one `mireg`. Maybe I'm still missing something?)

I think they should be enough. The only thing we need is to understand the process correctly. At least I have made a mistake.

ThinkOpenly: I now understand better, and believe your approach is correct. The crux is that the lookups are not specific to the current mode; they are specific to the CSR being used to access the actual CSR data. `vsireg` has an available namespace which is different than that of `sireg` and `mireg`. Looking up `cycle` (0x40) should work for `miselect=0x40, (mireg)` and `siselect=0x40, (sireg)` but not for `vsiselect=0x40, (vsireg)`. Passing the mode associated with the CSR (not the current mode) to the lookup function is one way to distinguish between (what are currently indistinguishable) calls to the same lookup function from each CSR's respective methods.

backends/cpp_hart_gen/templates/hart.hxx.erb
```diff
@@ -118,10 +118,17 @@ namespace udb {
             <%- end -%>
           },
           <%- if cfg_arch.not_prohibited_csrs.any?(&:indirect?) -%>
+          <%-
+            duplicate_indirect_keys = cfg_arch.not_prohibited_csrs.select(&:indirect?).group_by { |csr| [csr.priv_mode, csr.indirect_address, csr.indirect_slot] }.select { |_key, csrs| csrs.size > 1 }
+            unless duplicate_indirect_keys.empty?
+              duplicate_indirect_key_details = duplicate_indirect_keys.map { |(mode, addr, slot), csrs| "#{mode}:#{addr}:#{slot} => #{csrs.map(&:name).join(', ')}" }.join("; ")
+              raise "Duplicate indirect CSR keys: #{duplicate_indirect_key_details}"
+            end
+          -%>
```
ThinkOpenly, line 127: Putting this check for duplicated keys formed by [mode + indirect_address + slot] seems like a fairly late stage to make this determination. It seems like it could be detected before such a conflict is even merged into the repository, or possibly at "resolve" time if there is overlay data. Can we turn this into a test that is run in pre-commit or CI?

ThinkOpenly (changes requested):

spec/std/isa/csr/Smcsrind/mireg.yaml
inochisa: Yes, you are right

backends/cpp_hart_gen/templates/hart.hxx.erb
inochisa: I think it should, I will try which way is more clear. Thanks.

ThinkOpenly: Given the new test you added in `Rakefile`, this test now seems redundant. Let's remove it here.

inochisa: OK. Removed

backends/cpp_hart_gen/templates/hart.hxx.erb
ThinkOpenly, line 266: I know this is not all part of this PR, but it seems like this can be simplified:
```
        if (csr == m_csr_indirect_addr_map.end() || !csr->second->defined()) {
          csr_handle.valid = false;
          return csr_handle;
        }
        csr_handle.valid = true;
        csr_handle.name = csr->second->name();
        csr_handle.addr_type = CsrAddressType::Indirect;
        csr_handle.address = csr_indirect_addr;
        csr_handle.indirect_slot = window_slot;
        csr_handle.mode = csr->second->mode();
        csr_handle.writable = csr->second->writable();
        return csr_handle;
        <%- end -%>

ThinkOpenly (commented):
One final comment. Let me know what you think. Regardless, looks good to me! Thanks!

backends/cpp_hart_gen/templates/hart.hxx.erb
inochisa: I think it is acceptable, I fix two cases of the csr lookup. Thanks.

ThinkOpenly (approved):

PR #2470, opened 2026-08-13
fix(generators): interpret UDB definedBy format in extension filtering

### LINKED ISSUE #2344: parse_extension_requirements never matches UDB definedBy format, so --extensions filtering in Go/C-header/SystemVerilog generators is broken
**Bug description**
parse_extension_requirements() in backends/generators/generator.py:113 — shared by the Go (Go/go_generator.py), C-header (c_header/generate_encoding.py), and SystemVerilog (sverilog/sverilog_generator.py) generators — never matches the actual UDB definedBy format, so the documented --extensions filter produces wrong output in both directions.
The function expects definedBy as a plain string or a flat dict ({allOf: [...]}, {anyOf: [...]}, {name, version}). However, every instruction (1377) and CSR (396) in spec/std/isa uses the nested format {extension: {name: X}} or {allOf: [{xlen: N}, {extension: {name: Y}}]}. As a result:
- {extension: {name}} (1212 inst / 150 CSR) falls through to the permissive default (return lambda exts: True, line 193) → always included, even when the extension is not in --extensions.
- {allOf: [{xlen}, {extension: ...}]} (139 inst / 246 CSR) → always excluded, even when the extension is enabled, because check_requirement() returns False for the {xlen: N} term (no top-level name key).
- {extension: {anyOf: [...]}} (22 inst) → always included.
- top-level anyOf with nested allOf (4 inst: c.sd, c.sdsp, c.ld, c.ldsp) → always excluded.

**To Reproduce**
1. On branch main,
2. Run:
uv run backends/generators/Go/go_generator.py \
  --inst-dir=spec/std/isa/inst \
  --csr-dir=spec/std/isa/csr \
  --extensions=Zbb \
  --output=/tmp/inst.go
(omit --include-all so filtering is active).
Text output/Error Messages
No error; the tool succeeds silently. 

**Observed behavior:**
- Output contains instructions from M, V, F, D, A, etc., even though only Zbb was requested (the filter is a no-op for {extension: {name}} instructions).
- RV64-only Zbb instructions ctzw and clzw (defined via allOf: [{xlen: 64}, {extension: {name: Zbb}}]) are missing from the output.
A minimal check of the predicate against the actual data confirms the misclassification, e.g. {extension: {name: I}} → predicate returns True for any extension set; {allOf: [{xlen: 64}, {extension: {name: Zbb}}]} → predicate returns False even for ["Zbb"].

**Screenshots**
Not applicable.

**Additional context**
- Affects all three generators: Go/go_generator.py, c_header/generate_encoding.py, sverilog/sverilog_generator.py, plus the excludedBy path in load_instructions() (line 256), which uses the same parser (currently dormant no instruction uses excludedBy).
- The repo's own Rake tasks are unaffected because backends/generators/tasks.rake passes --include-all for c-header and sverilog, and go_generator.py:118 defaults to include_all when --extensions is absent. Only the explicit --extensions feature is broken.
- Function unchanged since introduced in #526 (Go), #553 (C header), #1090 (SystemVerilog).
- Suggested fix: rewrite the predicate to interpret the nested UDB structure — extension.name → name must be in the enabled set; extension.anyOf/oneOf → OR; allOf → AND; treat {xlen} terms as neutral (XLEN is already handled by the --arch flag). Add unit tests covering each shape.

AlgoArtist06:
Fixes #2344

### Summary

`parse_extension_requirements()` in `backends/generators/generator.py` expected `definedBy` to be a plain string or a flat dict (`{allOf: [...]}`, `{anyOf: [...]}`, `{name, version}`), but every instruction and CSR in `spec/std/isa` uses the nested UDB format, e.g. `{extension: {name: X}}` or `{allOf: [{xlen: N}, {extension: {name: Y}}]}`.

This broke the documented `--extensions` filter of the Go, C-header, and SystemVerilog generators in both directions:

- `{extension: {name: X}}` (1212 instructions / 150 CSRs) fell through to the permissive default, so instructions were included even when their extension was not in `--extensions`.
- `{allOf: [{xlen}, {extension: ...}]}` (139 instructions / 246 CSRs) always failed, so RV64-only instructions such as `ctzw` and `clzw` were dropped.
- `{extension: {anyOf: [...]}}` and top-level `anyOf` with nested `allOf` were likewise misclassified.

### Changes

- Rewrite `check_requirement()` to interpret the nested UDB structure recursively: `extension.name` must be in the enabled set, `anyOf`/`oneOf` behave as OR, `allOf` as AND, `not` negates (used by `zext.h`), and `xlen`/`param` terms are neutral.
- Unwrap the `{extension: ...}` term in `parse_extension_requirements()` and use the recursive helper uniformly for `allOf`/`oneOf`/`anyOf`; the plain `{name}` form is now also recognized.
- Add `backends/generators/test_generator.py` with parametrized unit tests covering each `definedBy` shape present in the repository.

### Validation

- New unit tests: 26 passed.
- End-to-end: `go_generator.py --extensions=Zbb --arch=RV64` now emits exactly the 24 Zbb instructions (previously 1234, with `ctzw`/`clzw`/`cpopw`/`zext.h` missing).
- Cross-checked the filter against every instruction (1377) and CSR (396) in `spec/std/isa` for multiple extension combinations with zero mismatches.
- `ruff format`/`ruff check` clean; all pre-commit hooks pass.
- `./bin/regress -n regress-gen-go`, `regress-gen-c-header`, `regress-gen-sverilog` all pass.

### DIFF
--- a/backends/generators/generator.py
+++ b/backends/generators/generator.py
@@ -17,9 +17,28 @@ LOGGER = logging.getLogger(__name__)
 def check_requirement(req, exts):
     if isinstance(req, str):
         return req in exts
-    elif isinstance(req, dict) and "name" in req:
-        # If it has a name field, just match the extension name and ignore version
-        return req["name"] in exts
+    elif isinstance(req, dict):
+        if "extension" in req and req["extension"] is not None:
+            # UDB wraps requirements in an {extension: ...} term
+            return check_requirement(req["extension"], exts)
+        if "anyOf" in req:
+            alternatives = req["anyOf"] if isinstance(req["anyOf"], list) else [req["anyOf"]]
+            return any(check_requirement(alt, exts) for alt in alternatives)
+        if "oneOf" in req:
+            alternatives = req["oneOf"] if isinstance(req["oneOf"], list) else [req["oneOf"]]
+            return any(check_requirement(alt, exts) for alt in alternatives)
+        if "allOf" in req:
+            required = req["allOf"] if isinstance(req["allOf"], list) else [req["allOf"]]
+            return all(check_requirement(r, exts) for r in required)
+        if "not" in req and isinstance(req["not"], dict):
+            # A "not" term requires the nested requirement to be absent
+            return not check_requirement(req["not"], exts)
+        if isinstance(req.get("name"), str):
+            # If it has a name field, just match the extension name and ignore version
+            return req["name"] in exts
+        if "xlen" in req or "param" in req:
+            # xlen and param terms are not extension requirements
+            return True
     return False
 
 
@@ -134,6 +153,10 @@ def parse_extension_requirements(extensions_spec):
             return lambda enabled_exts: any(ext_part in enabled_exts for ext_part in ext_parts)
         return lambda exts: extension in exts
 
+    if "extension" in extensions_spec and extensions_spec["extension"] is not None:
+        # UDB wraps requirements in an {extension: ...} term
+        extensions_spec = extensions_spec["extension"]
+
     # Handle complex cases with allOf/oneOf/anyOf
     if "allOf" in extensions_spec:
         required = extensions_spec["allOf"]
@@ -149,14 +172,7 @@ def parse_extension_requirements(extensions_spec):
             alternatives = [alternatives]
 
         # Process each alternative, which could be a string or a dict with name/version
-        def check_alternative_one_of(alt, exts):
-            if isinstance(alt, str):
-                return alt in exts
-            elif isinstance(alt, dict) and "name" in alt:
-                return alt["name"] in exts
-            return False
-
-        return lambda exts: any(check_alternative_one_of(alt, exts) for alt in alternatives)
+        return lambda exts: any(check_requirement(alt, exts) for alt in alternatives)
 
     # Handle anyOf case (most common in the error output)
     if "anyOf" in extensions_spec:
@@ -165,23 +181,10 @@ def parse_extension_requirements(extensions_spec):
             alternatives = [alternatives]
 
         # Process each alternative, which could be a string, dict with name/version, or nested allOf
-        def check_alternative(alt, exts):
-            if isinstance(alt, str):
-                return alt in exts
-            elif isinstance(alt, dict):
-                if "allOf" in alt:
-                    reqs = alt["allOf"]
-                    if isinstance(reqs, str):
-                        reqs = [reqs]
-                    return all(check_requirement(r, exts) for r in reqs)
-                elif "name" in alt:
-                    return alt["name"] in exts
-            return False
-
-        return lambda exts: any(check_alternative(alt, exts) for alt in alternatives)
+        return lambda exts: any(check_requirement(alt, exts) for alt in alternatives)
 
     # Handle direct name/version specification
-    if "name" in extensions_spec and "version" in extensions_spec:
+    if isinstance(extensions_spec.get("name"), str):
         extension = extensions_spec["name"]
         # We don't actually check the version, just the extension name
         return lambda exts: extension in exts
--- a/backends/generators/test_generator.py
+++ b/backends/generators/test_generator.py
@@ -1,0 +1,93 @@
+import pytest
+from generator import parse_extension_requirements
+
+
+@pytest.mark.parametrize(
+    "defined_by,enabled,expected",
+    [
+        # Single extension via the nested {extension: {name}} form used throughout UDB
+        ({"extension": {"name": "I"}}, ["I"], True),
+        ({"extension": {"name": "I"}}, ["M"], False),
+        # {allOf: [xlen, extension]} form: the xlen term is neutral
+        ({"allOf": [{"xlen": 64}, {"extension": {"name": "Zbb"}}]}, ["Zbb"], True),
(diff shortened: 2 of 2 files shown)


backends/generators/generator.py
```diff
@@ -17,9 +17,28 @@
 def check_requirement(req, exts):
     if isinstance(req, str):
         return req in exts
-    elif isinstance(req, dict) and "name" in req:
-        # If it has a name field, just match the extension name and ignore version
-        return req["name"] in exts
+    elif isinstance(req, dict):
+        if "extension" in req and req["extension"] is not None:
+            # UDB wraps requirements in an {extension: ...} term
```
ThinkOpenly, line 22: Optionally, remove this line or change it to:
```suggestion
            # UDB wraps extension requirements in an {extension: ...} term
```

backends/generators/generator.py
```diff
@@ -134,6 +153,10 @@ def parse_extension_requirements(extensions_spec):
             return lambda enabled_exts: any(ext_part in enabled_exts for ext_part in ext_parts)
         return lambda exts: extension in exts
```
ThinkOpenly, line 154: Since we're here, delete this code, as it can no longer trigger based on the current schema (changed in #891).

backends/generators/generator.py
ThinkOpenly, line 132: The parameter here is no longer just an "extensions_spec", and is a "definedBy condition".
```suggestion
def parse_extension_requirements(condition):
```
change here and elsewhere, of course.

ThinkOpenly, line 135:
```suggestion
   Parse the "definedBy" condition.
   Conditions can be a singleton YAML object like "extension" or a hierarchical dictionary including allOf/oneOf/anyOf aggregations of objects.
```

backends/generators/generator.py
```diff
@@ -134,6 +153,10 @@ def parse_extension_requirements(extensions_spec):
...
 
+    if "extension" in extensions_spec and extensions_spec["extension"] is not None:
+        # UDB wraps requirements in an {extension: ...} term
```
ThinkOpenly, line 157:
```suggestion
```

backends/generators/generator.py
```diff
@@ -165,23 +181,10 @@ def check_alternative_one_of(alt, exts):
...
-
-        return lambda exts: any(check_alternative(alt, exts) for alt in alternatives)
+        return lambda exts: any(check_requirement(alt, exts) for alt in alternatives)
 
     # Handle direct name/version specification
-    if "name" in extensions_spec and "version" in extensions_spec:
+    if isinstance(extensions_spec.get("name"), str):
         extension = extensions_spec["name"]
         # We don't actually check the version, just the extension name
         return lambda exts: extension in exts
```
ThinkOpenly, line 190: This code can only trigger when "extension" is found, so let's move it up under the conditional for "extension" above.

backends/generators/generator.py
ThinkOpenly, line 196: seems like we can just return True here?
```suggestion
    return True
```

ThinkOpenly, line 297: This has been changed (with #891) to now appear only in the "definedBy" for instructions by using "xlen: &lt;value&gt;".

That means it needs to be accommodated in `parse_extension_requirements`, which sounds non-trivial. Are you willing to implement that change?

ThinkOpenly, line 441: This also has been moved into `definedBy` condition.

ThinkOpenly (changes requested):

backends/generators/generator.py
AlgoArtist06: Done - that comment (and the helper it sat in) is gone in the rewrite; the unwrapping now happens in `condition_holds`.

backends/generators/generator.py
AlgoArtist06: Done - the plain-string handling is removed.

backends/generators/generator.py
AlgoArtist06: Done - the parameter is now `condition`. Since the result was always called right away, the function now returns a bool directly (`condition_holds(condition, enabled_extensions, target_arch)`) rather than building a lambda.

AlgoArtist06: Done - used your wording, with `xlen` added as another singleton example.

backends/generators/generator.py
AlgoArtist06: Done - removed.

backends/generators/generator.py
AlgoArtist06: Done - name matching now lives in `extension_condition_holds`, which is only reached from the `extension:` term.

backends/generators/generator.py
AlgoArtist06: Done - the fallthrough is now a plain `return True` (with a warning so unexpected shapes are visible).

AlgoArtist06: Yes, done. `xlen: N` is now checked against `--arch` inside `condition_holds`, and the old `base` block is deleted. This also fixed a real problem: RV32 and RV64 outputs were byte-identical before, since `base` no longer exists in the data. `--include-all` still enforces xlen, matching what `base` used to do.

AlgoArtist06: Done - same treatment for CSRs; the `base` block is deleted and xlen is evaluated from `definedBy`. RV64 now correctly drops the RV32-only `*h` CSRs (`mstatush`, `mcycleh`, ...).

PR #2471, opened 2026-08-14
fix(ruby): Preserve signedness through pruning and rendering to C++

ThinkOpenly:
A failure occurred when running riscv-tests for `clz` instruction.

The IDL for `clz` is:
```
X[xd] = (xlen() - 1) - $signed(highest_set_bit(X[xs1]));
```

The return code from `highest_set_bit(0)`, when cast with `$signed`, should be -1.

The value returned from `highest_set_bit(0)` is `-'sd1` (negative signed decimal 1),
but the sign bit was lost in translation.

An AI-assisted investigation concluded:

- The bug is in `IntLiteralAst#gen_cpp`, triggered specifically by how the prune pass const-folds `-'sd1`.

  Chain of events:

  1. `highest_set_bit`'s returns `-'sd1`. `'sd1` (no explicit width) gets type `Bits<64, signed>`.
     The unary `-` clones that type, so the whole expression `-'sd1` is `Bits<64,signed>`, value -1.

  2. The `func.prune(symtab)` pass const-folds this. `UnaryOperatorExpressionAst#prune` calls
     `PruneHelpers.create_literal(symtab, -1, Bits<64,signed>)` -> `create_int_literal(-1, forced_type:)`:
     ```
     width = forced_type.width          # 64
     v = value.to_s                     # "-1"
     str = "#{width}'#{v}"              # "64'-1"   <-- BUG: no 's' signed marker
     ```

  3. This synthesizes the text "64'-1" and re-parses it as a new `IntLiteralAst`. Since there's no "s" after the "'", the regex captures
     signed-group as empty -- the new literal is silently unsigned `Bits<64, false>`, even though it originated from a signed value. (The numeric value -1 survives only by
     accident, because the trailing `(.*)$` group in the regex is permissive enough that Ruby's `"-1".to_i(10)` still parses to -1.)

  4. `IntLiteralAst#gen_cpp` then renders this pruned literal as:
     "`_Bits<#{t.width}, #{t.signed?}>(#{v}_b)`"   # => "`_Bits<64, false>(-1_b)`"

  5. This is exactly what appears in the generated `return _Bits<64, false>(-1_b);`

  6. The real killer is the C++ token "-1_b" itself. "1_b" is a user-defined literal; `BitsStrHelpers::get_width` computes its width purely from the digit
     string "1" -> 1 bit. So "1_b" is `_Bits<1,false>(1)`. The leading "-" is a separate unary-minus applied after that: `_Bits<1,false>::operator-()`, which negates modulo 2,
     i.e. -1 mod 2 = 1. The result (still 1) is then zero-extended (not sign-extended, since it's unsigned) into `_Bits<64,false>`. This is precisely the footgun the library
     authors warned about in bits.hpp: "be careful with negative numbers here, since literals are always unsigned... -15_b will be +1, because 15_b is only four
     bits, and negation loses the sign bit."

  This matches every GDB observation: highest_set_bit(0) returns 1 instead of -1/0xFF, so clz(0) computes 63 - 1 = 62 (later confirmed to become xlen()-1 overall)
  instead of 63 - (-1) = 64.

  Fix location would be `PruneHelpers.create_int_literal` (needs to preserve the "s" marker / emit a correctly-signed literal string) and/or `IntLiteralAst#gen_cpp` (needs to
  avoid emitting bare "-N_b" for negative values -- e.g. construct the full-width unsigned two's-complement pattern directly, or use `-_Bits<width,signed>(N_b)` instead of
  `(-N)_b`).

Both fixes are included, and are AI-generated. AI-generated tests are also included as well as AI-assisted documentation improvements.

Signed-off-by: Paul A. Clarke <paul.clarke@qti.qualcomm.com>

### DIFF
--- a/backends/cpp_hart_gen/tasks.rake
+++ b/backends/cpp_hart_gen/tasks.rake
@@ -434,6 +434,7 @@ namespace :test do
       sh "make -j #{$jobs} test_bits_random"
       sh "make -j #{$jobs} test_softfloat_fp"
       sh "make -j #{$jobs} test_regfile"
+      sh "make -j #{$jobs} test_util"
       sh "ctest -T coverage -T test"
     end
   end
@@ -471,6 +472,7 @@ namespace :test do
       "srl", "srli", "srliw", "srlw",
       "sub", "subw",
       "xor", "xori"]
+    #rv64uiTests = ["add"]
 
     rv32umTests = ["div", "divu",
       "mul", "mulh", "mulhsu", "mulhu",
--- a/backends/cpp_hart_gen/CMakeLists.txt
+++ b/backends/cpp_hart_gen/CMakeLists.txt
@@ -261,6 +261,7 @@ add_dependencies(test_bits_random ${RANDOM_TESTS})
 
 add_executable(test_util
   ${CMAKE_SOURCE_DIR}/test/test_util.cpp
+  ${CMAKE_SOURCE_DIR}/src/NotificationHandler.cpp
 )
 target_include_directories(test_util PUBLIC ${CMAKE_SOURCE_DIR}/include)
 target_link_libraries(test_util PRIVATE hart Catch2::Catch2WithMain)
@@ -304,6 +305,7 @@ include(CTest)
 include(Catch)
 catch_discover_tests(test_bits_directed)
 catch_discover_tests(test_regfile)
+catch_discover_tests(test_util)
 
 foreach (random_test IN LISTS RANDOM_TESTS)
   catch_discover_tests(${random_test})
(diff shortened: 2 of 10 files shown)


backends/cpp_hart_gen/tasks.rake
```diff
@@ -471,6 +472,7 @@ namespace :test do
       "srl", "srli", "srliw", "srlw",
       "sub", "subw",
       "xor", "xori"]
+    #rv64uiTests = ["add"]
```
jordancarlin: Drop this.

henrikg-qc (approved):

PR #2481, opened 2026-08-16
Add missing Smctr CSR models and CTR depth parameter (#2478)

### LINKED ISSUE #2478: Smctr: five registers named in the normative rules have no model in UDB
UDB models the Smctr control CSRs and the instruction that operates on the CTR buffer, but not the buffer itself. Five registers named in the Smctr normative rules have no representation anywhere under `spec/`. References are to `main` at `a5b67eda`.

```
ctrsource   ctrtarget   ctrdata   sctrdepth   sctrstatus
```

```console
$ git grep -il "sctrdepth\|ctrsource\|ctrtarget" -- spec/
spec/std/isa/inst/Smctr/sctrclr.yaml
```

That single hit is prose, not a model. `sctrclr.yaml:17` describes the instruction's effect on registers UDB does not define:

> Any read of `ctrsource`, `ctrtarget`, or `ctrdata` that follows SCTRCLR, such that it precedes the next ...

What is present: `mctrctl`, `sctrctl`, `vsctrctl`, the `sctrclr` instruction, and seventeen `MCTRCTL_*_IMPLEMENTED` parameters. `ext/Ssctr.yaml` exists with no `csr/Ssctr/` directory at all.

Both extensions are `state: ratified`, `ratification_date: 2024-11`.

## Two different causes, worth separating

**`sctrdepth` and `sctrstatus`** are ordinary CSRs. Nothing blocks modelling them today.

**`ctrsource`, `ctrtarget` and `ctrdata`** are Entry Registers reached through the `Smcsrind` window. `csr_schema.json` defines `indirect_address` and `indirect_slot` for exactly this case and `csr/schema.adoc` documents them, but:

```console
$ git grep -l "indirect_address" -- spec/std/isa/csr/
spec/std/isa/csr/schema.adoc
```

No CSR uses the keys. These three would be the first, so this may be blocked on deciding how indirect CSRs should be authored in practice rather than on the data. (#2468 adds a privilege-mode key to that mechanism; it does not introduce it.)

## A related gap in the same extension

`ext/Smctr.yaml:27-28` states the record count depends on "the implementation (the maximum supported depth) and the CTR configuration (the software selected depth)", and the `SCTRDEPTH_DEPTH` normative rule says "It is implementation-specific which DEPTH value(s) are supported."

No parameter in `spec/std/isa/param/` mentions depth or buffer, so that implementation choice is currently inexpressible.

## Ask

Is the omission deliberate — pending a decision on authoring indirect CSRs — or simply not yet done? Happy to add `sctrdepth` and `sctrstatus` plus a depth parameter as a first step, since neither depends on the indirect mechanism.

Ganesh-403:
## Summary

This PR addresses the missing Smctr models identified in #2478.

Changes:
- Added UDB CSR model for `sctrdepth`.
- Added UDB CSR model for `sctrstatus`.
- Added implementation-specific `SCTRDEPTH_DEPTH_LEGAL_VALUES` parameter support.
- Left `ctrsource`, `ctrtarget`, and `ctrdata` unchanged pending clarification on indirect CSR authoring.

## Testing

Validated with:

- `./bin/pre-commit`
- `./do test:udb:unit`
- `./do test:schema`

All checks passed locally.

Fixes #2478

### DIFF
--- a/spec/std/isa/csr/Smctr/sctrstatus.yaml
+++ b/spec/std/isa/csr/Smctr/sctrstatus.yaml
@@ -1,0 +1,37 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: sctrstatus
+long_name: Supervisor Control Transfer Records Status
+address: 0x14F
+writable: true
+priv_mode: S
+length: 32
+description: |
+  The `sctrstatus` register provides access to CTR status information and is
+  updated by hardware whenever CTR is active.
+definedBy:
+  extension:
+    name: Smctr
+fields:
+  WRPTR:
+    location: 7-0
+    description: |
+      Write Pointer. Indicates the physical CTR buffer entry to be written next.
+
+      Note: Bits above depth-1 are read-only zero depending on the configured CTR depth.
+    type: RW-H
+    reset_value: UNDEFINED_LEGAL
+    sw_write(csr_value): |
+      Bits<8> mask = (9'd16 << CSR[sctrdepth].DEPTH) - 1;
+      return csr_value.WRPTR & mask;
+  FROZEN:
+    location: 31
+    description: |
+      When set, inhibits transfer recording.
+    type: RW-H
+    reset_value: UNDEFINED_LEGAL
--- a/spec/std/isa/param/SCTRDEPTH_DEPTH_LEGAL_VALUES.yaml
+++ b/spec/std/isa/param/SCTRDEPTH_DEPTH_LEGAL_VALUES.yaml
@@ -1,0 +1,27 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../schemas/param_schema.json
+
+$schema: param_schema.json#
+kind: parameter
+name: SCTRDEPTH_DEPTH_LEGAL_VALUES
+description: |
+  The set of supported actual depths for the Control Transfer Records (CTR) buffer.
+long_name: Supported CTR buffer depths
+schema:
+  type: array
+  items:
+    type: integer
+    enum:
+      - 16
+      - 32
+      - 64
+      - 128
+      - 256
+  maxItems: 5
+  minItems: 1
+  uniqueItems: true
+definedBy:
+  extension:
+    name: Smctr
--- a/spec/std/isa/csr/Smctr/sctrdepth.yaml
+++ b/spec/std/isa/csr/Smctr/sctrdepth.yaml
@@ -1,0 +1,43 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../../schemas/csr_schema.json
+
+$schema: "csr_schema.json#"
+kind: csr
+name: sctrdepth
+long_name: Supervisor Control Transfer Records Depth
+address: 0x15F
+writable: true
+priv_mode: S
+length: 32
+description: |
+  The `sctrdepth` register is a 32-bit read/write register used to select the number
+  of records that can be held in the CTR buffer.
+definedBy:
+  extension:
+    name: Smctr
+fields:
+  DEPTH:
+    location: 2-0
+    description: |
+      Selects the depth of the CTR buffer.
+    type(): |
+      return ($array_size(SCTRDEPTH_DEPTH_LEGAL_VALUES) == 1) ? CsrFieldType::RO : CsrFieldType::RW;
+    reset_value(): |
+      if ($array_size(SCTRDEPTH_DEPTH_LEGAL_VALUES) == 1) {
+        if (SCTRDEPTH_DEPTH_LEGAL_VALUES[0] == 16) { return 0; }
+        else if (SCTRDEPTH_DEPTH_LEGAL_VALUES[0] == 32) { return 1; }
+        else if (SCTRDEPTH_DEPTH_LEGAL_VALUES[0] == 64) { return 2; }
+        else if (SCTRDEPTH_DEPTH_LEGAL_VALUES[0] == 128) { return 3; }
+        else if (SCTRDEPTH_DEPTH_LEGAL_VALUES[0] == 256) { return 4; }
+        else { return UNDEFINED_LEGAL; }
+      } else {
+        return UNDEFINED_LEGAL;
+      }
+    sw_write(csr_value): |
+      if ($array_includes?(SCTRDEPTH_DEPTH_LEGAL_VALUES, 32'd16 << csr_value.DEPTH)) {
+        return csr_value.DEPTH;
+      } else {
+        return UNDEFINED_LEGAL_DETERMINISTIC;
+      }


titoatwork (commented):
Thanks for picking this up so quickly, and for splitting it the way the issue suggested.

Both addresses check out — `csrs.adoc:266-267` gives `0x14F` for `sctrstatus` and `0x15F` for `sctrdepth`. Everything below is checked against `ext/riscv-isa-manual` at `560dddf7`.

## Both registers are fixed 32-bit, not `SXLEN`

`smctr.adoc:182` — "The 32-bit `sctrdepth` register specifies the depth of the CTR buffer."
`smctr.adoc:232` — "The 32-bit `sctrstatus` register grants access to CTR status information..."

The normative rule definitions say the same (`normative_rule_defs/smctr.yaml:149`, `:171`), and `csr_schema.json` accepts a literal `32` in its `length` enum.

With `length: SXLEN` the model creates bits the spec does not define: `sctrdepth.WPRI` runs to 63 on RV64, and `sctrstatus.WPRI_2` at 63-32 exists only because of it. Setting `length: 32` lets every `location_rv32`/`location_rv64` pair collapse to a plain `location`, and `WPRI_2` goes away.

(`sctrctl` next door is `length: 64`, so the family isn't uniform — worth reading each one off its own spec text.)

## `sctrstatus` has a field the spec doesn't define

The register format at `smctr.adoc:239-241` is three fields:

```
{bits: 8, name: 'WRPTR'}, {bits: 23, name: 'WPRI'}, {bits: 1, name: 'FROZEN'}
```

That is WRPTR [7:0], WPRI [30:8], FROZEN [31]. The PR has `WPRI_1` at 29-8 and a `WRAP` bit at 30.

There is no WRAP field in Smctr. The only occurrence of the word is inside WRPTR's own description at `:250` — "WRPTR wraps to 0 on an increment when the value matches depth-1". Wrapping is behaviour of WRPTR, not a separate status bit, so `WPRI_1` should be 30-8 and `WRAP` should come out.

## `mctrstatus` is not a register in any ratified extension

The description says `sctrstatus` "is an S-mode and VS-mode alias to the `mctrstatus` register." The single mention of `mctrstatus` in the manual is a non-normative note at `smctr.adoc:281`:

> _If, in the future, it becomes desirable to remove this dependency, an extension **could add** `mctrdepth` and `mctrstatus` CSRs that reflect the same state as `sctrdepth` and `sctrstatus`._

So it's hypothetical future work, not something `sctrstatus` aliases today. Worth noting UDB does model real aliasing with an `alias:` key per field (`sctrctl.yaml` uses `alias: mctrctl.U` and so on), so if there were an alias here it would be expressed structurally rather than in prose.

## `SCTRDEPTH_DEPTH_LEGAL_VALUES` isn't wired to anything

This is the main one. The parameter is added, but `DEPTH` is a plain `type: RW` with `reset_value: UNDEFINED_LEGAL` and no `sw_write()`, so nothing in the tree reads the parameter. As written a config can declare `[0]` and still have the model accept a write of `4`.

`MSTATUS_FS_LEGAL_VALUES` is the in-tree precedent for exactly this shape, and consumes it at three sites on `mstatus.FS`:

```
type():        $array_size(MSTATUS_FS_LEGAL_VALUES) == 1 ? CsrFieldType::RO : CsrFieldType::RW
reset_value(): $array_size(MSTATUS_FS_LEGAL_VALUES) == 1 ? MSTATUS_FS_LEGAL_VALUES[0] : UNDEFINED_LEGAL
sw_write():    $array_includes?(MSTATUS_FS_LEGAL_VALUES, csr_value.FS) ? csr_value.FS : UNDEFINED_LEGAL_DETERMINISTIC
```

That maps onto `sctrdepth.DEPTH` almost unchanged, and it's what makes the field actually WARL rather than nominally so. `SCTRDEPTH_DEPTH_WARL` at `normative_rule_defs/smctr.yaml:159` carries `impl-def-category: WARL`, which is the rule the parameter exists to satisfy.

The enum `[0,1,2,3,4]` is a fair reading of the encoding table, where `depth = 2^(DEPTH+4)` gives 16 through 256. One caveat: the table lists `'11x - reserved`, which covers 6 and 7 but leaves `'101` unlisted. I think that's a typo for `'1xx` upstream and your set is the right one, but it's a genuine gap in the text rather than something you should have to guess at.

## Two things that may be follow-ups rather than this PR

**`WRPTR` narrows with depth.** `:250` says "Bits above those needed to represent depth-1 (e.g., bits 7:4 for a depth of 16) are read-only 0." So how much of WRPTR is writable depends on `sctrdepth.DEPTH`, a live CSR field. Plain `RO-H` doesn't capture that. It may be fine to leave for later, but it's worth a comment in the file so it isn't mistaken for complete.

**`definedBy` is split across the PR.** The two CSRs say `Ssctr` and the parameter says `Smctr`. Worth making consistent, and worth a maintainer's call on which: `sctrctl` is also an S-mode register but lives in `csr/Smctr/` with `definedBy: Smctr`, and the normative rule def for `SCTRDEPTH_DEPTH` carries `instance: Smctr`. I flagged the missing `csr/Ssctr/` directory in the issue, so I may have pointed you at the wrong convention there.

Happy to look again once these are in.

spec/std/isa/csr/Ssctr/sctrdepth.yaml
```diff
@@ -0,0 +1,44 @@
...
+        return csr_value.DEPTH;
+      } else {
+        return UNDEFINED_LEGAL_DETERMINISTIC;
+      }
+  WPRI:
+    location: 31-3
+    description: |
+      Reserved.
+    type: RO
+    reset_value: 0
```
ThinkOpenly: Default for bits left undefined is read-only-0, so there is no need for a definition here.
```suggestion
```

spec/std/isa/csr/Smctr/sctrstatus.yaml
```diff
@@ -0,0 +1,40 @@
...
+definedBy:
+  extension:
+    name: Smctr
+fields:
+  WRPTR:
+    location: 7-0
+    description: |
+      Write Pointer. Indicates the physical CTR buffer entry to be written next.
+
+      Note: Bits above depth-1 are read-only zero depending on the configured CTR depth.
```
ThinkOpenly, line 26: If you are willing and able, you could implement this restriction here with a `sw_write()` method.

spec/std/isa/csr/Ssctr/sctrstatus.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+      Note: Bits above depth-1 are read-only zero depending on the configured CTR depth.
+    type: RO-H
+    reset_value: UNDEFINED_LEGAL
+  WPRI:
+    location: 30-8
+    description: |
+      Reserved.
+    type: RO
+    reset_value: 0
```
ThinkOpenly:
```suggestion
```

spec/std/isa/param/SCTRDEPTH_DEPTH_LEGAL_VALUES.yaml
```diff
@@ -0,0 +1,29 @@
...
+schema:
+  type: array
+  items:
+    type: integer
+    enum:
+      - 0
+      - 1
+      - 2
+      - 3
+      - 4
```
ThinkOpenly: It would be nicer to offer the _depths_ here, rather than the encoding of the depths.
```suggestion
      - 16
      - 32
      - 64
      - 128
      - 256
```
Then, you'd need to update the `sw_write()` method to confirm the provided value is legal.

ThinkOpenly (changes requested):
Thank you for your contribution!

There are a couple Copilot comments that need to be addressed, and I have a few comments, too.

Ganesh-403:
I’ve addressed the remaining review details as well: removed the redundant sctrstatus WPRI definition and corrected the IDL shift expressions. The validation checks pass, and the updated branch is ready for another review.

spec/std/isa/csr/Ssctr/sctrdepth.yaml
Ganesh-403: Addressed — removed the explicit WPRI definition since undefined bits default to read-only zero.

spec/std/isa/csr/Smctr/sctrstatus.yaml
Ganesh-403: Addressed — implemented the depth-dependent WRPTR restriction using sw_write(). The mask is derived from sctrdepth.DEPTH, so bits above the configured depth-1 are cleared on software writes.

spec/std/isa/csr/Ssctr/sctrstatus.yaml
Ganesh-403: Addressed — removed the explicit WPRI definition; the reserved bits are now left undefined and therefore default to read-only zero.

spec/std/isa/param/SCTRDEPTH_DEPTH_LEGAL_VALUES.yaml
Ganesh-403: Addressed — changed SCTRDEPTH_DEPTH_LEGAL_VALUES to represent the supported buffer depths (16, 32, 64, 128, 256) and updated the DEPTH WARL logic accordingly.

spec/std/isa/csr/Ssctr/sctrdepth.yaml
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../../schemas/csr_schema.json
+
+$schema: "../../../../schemas/csr_schema.json#"
```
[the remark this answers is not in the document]
ThinkOpenly: Can you include this fix, please?

spec/std/isa/csr/Ssctr/sctrdepth.yaml
```diff
@@ -0,0 +1,43 @@
...
+        else if (SCTRDEPTH_DEPTH_LEGAL_VALUES[0] == 32) { return 1; }
+        else if (SCTRDEPTH_DEPTH_LEGAL_VALUES[0] == 64) { return 2; }
+        else if (SCTRDEPTH_DEPTH_LEGAL_VALUES[0] == 128) { return 3; }
+        else if (SCTRDEPTH_DEPTH_LEGAL_VALUES[0] == 256) { return 4; }
+        else { return UNDEFINED_LEGAL; }
+      } else {
+        return UNDEFINED_LEGAL;
+      }
+    sw_write(csr_value): |
+      if ($array_includes?(SCTRDEPTH_DEPTH_LEGAL_VALUES, 16 << csr_value.DEPTH)) {
```
ThinkOpenly: Any operation that must create a wider type to hold the result needs to either use a widening operator or explicitly specify a compile-time known width. The `16` here will be represented as a 5-bit quantity, and the shift won't widen that as needed. Widening operators need operands with compile-time known widths, which doesn't apply here. So, let's specify the bit width of the constant operand (somewhat arbitrarily, but large enough to hold any expected result):
```suggestion
      if ($array_includes?(SCTRDEPTH_DEPTH_LEGAL_VALUES, 32'd16 << csr_value.DEPTH)) {
```
(untested by me)

spec/std/isa/csr/Ssctr/sctrstatus.yaml
```diff
@@ -0,0 +1,53 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../../schemas/csr_schema.json
+
+$schema: "../../../../schemas/csr_schema.json#"
```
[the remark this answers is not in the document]
ThinkOpenly: adopt this fix, too.

spec/std/isa/csr/Ssctr/sctrstatus.yaml
```diff
@@ -0,0 +1,37 @@
...
+  WRPTR:
+    location: 7-0
+    description: |
+      Write Pointer. Indicates the physical CTR buffer entry to be written next.
+
+      Note: Bits above depth-1 are read-only zero depending on the configured CTR depth.
+    type: RW-H
+    reset_value: UNDEFINED_LEGAL
+    sw_write(csr_value): |
+      Bits<8> mask = (16 << CSR[sctrdepth].DEPTH) - 1;
```
ThinkOpenly: Need widening here, too. We need a width of 9 to handle the full 2^8 value before the subtraction.
```suggestion
      Bits<8> mask = (9'd16 << CSR[sctrdepth].DEPTH) - 1;
```

spec/std/isa/csr/Ssctr/sctrdepth.yaml
ThinkOpenly: I see you used an 8-bit width below (to match the width of `WRPTR`, logically), so we could use a width of 9 here (to hold "256"). `WRPTR` would have to be redefined larger to hold anything more than 255, anyway.

ThinkOpenly (changes requested):

spec/std/isa/csr/Ssctr/sctrstatus.yaml
Ganesh-403: Addressed, updated the WRPTR mask calculation to use a 9'd16 operand so the intermediate value can represent 256 before subtracting one. Validation passes with ./bin/pre-commit, ./do test:udb:unit, and ./do test:schema.

spec/std/isa/csr/Ssctr/sctrstatus.yaml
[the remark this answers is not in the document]
Ganesh-403: Addressed, updated the CSR schema references in sctrstatus.yaml to use csr_schema.json# following the repository convention.

spec/std/isa/csr/Ssctr/sctrdepth.yaml
Ganesh-403: Addressed, updated the constant operand width to 32'd16 so the shift operation has sufficient width for the supported depth values. Validation passes with ./bin/pre-commit, ./do test:udb:unit, and ./do test:schema.

spec/std/isa/csr/Ssctr/sctrdepth.yaml
```diff
@@ -0,0 +1,43 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=csr_schema.json
```
titoatwork: This one was right before the last commit. Copilot's note was about the `$schema:` key below it; the editor directive above wants the relative path, which is what every other file under `spec/std/isa/csr/` uses.

```suggestion
# yaml-language-server: $schema=../../../../schemas/csr_schema.json
```

spec/std/isa/csr/Ssctr/sctrstatus.yaml
```diff
@@ -0,0 +1,37 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=csr_schema.json
```
titoatwork: This one was right before the last commit. Copilot's note was about the `$schema:` key below it; the editor directive above wants the relative path, which is what every other file under `spec/std/isa/csr/` uses.

```suggestion
# yaml-language-server: $schema=../../../../schemas/csr_schema.json
```

titoatwork (commented):
Two small things, both worth folding into the force-push you are doing anyway.

One is inline: the `yaml-language-server` line got changed in the last commit, and it was right before.

The other is the directory. It still says `csr/Ssctr/` while `definedBy` says `Smctr`. I left that open in my earlier review when I should not have: `sctrctl` and `vsctrctl` are both S-mode and VS-mode registers and both sit in `csr/Smctr/`, so that is the precedent, and nothing in the loader reads the directory name either way. Your call, @ThinkOpenly, if you would rather they stay under `Ssctr`.

Separately, `./do test:idl CFG=_` passes on this. Worth knowing because it is the one that compiles the `sw_write()` bodies: I broke one deliberately to check, and `test:schema` and `test:csrs` both still went green.

ThinkOpenly (approved):

Issue #2463, opened 2026-08-12
CSR length is computed without U or VU mode, and `max_length` raises on a core without the M extension


titoatwork:
`ConfiguredArchitecture#multi_xlen_in_mode?` (`tools/ruby-gems/udb/lib/udb/cfg_arch.rb:103`) answers "can the effective XLEN change in this mode" for `M`, `S`, `U`, `VS` and `VU`, with explicit handling for unconfigured, partially configured and fully configured architectures. `CsrField#dynamic_location?` (`csr_field.rb:380`) uses it, over `Csr#modes_with_access`.

`Csr` answers the same question itself, from the `length` token, in five places: `dynamic_length?`, `max_length`, `length_cond32`, `length_cond64` and `length_pretty`. None of the five considers U or VU mode.

Everything below is on `main` at `858ae306`, reproduced with the config at the end.

### 1. A register and its own fields disagree

In a config where the only mutable mode XLEN is `UXLEN`, 8 of 396 CSRs report `dynamic_length? == false` while at least one of their fields reports `dynamic_location? == true`:

`ssp`, `jvt`, `vl`, `vlenb`, `vstart`, `vtype`, `vxrm`, `vxsat`

All eight are `priv_mode: U`, so `modes_with_access` returns `["M", "S", "U", "VS", "VU"]`. The field asks about five modes; the register asks about three.

The conditions that get emitted are wrong in U-mode:

- `ssp` and `jvt` are `length: XLEN`, and `length_cond32` produces
  `(priv_mode() == PrivilegeMode::M && CSR[misa].MXL == 0) || (priv_mode() == PrivilegeMode::S && CSR[mstatus].SXL == 0) || (priv_mode() == PrivilegeMode::VS && CSR[hstatus].VSXL == 0)`.
  Executing in U-mode every disjunct is false regardless of `mstatus.UXL`.
- `vl`, `vlenb`, `vstart`, `vtype`, `vxrm`, `vxsat` are `length: MXLEN`, and `length_cond32` produces `CSR[misa].MXL == 0`, which is MXLEN for a register whose width follows the current mode.

[...]

PR #2489, opened 2026-08-18
fix(csr): seven vector CSRs take XLEN, not MXLEN

titoatwork:
Fix defined widths of `vcsr`, `vl`, `vlenb`, `vstart`, `vtype`, `vxrm` and `vxsat` to match ISA definitions.

Partially addresses #2463.


ThinkOpenly (approved):
Thank you!

Issue #2282, opened 2026-07-31
`ld` and `sd` for RV32 are broken


ThinkOpenly:
I think `ld` (RV32) is broken:

It attempts to read 64 bits of data:
```
  if (xlen() == 32) {
    if (implemented?(ExtensionName::Zilsd)) {
      Bits<64> data = read_memory(64, virtual_address, $encoding);
```

However, `read_memory`'s signature is:
```
function read_memory {
  returns XReg
```
`XReg` is equivalent to `Bits<MXLEN>` and MXLEN for RV32 is 32.

So, whatever `read_memory` actually reads, only a 32 bit result is returned.

Maybe `read_memory` should always return `Bits<64>`?  (@henrikg-qc?)

_Originally posted by @ThinkOpenly in https://github.com/riscv/riscv-unified-db/pull/2256#discussion_r3686723500_

ThinkOpenly:
Also, @titoatwork noted in https://github.com/riscv/riscv-unified-db/pull/2256#issuecomment-5144714643, `sd` (RV32) and `write_memory` have a very similar problem.

henrikg-qc:
read_memory() interface return to XReg of width in multiples of 8bits. This is the first argument forwarded to the SoC model also for discrete read_physical_memeory_8/16/32/64. 
I'm not sure this is broken. Can we look at the generated code for ld in 32bit config with Zilsd?

ThinkOpenly:
> I'm not sure this is broken.

@henrikg-qc, How does `read_memory` return a 64-bit result in an RV32 `XReg`? `XReg` is only 32 bits in RV32.

iakshkhurana:
Sir please assign me this issue.
Thanks.

ThinkOpenly:
> Sir please assign me this issue. Thanks.

@iakshkhurana, I'd prefer to get agreement that this issue is valid first. Stand by.

iakshkhurana:
Okay sir

henrikg-qc:
> > I'm not sure this is broken.
> 
> [Henrik Gustafsson (@henrikg-qc)](https://github.com/henrikg-qc), How does `read_memory` return a 64-bit result in an RV32 `XReg`? `XReg` is only 32 bits in RV32.

Let me clarify, read_memory is not broken,. the IDL implementation is not correct for 32 bit ld/sd.  
[...]

PR #2491, opened 2026-08-18
fix(isa): compressed FP load/stores must use the float register file

### LINKED ISSUE #2490: Compressed FP load/stores use the integer register file, and `c.fld` skips `creg2reg`
The four non-`sp` compressed floating-point load/stores under `Zcf` and `Zcd` do not touch the floating-point register file the way their `sp` siblings do. Three of them use `X[]`, and the fourth indexes `F[]` without the compressed-register mapping. Line numbers are on `main` at `22776b21`.

| instruction | register file | index | correct? |
|---|---|---|---|
| `c.fld` | `F` | `fd` | **no**, `fd` is 3 bits |
| `c.fldsp` | `F` | `fd` | yes, `fd` is 5 bits here |
| `c.fsd` | **`X`** | `creg2reg(fs2)` | **no** |
| `c.fsdsp` | `F` | `fs2` | yes |
| `c.flw` | **`X`** | `creg2reg(fd)` | **no** |
| `c.flwsp` | `F` | `fd` | yes |
| `c.fsw` | **`X`** | `creg2reg(fs2)` | **no** |
| `c.fswsp` | `F` | `fs2` | yes |

All four `sp` forms are right and all four non-`sp` forms are wrong.

## The three that use the integer register file

`spec/std/isa/inst/Zcd/c.fsd.yaml:41`

```idl
write_memory(64, virtual_address, X[creg2reg(fs2)], $encoding);
```

`spec/std/isa/inst/Zcf/c.fsw.yaml:41`

```idl
write_memory(32, virtual_address, X[creg2reg(fs2)][31:0], $encoding);
```

`spec/std/isa/inst/Zcf/c.flw.yaml:41`

```idl
X[creg2reg(fd)] = sext(read_memory(32, virtual_address, $encoding), 32);
```

Each stores from, or loads into, an integer register whose index happens to equal the intended float register index. `c.fsd` writes eight bytes of `x8`-`x15` to memory where the instruction is defined to write `f8`-`f15`, and `c.flw` overwrites an integer register instead of a float one.

The `sp` siblings show the intended form. `c.fsdsp.yaml:42` is `write_memory(64, virtual_address, F[fs2][63:0], $encoding)`, `c.fswsp.yaml:42` is `write_memory(32, virtual_address, F[fs2][31:0], $encoding)`, and `c.flwsp.yaml:42` is `F[fd] = read_memory(32, virtual_address, $encoding)`. The uncompressed `F/fsw.yaml:39` also uses `F[fs2][31:0]`.

The files themselves say what they mean. `c.fsd.yaml` describes itself as "Stores a double precision floating-point value in register fs2" and "It expands to `fsd` `fs2, offset(xs1)`", and its `encoding` declares a variable named `fs2`, not `xs2`.

## The one with the wrong index

`spec/std/isa/inst/Zcd/c.fld.yaml:41`

```idl
F[fd] = sext(read_memory(64, virtual_address, $encoding), 64);
```

Here the register file is right and the index is not. `c.fld` is a CL-format instruction, so its `fd` variable sits at `location: 4-2`, three bits, and needs `creg2reg` to reach `f8`-`f15`. As written it targets `f0`-`f7`. `c.fldsp` needs no mapping because its `fd` is at `location: 11-7`, five bits, and that is why the same expression is correct there and wrong here.

This is the same shape as `c.sdsp` in #2418, where a five-bit field was passed through `creg2reg`. This is the inverse: a three-bit field that is not.

## How I came across it

Counting `misa` guard sites for #2434 and #2435. `Zcf` and `Zcd` are the only directories carrying both a `misa.C` guard and a `misa.F` or `misa.D` guard, so I read all eight files to check which bit each guard belonged to, and the register expressions did not match between the `sp` and non-`sp` pairs.

## Suggested fix

Bring each non-`sp` form in line with its `sp` sibling: `F[creg2reg(fd)]` and `F[creg2reg(fs2)]`. I will put up a PR.

## Not part of this

While reading these eight files I also noticed that none of them calls `check_f_ok()`, which `spec/std/isa/isa/fp.idl:143` defines and which 39 of the 44 instruction files touching `F[]` do call. It tests `MUTABLE_MISA_F && CSR[misa].F == 0` and `CSR[mstatus].FS == 0`, and the compressed forms test neither. The uncompressed `F/flw.yaml` also NaN-boxes its result and calls `mark_f_state_dirty()`, and none of the compressed forms do either. That is a different defect with a different fix, and I will file it separately rather than widen this one.

titoatwork:
`c.fsd` and `c.fsw` store from `X[creg2reg(fs2)]` and `c.flw` loads into `X[creg2reg(fd)]`, so all three use the integer register file where their `sp` siblings use `F[]`. `c.fld` uses `F[]` but indexes it with its raw three-bit `fd` field, reaching `f0`-`f7` instead of `f8`-`f15`.

Four one-line changes, each matching the corresponding `sp` form.

Closes #2490

### DIFF
--- a/spec/std/isa/inst/Zcd/c.fsd.yaml
+++ b/spec/std/isa/inst/Zcd/c.fsd.yaml
@@ -38,4 +38,4 @@ operation(): |
 
   XReg virtual_address = X[creg2reg(xs1)] + imm;
 
-  write_memory(64, virtual_address, X[creg2reg(fs2)], $encoding);
+  write_memory(64, virtual_address, F[creg2reg(fs2)][63:0], $encoding);
--- a/spec/std/isa/inst/Zcd/c.fld.yaml
+++ b/spec/std/isa/inst/Zcd/c.fld.yaml
@@ -38,4 +38,4 @@ operation(): |
 
   XReg virtual_address = X[creg2reg(xs1)] + imm;
 
-  F[fd] = sext(read_memory(64, virtual_address, $encoding), 64);
+  F[creg2reg(fd)] = sext(read_memory(64, virtual_address, $encoding), 64);
--- a/spec/std/isa/inst/Zcf/c.flw.yaml
+++ b/spec/std/isa/inst/Zcf/c.flw.yaml
@@ -38,4 +38,4 @@ operation(): |
 
   XReg virtual_address = X[creg2reg(xs1)] + imm;
 
-  X[creg2reg(fd)] = sext(read_memory(32, virtual_address, $encoding), 32);
+  F[creg2reg(fd)] = read_memory(32, virtual_address, $encoding);
--- a/spec/std/isa/inst/Zcf/c.fsw.yaml
+++ b/spec/std/isa/inst/Zcf/c.fsw.yaml
@@ -38,4 +38,4 @@ operation(): |
 
   XReg virtual_address = X[creg2reg(xs1)] + imm;
 
-  write_memory(32, virtual_address, X[creg2reg(fs2)][31:0], $encoding);
+  write_memory(32, virtual_address, F[creg2reg(fs2)][31:0], $encoding);


spec/std/isa/inst/Zcd/c.fsd.yaml
```diff
@@ -38,4 +38,4 @@ operation(): |
 
   XReg virtual_address = X[creg2reg(xs1)] + imm;
 
-  write_memory(64, virtual_address, X[creg2reg(fs2)], $encoding);
+  write_memory(64, virtual_address, F[creg2reg(fs2)][63:0], $encoding);
```
ThinkOpenly, line 41: Why is the `[63:0]` added here?

titoatwork: Because `F[]` is not always 64 bits wide. `spec/std/isa/register_file/F.yaml` defines it as

```
register_length(): |
  return implemented?(ExtensionName::Q) ? 128 : (implemented?(ExtensionName::D) ? 64 : 32);
```

so in a configuration implementing Q it is 128. `write_memory` takes `XReg value` (`globals.isa:3023-3028`), so the operand has to be narrowed either way, and `[63:0]` is what makes the intended 64 bits explicit rather than leaving it to a truncation.

It also matches what every other float store in the tree already does:

```
Zcd/c.fsdsp.yaml:42   write_memory(64, virtual_address, F[fs2][63:0], $encoding);
Zcf/c.fswsp.yaml:42   write_memory(32, virtual_address, F[fs2][31:0], $encoding);
F/fsw.yaml:39         write_memory(32, virtual_address, F[fs2][31:0], $encoding);
```

`c.fsw` in this PR takes `[31:0]` for the same reason.

Happy to drop both slices if you would rather, though then the three above want the same treatment so the family stays consistent.

One thing I noticed while checking this, and it is not for this PR: `write_memory`'s `value` is `XReg`, which is 32 bits on RV32, and `Zcd` is available there. So `c.fsd` on RV32 passes a 64-bit operand through a 32-bit parameter. That is the same shape as #2282 rather than anything this change introduces, and it is untouched here.

ThinkOpenly: > Because `F[]` is not always 64 bits wide. `spec/std/isa/register_file/F.yaml` defines it as
> 
> ```
> register_length(): |
>   return implemented?(ExtensionName::Q) ? 128 : (implemented?(ExtensionName::D) ? 64 : 32);
> ```

Yep, point proven. :-) Thanks.

> `write_memory`'s `value` is `XReg`, which is 32 bits on RV32, and `Zcd` is available there. So `c.fsd` on RV32 passes a 64-bit operand through a 32-bit parameter. That is the same shape as #2282 rather than anything this change introduces, and it is untouched here.

Sigh. Yep. I added comment about this in #2282.

ThinkOpenly (approved):

PR #2496, opened 2026-08-19
fix(udb): max_length tests the M extension where it means M-mode

titoatwork:
`Csr#max_length` guards its MXLEN branch with `include?("M")`. `M` is integer multiply and divide; machine mode is `Sm`. The branch derives the width from MXLEN, so the test means M-mode and names the wrong extension.

related to #2463

### DIFF
--- a/tools/ruby-gems/udb/lib/udb/obj/csr.rb
+++ b/tools/ruby-gems/udb/lib/udb/obj/csr.rb
@@ -276,11 +276,7 @@ module Udb
       return T.must(base) unless base.nil?
 
       case @data["length"]
-      when "MXLEN", "XLEN"
-        # The effective XLEN of a mode is never larger than MXLEN. The requirements
-        # on SXLEN, UXLEN, VSXLEN and VUXLEN in spec/std/isa/param state this, each
-        # transitively bounding its mode by M-mode. So MXLEN is the widest a CSR can
-        # be whether its length follows M-mode or the executing mode.
+      when "MXLEN"
         cfg_arch.mxlen || 64
       when "SXLEN"
         if cfg_arch.param_values.key?("SXLEN")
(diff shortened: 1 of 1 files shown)


tools/ruby-gems/udb/lib/udb/obj/csr.rb
```diff
@@ -276,7 +276,11 @@ def max_length
       return T.must(base) unless base.nil?
 
       case @data["length"]
-      when "MXLEN"
+      when "MXLEN", "XLEN"
```
ThinkOpenly: I think the original code, below, exists in order to properly handle configurations which do not include M mode. In these cases, the bounding XLEN is based on the modes available in the configuration, in priority order as described in the original code. I suggest the original code be retained with the change from checking extension "M" (multiply/divide) to extension "Sm" (machine mode).

As an aside, and possibly for a separate PR, the elaborate tests here like:
```
if size(array) > 1
  array.max
else
  array[0]
end
```
Seem completely unnecessary and could be replaced with:
```
array.max
```

ThinkOpenly (changes requested):

tools/ruby-gems/udb/lib/udb/obj/csr.rb
titoatwork: You are right, and I have taken the suggestion. The cascade is back exactly as it was and the only change is the guard, `M` to `Sm`. The diff is one line now.

For the record on why the cascade matters, the fall-through needs all three of `M`, `S` and `H` to be absent, so the config that reaches `raise "Unexpected"` is an M-mode-only core with no supervisor, no hypervisor and no multiply. On a partially configured RV32 core with those three prohibited, all 16 `length: XLEN` CSRs raise today and resolve to 32 with the guard fixed. `max_length` across the 10 configs in `cfgs/` is unchanged.

Noted on the `size > 1` tests. I will send that separately rather than widen this one.

ThinkOpenly (approved):

PR #2498, opened 2026-08-20
fix(zaamo,zabha): sign-extend narrow AMO results into xd

aashu2006:
The nine `Zaamo/amo*.SIZE.AQRL.layout` templates assigned the result of `amo()` straight to `X[xd]`. `amo()` is declared to return `XReg` but produces its value from `atomic_read_modify_write_32/64`, and widening a narrower Bits is a zero-extension, so `amoadd.w` and friends put a zero-extended word in `xd` on RV64, and the Zabha `.b/.h` variants never sign-extended at all.

The spec requires the opposite in both cases: "For RV64, 32-bit AMOs always sign-extend the value placed in rd" (Zaamo) and "Byte and halfword AMOs always sign-extend the value placed in `rd`" (Zabha). Each instruction's own description already said "Write the sign-extended value into xd". Wrapping the call in `sext(..., <access width>)` for the b/h/w sizes matches how `lw` is written; `.d` is unchanged.

Only the 9 `.layout` files are hand-edited; the other 108 YAML files were regenerated with `./do gen:arch` (3 sizes × 9 ops × 4 ordering variants). This corrects the value written to `xd`; the `.b/.h` variants still perform a 64-bit read-modify-write until part 3 is fixed.

Part 2 of #2432.

### DIFF
--- a/spec/std/isa/inst/Zaamo/amoadd.SIZE.AQRL.layout
+++ b/spec/std/isa/inst/Zaamo/amoadd.SIZE.AQRL.layout
@@ -74,7 +74,7 @@ operation(): |
 
 <% end -%>
   XReg virtual_address = X[xs1];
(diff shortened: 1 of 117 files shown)


ThinkOpenly (approved):
Good catch! Thanks!

Issue #210, opened 2024-10-30
Implement V in IDL


dhower-qc:
The entire Vector (V) extension needs an IDL implementation.

AFOliveira:
If I understand correctly this is currently being done in #467 by @jmawet .  I added the correct tags so we can just mange all active development better. If I am wrong, please correct me

ThinkOpenly:
It's a really big task, is it something that could be a candidate for "divide and conquer"? We have 2 IDL Writer mentees that might be helpful.

jmawet:
Yeah I've been spending a lot of my free time reading through the spec trying to understand the vector instructions. I definitely get the purpose of this udb project, the spec is so confusing haha, especially trying to figure out each individual instruction. My progress is moving at a snail's pace I feel. 

Combine that with the huge task of creating a vector implementation in IDL... maybe I've bitten off more than I can chew. If we have IDL writer mentees, they might be better suited to this, especially implementing idl vector functions

latifbhatti:
Hi @jmawet ,
First of all, I really appreciate the work you’re doing on the RISC-V vector extension for the IDL UDB project! It’s great to see this being developed, and I’m excited to contribute.

I wanted to touch base regarding the RISC-V vector extension work in the IDL UDB project. Abdullah and I have been selected as mentees for this project, and I’d like to ensure that our contributions align with the existing implementation.

I have experience with the RISC-V vector extension and have developed an open-source RISC-V vector co-processor (Vaquita). 

However, I’m not sure how much progress has already been made on the vector extension in IDL. To avoid duplicate work and ensure an efficient divide and conquer approach, I’d love to discuss:

    The current status of the vector extension implementation in IDL

    Areas where contributions are needed

    How we can best divide tasks to accelerate progress

Would it be possible to have a quick sync to go over these points? Alternatively, if there are any documents or ongoing discussions I should review, please point me in the right direction.

Looking forward to collaborating!

latifbhatti:
@dhower-qc , @ThinkOpenly and @jmawet 

I appreciate the ongoing development of the RISC-V vector extension for IDL UDB! Abdullah and I have been selected as mentees for this project, and we want to ensure our contributions align with the existing implementation. Could you share the current status, areas that need contributions, and how we can effectively divide tasks? Also, what’s the best way to get started with the IDL project? Looking forward to collaborating!

jmawet:
Hi! 
[...]

PR #2502, opened 2026-08-20
Add Zvfbfa extension and update instruction definitions

abkshaqster:
Add the draft `Zvfbfa` extension and update `definedBy` for the 81 existing instructions the specification refines under `vtype.altfmt`.

riscv/riscv-isa-manual#2743

Zvfbfa adds a 1-bit `altfmt` field to `vtype` (bit 8) that enables BF16 vector compute across a defined set of existing instructions. This field is NOT added in this PR, nor are any semantic functions added or changed.

Signed-off-by: Abhik Gude <abhigude@qti.qualcomm.com>


spec/std/isa/ext/Zvfbfa.yaml
```diff
@@ -0,0 +1,24 @@
...
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+# yaml-language-server: $schema=../../../schemas/ext_schema.json
+$schema: "ext_schema.json#"
+kind: extension
+name: Zvfbfa
+long_name: Vector Extension for Additional BF16 Computation
+type: unprivileged
+description: |
+ The Zvfbfa extension adds more complete BF16 vector compute support.
+ Zvfbfa adds a 1-bit field, altfmt, to the vtype CSR in bit position 8.
```
ThinkOpenly: This information will be evident in that CSR definition (when it is added there, of course, presumably in a new PR).
```suggestion
```

spec/std/isa/ext/Zvfbfa.yaml
```diff
@@ -0,0 +1,24 @@
...
+ Attempting to set altfmt=1 and SEW >= 32 is reserved.
```
ThinkOpenly: similarly
```suggestion
```

spec/std/isa/ext/Zvfbfa.yaml
```diff
@@ -0,0 +1,24 @@
...
+ Zvfbfa is compatible with, but does not require, Zvfbfwma and Zvfbfmin.
+versions:
+ - version: "0.1.0"
```
ThinkOpenly: maybe add some comments here
```suggestion
 - version: "0.1.0"
   # date: 2026-06-08
   # commit: f15df74e95e56ad5fabdc9edcf0f1c907d29040a
```

spec/std/isa/inst/Zvfbfwma/vfwmaccbf16.vf.yaml
```diff
@@ -11,7 +11,9 @@ description: |
   Performs a BFloat16 multiply-accumulate operation on each element in vector register vs2 with scalar register rs1, accumulating results in vector register vd according to mask vm. This is a widening operation that produces single-precision results.
 definedBy:
   extension:
-    name: Zvfbfwma
+    anyOf:
+      - name: Zvfbfwma
+      - name: Zvfbfa
```
ThinkOpenly: I don't think this is needed, or correct. The spec says:
> `vfwmaccbf16.vv`, `vfwmaccbf16.vf` (only if Zvfbfwma is implemented)

So, these instructions are defined if and only if Zvfbfwma is defined. Zvfbfa just adds some semantics.
```suggestion
    name: Zvfbfwma
```

spec/std/isa/inst/Zvfbfwma/vfwmaccbf16.vv.yaml
```diff
@@ -11,7 +11,9 @@ description: |
   Performs a BFloat16 multiply-accumulate operation on each element in vector register vs2 with the corresponding element in vector register vs1, accumulating results in vector register vd according to mask vm. This is a widening operation that produces single-precision results.
 definedBy:
   extension:
-    name: Zvfbfwma
+    anyOf:
+      - name: Zvfbfwma
+      - name: Zvfbfa
```
ThinkOpenly: same
```suggestion
    name: Zvfbfwma
```

ThinkOpenly (changes requested):
I think 2 instructions are missing similar changes: `vfncvtbf16.f.f.w` and `vfwcvtbf16.f.f.v`.

Eventually, when the vector floating-point instructions are semantically defined, we'll need to accommodate this from the extension spec:
> When csr::[altfmt]=1 and SEW=16, all vector floating-point instructions become reserved, except for the [_the set which accommodates this extension_], which are redefined to use the BF16 format for any operand that would otherwise have used the FP16 format

I've added a [comment](https://github.com/riscv/riscv-unified-db/issues/210#issuecomment-5360003774) to #210.

spec/std/isa/ext/Zvfbfa.yaml
ThinkOpenly: @jordancarlin says it's "frozen" now, which equates to "0.7.0"

jordancarlin: Looks like it is not actually frozen but is "in freeze". See https://riscv.atlassian.net/browse/RVS-3748.

ThinkOpenly: As in, there isn't a "frozen" state but there is an "in freeze" state? I think we're in "close enough" state, or we need to change UDB https://github.com/riscv/riscv-unified-db/blob/e3dd262ead8ee6ecb22eb962a54e31c279b90d41/spec/schemas/schema_defs.json#L119-L126

jordancarlin: No. As in the extension is in the process of being frozen but is not yet frozen. At least, I think that is what it means.

ThinkOpenly: There's only "in freeze" in this diagram:
image
So, this seems like the steps to ratification:
1. Inception
2. Planning ("In planning")
3. Development ("Under development")
4. Stabilization ("Under stabilization")
5. Frozen ("In freeze")
6. Ratification-ready ("In ratification-ready")
7. Publication
8. Ratified

(Admittedly, those don't perfectly match the UDB states above.)

jordancarlin: The image I usually refer to is the following:
image

This differentiates between the process of freezing the spec and a spec being frozen. Also, apparently frozen is actually 0.9.

ThinkOpenly: Kinda inconsistent, to my eyes. Given all of the above, I view the current state as "frozen".

@abkshaqster, please set the version to "0.9.0". (Leave the date and commit hash.)

jordancarlin: @ThinkOpenly @abkshaqster This came up for a different extension over in sail-riscv, and @rpsene said that an extension that is still “in freeze” (has not completed all the freeze requirements yet) should be v0.8, not v0.9. See https://github.com/riscv/sail-riscv/pull/1872#discussion_r3855863964. Seems like the same would apply to this extension.

spec/std/isa/inst/Zvfbfmin/vfncvtbf16.f.f.w.yaml
ThinkOpenly, line 1: Actually, this doesn't need to change. Similar to `vfwmaccbf16`, the instruction itself is defined if and only if Zvfbfmin is defined. Zvfbfa only adds some semantic changes.

spec/std/isa/inst/Zvfbfmin/vfwcvtbf16.f.f.v.yaml
ThinkOpenly, line 1: same

ThinkOpenly (changes requested):

spec/std/isa/ext/Zvfbfa.yaml
ThinkOpenly: OK.

But, that explanation doesn't exactly match the diagram that I posted above (which can be found on any of the extension Atlassian pages by clicking on the state in the upper right, then "View workflow"). There, the "back to freeze" transition goes to the "Specification in Freeze" state, and there is no separate "frozen" state. The major transition out of "in freeze" goes to "ratification ready".

Even the diagram you posted only has "freezing" and "frozen" (no "in freeze").

jordancarlin: @jjscheel @rpsene can either of you provide any insight?

rpsene: @jordancarlin @ThinkOpenly @abkshaqster happy to settle this.

**Short answer: Zvfbfa should be `0.8.0`.** Jordan is right, and the same rule I gave on sail-riscv applies here.

### The version ladder

These are the milestone labels our planning tool emits for every specification, so this is the authoritative mapping:

| Milestone | Version |
|---|---|
| Development Completed | v0.6 |
| Internal Review Start | v0.6 |
| **Specification Stabilized** | **v0.8** |
| ARC Freeze Approval Request | (no version change) |
| **Specification Frozen** | **v0.9** |
| Public Review Start | v0.9 |
| TSC Approval | v0.99 |
| Specification Ratified | v1.0 |

A specification keeps the version of the last milestone it actually achieved. Zvfbfa achieved Stabilized on 2025-09-03 and has not achieved Frozen, so it is v0.8.

### Why there is no "Frozen" box in the workflow diagram

Paul, your diagram is correct and so is Jordan's. They are showing different things.

The Jira workflow shows **phases**. `Specification in Freeze` is the phase during which the freeze work is done: ARC review, compliance check, ACTs, Sail model, proof of concept, and the Governing Committee approval.

`Frozen` is not a phase, it is the **milestone that ends that phase**. The moment it is achieved, the specification transitions to `Specification in Ratification-Ready`. That is why there is no Frozen box: the state immediately after Frozen is Ratification-Ready. 

So "in freeze" means the work is under way, and "frozen" means it is finished. They are not two names for one thing, and a specification can sit in the freeze phase for a long time without being frozen.

### Zvfbfa specifically

RVS-3748 is in the Freeze phase, entered 2025-09-03. Its Frozen date is empty and five freeze subtasks are still open, including the Governing Committee Approval, which has not started. It is not frozen by any reading.

### On the UDB schema, which I think is the real issue here

Paul flagged this early and I think he is right. `spec_state` in `spec/schemas/schema_defs.json` has `development`, `frozen`, `public-review`, `ratification-ready`, `ratified`, `nonstandard-released`. There is no value for a specification that is in the freeze phase but not yet frozen, which is exactly the case being argued about.

Today the honest mapping is `development` until the Frozen milestone is achieved, then `frozen`. That is accurate but it does undersell how far along something like Zvfbfa is. If UDB wants to model the lifecycle faithfully, adding a value for the freeze phase would close the gap. I do not want to make that call unilaterally on a UDB schema, so I am happy to bring it to the UnifiedDB SIG if you want it decided properly.

ThinkOpenly: Thanks, @rpsene! This is helpful, especially the "authoritative mapping". However, that lists milestones not states although they are certainly related. Remapping that milestone table to states, I see:
| State | Version |
|---|---|
| development | `( 0.0 - 0.6 )` |
| stabilization | `[ 0.6 - 0.8 )` |
| freezing | `[ 0.8 - 0.9 )` |
| ratification | `[ 0.9 - 0.99 )` |
| publication | `[ 0.99 - 1.0 )` |
| ratified | `1.0` |

So, the UDB schema could change to the following and faithfully represent the states in which extensions exist per the current process definition:
```
 "enum": [ 
   "development", 
   "stabilization", 
   "freezing", 
   "ratification", 
   "publication", 
   "ratified", 
   "nonstandard-released" 
 ]
```

Is that correct?

rpsene: > Thanks, Rafael Sene (@rpsene)! This is helpful, especially the "authoritative mapping". However, that lists milestones not states although they are certainly related. Remapping that milestone table to states, I see:
> 
> State	Version
> development	`( 0.0 - 0.6 )`
> stabilization	`[ 0.6 - 0.8 )`
> freezing	`[ 0.8 - 0.9 )`
> ratification	`[ 0.9 - 0.99 )`
> publication	`[ 0.99 - 1.0 )`
> ratified	`1.0`
> So, the UDB schema could change to the following and faithfully represent the states in which extensions exist per the current process definition:
> 
> ```
>  "enum": [ 
>    "development", 
>    "stabilization", 
>    "freezing", 
>    "ratification", 
>    "publication", 
>    "ratified", 
>    "nonstandard-released" 
>  ]
> ```
> 
> Is that correct?

Yes!

ThinkOpenly: > Today the honest mapping is `development` until the Frozen milestone is achieved, then `frozen`.

That ignores the "stable" milestone. Is that intentional?

rpsene: > > Today the honest mapping is `development` until the Frozen milestone is achieved, then `frozen`.
> 
> That ignores the "stable" milestone. Is that intentional?

Nope, missed that. Good catch, @ThinkOpenly !

ThinkOpenly: I opened #2528. Until that is addressed, we can stick with "development" here and version "0.8". Thanks, all!

ThinkOpenly (approved):

PR #1817, opened 2026-04-30
Add Sspmp, Sspmpen, and Smpmpdeleg extension definitions

ybc-alkaid:
This PR adds Unified Database extension definitions for RISC-V S-level Physical Memory Protection (SPMP), consisting of three extensions:

Sspmp: Core SPMP mechanism for S-mode memory protection.
Sspmpen: Per-entry enable bits via the spmpen CSR.
Smpmpdeleg: PMP-to-SPMP resource sharing via the mpmpdeleg CSR.

References:
SPMP specification: https://github.com/riscv/riscv-isa-manual/pull/2573
SPMP arch test suites: https://github.com/riscv/riscv-arch-test/pull/1166

PR #2510, opened 2026-08-22
feat(ext): add Smpmpind, Smcfiss, Smucfiss and Sspmpss definitions

michalchm:
Adds definitions for four pre-ratification extensions that together provide Zicfiss shadow stacks where page
tables are not available. Smpmpind reaches the PMP registers through the Smcsrind indirect window and adds an
enable bit that the direct CSRs do not encode. Smcfiss and Smucfiss use that bit to tag a PMP entry as an
M-mode or U-mode shadow-stack region. Sspmpss instead defines shadow-stack regions through SPMP for
S-mode and U-mode.

This also removes the S requirement from Smcsrind, which should work on M+U harts as well.

### DIFF
--- a/spec/std/isa/ext/Smucfiss.yaml
+++ b/spec/std/isa/ext/Smucfiss.yaml
@@ -1,0 +1,30 @@
+# Copyright (c) NVIDIA Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../schemas/ext_schema.json
+
+$schema: "ext_schema.json#"
+kind: extension
+name: Smucfiss
+long_name: U-mode Shadow Stack on M+U-only systems
+description: |
+  The Smucfiss extension enables Zicfiss shadow stacks in U-mode on harts with
+  no S-mode, where page tables are not available to mark shadow-stack pages.
+
+  It reuses the PMP region tagging introduced by Smpmpind and Smcfiss. A PMP
+  entry with E=1 and XWR=110 marks a U-mode shadow-stack region, which M-mode
+  enables by setting menvcfg.SSE.
+type: privileged
+versions:
+  - version: "0.1.0"
+    state: development
+    url: "https://github.com/ved-rivos/riscv-isa-manual/blob/smpmpss/src/smcfiss.adoc"
+requirements:
+  extension:
+    allOf:
+      - name: Zicfiss
+        version: "~> 1.0"
+      - name: Smcfiss
+        version: "~> 0.1"
+      - not:
+          name: S
--- a/spec/std/isa/ext/Sspmpss.yaml
+++ b/spec/std/isa/ext/Sspmpss.yaml
@@ -1,0 +1,27 @@
+# Copyright (c) NVIDIA Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# yaml-language-server: $schema=../../../schemas/ext_schema.json
+
+$schema: "ext_schema.json#"
+kind: extension
+name: Sspmpss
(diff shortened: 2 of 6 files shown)


spec/std/isa/ext/Smucfiss.yaml
```diff
@@ -0,0 +1,23 @@
...
+  It reuses the PMP region tagging introduced by Smpmpind and Smcfiss. A PMP
+  entry with E=1 and XWR=110 marks a U-mode shadow-stack region, which M-mode
+  enables by setting menvcfg.SSE.
+type: privileged
+versions:
+  - version: "0.1.0"
+    state: development
+    url: "https://github.com/ved-rivos/riscv-isa-manual/blob/smpmpss/src/smcfiss.adoc"
+requirements:
+  extension: { name: Smcfiss, version: "~> 0.1" }
```
ThinkOpenly: I think this needs Zicfiss and "not S", right? Per the spec:
> The Smucfiss extension depends on the Zicfiss and Smcfiss extensions. The Smucfiss extension is not supported when the S-mode is implemented.

spec/std/isa/ext/Sspmpss.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - version: "0.1.0"
+    state: development
+    url: "https://github.com/ved-rivos/riscv-isa-manual/blob/smpmpss/src/smcfiss.adoc"
+requirements:
+  extension:
+    allOf:
+      - name: Zicfiss
+        version: "~> 1.0"
+      - name: S
+        version: "~> 1.13"
```
ThinkOpenly, line 27: This should be Sspmp, right?
> The Sspmpss extension depends on the Zicfiss and Sspmp extensions.

ThinkOpenly (changes requested):
Thanks for this submission! I've a couple of questions...

spec/std/isa/ext/Sspmpss.yaml
michalchm: Yes, Sspmp is the correct dependency. Since UDB does not define it yet, should I add it as a dependency now? If so, should it target the ratified version 1.0?

spec/std/isa/ext/Smucfiss.yaml
michalchm: Correct. Smucfiss should require Zicfiss and exclude S. The problem is that Zicfiss currently requires S whenever U is present, so encoding these requirements directly would make Smucfiss unsatisfiable. How would you prefer to handle this?

ThinkOpenly: So this would need to be something like (for Zicfiss, untested):
```
  if:
    allOf:
      extension:
        name: U
      not:
        extension:
          name: Smucfiss
  then:
    extension:
      name: S
```
Does that look right?

spec/std/isa/ext/Sspmpss.yaml
ThinkOpenly: I'll try to get Sspmp merged.

ThinkOpenly: reference to #1817

spec/std/isa/ext/Smucfiss.yaml
michalchm: Yes, the logic is correct. I updated this PR with Smucfiss requirements and the corresponding Zicfiss adjustment.

spec/std/isa/ext/Sspmpss.yaml
michalchm: Thanks. I will update Sspmpss to depend on Sspmp once #1817 is merged.

PR #2512, opened 2026-08-22
fix(udb): CSR length condition omits U and VU mode

titoatwork:
`Csr` keeps a second copy of the "can the effective XLEN change here" question, hardcoded to M, S and VS. `CsrField#dynamic_location?` already asks it over `modes_with_access` via `multi_xlen_in_mode?`, which covers every mode. This routes `Csr` through the same helper and builds the condition string from that mode list instead of hardcoding it.

Three things the hardcoded string got wrong:

- U and VU are never named, so the nine `priv_mode: U` CSRs with `length: XLEN` (`jvt`, `ssp` and the seven vector CSRs) emit a condition that is false whenever the hart is executing in U-mode.
- `tdata1`, `tdata2`, `tdata3` and `tselect` are M-mode only, but got the S and VS disjuncts anyway.
- `length_pretty` filled the encoding in with `String#sub`, which substitutes only the first of the three placeholders and leaves a literal `%%` in the rendered length line.

The per-mode field mapping follows `base32?` in `spec/std/isa/isa/globals.isa`.

`test_csr.rb` stubs `multi_xlen_in_mode?` rather than `param_values`, since that is the collaborator being consulted now. All five existing cases are kept, with six added for the cases above.

Second of the two PRs from #2463.

### DIFF
--- a/tools/ruby-gems/udb/lib/udb/obj/csr.rb
+++ b/tools/ruby-gems/udb/lib/udb/obj/csr.rb
@@ -21,9 +21,8 @@ module Udb
     BITS128_TYPE = Idl::Type.new(:bits, width: 128).freeze
     ENCODING_SIZE_VAR = Idl::Var.new("__instruction_encoding_size", BITS6_TYPE, 32).freeze
     # Privilege mode => IDL expression naming the field that selects the effective
-    # XLEN in that mode. Mirrors the mode dispatch in `base32?` in globals.isa.
-    # Debug mode executes at MXLEN, so it shares M-mode's field.
-    XLEN_FIELD_BY_MODE = {
+    # XLEN in that mode. Debug mode executes at MXLEN, so it shares M-mode's field.
+    MODE_TO_XLEN_FIELD = {
       "M" => "CSR[misa].MXL",
       "D" => "CSR[misa].MXL",
       "S" => "CSR[mstatus].SXL",
@@ -31,6 +30,7 @@ module Udb
       "VS" => "CSR[hstatus].VSXL",
       "VU" => "CSR[vsstatus].UXL"
     }.freeze
+    XLEN_TO_ENCODING = { 32 => 0, 64 => 1 }.freeze
     # Add all methods in this module to this type of database object.
     include HasFields
 
@@ -204,9 +204,6 @@ module Udb
         # dynamic if either we don't know VSXLEN or VSXLEN is explicitly mutable
         !cfg_arch.param_values.key?("VSXLEN") || cfg_arch.param_values["VSXLEN"].size > 1
       when "XLEN"
-        # The effective XLEN follows whichever mode is executing, so the length is
-        # dynamic when any mode that can reach this CSR has a variable XLEN.
-        # CsrField#dynamic_location? already asks the question this way.
         modes_with_dynamic_xlen.any?
       else
         raise "Unexpected length"
@@ -354,22 +351,23 @@ module Udb
       modes_with_access.select { |mode| cfg_arch.multi_xlen_in_mode?(mode) }
     end
 
-    # @param encoding [String] "0" for 32-bit, "1" for 64-bit
-    # @return [String] IDL condition of when the effective xlen has the given encoding
-    def length_cond(encoding)
+    # @param xlen [Integer] 32 or 64
+    # @return [String] IDL condition of when the effective xlen is +xlen+
+    def length_cond(xlen)
+      encoding = XLEN_TO_ENCODING.fetch(xlen)
       case @data["length"]
       when "MXLEN"
-        "CSR[misa].MXL == #{encoding}"
+        "#{MODE_TO_XLEN_FIELD['M']} == #{encoding}"
       when "SXLEN"
-        "CSR[mstatus].SXL == #{encoding}"
+        "#{MODE_TO_XLEN_FIELD['S']} == #{encoding}"
       when "VSXLEN"
-        "CSR[hstatus].VSXL == #{encoding}"
+        "#{MODE_TO_XLEN_FIELD['VS']} == #{encoding}"
       when "XLEN"
         modes = modes_with_dynamic_xlen
         raise "No mode with access to #{name} has a dynamic XLEN" if modes.empty?
 
         modes.map { |mode|
-          "(priv_mode() == PrivilegeMode::#{mode} && #{XLEN_FIELD_BY_MODE.fetch(mode)} == #{encoding})"
+          "(priv_mode() == PrivilegeMode::#{mode} && #{MODE_TO_XLEN_FIELD.fetch(mode)} == #{encoding})"
         }.join(" || ")
       else
         raise "Unexpected length #{@data['length']} for #{name}"
@@ -377,10 +375,10 @@ module Udb
     end
 
     # @return [String] IDL condition of when the effective xlen is 32
-    def length_cond32 = length_cond("0")
+    def length_cond32 = length_cond(32)
 
     # @return [String] IDL condition of when the effective xlen is 64
-    def length_cond64 = length_cond("1")
+    def length_cond64 = length_cond(64)
 
     # @param effective_xlen [Integer or nil] 32 or 64 for fixed xlen, nil for dynamic
     # @return [String] Pretty-printed length string
@@ -389,8 +387,8 @@ module Udb
       if dynamic_length?
         if effective_xlen.nil?
           [
-            "* #{length(32)} when #{length_cond('0')}",
-            "* #{length(64)} when #{length_cond('1')}"
+            "* #{length(32)} when #{length_cond(32)}",
+            "* #{length(64)} when #{length_cond(64)}"
           ].join("\n")
         else
           "#{length(effective_xlen)}-bit"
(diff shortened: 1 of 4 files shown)


tools/ruby-gems/udb/lib/udb/obj/csr.rb
```diff
@@ -341,61 +348,49 @@ def max_length
...
-        "CSR[mstatus].SXL == 1"
-      when "VSXLEN"
-        "CSR[hstatus].VSXL == 1"
-      when "XLEN"
-        "(priv_mode() == PrivilegeMode::M && CSR[misa].MXL == 1) || (priv_mode() == PrivilegeMode::S && CSR[mstatus].SXL == 1) || (priv_mode() == PrivilegeMode::VS && CSR[hstatus].VSXL == 1)"
-      else
-        raise "Unexpected length"
-      end
-    end
+    def length_cond64 = length_cond("1")
```
ThinkOpenly: At least these would read better if the arguments had some intrinsic meaning:
```suggestion
    def length_cond32 = length_cond(32)

# @return [String] IDL condition of when the effective xlen is 64
    def length_cond64 = length_cond(64)
```
`length_cond` would have to change to map those inputs to the appropriate encoding, of course.

tools/ruby-gems/udb/lib/udb/obj/csr.rb
```diff
@@ -341,61 +348,49 @@ def max_length
...
       case @data["length"]
       when "MXLEN"
-        "CSR[misa].MXL == 0"
+        "CSR[misa].MXL == #{encoding}"
       when "SXLEN"
-        "CSR[mstatus].SXL == 0"
+        "CSR[mstatus].SXL == #{encoding}"
       when "VSXLEN"
-        "CSR[hstatus].VSXL == 0"
+        "CSR[hstatus].VSXL == #{encoding}"
```
ThinkOpenly: Should we use our fancy mode-to-field mapping here, too?
```suggestion
        "#{XLEN_FIELD_BY_MODE['M']} == #{encoding}"
when "SXLEN"
        "#{XLEN_FIELD_BY_MODE['S']} == #{encoding}"
when "VSXLEN"
        "#{XLEN_FIELD_BY_MODE['VS']} == #{encoding}"
```

tools/ruby-gems/udb/lib/udb/obj/csr.rb
```diff
@@ -20,6 +20,17 @@ class Csr < TopLevelDatabaseObject
     BITS6_CONST_TYPE = Idl::Type.new(:bits, width: 6, qualifiers: [:const]).freeze
     BITS128_TYPE = Idl::Type.new(:bits, width: 128).freeze
     ENCODING_SIZE_VAR = Idl::Var.new("__instruction_encoding_size", BITS6_TYPE, 32).freeze
+    # Privilege mode => IDL expression naming the field that selects the effective
+    # XLEN in that mode. Mirrors the mode dispatch in `base32?` in globals.isa.
+    # Debug mode executes at MXLEN, so it shares M-mode's field.
+    XLEN_FIELD_BY_MODE = {
```
ThinkOpenly:
```suggestion
    MODE_TO_XLEN_FIELD = {
```

tools/ruby-gems/udb/lib/udb/obj/csr.rb
```diff
@@ -193,14 +204,10 @@ def dynamic_length?
...
-        # SXLEN condition applies if S-mode is possible
-        # VSXLEN condition applies if VS-mode is possible
-        (cfg_arch.mxlen.nil?) || \
-        (cfg_arch.possible_extensions.map(&:name).include?("S") && \
-          (cfg_arch.param_values["SXLEN"].nil? || cfg_arch.param_values["SXLEN"].size > 1)) || \
-        (cfg_arch.possible_extensions.map(&:name).include?("H") && \
-          (cfg_arch.param_values["VSXLEN"].nil? || cfg_arch.param_values["VSXLEN"].size > 1))
+        # The effective XLEN follows whichever mode is executing, so the length is
+        # dynamic when any mode that can reach this CSR has a variable XLEN.
+        # CsrField#dynamic_location? already asks the question this way.
```
ThinkOpenly: Comments that sortof reference other code that itself might change (and therefore must be independently maintained) are discouraged. Written by AI?
```suggestion
```

tools/ruby-gems/udb/test/test_csr.rb
```diff
@@ -12,93 +12,173 @@
...
     csr
   end
 
-  # M + S, MXLEN known, SXLEN mutable, NO H extension.
   # S-mode alone makes the effective XLEN variable, so the CSR IS dynamic.
   # (Regression pin: the buggy `A || B && C` precedence returns false here.)
   def test_xlen_dynamic_when_only_sxlen_mutable
-    arch = make_cfg_arch(mxlen: 32, exts: %w[A S], param_values: { "SXLEN" => [32, 64] })
-    csr = make_csr(arch, length: "XLEN")
+    csr = make_csr(make_cfg_arch(dynamic_modes: %w[S]), length: "XLEN", priv_mode: "S")
```
ThinkOpenly: Combined changes make PRs difficult to review. Please split back to 2 lines. Only make changes with good reason. If you must refactor code, do that in a separate commit that has no functional impact.

tools/ruby-gems/udb/test/test_csr.rb
```diff
@@ -12,93 +12,173 @@
...
       "XLEN CSR must be dynamic when S-mode is possible and SXLEN is mutable, even without H"
   end
 
-  # M + H, MXLEN known, VSXLEN mutable, NO S extension.
-  # H-mode alone makes the effective XLEN variable, so the CSR IS dynamic.
+  # VS-mode alone makes the effective XLEN variable, so the CSR IS dynamic.
   def test_xlen_dynamic_when_only_vsxlen_mutable
-    arch = make_cfg_arch(mxlen: 32, exts: %w[A H], param_values: { "VSXLEN" => [32, 64] })
-    csr = make_csr(arch, length: "XLEN")
+    csr = make_csr(make_cfg_arch(dynamic_modes: %w[VS]), length: "XLEN", priv_mode: "S")
```
ThinkOpenly: same. this construct appears many time below as well.

ThinkOpenly (changes requested):
This is a very large PR for what could have been (IMHO) a fairly small fix. I'm not objecting to some refactoring, but the caveat is that a reviewer could punt instead of spending the time on it. Advice: keep things small and isolated. Multiple PRs if necessary. Make the reviewers job as easy as possible. Other comments inline...

tools/ruby-gems/udb/lib/udb/obj/csr.rb
titoatwork: Agreed, done. Reads better next to `length(32)` in `length_pretty` as well.

tools/ruby-gems/udb/lib/udb/obj/csr.rb
titoatwork: Yes, done. The three map one to one onto M, S and VS.

tools/ruby-gems/udb/lib/udb/obj/csr.rb
titoatwork: Done. Matches `OP_TO_KEY` and `TYPE_VALUE_TO_CPP_NAME` too, so that's the more consistent name.

tools/ruby-gems/udb/lib/udb/obj/csr.rb
titoatwork: Fair question. I use it for first drafts and verify the substance myself before anything goes out, which is why the tests came with the change rather than after it. What slipped through here is the prose habit rather than the logic, and that is the part I need to tighten.

Gone, along with two others of the same kind elsewhere in the diff: the `base32?` line above the constant and the header block in `test_csr.rb`.

tools/ruby-gems/udb/test/test_csr.rb
titoatwork: Split back to two lines.

tools/ruby-gems/udb/test/test_csr.rb
titoatwork: Same here. There were 13 of these, all back to two lines now.

tools/ruby-gems/udb/test/test_csr.rb
```diff
@@ -12,93 +12,183 @@
...
+  #
+  # @param dynamic_modes [Array<String>] modes whose effective XLEN can vary
+  def make_cfg_arch(dynamic_modes:)
     arch = Udb::ConfiguredArchitecture.allocate
     arch.instance_variable_set(:@objects, Concurrent::Hash.new)
     arch.instance_variable_set(:@object_hashes, Concurrent::Hash.new)
-    ext_objs = exts.map { |name| OpenStruct.new(name: name) }
-    arch.define_singleton_method(:mxlen) { mxlen }
-    arch.define_singleton_method(:possible_extensions) { ext_objs }
-    arch.define_singleton_method(:param_values) { param_values }
```
ThinkOpenly, line 37: So, these things were found to be unnecessary for these tests?

titoatwork: Yes. The XLEN branch of `dynamic_length?` now calls `multi_xlen_in_mode?` and nothing else, so the other three stubs are dead. Confirmed by stubbing `mxlen`, `param_values` and `possible_extensions` to raise: all 14 cases still pass.

The `require "ostruct"` at the top was only there for that `OpenStruct` line, so I dropped that as well.

Worth flagging separately: `multi_xlen_in_mode?` has no direct test of its own, here or on main. Stubbing it narrows what this file covers without handing that coverage anywhere else. Happy to add tests for it in its own PR.

ThinkOpenly (approved):

Issue #2511, opened 2026-08-22
Relax COUNTINHIBIT_EN


davidharrishmc:
Presently UDB requires COUNTINHIBIT_EN[n] to be false when HPM_COUNTER_EN[n] is false.  
However, this is not a requirement of the ratified spec. Whisper and spike allow mcountinhibit bits
to be written even for non-counting HPM counters.

titoatwork:
Line numbers below are at `5f91ed1f` here and `1d472b80` on riscv/riscv-isa-manual.

Confirmed, and it is enforced rather than advisory. `spec/std/isa/param/COUNTINHIBIT_EN.yaml:29` carries

```yaml
requirements:
  idl(): |
    for (U32 i = 3; i < 32; i++) {
      (!HPM_COUNTER_EN[i]) -> !COUNTINHIBIT_EN[i];
    }
```

and `cfg_arch.rb:330` evaluates it for every fully configured cfg, failing with "Parameter requirements not met". So a config that models spike or Whisper honestly cannot validate.

The manual is on your side here. The `mcountinhibit` section at `machine.adoc:1617` carries five normative rules, `sz_warl_op1`, `only_inc`, `op2`, `cy_shared` and `not_impl`, and none of them constrains bits for unimplemented counters. WARL allows hardwiring the bit to zero and equally allows making it writable, so forbidding the second is UDB's choice rather than the spec's.

One thing worth flagging before someone implements this. Dropping the `requirements:` block on its own makes the combination legal but inert. Each of the 29 HPM fields is separately guarded by `definedBy: param HPM_COUNTER_EN index n equal true`, and `CsrField#exists_in_cfg?` at `csr_field.rb:86` requires that to be `SatisfiedResult::Yes`, so with `HPM_COUNTER_EN[n]` false the field does not exist at all, whatever `COUNTINHIBIT_EN[n]` says. `type()` already keys on `COUNTINHIBIT_EN` alone, so the outstanding change is the `definedBy`. That half does churn generated output for the seven fully configured cfgs, whereas removing the requirement cannot invalidate any config that validates today.

Is exposing the field when the counter is hardwired zero what you are after?

Separately, all 29 of those fields carry this description at `spec/std/isa/csr/Zicntr/mcountinhibit.layout:90`, under a `[when="COUNTINHIBIT_EN[<%= hpm_num %>] == false"]` guard:

```
Since hpmcounter<%= hpm_num %> is not implemented, this field is read-only zero.
```

[...]

PR #2523, opened 2026-08-24
docs(csr): correct the read-only reason on mcountinhibit HPM fields

titoatwork:
Each `mcountinhibit.HPM<n>` field is `definedBy` `HPM_COUNTER_EN[n] == true`, so it exists only where `hpmcounter<n>` is implemented. Its `COUNTINHIBIT_EN[n] == false` branch reads `Since hpmcounter<n> is not implemented, this field is read-only zero`, which names the one condition under which the field does not exist.

`COUNTINHIBIT_EN[n] == false` means the counter is implemented but cannot be inhibited, which is what `type()` already returns.

One line in the layout, regenerated across all 29 fields in `mcountinhibit.yaml`. Descriptions only, no `definedBy` or `type()` change.

related to #2511

### DIFF
--- a/spec/std/isa/csr/Zicntr/mcountinhibit.layout
+++ b/spec/std/isa/csr/Zicntr/mcountinhibit.layout
@@ -87,7 +87,7 @@ fields:
       When set, `hpmcounter<%= hpm_num %>.COUNT` stops counting in all privilege modes.
 
       [when="COUNTINHIBIT_EN[<%= hpm_num %>] == false"]
-      Since `hpmcounter<%= hpm_num %>.COUNT` cannot be inhibited, this field is read-only zero.
+      Since `hpmcounter<%= hpm_num %>` cannot be inhibited, this field is read-only zero.
     type(): |
       return COUNTINHIBIT_EN[<%= hpm_num %>] ? CsrFieldType::RW : CsrFieldType::RO;
     reset_value(): |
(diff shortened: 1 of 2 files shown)


spec/std/isa/csr/Zicntr/mcountinhibit.layout
```diff
@@ -87,7 +87,7 @@ fields:
       When set, `hpmcounter<%= hpm_num %>.COUNT` stops counting in all privilege modes.
 
       [when="COUNTINHIBIT_EN[<%= hpm_num %>] == false"]
-      Since hpmcounter<%= hpm_num %> is not implemented, this field is read-only zero.
+      Since `hpmcounter<%= hpm_num %>.COUNT` cannot be inhibited, this field is read-only zero.
```
ThinkOpenly:
```suggestion
      Since `hpmcounter<%= hpm_num %>` cannot be inhibited, this field is read-only zero.
```

ThinkOpenly (changes requested):

spec/std/isa/csr/Zicntr/mcountinhibit.layout
titoatwork: Thanks, applied in 5d26e1e. Regenerated `mcountinhibit.yaml` from the layout so all 29 HPM fields carry the new wording.

ThinkOpenly (approved):

PR #2526, opened 2026-08-26
feat: Add parameters for WFI timeout and U-Mode support

### LINKED ISSUE #2520: Add UDB parameters for WFI timeout and WFI support in U-mode
Some WFI tests in the `riscv-arch-test` suite require DUT-specific WFI behavior that is not supported by all implementations.

1. **WFI timeout:** The `wfi_timeout` test can stall indefinitely on DUTs that do not define a WFI timeout behavior. Such DUTs cannot complete the test and may eventually terminate due to the infinite stall.

2. **WFI in U-mode:** The WFI tests in U-mode should only run on DUTs that support WFI in U-mode. For example, Spike and QEMU do not support WFI in U-mode and generate an illegal-instruction exception, resulting in a signature mismatch.

These tests should therefore be guarded by UDB parameters so that DUTs can indicate whether they support the required behavior.

**Add the following two UDB parameters:**

1. `wfi_timeout` — indicates whether the DUT defines a finite timeout/fall-through behavior for WFI. The WFI timeout test should be guarded by this parameter so that DUTs without such behavior can exclude the test.

2. `wfi_u_mode` — indicates whether the DUT supports WFI in U-mode. The WFI-in-U-mode test should be guarded by this parameter so that DUTs that do not support WFI in U-mode can exclude the test.

neelkumar01:
### Summary

This PR adds two new UDB parameters for WFI timeout behavious and U mode support

- `WFI_TIMEOUT` : Indicates whether execution resumes after WFI within finite amount of time
- `WFI_U_MODE` : Indicates whether WFI is supported in U mode

### Why is this needed?

Without these parameters:

- The WFI timeout test may wait indefinitely on implementations that do not provide a finite timeout behavior
- the U mode WFI test may fail on implementations where WFI is not supported in U mode

These parameters make that support explicit so tests can avoid running cases that may hang or fail on unsupported implementations

### Changes

Added:

- `spec/std/isa/param/WFI_TIMEOUT.yaml`
- `spec/std/isa/param/WFI_U_MODE.yaml`

Both parameters are boolean flags and follow the existing UDB parameter schema and file structure

Closes #2520


ThinkOpenly (changes requested):
A few comments:
- Could you fix the "end of file" condition which GitHub is complaining about? I think the CI tests will also complain/fail.
- Could you add support for `WFI_U_MODE` to `wfi.yaml`? Currently, it fails immediately in U mode:
  ```
  if (mode() == PrivilegeMode::U) {
  }
  ```
- Can we change "WFI_TIMEOUT" to something like "WFI_FINITE"? ("TIMEOUT" to me sounds like it should represent some sort of time unit, instead of a boolean.)

ThinkOpenly (approved):

ThinkOpenly:
@neelkumar01, any objection to reopening this?

spec/std/isa/param/WFI_U_MODE.yaml
```diff
@@ -0,0 +1,20 @@
...
+long_name: |
+  Whether WFI is supported in U mode
+definedBy:
+  extension:
+    name: Sm
+description: |
+  Options:
+
+    * true: WFI can be executed in U mode
+    * false: WFI cannot be executed in U mode
```
davidharrishmc, line 18: Should this be "WFI throws an illegal instruction immediately if executed in U-mode"?

PR #2566, opened 2026-09-02
fix(idl,cpp): add 8- and 16-bit atomic RMW paths for Zabha

### LINKED ISSUE #2432: Atomics narrower than XLEN are modeled incorrectly

aashu2006:
`amo()` only checked for `N == 32` and sent everything else to the 64-bit read-modify-write, because those were the only two RMW builtins that existed. But all 72 non-CAS Zabha instructions call `amo(8, ...)` or `amo(16, ...)`, so every one of them did a 64-bit RMW: reading and writing 8 bytes at a byte address, wrecking up to 7 bytes next to it. `amomax.b`/`amomin.b` were worse, comparing a full doubleword signed against a zero-extended byte.

This adds `atomic_read_modify_write_8` and `_16` next to the existing pair, and makes `amo()` pick between all four widths. On the C++ side that means the two new `SocModel` requirements, the `hart.hpp` wrappers, the real implementations in `iss_soc_model.hpp`, and stubs in the renode and softfloat-test models. Signed `max`/`min` cast to `int8_t`/`int16_t` so the compare happens at the right width. `DenseMemory` already read and wrote 1 and 2 bytes, so nothing changed there.

With #2498 the Zabha byte/halfword AMOs are now right end to end - the RMW happens at the real width, and the `sext()` from that PR puts the correct value in `xd`.

Closes #2432.

### DIFF
--- a/spec/std/isa/isa/globals.isa
+++ b/spec/std/isa/isa/globals.isa
@@ -2978,8 +2978,10 @@ function amo {
       return atomic_read_modify_write_16(physical_address, value, op);
     } else if (N == 32) {
       return atomic_read_modify_write_32(physical_address, value, op);
-    } else {
+    } else if (N == 64) {
       return atomic_read_modify_write_64(physical_address, value, op);
+    } else {
+      unreachable();
     }
 
   }
--- a/spec/std/isa/isa/globals.isa
+++ b/spec/std/isa/isa/globals.isa
@@ -2972,7 +2972,11 @@ function amo {
       raise (ExceptionCode::StoreAmoAddressMisaligned, mode(), virtual_address);
     }
 
-    if (N == 32) {
+    if (N == 8) {
+      return atomic_read_modify_write_8(physical_address, value, op);
+    } else if (N == 16) {
+      return atomic_read_modify_write_16(physical_address, value, op);
+    } else if (N == 32) {
       return atomic_read_modify_write_32(physical_address, value, op);
     } else {
       return atomic_read_modify_write_64(physical_address, value, op);
--- a/backends/cpp_hart_gen/cpp/include/udb/hart.hpp
+++ b/backends/cpp_hart_gen/cpp/include/udb/hart.hpp
@@ -238,6 +238,14 @@ namespace udb {
                                 const PossiblyUnknownBits<32>& pte_len) {
       return atomically_set_pte_a_d(pte_paddr.get(), pte_value.get(), pte_len.get());
     }
+    Bits<8> atomic_read_modify_write_8(const PossiblyUnknownBits<64>& paddr, const PossiblyUnknownBits<8>& value,
+                                       AmoOperation op) {
+      return Bits<8>{m_soc.atomic_read_modify_write_8(paddr.get(), value.get(), op)};
+    }
+    Bits<16> atomic_read_modify_write_16(const PossiblyUnknownBits<64>& paddr, const PossiblyUnknownBits<16>& value,
+                                         AmoOperation op) {
+      return Bits<16>{m_soc.atomic_read_modify_write_16(paddr.get(), value.get(), op)};
+    }
     Bits<32> atomic_read_modify_write_32(const PossiblyUnknownBits<64>& paddr, const PossiblyUnknownBits<32>& value,
                                          AmoOperation op) {
       return Bits<32>{m_soc.atomic_read_modify_write_32(paddr.get(), value.get(), op)};
--- a/backends/cpp_hart_gen/cpp/include/udb/iss_soc_model.hpp
+++ b/backends/cpp_hart_gen/cpp/include/udb/iss_soc_model.hpp
@@ -216,6 +216,128 @@ namespace udb {
                                    uint32_t pte_len) {
       return true;
     }
+    uint64_t atomic_read_modify_write_8(uint64_t phys_addr, uint64_t value,
+                                        AmoOperation op) {
+      switch (op.value()) {
+        case AmoOperation::Swap: {
+          uint8_t orig = m_memory.read(phys_addr, 1);
+          m_memory.write(phys_addr, value, 1);
+          return orig;
+        }
+        case AmoOperation::Add: {
+          uint8_t orig = m_memory.read(phys_addr, 1);
+          m_memory.write(phys_addr, orig + value, 1);
+          return orig;
+        }
+        case AmoOperation::And: {
+          uint8_t orig = m_memory.read(phys_addr, 1);
+          m_memory.write(phys_addr, orig & value, 1);
+          return orig;
+        }
+        case AmoOperation::Or: {
+          uint8_t orig = m_memory.read(phys_addr, 1);
+          m_memory.write(phys_addr, orig | value, 1);
+          return orig;
+        }
+        case AmoOperation::Xor: {
+          uint8_t orig = m_memory.read(phys_addr, 1);
+          m_memory.write(phys_addr, orig ^ value, 1);
+          return orig;
+        }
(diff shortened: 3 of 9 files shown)


spec/std/isa/isa/globals.isa
```diff
@@ -2972,10 +2972,16 @@ function amo {
...
+      return atomic_read_modify_write_8(physical_address, value, op);
+    } else if (N == 16) {
+      return atomic_read_modify_write_16(physical_address, value, op);
+    } else if (N == 32) {
       return atomic_read_modify_write_32(physical_address, value, op);
-    } else {
+    } else if (N == 64) {
       return atomic_read_modify_write_64(physical_address, value, op);
+    } else {
+      unreachable();
```
ThinkOpenly, line 2984: This should `raise` an exception of some sort.

ThinkOpenly (commented):
With a fairly quick look, this looks pretty good to me. I can't claim to understand `soc_model.hpp`. I have one comment below.

I've asked @henrikg-qc to review.

Issue #2567, opened 2026-09-03
regress-riscv-tests-64 aborts on rv64ui-p-add with UndefinedValueError


DengZhiyuan-math:
**Describe the bug**

On current `main` (`a3ba3e6fbf8f38bd4adc4f9aeebc3fcce049525c`), the `regress-riscv-tests-64` regression generates and builds the RV64 ISS successfully, then aborts on the first test binary, `rv64ui-p-add`.

**To Reproduce**

1. Check out `a3ba3e6fbf8f38bd4adc4f9aeebc3fcce049525c`.
2. Complete the normal repository setup and ensure the RISC-V cross-toolchain is available.
3. Run `./bin/regress -n regress-riscv-tests-64`.

I reproduced this in the official arm64 toolchain image `ghcr.io/riscv/udb-toolchain:b320716f4dd643be` with GCC 14.2.1. The image currently lacks `which`, so I supplied a minimal PATH-compatible `which` shim to get past that separate runner prerequisite.

**Text output/Error Messages**

```text
Running /workspace/gen/cpp_hart_gen/rv64_Debug/build/iss -m rv64 -c /workspace/cfgs/rv64-riscv-tests.yaml ext/riscv-tests/isa/rv64ui-p-add
terminate called after throwing an instance of 'udb::UndefinedValueError'
  what():  Cannot compare unknown value
Finished ... with exit status 6 (failed)
```

The regression command exits 1. A repeated run directly on the exact upstream commit reproduced the same failure after the generated ISS and all build targets completed successfully.

**Additional context**

Issue #1047 contains the same exception text, but describes a decode-tree construction failure for base-specific instructions. This report reaches runtime execution of `rv64ui-p-add`, so it appears to be a distinct failure stage and reproduction.

ThinkOpenly:
> 1. Check out `a3ba3e6fbf8f38bd4adc4f9aeebc3fcce049525c`.
> 2. Complete the normal repository setup and ensure the RISC-V cross-toolchain is available.
> 3. Run `./bin/regress -n regress-riscv-tests-64`.

I cannot reproduce this failure.

```
$ git log -1 --oneline
a3ba3e6fbf8f (HEAD -> main, upstream/main, origin/main, origin/HEAD) fix(data): correct Zaamo long name (#2565)
$ ./bin/regress -n regress-riscv-tests-64
[...]
-- The CXX compiler identification is GNU 13.3.0
[...]

PR #2575, opened 2026-09-04
fix(test): isolate C++ unit build state

DengZhiyuan-math:
The C++ unit regression and RV64 ISS regression both used `rv64_Debug`, but configure that shared CMake cache with incompatible `IGNOREUNDEFINED` values. Running C++ unit first could therefore leave RV64 with stale build state.

Reuse the existing `BUILD_NAME` input to place C++ unit in `rv64-unit_Debug`, while RV64 continues to use `rv64_Debug`. The generated regression workflow is updated from the source YAML.

Validated in an arm64 Linux fresh clone at `4cba519e9cbcd20a3e5eb818518a6d1f46c6391e`: C++ unit followed immediately by RV64 passes without cleaning, the caches retain `IGNOREUNDEFINED=NO` and `YES` respectively, and `./bin/regress --all --jobs 1` passes 67/67. Related to #2567.

### DIFF
--- a/.github/workflows/regress.yml
+++ b/.github/workflows/regress.yml
@@ -576,7 +576,7 @@ jobs:
         with:
           toolchain: true
       - name: Run cpp unit tests
-        run: ./do test:cpp_hart CONFIG=rv64 JOBS=4 BUILD_TYPE=debug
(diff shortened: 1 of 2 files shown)


henrikg-qc (approved):

DengZhiyuan-math:
This PR has received review approval, but CI/CD and autofix.ci are still awaiting workflow approval. Could a maintainer with the necessary permissions please approve the runs? Thank you.

PR #2577, opened 2026-09-05
fix(data): make Zbc require Zbkc and define clmul/clmulh in Zbkc

### LINKED ISSUE #2535: Should Zbc require Zbkc?
The M extension requires Zmmul, here:

https://github.com/riscv/riscv-unified-db/blob/cf4fe2a724addfcb3bc025e3a1ae4d265144a7a1/spec/std/isa/ext/M.yaml#L8-L18

This was added by #2299, which justifies the change by Zmmul being a proper subset of M.

There's no such requirement for Zbc -> Zbkc, even though Zbkc is a proper subset of Zbc:

https://github.com/riscv/riscv-unified-db/blob/cf4fe2a724addfcb3bc025e3a1ae4d265144a7a1/spec/std/isa/ext/Zbc.yaml#L8-L15

Should this have a `requirements`?

I noticed this because the new ACTs use `requirements` to add implied extensions and their tests. Zbc and Zbkc have separate test suites, but if you only specify Zbc for your implementation, it doesn't run the Zbkc tests (even though you have the instructions).

This is pretty minor as they have the same ACT coverpoints, but I wanted to flag the inconsistency.

YousufFFFF:
Zbkc is a proper subset of Zbc — both define clmul and clmulh, and Zbc adds clmulr — but the database did not record that relationship. Anything that walks requirements to derive implied extensions therefore missed Zbkc for a Zbc implementation; the new ACTs hit this concretely, running only the Zbc test suite for a Zbc implementation even though the hart has the Zbkc instructions.

This applies the same modelling the database already uses for M → Zmmul (added in #2299 for the same "proper subset" reason):

clmul and clmulh move to spec/std/isa/inst/Zbkc/ and are definedBy Zbkc only, instead of anyOf: [Zbc, Zbkc].
Zbc 1.0.0 gains requirements: extension: {name: Zbkc, version: "~> 1.0.0"}, which covers both ratified Zbkc versions (1.0.0 and 1.0.1).
clmulr is untouched — it stays definedBy Zbc, which is where it belongs.
A Zbc implementation still has all three instructions, now by way of the required Zbkc. As with M, Zbc's own directly-defined instruction list is reduced to just its unique instruction.

Verified with ./do test:schema, ./do test:inst_encodings, ./bin/regress --tag smoke (17/17), and pre-commit hooks. ./do gen:cfg produces no changes to the generated profile configs.

Closes #2535

### DIFF
--- a/spec/std/isa/ext/Zbc.yaml
+++ b/spec/std/isa/ext/Zbc.yaml
@@ -17,6 +17,10 @@ versions:
   - version: "1.0.0"
     state: ratified
     ratification_date: 2021-06
+    requirements:
+      extension:
+        name: Zbkc
+        version: "~> 1.0.0"
     repositories:
       - url: https://github.com/riscv/riscv-bitmanip
         branch: main
--- a/spec/std/isa/inst/Zbkc/clmul.yaml
+++ b/spec/std/isa/inst/Zbkc/clmul.yaml
@@ -11,9 +11,7 @@ description: |
   `clmul` produces the lower half of the 2*XLEN carry-less product
 definedBy:
   extension:
-    anyOf:
-      - name: Zbc
-      - name: Zbkc
+    name: Zbkc
 assembly: xd, xs1, xs2
 encoding:
   match: 0000101----------001-----0110011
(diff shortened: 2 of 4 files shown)


ThinkOpenly (approved):
Thank you!

PR #2596, opened 2026-09-10
fix: Zilsd IDL implementation 

### LINKED ISSUE #2282: `ld` and `sd` for RV32 are broken

henrikg-qc:
includes:

- Zilsd IDL implementation
- Load and store in endianness of the current mode.

Resolves #2282


spec/std/isa/isa/globals.isa
```diff
@@ -945,12 +945,116 @@ function xlen {
...
+      } else if (M_MODE_ENDIANNESS == "dynamic") {
+        if (CSR[mstatus].MBE == 1'b1) {
+          return Endian::Big;
+        } else {
+          return Endian::Little;
+        }
+      } else {
+        assert(false, "Invalid M_MODE_ENDIANNESS value");
+        unreachable();
+      }
```
jordancarlin, line 973: Could we simplify this to the following? The `mstatus.MBE` bit should be set to the correct value even when only `Big` or only `Little` is supported.

```suggestion
    if (mode() == PrivilegeMode::M) {
        if (CSR[mstatus].MBE == 1'b1) {
          return Endian::Big;
        } else {
          return Endian::Little;
        }
      }
```

spec/std/isa/isa/globals.isa
```diff
@@ -945,12 +945,116 @@ function xlen {
...
+
 function virtual_mode? {
   returns Boolean
   description {
     Returns True if the current mode is virtual (VS or VU).
   }
   body {
+    if(endianness() == Endian::Big) {
+     return false;
+    }
```
jordancarlin, line 1056: Why does the endianness impact if we are in a virtual mode?

spec/std/isa/isa/globals.isa
```diff
@@ -2627,9 +2731,19 @@ function read_memory {
...
     if (aligned) {
-      return read_memory_aligned(LEN, virtual_address, encoding, 1'b0, 1'b0);
+      result =  read_memory_aligned(LEN, virtual_address, encoding, 1'b0, 1'b0);
+      if(LEN > 8 && endianness() == Endian::Big ) {
+        U32 tmp;
+        for (U32 i=0; i < LEN/2; i = i+8) {
+          tmp = result[(i+7):i];
+          result[(i+7):i] = result[(LEN-1-i):(LEN-8-i)];
+          result[(LEN-1-i):(LEN-8-i)] = tmp[7:0];
+        }
```
jordancarlin, line 2744: How about defining a byte-swap helper function (or something similar) and using that everywhere instead of duplicating this logic inline in multiple places?

spec/std/isa/isa/globals.isa
```diff
@@ -945,12 +945,116 @@ function xlen {
   }
 }
 
+enum Endian {
+  Little  0
+  Big     1
+}
+
+function endianness {
```
jordancarlin, line 953: There might be code elsewhere that already handles this, but how is the concept of "effective privilege mode" handled? For example, when in M-mode but `mstatus.MPRV` is set?

jordancarlin (changes requested):
Looks like you need to sign-off on your commits and also sign your commits (two similar sounding but distinct requirements).

spec/std/isa/isa/globals.isa
henrikg-qc: Good catch, that's a test artifact never removed

spec/std/isa/inst/I/ld.yaml
```diff
@@ -60,10 +60,8 @@ operation(): |
 
   if (xlen() == 32) {
     if (implemented?(ExtensionName::Zilsd)) {
```
ThinkOpenly, line 62: Outside of the scope of this PR, but this check isn't needed because of the "requirements" above that forces Zilsd support when XLEN=32.

spec/std/isa/inst/I/ld.yaml
```diff
@@ -60,10 +60,8 @@ operation(): |
...
-      Bits<64> data = read_memory(64, virtual_address, $encoding);
-
-      X[xd] = data[31:0];
-      X[xd+1] = data[63:32];
+      X[xd] = read_memory(32, virtual_address, $encoding);
+      X[xd+1] = read_memory(32, virtual_address + 4, $encoding);
```
ThinkOpenly, line 64: These need to be reversed in big endian.

spec/std/isa/inst/I/sd.yaml
```diff
@@ -52,17 +52,15 @@ access:
...
   vu: always
 operation(): |
-  Bits<64> data;
   XReg virtual_address = X[xs1] + $signed(imm);
 
   if (xlen() == 32) {
     if (implemented?(ExtensionName::Zilsd)) {
-      data = {X[xs2 + 1], X[xs2]};
+      write_memory(32, virtual_address, X[xs2], $encoding);
+      write_memory(32, virtual_address + 4, X[xs2 + 1], $encoding);
```
ThinkOpenly, line 60: These need to be reversed in big endian.

spec/std/isa/inst/I/sd.yaml
```diff
@@ -52,17 +52,15 @@ access:
   vs: always
   vu: always
 operation(): |
-  Bits<64> data;
   XReg virtual_address = X[xs1] + $signed(imm);
 
   if (xlen() == 32) {
     if (implemented?(ExtensionName::Zilsd)) {
```
ThinkOpenly, line 58: This isn't needed given the "requirements".

spec/std/isa/isa/globals.isa
ThinkOpenly: Why are we byte-swapping here at all?

spec/std/isa/isa/globals.isa
ThinkOpenly: True for the other stanzas below as well, presumably.

ThinkOpenly (changes requested):

spec/std/isa/inst/I/ld.yaml
henrikg-qc: From the spec Chapter34:

"Regardless of endianness, the lower-numbered register holds the low-order bits, and the highernumbered
register holds the high-order bits: e.g., bits 31:0 of an operand in Zilsd might be held in
register x14, with bits 63:32 of that operand held in x15."

spec/std/isa/inst/I/ld.yaml
henrikg-qc: Correct. I can optimize this block

spec/std/isa/inst/I/sd.yaml
henrikg-qc: "Regardless of endianness, the lower-numbered register holds the low-order bits, and the highernumbered
register holds the high-order bits: e.g., bits 31:0 of an operand in Zilsd might be held in
register x14, with bits 63:32 of that operand held in x15."

spec/std/isa/isa/globals.isa
henrikg-qc: read_memory_aligned return memory content read as LE always. If BE we swap byte order, Functions read/write_memory are the only places LE/BE conversion is taking place

henrikg-qc: Yes, we can create a byteswapper, I was first not in favor as return arg can be of different size but I can return XReg, need to bounds check though based on xlen()

tests/data/golden/Xqci@0.13.0.adoc
```diff
@@ -39489,6 +39489,109 @@ if pass:[(]xref:#udb-function-mode[mode]pass:[(]) == PrivilegeMode::M) {
 return xref:#udb-function-mode[mode]pass:[(]);
 ----
 
+[#udb-function-endianness]
```
henrikg-qc, line 39492: What is this? I never comitted this file. I assume this is generated for Xcqi and all is well?

spec/std/isa/inst/I/ld.yaml
ThinkOpenly: Correct, but bits 63:32 are in the lower memory addresses in BE and in the higher memory addresses in LE.

henrikg-qc (commented):
As I am putting in the sign-off and signing I may have to force push. I hope that will maintain our comments. 
If the PR changes identity I will let you know and summarize what has been covered if we lose PR review comments

spec/std/isa/isa/globals.isa
henrikg-qc: It is not, let me check the spec again on that and update the implementation

jordancarlin: That seems like it might be a much larger issue because "effective privilege mode" impacts a lot of things beyond just endianness (PMP, address translation, and pointer masking just to name a few). There are also extra implications with what the effective privilege mode is when Hypervisor support is implemented. The approach the Sail model took (and is likely to be unavoidable for UDB too) was to make all of the memory access functions take an explicit privilege mode as an argument instead of using the current priv mode. Then the calling site can determine the appropriate effective privilege mode.

tests/data/golden/Xqci@0.13.0.adoc
jordancarlin: It is a file that is automatically updated by auto-fix. Not sure why Xqci includes a listing off the internal IDL functions, but that is a question for a different issue.

spec/std/isa/isa/globals.isa
ThinkOpenly: > read_memory_aligned return memory content read as LE always.

Really!? How is that decided? We should really document that if true. What if I'm running natively in a BE environment? (It seems to me that a native "read memory" function should return data in the current endian mode.)

I see Jordan has brought up "effective privilege mode" in another thread here. I wasn't aware of that. So, yeah, things get weird.

spec/std/isa/isa/globals.isa
ThinkOpenly: Interesting. Since we probably want this behavior "exposed" in IDL, I suggest we add a parameter to read_memory_aligned (and any peers) to specify the endianness of the accesses, and let it handle any byte swapping under the covers. (Those functions might need to byte swap anyway, to match a native BE access to a current LE mode or vice-versa, and I think it preferable to hide the native endianness from the implementation here.)

henrikg-qc: effective_ldst_mode() solves this problem. All read/writes where endianness matters goes through read/write_memory, so byteswapping will take place there

spec/std/isa/isa/globals.isa
henrikg-qc: Let me clarify: read_memory_aligned is always called through read_memory. read_mememory does the endianess swap, same thing for writes.

henrikg-qc: The IDL compiler hangs when I create a byte_swap function returning XReg and taking len and register value as argument. I cannot figure out why. The call stack after ctrl-C is massive

```
function byte_swap {
  returns XReg
  arguments
    U32 LEN,
    XReg value    
  description {
    Swaps bytes in register based on LEN.
  }
  body {    
    XReg result;
    
    result = 0;
    for (U32 i=0; i < LEN; i = i+8) {
      result[(i+7):i] = value[(LEN-1-i):(LEN-8-i)];      
    }
    return result;
  }
}
```

PR #2607, opened 2026-09-14
ci: add udb-64-max vector arch-test regression

jpak-qc:
Add a manual integration regression that runs the riscv-arch-test vector suites against the fully configured udb-64-max C++ hart.

The runner pins riscv-arch-test to edfa48d3, which includes the upstream fix for the RV32-only indexed truncation generator case, and resets reused checkouts to that exact revision.


tools/test/run_udb_64_max_vector_arch_test.sh
jordancarlin, line 1: Let's keep the actual config files (linker script, rvmodel macros, sail config, etc) in this repo rather than a sort of hacky shell script that creates them.

tools/test/gen_regress.py
```diff
@@ -32,7 +32,7 @@ def load_yaml(path: Path) -> dict[str, Any]:
 
 def validate_test(test_name: str, test_data: dict[str, Any]) -> None:
     ci_stage = test_data["ci_stage"]
-    if ci_stage not in {"pr", "merge_queue"}:
+    if ci_stage not in {"pr", "merge_queue", "manual"}:
```
jordancarlin: What is the point of adding manual? Why don't we want this running in CI?

jordancarlin (changes requested):

tools/test/run_udb_64_max_vector_arch_test.sh
jpak-qc: Done in 18871fed. The linker script, RV model macros, Sail configuration, and test configuration now live under cfgs/arch-test/udb-64-max. The runner copies that repository-owned configuration into the pinned riscv-arch-test checkout; it only generates run_cmd.txt because it contains the local ISS build path.

tools/test/gen_regress.py
jpak-qc: I thought I would have it as manual until everything landed, but we definitely should have it running in CI. Moved this regression to merge-queue CI in 18871fed and removed the manual CI stage support. The full architecture-test run has a 120-minute timeout, so merge queue makes it mandatory before merge without adding that cost to every PR update.

jordancarlin: Having this kind of test only run in the merge queue will be confusing for users when the tests fail. It should run as part of the `pr` step. If it is taking too long, we should split it across multiple CI runners. For example, have one CI job that runs Vx* and one that runs Vls*. Split it into whatever groups are necessary to get a reasonable runtime.

tools/test/run_udb_64_max_vector_arch_test.sh
jordancarlin: You shouldn't need to copy the files. Just use their existing location as the value passed to CONFIG_FILES.

tools/test/gen_regress.py
jordancarlin, line 1: Why are the changes in this file needed?

tools/test/gen_regress.py
jpak-qc: That makes sense, made changes accordingly then. The regression now runs as three separate jobs Vx, Vls, and Vf. This should keep failures visible while getting reasonable runtimes.

tools/test/gen_regress.py
jpak-qc: Yeah they were not needed, changes have been removed and restored back to how it was.

tools/test/run_udb_64_max_vector_arch_test.sh
jpak-qc: Now directly passes to CONFIG_FILES.

PR #2611, opened 2026-09-15
Add the Zkr seed CSR, with entropy source support in the C++ backend

### LINKED ISSUE #561: Add Zkr CSR

abkshaqster:
Adds the seed CSR (0x015) defined by the Zkr entropy source extension, along with the mseccfg access-control bits that gate it and a read_entropy builtin implemented in the cpp_hart_gen backend.

This supersedes #2236 and #913. @jordancarlin handed #2236 over to me; his three commits are cherry-picked here with authorship preserved. @henrikg-qc's ISS entropy implementation from the henrikg-qc/add-zkr-seed-csr branch is picked in the same way — it was offered for cherry-picking on #913 in October 2025 and hadn't been taken up. Neither the YAML nor the Zkr.yaml changes from #913's lineage are included.

Closes #561.

What's here

Specification data

- spec/std/isa/csr/Zkr/seed.yaml — the seed CSR: length: 32, priv_mode: U, writable: true, fields OPST (31:30) and ENTROPY (15:0)
- spec/std/isa/csr/mseccfg.yaml — the USEED (bit 8) and SSEED (bit 9) access-control fields
- spec/std/isa/param/{USEED,SSEED}_READ_ONLY_ZERO.yaml — parameters for implementations that hardwire those bits to zero, which the spec explicitly permits so software can probe whether seed access can be enabled
- spec/std/isa/isa/builtin_functions.idl — the read_entropy builtin, returning Bits<16>

C++ backend — read_entropy() plumbed through hart.hpp, the SocModel concept, both concrete SoC models, and the FnPointerSocModel struct in the ERB templates.

Access control

sw_read() implements all seven rows of the access-control table. The read-only-instruction check runs first, so it applies in every mode regardless of mseccfg, and a read-only access from VS/VU with sseed=1 correctly raises illegal rather than virtual.

The encoding test is $encoding[13] == 1'b1 && $encoding[19:15] == 5'b00000. Bit 13 is funct3[1], set for exactly CSRRS/CSRRC/CSRRSI/CSRRCI; bits 19:15 hold rs1 in the register forms and uimm in the immediate forms, so one zero check covers both rs1 == x0 and uimm == 0. There's a comment to that effect in the file.

Validation

./bin/prek run --all-files clean across the tree, and ./bin/regress --tag smoke passes 17/17, including regress-idl-typecheck-smoke for both rv64 and qc_iu.


spec/std/isa/csr/Zkr/seed.yaml
```diff
@@ -0,0 +1,70 @@
...
+  } else if (mode() == PrivilegeMode::VS || mode() == PrivilegeMode::VU) {
+    if (CSR[mseccfg].SSEED == 1'b1) {
+      raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
+    } else {
+      raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+    }
+  }
+
+    # The entropy source always reports ES16 with freshly polled entropy.
+    return {2'b10, 14'd0, read_entropy()};
```
jordancarlin: I haven't looked through this in detail yet, but should OPST come from a builtin too? Our default ISS implementation could always return ES16, but other users may want to model WAIT or BIST states as well. It also models the actual spec better that way.

spec/std/isa/csr/Zkr/seed.yaml
```diff
@@ -0,0 +1,70 @@
...
+  } else if (mode() == PrivilegeMode::VS || mode() == PrivilegeMode::VU) {
+    if (CSR[mseccfg].SSEED == 1'b1) {
+      raise(ExceptionCode::VirtualInstruction, mode(), $encoding);
+    } else {
+      raise(ExceptionCode::IllegalInstruction, mode(), $encoding);
+    }
+  }
+
+    # A single poll returns both the status and the entropy it goes with.
+    return read_seed();
```
ThinkOpenly: indentation is off here

backends/cpp_hart_gen/cpp/test/test_softfloat_fp.cpp
```diff
@@ -70,6 +70,7 @@ struct NullSocModel {
   void write_device_32(uint64_t, uint32_t) {}
   void sync_read_after_write_device(bool, uint32_t) {}
   void sync_write_after_read_device(bool, uint32_t) {}
+  uint32_t read_seed() { return 0x2u << 30; }
```
ThinkOpenly: Should we add a comment why we're returning a ENTROPY value of zero here?

backends/cpp_hart_gen/cpp/src/libhart_renode.cpp
```diff
@@ -103,6 +104,12 @@ struct RenodeSocModel {
                          uint32_t len) {
     return 0;
   }
+  uint32_t read_seed() {
+    // Always report ES16 (0b10) with fresh entropy in the low 16 bits.
+    // Seeded deterministically so simulation runs are reproducible.
+    static std::mt19937 gen{0x5EEDu};
+    return (0x2u << 30) | (gen() & 0xffffu);
```
ThinkOpenly: The comment uses `0b10`, so let's use that here, too.
```suggestion
    return (0b10u << 30) | (gen() & 0xffffu);
```

backends/cpp_hart_gen/cpp/include/udb/iss_soc_model.hpp
```diff
@@ -355,6 +356,12 @@ namespace udb {
     void sync_read_after_write_device(bool, uint32_t) {}
 
     void sync_write_after_read_device(bool, uint32_t) {}
+    uint32_t read_seed() {
+      // Always report ES16 (0b10) with fresh entropy in the low 16 bits.
+      // Seeded deterministically so simulation runs are reproducible.
+      static std::mt19937 gen{0x5EEDu};
+      return (0x2u << 30) | (gen() & 0xffffu);
```
ThinkOpenly:
```suggestion
      return (0b10u << 30) | (gen() & 0xffffu);
```

backends/cpp_hart_gen/cpp/include/udb/iss_soc_model.hpp
```diff
@@ -355,6 +356,12 @@ namespace udb {
     void sync_read_after_write_device(bool, uint32_t) {}
 
     void sync_write_after_read_device(bool, uint32_t) {}
+    uint32_t read_seed() {
```
ThinkOpenly: by convention, add a blank line just above
```suggestion

    uint32_t read_seed() {
```

backends/cpp_hart_gen/cpp/include/udb/hart.hpp
```diff
@@ -250,6 +250,9 @@ namespace udb {
                         PossiblyUnknownBits<64> len) {
       return m_soc.pma_applies_Q_(attr, start_paddr.get(), len.get());
     }
+    PossiblyUnknownBits<32> read_seed() {
```
ThinkOpenly: Why `PossiblyUnknownBits` here?

backends/cpp_hart_gen/cpp/include/udb/iss_soc_model.hpp
henrikg-qc: uint32 has bits 32 bits 31:0 if you shift 0x2/0b10 left 30 bits you now have bit 32 set. What's the reason for this? It will be null and void as I see it

PR #2612, opened 2026-09-15
ast.rb: more toleration of unknown CSR widths

ThinkOpenly:
The following error can be reported when IDL has a variable of a width that may be unknown at compile time (like "SXLEN"):
```
ArgumentError: comparison of Integer with :unknown failed
```

In this case, "unknown" is OK and this case is not a catastrophic error. Ignore this case and continue.

Also, emit a better error message for incorrect field names. Turn this:
```
TypeError: Passed `nil` into T.must
```
into
```
{@field_name} is not a field of CSR[{csr_name}]"
```


tools/ruby-gems/idlc/lib/idlc/ast.rb
```diff
@@ -2967,7 +2967,10 @@ def type_check(symtab, strict:)
 
       value_result = value_try do
         idx_value = idx.value(symtab)
-        type_error "Array index (#{idx.text_value} = #{idx_value}) out of range (< #{lhs.type(symtab).width})" if idx_value >= lhs.type(symtab).width
+        lhs_width = lhs.type(symtab).width
+        if lhs_width != :unknown && idx_value >= lhs_width
+          type_error "Array index (#{idx.text_value} = #{idx_value}) out of range (< #{lhs_width})"
+        end
```
jordancarlin, line 2973: I assume this "unknown" case is handled later? Assuming that is true, then it seems fine to defer the check here.

jordancarlin (approved):

tools/ruby-gems/idlc/lib/idlc/ast.rb
ThinkOpenly: My admittedly limited understanding is that we're looking for hard errors at this point -- things that can be determined at compile time.

PR #2616, opened 2026-09-16
fix(ci): add native macOS build paths for UDB dependencies (#2583)

### LINKED ISSUE #2583: udb gem installs Linux (ELF) z3 binary instead of macOS (Mach-O) on arm64
**Describe the bug**
udb installs Linux (ELF) z3 binary instead of macOS (Mach-O) on arm64

**To Reproduce**
Environment:

    macOS, Apple Silicon (arm64)
    Ruby 3.4.10
    udb gem 0.1.16, bundler 4.0.19

Steps to reproduce:

1. Fresh bundle install that pulls in the udb gem (0.1.16) on an arm64 Mac
2. Run any command that triggers UDB config validation (e.g. make)

Note: Absolute paths in the error message have been replaced with `~`.

```
$ make
Could not open library '~/.cache/udb/z3/z3-5.0.0/arm64/libz3.so': dlopen(~/.cache/udb/z3/z3-5.0.0/arm64/libz3.so, 0x0005): tried: '~/.cache/udb/z3/z3-5.0.0/arm64/libz3.so' (slice is not valid mach-o file), '/System/Volumes/Preboot/Cryptexes/OS/Users/user/.cache/udb/z3/z3-5.0.0/arm64/libz3.so' (no such file), '~/.cache/udb/z3/z3-5.0.0/arm64/libz3.so' (slice is not valid mach-o file).
```

**Additional context**
Workaround:
brew install z3
cp "$(brew --prefix z3)/lib/libz3.dylib" ~/.cache/udb/z3/z3-5.0.0/arm64/libz3.so

drona-gyawali:
Fixes #2583

## Problem
extconf.rb was downloading Linux ELF binaries (.so) on all platforms.
On macOS.  So, we are solving this by extending the build  pipeline

## Changes

**extconf.rb** : added host_os detection, all 4 binaries now use
OS-aware download URLs and correct library extension (.so/.dylib)

**update.sh** : OS detection added to all 4 build functions; on Linux
calls existing _with_docker.sh scripts; on Mac calls new native scripts

**4 new native macOS build scripts** : build_espresso_mac.sh,
build_must_mac.sh, build_eqntott_mac.sh, build_z3_mac.sh,  build
natively using the host toolchain (no Docker), produce real Mach-O
binaries, and verify with file() before finishing

**release_udb_deps.yml** : added macos-15 (x64) and macos-15-xlarge
(arm64) runners to all 4 build jobs

**z3_loader.rb** : fixed library_name being private while called
externally; fixed FFI override to use library_name() instead of
the raw incoming name

**Note:**
I didn't test this PR locally since it heavily involves Mac hardware, release files, etc. Just giving a heads-up!


.github/workflows/release_udb_deps.yml
```diff
@@ -101,6 +101,10 @@ jobs:
             arch: x64
           - runner: ubuntu-24.04-arm
             arch: arm64
+          - runner: macos-15
+            arch: x64
+          - runner: macos-15-xlarge
+            arch: arm64
```
jordancarlin: Based on https://github.com/actions/runner-images#available-images
```suggestion
          - runner: macos-15-intel
            arch: x64
          - runner: macos-15
            arch: arm64
```

.github/workflows/release_udb_deps.yml
```diff
@@ -116,10 +120,10 @@ jobs:
             ./bin/chore update eqntott -n
           fi
 
-  build-z3:
+  build-espresso:
```
jordancarlin: Please don't change the order of the tool builds. It makes this much harder to review because GitHub shows a bunch of changes that aren't significant.

tools/ruby-gems/udb/ext/udb_download/extconf.rb
```diff
@@ -40,6 +40,20 @@ def find_executable(name)
           "Only x64 and arm64 are supported."
   end
 
+# ---- detect host OS ----
```
jordancarlin: The arch detection doesn't have this comment banner. Let's stay consistent.

tools/ruby-gems/udb/ext/udb_download/extconf.rb
```diff
@@ -40,6 +40,20 @@ def find_executable(name)
...
+os_ext =
```
jordancarlin: Why `os_ext`? This isn't an extension, it is the os.

jordancarlin (changes requested):
Are separate build scripts for macOS really needed? I assume most of the build process is the same for Linux and macOS. Can we just have a few ifs in the existing scripts instead?

drona-gyawali:
> Are separate build scripts for macOS really needed? I assume most of the build process is the same for Linux and macOS. Can we just have a few ifs in the existing scripts instead?

Thanks for the feedback! I was also thinking along the same lines , the existing build script filenames say "with_docker" and they install a Linux-based toolchain (musl + AlmaLinux), which isn't supported on macOS runners. That was my reason for keeping them separate. I'm not sure we can cleanly stitch the two together, but happy to consolidate if you prefer.

.github/workflows/release_udb_deps.yml
drona-gyawali: done , changes has been updated.

.github/workflows/release_udb_deps.yml
drona-gyawali: sorry for that. now it looks clean, i guess.

tools/ruby-gems/udb/ext/udb_download/extconf.rb
drona-gyawali: done

jordancarlin:
> Thanks for the feedback! I was also thinking along the same lines , the existing build script filenames say "with_docker" and they install a Linux-based toolchain (musl + AlmaLinux), which isn't supported on macOS runners. That was my reason for keeping them separate. I'm not sure we can cleanly stitch the two together, but happy to consolidate if you prefer.

We should do a bit of an overhaul of those and dramatically simplify the builds, but this seems fine for now and I agree we should keep them separate.

bin/.chore/update.sh
```diff
@@ -129,74 +148,89 @@ do_update_espresso() {
...
+    if [ "${current_os}" = "Mac" ]; then
+      "${UDB_ROOT}"/tools/scripts/build_espresso_mac.sh "${work_dir}/espresso-build" "${native_arch}" || exit 1
+    else
+      "${UDB_ROOT}"/tools/scripts/build_espresso_with_docker.sh "${work_dir}/espresso-build" "${native_arch}" || exit 1
+    fi
 
-    # Move the binary to the asset name expected by the gem
-    mv "${work_dir}/espresso-build/espresso" "${work_dir}/espresso-${native_arch}"
+    # include OS in asset name
+    mv "${work_dir}/espresso-build/espresso" "${work_dir}/espresso-${current_os}-${native_arch}"
```
jordancarlin: Revert this comment change.

bin/.chore/update.sh
```diff
@@ -50,7 +50,7 @@ read_git_pinned_tool_version() {
     exit 1
   fi
 
-  commit="${version#${tool}-}"
+  commit="${version#"${tool}"-}"
```
jordancarlin: This quoting seems wrong.

bin/.chore/update.sh
```diff
@@ -129,74 +148,89 @@ do_update_espresso() {
...
+    else
+      "${UDB_ROOT}"/tools/scripts/build_espresso_with_docker.sh "${work_dir}/espresso-build" "${native_arch}" || exit 1
+    fi
 
-    # Move the binary to the asset name expected by the gem
-    mv "${work_dir}/espresso-build/espresso" "${work_dir}/espresso-${native_arch}"
+    # include OS in asset name
+    mv "${work_dir}/espresso-build/espresso" "${work_dir}/espresso-${current_os}-${native_arch}"
 
-    # Generate checksum
```
jordancarlin, line 138: Why was this comment deleted?

tools/scripts/build_eqntott_mac.sh
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env bash
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
jordancarlin: Include your own copyright in new files.

jordancarlin (changes requested):
A few small changes. You also need to cryptographically sign all of your commits. Let me know if you need help with this.

Have you tested the macOS scripts on a Mac?

We should eventually also add some level of macOS CI to ensure everything continues to work, but that will need to come it a future PR because it will not work until these binaries are published, and that can't happen until this is merged.

bin/.chore/update.sh
drona-gyawali: quoting inside parameter expansion like ${version#"${tool}"-} is valid syntax in Bash/POSIX, it basically disables pattern matching so variables with characters like * or ? are treated as literal strings. However, since ${tool} is just a plain name here, the quotes aren't strictly necessary. Reverting to ${version#${tool}-} makes sense to keep the code clean

bin/.chore/update.sh
drona-gyawali: I removed it because I felt the echo "Generating checksum..." log line right below it made the comment redundant. However, I wil revert the change :)

drona-gyawali:
> A few small changes. You also need to cryptographically sign all of your commits. Let me know if you need help with this.

Thanks, just recently knew about this i will add signed commits from onward.

> Have you tested the macOS scripts on a Mac?

To be frank, I don't have access to a Mac, so I wasn't able to test the scripts locally. I mentioned this in the  PR description. The scripts follow the same pattern as the existing Docker scripts and use standard Xcode/Clang toolchain commands that are available on macOS  runners. if you're able to test them on a Mac that would be very helpful :).

> We should eventually also add some level of macOS CI to ensure everything continues to work, but that will need to come it a future PR because it will not work until these binaries are published, and that can't happen until this is merged.

Understood,  agreed that macOS CI testing makes sense as a follow-up once the binaries are published. Happy to open a follow-up issue to track that if it would be helpful

jordancarlin:
I don't want to merge this without testing it on a Mac first. I will try to do that soon, but it may have to wait a few days.

drona-gyawali:
> I don't want to merge this without testing it on a Mac first. I will try to do that soon, but it may have to wait a few days.

Hi  jordan,

To test on actual macOS hardware, I set up a GitHub Actions test workflow running on macos-latest: https://github.com/drona-gyawali/mac-script-harness

While running the builds, I caught a few platform edge cases,  which I've updated in this PR.

I'm not 100% sure if a CI harness fully replaces a local manual run for you, but all four builds passed cleanly on the macOS runner, feel free to check the workflow logs!

PR #2620, opened 2026-09-17
feat(zacas): implement amocas.w and amocas.d operation()

JulienKervarrec:
Part of #517.

Opened as a draft on purpose: I was not able to run `./do test:idl` in my environment (no access to rubygems.org, so `bundle install` fails), which means CI here is the first type-check this has had. The design question at the bottom is also worth settling before a line-by-line review.

## What this adds

`amo_cas()` in `spec/std/isa/isa/globals.isa`, plus `operation()` bodies for `amocas.w` and `amocas.d` (all four aq/rl variants each) in place of the `# TODO` stubs.

`amo_cas()` is modelled on `amo()`: alignment check, address translation for a `ReadModifyWrite`, the PMA checks, then the atomic access — here `atomic_check_then_write_{32,64}`, as suggested in #517.

## Semantics

From `zacas.adoc`:

* `amocas.w` — compare against the least-significant word of `xd` (not `xs2`), store the least-significant word of `xs2`, write the sign-extended loaded word into `xd`.
* `amocas.d` on RV64 — the same shape, 64-bit, single registers.
* `amocas.d` on RV32 — register pairs `(xd, xd+1)` and `(xs2, xs2+1)`; odd-numbered first registers are reserved; when the first register of a source pair is `x0` both halves read as zero; when the first register of the destination pair is `x0` the result is discarded.
* The address must be naturally aligned.
* Write permission is required whether or not the swap occurs.
* `aq` orders the operation whether or not it succeeded; `rl` applies only to a **successful** amocas.

## One data fix that came with it

The `description` of `amocas.w` and `amocas.d` said the comparison was against `xs2` and that the stored value came from `xs2+1`. Both contradict `zacas.adoc` and the code this PR adds, so the description is corrected for those two sizes in the same change. `b`, `h` and `q` are left alone.

## Design question: the read before the compare-and-swap

`atomic_check_then_write_{32,64}` returns only whether the write happened, but `amocas` has to place the **loaded** value in `xd` in both the success and the failure case, and it has to decide whether to perform the release, which the spec ties to success. So `amo_cas()` reads the location first, uses that value for both decisions, and still lets the builtin perform the architecturally atomic compare-and-store.

If you would rather have a builtin that returns the pre-existing value (`atomic_compare_exchange_{32,64}`), I am happy to add it together with its `backends/cpp_hart_gen` implementation instead; that would remove the extra read.

## Out of scope

* `amocas.b` / `amocas.h` (Zabha) — these need 8/16-bit CAS builtins, and the 8/16-bit atomic path is being added in #2566.
* `amocas.q` — needs a 128-bit CAS builtin.

Their layout bodies keep the existing `# TODO` and their generated files are unchanged.

## Generated files

`spec/std/isa/inst/Zacas/amocas.SIZE.AQRL.layout` is the source; the eight `amocas.{w,d}[.aq|.rl|.aqrl].yaml` files are its `./do gen:arch` output.


JulienKervarrec:
While this waits for the workflow runs to be approved, here is what I could check locally. I still can't run `./do test:idl`, because rubygems.org is unreachable from my environment.

- I regenerated all 20 `amocas.*.yaml` files from `amocas.SIZE.AQRL.layout`, using the same ERB + `insert_warning` step as the Rakefile. The output is byte-identical to the committed files.
- The branch merges cleanly with current `main`, and with #2619 in either order.
- The aq/rl, write-permission and RV32 register-pair handling follows the normative statements in `zacas.adoc` (`Zacas_amocas_mem_op_fail_aq_rl`, `Zacas_amocas_w_permission`, and the `Zacas_rv32_amocas-d_*` rules).

Could a maintainer approve the workflow runs? The IDL type-check is the one check I can't reproduce here. I'll mark this ready for review once it's green.

